PR #365, opened 2020-11-04
Add kine smoke test

ncopa:
---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**What this PR Includes**

Add a simple smoke test with kine as storage backend.

### DIFF
--- a/inttest/.gitignore
+++ b/inttest/.gitignore
@@ -1,2 +1,3 @@
 **/*.tar.gz
 bin/
+kine.conf
--- a/inttest/Makefile
+++ b/inttest/Makefile
@@ -1,6 +1,7 @@
 
 ARCH = $(shell go env GOARCH)
 OS = $(shell go env GOOS)
+K0S_PATH := $(realpath ../k0s)
 
 sonobuoy_url = https://github.com/vmware-tanzu/sonobuoy/releases/download/v0.19.0/sonobuoy_0.19.0_$(OS)_$(ARCH).tar.gz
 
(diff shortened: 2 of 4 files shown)


inttest/.gitignore
```diff
@@ -1,2 +1,3 @@
 **/*.tar.gz
-bin/
\ No newline at end of file
+bin/
+kine.conf
```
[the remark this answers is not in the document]
ncopa: `kine.conf` is generated from Makefile and deleted with `make clean`. I don't think we want `kine.conf` checked in to git?

inttest/Makefile
```diff
@@ -17,23 +18,30 @@ bin:
 bin/sonobuoy: | bin
 	$(curl) $(sonobuoy_url) | tar -C bin/ -zxv $(notdir $@)
 
+kine.conf:
```
trawler, line 21: shouldn't this be `k0s.yaml` (or `k0s.conf`)? since it's not actually kine's config, but k0s' config.

trawler (changes requested):

inttest/common/footloosesuite.go
```diff
@@ -189,6 +189,13 @@ func (s *FootlooseSuite) keepEnvironment() bool {
 	}
 }
 
+func k0sConfigFlag() string {
+	if os.Getenv("K0S_CONFIG") != "" {
+		return "--config /etc/k0s.conf"
```
kke, line 194: Always `/etc/k0s.conf`? I would expect it to use whatever path it finds in `K0S_CONFIG`.

kke, line 194: Is this supposed to always return the hardcoded path? The code below will then have the same source and destination:

```go
		volumes = append(volumes, config.Volume{
			Type:        "bind",
			Source:      configFile, // <= will be /etc/k0s.conf when K0S_CONFIG is set.
			Destination: "/etc/k0s.conf",
		})
```

inttest/Makefile
ncopa: We could use `k0s.yaml` or `k0s.conf` but this is config for check-basic testing kine specifically. We could other configs for other tests in future too.

inttest/common/footloosesuite.go
ncopa: No, the `Source` will be `osGetenv("K0S_CONFIG")`, which is a path on the host running the test. This will be bind mounted to `/etc/k0s.conf` in the footloose container, and `k0sConfigFlag()` will append the `--config /etc/k0s.conf` to `k0s` if the bindmount was added to footloose regardless of what the env var was set to. (i.e the test runs with a specified config instead of the defaults).

```diff
@@ -402,6 +410,15 @@ func (s *FootlooseSuite) createConfig() config.Config {
 
        volumes = append(volumes, s.ExtraVolumes...)
 
+       configFile := os.Getenv("K0S_CONFIG")
+       if configFile != "" {
+               volumes = append(volumes, config.Volume{
+                       Type:        "bind",
+                       Source:      configFile,
+                       Destination: "/etc/k0s.conf",
+               })
+       }
+
        portMaps := []config.PortMapping{
                {
                        ContainerPort: 22,

```

inttest/Makefile
ncopa: I renamed it to `k0s-kine.conf`

ncopa:
This is followed up in #383

PR #367, opened 2020-11-05
verify SANs with regenerateCert func

### LINKED ISSUE #345: Changing SANs is not reflected upon mke restart
**Version**
```
current main
```
**Platform**
Which platform did you run mke on?
```
any
```
**What happened?**

Changing e.g. SANs in cluster settings will not make mke to re-create the api serving cert

**How To Reproduce**

Add SAN attribute and re-run mke, the serving cert is not re-generated.

**Expected behavior**

mke should re-create the serving cert if the relevant config changes.

**Screenshots & Logs**
N/A

trawler:

**Issue**
Fixes #345

**What this PR Includes**
This PR adds an regenerateCert function to the certificate manager in order to detect changes in the configured SANs and re-generate the certificate, if exist.

**e2e Tests**
https://github.com/k0sproject/k0s/runs/1363701816?check_suite_focus=true ✅

### DIFF
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -21,16 +21,19 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/pkg/errors"
+
+	"github.com/cloudflare/cfssl/certinfo"
 	"github.com/cloudflare/cfssl/cli"
 	"github.com/cloudflare/cfssl/cli/genkey"
 	"github.com/cloudflare/cfssl/cli/sign"
 	"github.com/cloudflare/cfssl/csr"
 	"github.com/cloudflare/cfssl/initca"
 	"github.com/cloudflare/cfssl/signer"
+	"github.com/sirupsen/logrus"
+
 	"github.com/k0sproject/k0s/pkg/constant"
 	"github.com/k0sproject/k0s/pkg/util"
-	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/certificate/manager.go
```diff
@@ -111,23 +114,34 @@ func (m *Manager) EnsureCertificate(certReq Request, ownerName string) (Certific
...
-			return Certificate{}, errors.Wrapf(err, "failed to read ca cert %s for %s", certFile, certReq.Name)
-		}
-		key, err := ioutil.ReadFile(keyFile)
-		if err != nil {
-			return Certificate{}, errors.Wrapf(err, "failed to read ca key %s for %s", keyFile, certReq.Name)
-		}
+	// this loop lets regenerateCert asses if a certificate needs to be generated
+	// certsExist is a dummy value for a single-run loop
+
+	for certsExist = true; certsExist; {
```
jnummelin, line 122: Couldn't we achieve the same logic without the loop around if we do something like:
```
if !m.regenerateCert(certReq, certFile) {
   // Load and return here, the same stuff that happens in the else block now
}

```

ncopa:
I'd like to have an explanation in the commit message on why the change is needed, what problem it solves.

trawler:
> I'd like to have an explanation in the commit message on why the change is needed, what problem it solves.

done

jnummelin (approved):

Issue #207, opened 2020-09-25
Logging levels should be configurable at runtime


trawler:
Right now our log levels are fixed in the code. We should make it configurable at runtime, per component.

PR #372, opened 2020-11-06
Cobra CLI + Granulated Logging

### LINKED ISSUE #344: Adjust logging in the `k0s` supervisor process
Now we have a lot of logs from supervised processed. It would be nice to have way to adjust log levels for them.
Also, it would be good to stop logging at all during shutdown and describe some friendly message to the user, that we are working on shutting down.

trawler:

**Issues**
Fixes #344 and #207 

**What this PR Includes**
* Refactor the cmd package to use cobra cli instead of urfave/cli, in order to allow more structured flags (such as maps)
* Add a logging flag for the different components
* Sets default log levels
* Add docs command for quick generation of command line documentation

**e2e Tests**
https://github.com/k0sproject/k0s/runs/1379420641?check_suite_focus=true


cmd/root.go
```diff
@@ -0,0 +1,134 @@
...
+	"kube-apiserver":          "1",
+	"kube-controller-manager": "1",
+	"kube-scheduler":          "1",
+	"kubelet":                 "1",
+}
+
+func init() {
+	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", "config file (default is $HOME/k0s.yaml)")
+	rootCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "Debug logging (default is false)")
+	rootCmd.PersistentFlags().StringToStringVarP(&cmdLogLevels, "logging", "l", defaultLogLevels, "Logging Levels for the different components")
```
jnummelin, line 37: Again, why separately set here and not part of the command object creation?

cmd/server.go
```diff
@@ -45,46 +45,31 @@ import (
...
+		Short: "Run server",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				serverToken = args[0]
+			}
+			err := startServer(serverToken)
+			if err != nil {
+				return err
+			}
+			return nil
```
jnummelin, line 69:
```suggestion
			return startServer(serverToken)
```

cmd/server.go
```diff
@@ -45,46 +45,31 @@ import (
...
-
-	errors := clusterConfig.Validate()
-	if len(errors) > 0 {
-		messages := make([]string, len(errors))
-		for _, e := range errors {
-			messages = append(messages, e.Error())
-		}
-		return nil, fmt.Errorf("config yaml does not pass validation, following errors found:%s", strings.Join(messages, "\n"))
+var (
+	serverProfile string
```
jnummelin, line 54: should be `workerProfile`, also need to be changed for the flag

cmd/token.go
```diff
@@ -24,39 +24,36 @@ import (
...
-			&cli.StringFlag{
-				Name:      "kubeconfig",
-				Usage:     "path to kubeconfig",
-				Value:     constant.AdminKubeconfigConfigPath,
-				EnvVars:   []string{"KUBECONFIG"},
-				TakesFile: true,
-			},
-		},
+func init() {
+	tokenCmd.Flags().StringVar(&kubeConfig, "kubeconfig", "/var/lib/k0s/pki/admin.conf", "path to kubeconfig file [$KUBECONFIG]")
```
jnummelin, line 40: Why not use the const for the admin config path?

cmd/token.go
```diff
@@ -24,39 +24,36 @@ import (
...
+	tokenCreateCmd.Flags().StringVar(&workerRole, "role", "worker", "Either worker or controller")
+	tokenCreateCmd.Flags().BoolVar(&waitCreate, "wait", false, "wait forever (default false)")
+
+	tokenCmd.AddCommand(tokenCreateCmd)
 }
 
 var (
+	kubeConfig  string
+	tokenExpiry string
+	workerRole  string
```
jnummelin, line 54: maybe `tokenRole` as it's the role for the token usage

cmd/root.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+)
+
+func initConfig() error {
+	if cfgFile == "" {
+		home, err := util.HomeDir()
+		if err != nil {
+			return err
+		}
+		cfgFile = filepath.Join(home, "k0s.yaml")
```
jnummelin, line 100: This actually changes the behaviour to try to load the config file from users home by default. IMO not good idea, not very common place to look configs from. I'd rather preserve the existing behaviour where we try to load it from the current dir.

jnummelin (changes requested):

cmd/api.go
```diff
@@ -27,37 +27,37 @@ import (
...
 	}
-}
-
-var kubeClient k8s.Interface
+)
 
-func startAPI(ctx *cli.Context) error {
-	clusterConfig, err := configFromCmdFlag(ctx)
+func startAPI() error {
+	clusterConfig, err := ConfigFromYaml(cfgFile)
```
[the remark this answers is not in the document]
trawler: it comes from the global flag here: https://github.com/k0sproject/k0s/blob/a85972c6529b264d1f622ab2d48fa8f92dfa9337/cmd/root.go#L35

and initialized here: https://github.com/k0sproject/k0s/blob/a85972c6529b264d1f622ab2d48fa8f92dfa9337/cmd/root.go#L94-L106

cmd/root.go
trawler: these are global flags, so they don't have to be set separately for each command.

jnummelin: yep. should've read better the cobra way of defining these before commenting :)

cmd/root.go
trawler: fixed

cmd/server.go
trawler: changed to `serverWorkerProfile` so as to not collide with `workerProfile` in `worker.go`

jnummelin (approved):
LGTM once the smokes pass

Issue #277, opened 2020-10-12
Automate conformance test runs


jnummelin:
In similar way as we can currently run the sig-network tests against VMs

jnummelin:
it's almost fully automated in #377

PR #377, opened 2020-11-09
Terraform setup for sonobuoy

jasmingacic:
Signed-off-by: jasmingacic <jasmin.gacic@gmail.com>

---
name: Pull Request
about: Create a Pull Request
title: 'Terraform setup for sonobuoy'
labels: ''
assignees: ''

---

**What this PR Includes**
Simplified setup for deploying and running sonobuoy

### DIFF
--- a/inttest/terraform/test-cluster/main.tf
+++ b/inttest/terraform/test-cluster/main.tf
@@ -14,6 +14,7 @@ resource "aws_key_pair" "cluster-key" {
 
 // Save the private key to filesystem
 resource "local_file" "aws_private_pem" {
+  file_permission = "700"
   filename = format("%s/%s", path.module, "aws_private.pem")
   content = tls_private_key.k8s-conformance-key.private_key_pem
 }
--- a/inttest/terraform/test-cluster/output.tf
+++ b/inttest/terraform/test-cluster/output.tf
@@ -5,3 +5,8 @@ output "controller_external_ip" {
 output "worker_external_ip" {
   value = aws_instance.cluster-workers.*.public_ip
 }
+
+
+output "controller_pem" {
+  value = local_file.aws_private_pem
+}
--- a/inttest/conformance/README.md
+++ b/inttest/conformance/README.md
@@ -1,0 +1,88 @@
+# Sonobuoy 
+
+This is a TF configuration to easily setup a K0s cluster with AWS and start sonobuoy.
+Requirements:
+
+1. Terraform
+2. AWS credentials
+
+To provision required setup follow next steps:
+
+```
+$ cd terraform
+$ terraform init
+$ terraform apply
+module.k0s-sonobuoy.data.aws_ami.ubuntu: Refreshing state...
+module.k0s-sonobuoy.tls_private_key.k8s-conformance-key: Creating...
+module.k0s-sonobuoy.tls_private_key.k8s-conformance-key: Creation complete after 1s [id=07fae4a3e454177b6156c3342d6d92008426d703]
+module.k0s-sonobuoy.local_file.aws_private_pem: Creating...
+...
+Apply complete! Resources: 17 added, 0 changed, 0 destroyed.
+
+Outputs:
+
+controller_ip = [
+  "54.73.141.241",
+]
+```
+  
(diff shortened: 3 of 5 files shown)


inttest/terraform/test-cluster/main.tf
```diff
@@ -14,6 +14,7 @@ resource "aws_key_pair" "cluster-key" {
 
 // Save the private key to filesystem
 resource "local_file" "aws_private_pem" {
+  file_permission = "700"
```
jasmingacic, line 17: If you want to use this key from the terminal this permission level is required. Otherwise it will not connect to the server.

inttest/terraform/test-cluster/output.tf
```diff
@@ -5,3 +5,8 @@ output "controller_external_ip" {
 output "worker_external_ip" {
   value = aws_instance.cluster-workers.*.public_ip
 }
+
+
+output "controller_pem" {
+  value = local_file.aws_private_pem
```
jasmingacic, line 11: I needed this value to setup connection for file and remote provisioners when using it in the module.

inttest/sonobuoy/terraform/README.md
```diff
@@ -0,0 +1,43 @@
...
+ubuntu@controller-0:~$ sonobuoy status
+ubuntu@controller-0:~$ export KUBECONFIG=/var/lib/k0s/pki/admin.conf
+ubuntu@controller-0:~$ sonobuoy status
+         PLUGIN     STATUS   RESULT   COUNT
+            e2e    running                1
+   systemd-logs   complete                2
+   systemd-logs    running                1
+
+Sonobuoy is still running. Runs can take up to 60 minutes.
+```
```
jnummelin, line 43: Add some instructions how to get the sonobuoy result tarball. That is needed to be able to submit for the official CNCF conformance. Probably is something like:
- `sonobuoy results`
- from local machine: `scp ubuntu@controller:.<result file> .`

inttest/sonobuoy/terraform/gh.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+set -e
+sleep 30
+
+echo "prepare environment"
+sudo snap install jq
+sudo snap install kubectl --classic
+
+# Parse CLI args
+readonly git_tag="v0.7.0-beta1"
```
jnummelin, line 10: Could we take this as input? And feed it via TF variables? This way it would be super easy to run conformance for any tag/version.

inttest/sonobuoy/terraform/gh.sh
```diff
@@ -0,0 +1,46 @@
...
+readonly github_repo_owner="k0sproject"
+readonly github_repo_name="k0s"
+readonly release_asset_filename="k0s-v0.7.0-beta1-amd64"
```
jnummelin, line 13: This we should be able to concat based on the version variable

jasmingacic: I have removed this script entirely

inttest/sonobuoy/terraform/README.md
jasmingacic: Updated it

jnummelin (approved):
LGTM

jasmingacic:
This should close #277

jnummelin (dismissed):

trawler (changes requested):
small nitpick:
this PR then creates the path `inttest/sonobuoy/terraform`, while we also have `inttest/terraform`.
To the casual observer, these two paths would look the same, and there is no documentation to point out what each paths do.
Perhaps it would be best to "clean up" the paths, or use a more distinguished name.

jasmingacic:
@trawler yeah it makes sense, but I also wanted to keep it grouped. What are you suggesting we should do?

Also I've added README which assumes that you are going to be running this from inttest/sonobuoy/terraform.

jasmingacic:
As per our cafe conversation I've moved this under inttest/conformance.

Squashed and ready to be merged

trawler (approved):
lgtm! 👍

PR #378, opened 2020-11-10
Issue/184/helm integration

### LINKED ISSUE #184: Manifest applier needs to support Helm charts
Flux seems to have good abstractions for this.

Other option is to bundle Helm and use it through golang API.

Probably needs #182 first though

soider:
---
name: Pull Request
about: Create a Pull Request
title: Helm based addons
labels: 
assignees: ''

---

**Issue**
Fixes #184 

**What this PR Includes**
Implementation of the helm based addons for the k0s cluster


docs/configuration.md
```diff
@@ -178,10 +191,31 @@ telemetry:
...
+    url: https://prometheus-community.github.io/helm-charts
+  charts:
+  - name: prometheus-stack
+    chartname: prometheus-community/prometheus
+    version: "11.16.8"
+    values: |2
+        values: "for overriding"
+    namespace: default
+```
+
```
jnummelin, line 213: I think we should also document the CRD structure and how we pick those up from `kube-system` namespace.

pkg/apis/v1beta1/cluisterhelmaddons.go
```diff
@@ -0,0 +1,28 @@
+package v1beta1
```
jnummelin, line 1: Needs Apache 2 "header"

jnummelin (changes requested):
Couple minor change requests. Other than that fantastic stuff.

cmd/server.go
```diff
@@ -279,6 +280,13 @@ func createClusterReconcilers(clusterConf *config.ClusterConfig) map[string]comp
 
 	initNetwork(reconcilers, clusterConf)
 
+	manifestsSaver, err := server.NewManifestsSaver("helm")
+	if err != nil {
+		logrus.Warnf("failed to initialize reconcilers manifests saver: %s", err.Error())
+	}
+	reconcilers["crd"] = server.NewCRD(manifestsSaver)
```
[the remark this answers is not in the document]
soider: yes

.golangci.yml
```diff
@@ -2,8 +2,9 @@
 run:
 
   timeout: 5m
-  skip-dirs-use-default: true
-
+  skip-dirs-use-default: false
+  skip-dirs:
+    - pkg/helm
```
jnummelin, line 7: Why are we skipping this? Is it because generated code? If so, I'd appreciate a comment here on the reasons.

jnummelin (approved):

PR #382, opened 2020-11-10
Add fallback for MachineID based on hostname

### LINKED ISSUE #353: Add fallback for missing /etc/machine-id
There is no guarantee that `/etc/machine-id` exists on the host system. Would be convenient with a fallback if it doesn't.

Would be nice if mke would:
- use `/var/lib/mke/machine-id` if it exists. If not:
  - copy `/etc/machine-id` to `/var/lib/mke/machine-id` if it exists, otherwise:
    - generate our own machine id and store as `/var/lib/mke/machine-id`

Other option would be to use checksum(hostname) as fallback.

Alternatively we could also let `mke` create `/etc/machine-id` if its missing. It shouldn't hurt anything.

**Additional context**
I think `kubelet` may require a working `/etc/machine-id` so maybe the best idea is to generate it if it is missing.

ncopa:
Generate a machine id from md5sum of hostname in case /etc/machine-id is missing.

fixes #353

### DIFF
--- a/pkg/util/machineid.go
+++ b/pkg/util/machineid.go
@@ -1,8 +1,23 @@
 package util
 
-import "github.com/denisbrodbeck/machineid"
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"os"
+
+	"github.com/denisbrodbeck/machineid"
+)
 
 // MachineID returns protected id for the current machine
 func MachineID() (string, error) {
-	return machineid.ProtectedID("k0sproject-k0s")
+	id, err := machineid.ProtectedID("k0sproject-k0s")
+	if err != nil {
+		name, err := os.Hostname()
+		if err != nil {
+			return "", err
+		}
+		sum := md5.Sum([]byte(name))
+		id = hex.EncodeToString(sum[:])
+	}
+	return id, err
 }


ncopa:
Creating `/etc/machine-id` if its missing may be a bit controversial.

jnummelin:
> Creating /etc/machine-id if its missing may be a bit controversial.

I agree that it feels something k0s should not be doing.

What if instead we create some similar looking random ID in e.g. `/var/lib/k0s/machine-id`? This way we do pretty much the same thing, but create it purely for k0s purposes rather than trying to be general machineId.

ncopa:
> What if instead we create some similar looking random ID in e.g. `/var/lib/k0s/machine-id`? This way we do pretty much the same thing, but create it purely for k0s purposes rather than trying to be general machineId.

That was my original thought too, but also kubelet needs /etc/machine-id so I thought we could fix both in one go. However, I think the unique id is only used by the controllers and kubelet only needs machine-id on workers, so it might be better to not write to /etc/machine-id.

I wonder if we should do it as simple as use an md5 sum of hostname as fallback. Then we don't need to write anything at all.

s0j:
> I wonder if we should do it as simple as use an md5 sum of hostname as fallback. Then we don't need to write anything at all.

This makes the `machine-id` deterministic, which seems to be at odds with `man 5 machine-id` that says this should be treated as 'confidential'.

> This ID uniquely identifies the host. It should be considered "confidential", and must not be
> exposed in untrusted environments, in particular on the network. If a stable unique identifier
> that is tied to the machine is needed for some application, the machine ID or any part of it must
> not be used directly. Instead the machine ID should be hashed with a cryptographic, keyed hash
> function, using a fixed, application-specific key. That way the ID will be properly unique, and
> derived in a constant way from the machine ID but there will be no way to retrieve the original
> machine ID from the application-specific one. The `sd_id128_get_machine_app_specific(3)` API
> provides an implementation of such an algorithm.

ncopa:
> > I wonder if we should do it as simple as use an md5 sum of hostname as fallback. Then we don't need to write anything at all.
> 
> This makes the `machine-id` deterministic, which seems to be at odds with `man 5 machine-id` that says this should be treated as 'confidential'.

This is a fallback for handling the case when there are no `/etc/machine-id` on the node, so the leader election works instead of fail with error. It is also used for the telemetry as fallback. The thinking here is that if you need a proper machine id, the create `/etc/machine-id`, but don't make it a fatal error if it does not exist.

Do you think we should fail with error instead?

s0j:
> This is a fallback for handling the case when there are no `/etc/machine-id` on the node, so the leader election works instead of fail with error. It is also used for the telemetry as fallback. The thinking here is that if you need a proper machine id, the create `/etc/machine-id`, but don't make it a fatal error if it does not exist.
> 
> Do you think we should fail with error instead?

What about failing with an error that instructs to re-run using some `--generate-machine-id` argument?  This makes it explicit to the user that something required is missing.  They can either fix the machine-id themselves (if this was accidental), or just add the argument.

If `/etc/machine-id` is being created, I think it should be random bytes just to align better with what the manpage states.

ncopa:
> If `/etc/machine-id` is being created, I think it should be random bytes just to align better with what the manpage states.

It no longer creates `/etc/machine-id`. I refactored it to not leave it to the user or OS to deal with.

pkg/util/machineid.go
```diff
@@ -1,8 +1,24 @@
...
 // MachineID returns protected id for the current machine
 func MachineID() (string, error) {
-	return machineid.ProtectedID("k0sproject-k0s")
+	id, err := machineid.ProtectedID("k0sproject-k0s")
+	if err != nil {
+		name, err := os.Hostname()
+		if err != nil {
+			return "", err
+		}
+		hasher := md5.New()
```
s0j, line 19: You could just use `md5.Sum()` directly.

s0j (dismissed):

soider (approved):
lgtm

PR #395, opened 2020-11-13
Add "resolv-conf" flag to Kubelet for CoreDNS

### LINKED ISSUE #390: Fix CoreDNS loop problem
When a system uses systemd-resolvd, coreDNS will detect a loop interface and will fail to load.

We can detect if a system uses systemd-resolvd (with https://github.com/moby/libnetwork/blob/3c26b4e7495ef79690ab876eb0400fd5793516e6/resolvconf/resolvconf.go#L40, for example) and if so, add a `--resolv-conf` flag to kubelet with the "real" resolv.conf path.

trawler:
**Issue**
Fixes  #390

**What this PR Includes**
This PR removes the need for "hacking" the `resolv.conf` file in systemd-resolvd systems, by finding out the "real" resolv.conf file and passing it as the value for the `--resolv-conf` flag in kubelet.

**e2e**
https://github.com/k0sproject/k0s/actions/runs/361469902 ✅

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/cloudflare/cfssl v1.4.1
 	github.com/containerd/containerd v1.4.1 // indirect
 	github.com/denisbrodbeck/machineid v1.0.1
+	github.com/docker/libnetwork v0.5.6
 	github.com/evanphx/json-patch v4.9.0+incompatible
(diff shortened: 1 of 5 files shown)


jasmingacic (approved):
Tested it with multipass and it worked out of the box. LGTM

PR #407, opened 2020-11-16
Clarify readme

ncopa:
**What this PR Includes**

Try make it slightly clearer what k0s is and does, and how it is different from other kubernetes distros by
    
- Remove redundant and non essential text
- List key features early (above the demo)
- Move build and smoke tests to the bottom

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,38 +77,6 @@ While some Kubernetes distros package everything and the kitchen sink in, k0s tr
 
 With strong enough arguments we might take in new addons but in general those should be something that are essential for the "core" of k0s.
 
-## Build
-
-`k0s` can be built in 3 different ways:
-
-Fetch official binaries (except `kine` and `konnectivity-server`, which are built from source):
-```
-make EMBEDDED_BINS_BUILDMODE=fetch
-```
-
-Build Kubernetes components from source as static binaries (requires docker):
-```
-make EMBEDDED_BINS_BUILDMODE=docker
-```
-
-Build k0s without any embedded binaries (requires that Kubernetes
-binaries are pre-installed on the runtime system):
-```
-make EMBEDDED_BINS_BUILDMODE=none
-```
-
-Builds can be done in parallel:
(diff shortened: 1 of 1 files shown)


ncopa:
I think I forgot to add a line about where to find the `/var/lib/k0s/pki/admin.conf`

README.md
```diff
@@ -75,38 +77,6 @@ While some Kubernetes distros package everything and the kitchen sink in, k0s tr
...
-## Smoke test
-
-To run a smoke test after build:
-```
-make check-basic
-```
-
 ## Cluster bootstrapping
 
 Move the built `k0s` binary to each of the nodes.
```
jasmingacic, line 82: I'd suggest adding this too:
```
curl -SsLf get.k0s.sh 
```

README.md
```diff
@@ -128,6 +98,8 @@ Join a new worker node to the cluster by running:
 k0s worker "superlongtokenfrompreviousphase"
 ```
 
+The client config (for kubectl or lens) can be found in `/var/lib/k0s/pki/admin.conf` on the controller.
+
```
jasmingacic, line 102: We should also point out that "localhost" needs to be replaced if hosted elsewhere than cloud.

jasmingacic (changes requested):

README.md
jasmingacic: I can't comment on the line below but I'd also give an example with `enable-worker`

ncopa: Do we need `--enable-worker` example in the readme when we have it in the [FAQ](https://github.com/k0sproject/k0s/blob/main/docs/FAQ.md#how-do-i-run-a-single-node-cluster)?

jasmingacic: I'm thinking it is most likely how someone is going to try k0s

jasmingacic (approved):
looking good, let's ship it

jnummelin (approved):

PR #412, opened 2020-11-17
Refactor config doc structure

jnummelin:

**Issue**
N/A

**What this PR Includes**
Refactor config doc structure. So to not overwhelm new users now we first show the typical things user should be looking at in the config. Only at the end there's full config reference.

This also fixes some bits on the extension docs, they were not fully in sync with code. :)


jnummelin:
Marked as draft, I'll add some more config docs refactoring to this at the same time.

trawler (approved):

PR #423, opened 2020-11-18
Add data-dir flag to k0s

### LINKED ISSUE #405: Ability to configure the data directory location of k0s
This is a follow-up of https://github.com/k0sproject/k0s/pull/399#issuecomment-727948675.

Currently, k0s picks `/var/lib/k0s` and `/run/k0s` as the main directories for storing its files but the user is unable to change them.

The ability to set both `/var/lib/k0s` and `/run/k0s` to a new location. An example of the command line parameters could be `--data-dir` for changing the location of the directory `/var/lib/k0s` and `--run-dir` for changing the location of the directory `/run/k0s`.

N/A

**Additional context**
N/A

trawler:
**Issue**
Fixes #405

**What this PR Includes**
* Moves RunDir to be under DataDir (instead of using `/run/k0s`)
* Adds a data-dir flag, and make all resulting dir paths configurable as a result


pkg/component/server/etcd.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
```
ncopa, line 23: I think this was non-intentional due to a resolved merge conflict. I think we shouldn't use "path" here.

pkg/component/server/etcd.go
```diff
@@ -66,28 +59,45 @@ func (e *Etcd) Init() error {
...
+		"ca.crt",
+		"server.crt",
+		"server.key",
+	} {
+		if err := os.Chown(path.Join(e.K0sVars.EtcdCertDir, f), e.uid, e.gid); err != nil && os.Geteuid() == 0 {
+			// TODO: directory may not yet exist. log it and wait for retry for now
+			logrus.Errorf("failed to chown %s: %s", f, err)
+		}
+	}
+	return assets.Stage(e.K0sVars.BinDir, "etcd", constant.BinDirMode)
```
ncopa, line 89: I don't think we should re-introduce this. See commit 3daba31160eb1c83ae4abf6c0e6a6553a6ba5182

ncopa (changes requested):

jnummelin (approved):
I'd like to see a warning on the data-dir option that changing it for existing setup WILL break things badly.

trawler:
> I'd like to see a warning on the data-dir option that changing it for existing setup WILL break things badly.

added + typo fix

Issue #413, opened 2020-11-17
add --token-file $file for joining nodes


curx:
**Is your feature request related to a problem? Please describe.**

for joining nodes to the cluster we use the set token on cmdline

**Describe the solution you'd like**

Add an option --token-file /path/to/token-file for joining nodes

**Additional context**

If I get the `ps -C k0s -o cmd` we can see the whole token in proc

```
# join a node k8s server $token
$ ps -C k0s -o cmd
CMD
k0s server A....long...token
```

jasmingacic:
@curx Thank you for the suggestion.

In order to join a node to the cluster you should run `k0s worker [token]` you can find more details [here](https://github.com/k0sproject/k0s/blob/main/docs/create-cluster.md#joining-workers-to-cluster)

jasmingacic:
@curx we just adde ability to pass in token-file to `k0s worker` [here](https://github.com/k0sproject/k0s/pull/420)

PR #426, opened 2020-11-18
Added --token-file argument to server command

jasmingacic:
**What this PR Includes**
Added --token-file argument to server command
I also extended what was done in #413 
Plus I added examples for the command.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/signal"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


cmd/server.go
```diff
@@ -58,10 +60,30 @@ var (
 	serverCmd = &cobra.Command{
 		Use:   "server [join-token]",
 		Short: "Run server",
+		Example: `	Command to associate master nodes:
```
soider, line 63: I don't know if it was an intention, but there is tab character here and later in the example strings

soider (approved):
lgtm. 

There are some tab characters in the string literals, if it's part of the plan, go ahead, please.

cmd/server.go
jasmingacic: It is required so when the example is printed it aligns correctly 
```
Usage:
  k0s server [join-token] [flags]

Examples:
	Command to associate master nodes:
	CLI agument:
	$ k0s server [join-token]

	or CLI flag:
	$ k0s server --token-file [path_to_file]
	Note: Token can be passed either as a CLI argument or as a flag
```
vs without tab
```
Usage:
  k0s server [join-token] [flags]

Examples:
Command to associate master nodes:
	CLI agument:
	$ k0s server [join-token]

	or CLI flag:
	$ k0s server --token-file [path_to_file]
	Note: Token can be passed either as a CLI argument or as a flag
```

Issue #389, opened 2020-11-11
`k0s user create`


jnummelin:
**Is your feature request related to a problem? Please describe.**

By default k0s configures only cert based auth for API server and only the `admin.conf` with cluster admin role bound to it.

We should make available some subcommand to easily create a new cert and kubeconfig for a new user.

**Describe the solution you'd like**
`k0s user create jussi --group foo,bar`

That would dump out a kubeconfig with embedded certs for a user called `jussi`. In the cert there's groups for `foo` and `bar`.

**Describe alternatives you've considered**

**Additional context**

binoychitale:
Hi @jnummelin , I would like to have a go at this issue (since it is labeled as a good first issue).
From what I understand, we would need to do the following:

1. Create a cobra subcommand for the "user create" functionality 
2. Within this new subcommand, we would create a signed certificate (using the existing certificate library within the project) + pub key.
3.  Dump the **certificate** and **public key** into a kubeConfig template (the way its being done for admin.conf) and write it to stdout.

Does this look right to you? If yes, can I begin implementing the same?

Thanks.

jnummelin:
@binoychitale That sounds like exactly what I've had in mind for this. 👍 

One thing we need to make clear both in docs and in command description is that there's absolutely no way to revoke those certs. That's just how kube works currently.

PR #436, opened 2020-11-19
Add user certs

binoychitale:
---
name: Adds subcommand to create a signed certificate for a particular user.

---

**Issue**
Implements #389 

**What this PR Includes**
1) Creates a subcommand for user certificate creation
2) Uses internal `certificate` package to generate a signed certificate and public key for a user
3) Dumps generated kubeConfig to stdout


binoychitale:
@jnummelin I have implemented the features needed for this task. However I ran into a couple of snags.

1. `make check-basic`and make `check-hacontrolplane` test suites are failing due to some problems SSHing into the container.
2. When we print the kubeConfig to stdout, there are also some info logs printed internally by the cert generation package.
It looks something like -
```
2020-11-19 16:12:30.539094 I | [INFO] generate received request
2020-11-19 16:12:30.540157 I | [INFO] received CSR
2020-11-19 16:12:30.541662 I | [INFO] generating key: rsa-2048
2020-11-19 16:12:30.828439 I | [INFO] encoded CSR
2020-11-19 16:12:30.832278 I | [INFO] signed certificate with serial number 650774980094702345245377096345971566074187296622

apiVersion: v1
clusters:
...
```
Is there something we should do about this or leave it as is?

jnummelin:
First of all 👏  for the PR. We'll take a deeper look at the code tomorrow.

> make check-basicand make check-hacontrolplane test suites are failing due to some problems SSHing into the container. Is this a known issue?(FYI I am on MacOS)

@soider I think you mentioned hitting into something similar on mac. I'm on Linux and have not seen this kinda flakes.

> When we print the kubeConfig to stdout, there are also some info logs printed internally by the cert generation package.

Maybe we could explicitly put the log level into warn/error on the command? I think we do it already on some other commands too.

soider:
@binoychitale I had the same issues when I used default docker desktop settings with pretty limited CPU and Memory limits. Locally it works for me if I allow docker to use more cores, could you try the same?

Thank you for the contribution, we also need you do the sign off process as described  https://github.com/k0sproject/k0s/blob/main/docs/contributors/testing.md

binoychitale:
>Maybe we could explicitly put the log level into warn/error on the command?

Thanks @jnummelin, I have set up the log level to `Fatal` for command execution.

binoychitale:
>@binoychitale I had the same issues when I used default docker desktop settings with pretty limited CPU and Memory limits. Locally it works for me if I allow docker to use more cores, could you try the same?

@soider thanks for the suggestion, increasing CPU and memory worked! However, now I am getting another error post SSH keygen
```
=== RUN   TestHAControlplaneSuite
time="2020-11-20T14:41:42+05:30" level=info msg="Machine TestHAControlplaneSuite-controller0 hasn't been created..."
time="2020-11-20T14:41:42+05:30" level=info msg="Machine TestHAControlplaneSuite-controller1 hasn't been created..."
time="2020-11-20T14:41:42+05:30" level=info msg="Creating SSH key: /var/folders/9l/8djp461s6x5gzq74g3x4y2v40000gn/T/footloose-keys354652257/id_rsa ..."
time="2020-11-20T14:41:44+05:30" level=info msg="Pulling image: footloose-alpine ..."
Error response from daemon: pull access denied for footloose-alpine, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
I'm looking into possible causes for this, but do let me know if you've faced the same before

I see that the same checks have passed on the PR hook, so moving this out of draft mode.

cmd/user.go
```diff
@@ -0,0 +1,122 @@
...
+			if len(args) == 0 {
+				return errors.New("Username is mandatory")
+			}
+			var username = args[0]
+			clusterConfig, err := ConfigFromYaml(cfgFile)
+			if err != nil {
+				return err
+			}
+
+			caCertPath, caCertKey := path.Join(constant.CertRootDir, "ca.crt"), path.Join(constant.CertRootDir, "ca.key")
```
trawler, line 78: When the control plane isn't started yet, this will result in an error. I'd like to see a more meaningful error, same as in https://github.com/k0sproject/k0s/blob/main/cmd/token.go#L129-L132:
```suggestion
caCert, err := ioutil.ReadFile(path.Join(constant.CertRootDir, "ca.crt"))
if err != nil {
	return "", errors.Wrapf(err, "failed to read cluster ca certificate, is the control plane initialized on this node?")
}
caCertPath, caCertKey := path.Join(constant.CertRootDir, "ca.crt"), path.Join(constant.CertRootDir, "ca.key")
```

cmd/user.go
```diff
@@ -0,0 +1,122 @@
...
+- name: {{.User}}
+  user:
+	client-certificate-data: {{.ClientCert}}
+	client-key-data: {{.ClientKey}}
+`))
+
+	// userCmd creates new certs and kubeConfig for a user
+	userCmd = &cobra.Command{
+		Use:   "user",
+		Short: "Manage users",
```
trawler, line 53:
```suggestion
		Short: "Manage user access",
```

cmd/user.go
```diff
@@ -0,0 +1,122 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	userCreateCmd = &cobra.Command{
+		Use:   "create [username]",
+		Short: "Create a signed certificate for a user.",
```
trawler, line 61:
```suggestion
		Short: "Create a kubeconfig for a user.",
```

cmd/user.go
```diff
@@ -0,0 +1,122 @@
...
+		Long: `Create a signed certificate and public key for a given user (and optionally user groups). <br>
```
trawler, line 62: `<br` doesn't work  in Golang. Please fix the output:
```
ubuntu@ip-10-0-45-34:~/k0s_test/k0s (add-user-certs) [ 959a5c2 ] $ ./k0s user create --help
Create a signed certificate and public key for a given user (and optionally user groups). <br>
                *NOTE*: A certificate once signed cannot be revoked for a particular user.

```

cmd/user.go
```diff
@@ -0,0 +1,122 @@
...
+				ClientCert: base64.StdEncoding.EncodeToString([]byte(userCert.Cert)),
+				ClientKey:  base64.StdEncoding.EncodeToString([]byte(userCert.Key)),
+				User:       username,
+				JoinURL:    clusterConfig.Spec.API.ControllerJoinAddress(),
+			}
+
+			var buf bytes.Buffer
+
+			err = userKubeconfigTemplate.Execute(&buf, &data)
+			os.Stdout.Write(buf.Bytes())
```
trawler, line 115: > `os.Stdout.Write(buf.Bytes())`

This line needs to be after the err clause

trawler (changes requested):
Thank you for your contribution!
I have just a few remarks that need attention to. Other than that, looks good 👍

binoychitale:
Hi @trawler , thanks for the review. I have made the following changes -
- Implement changes requested in review
- Rebase onto main and use refactored APIs from the constants library

cmd/user.go
```diff
@@ -0,0 +1,125 @@
...
+    user: {{.User}}
+  name: k0s
+current-context: k0s
+kind: Config
+preferences: {}
+users:
+- name: {{.User}}
+  user:
+	client-certificate-data: {{.ClientCert}}
+	client-key-data: {{.ClientKey}}
```
trawler, line 47: When testing this template, kubectl fails, because the indentation is used generated using tabs instead of spaces:
```
error: error loading config file "/home/ubuntu/.kube/config": yaml: line 19: found character that cannot start any token
```
Please verify that the resulting Yaml file does not include tabs, since YAML forbids it.

cmd/user.go
```diff
@@ -0,0 +1,125 @@
...
+		Short: "Manage user access",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	userCreateCmd = &cobra.Command{
+		Use:   "create [username]",
+		Short: "Create a kubeconfig for a user.",
+		Long:  "Create a kubeconfig with a signed certificate and public key for a given user (and optionally user groups).\nNOTE: A certificate once signed cannot be revoked for a particular user.",
```
trawler, line 62: command line description should not end with a `.`

trawler (changes requested):
Since this is a new feature, it should be added to the docs, in a new section: 
## Adding a Cluster User
under: https://github.com/k0sproject/k0s/blob/main/docs/create-cluster.md

cmd/user.go
jasmingacic: Also I would add an example here too

trawler (approved):
lgtm 💯

PR #459, opened 2020-11-25
Fix finding path to the k0s binary when not in $PATH

ncopa:
Fixes the error when k0s is not in $PATH:
 Failed to start: fork/exec ./k0s: no such file or directory  component=k0s-control-api


trawler (approved):
lgtm

PR #462, opened 2020-11-26
Pin main README to v0.7.0

trawler:

**What this PR Includes**
This PR pins the linked docs in the main README to the latest release


ncopa:
I'm not sure this is the right way to deal with it. We'd end up need update this for every tag we create.

PR #471, opened 2020-11-30
Bump containerd to 1.4.2

ncopa:
Use go 1.15 to build containerd since that is what upstream containerd
uses.

See https://github.com/containerd/containerd/releases/tag/v1.4.2


ncopa:
It passed network test: https://github.com/k0sproject/k0s/actions/runs/391290961

jnummelin (approved):

PR #474, opened 2020-11-30
Fix Helm values parsing

### LINKED ISSUE #447: Helm extension not applying values
**Version**
```
v0.7.0
```
**Platform**
```
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.1 LTS
Release:        20.04
Codename:       focal
```
**What happened?**
First, the docs specify some yaml I'm unfamiliar with and was looking for an explanation (the `2` after the pipe character '|'' in the `values` definition).   After playing around with a yaml validator, the `2` seems related to consolidating the indentation when converting everything after the `|` into a single string.  Maybe this is more familiar to other people, but if it is obscure, a mention of what it means would be nice.

```
helm:
 repositories:
 - name: stable
   url: https://charts.helm.sh/stable
 - name: prometheus-community
   url: https://prometheus-community.github.io/helm-charts
 charts:
 - name: prometheus-stack
   chartname: prometheus-community/prometheus
   version: "11.16.8"
   values: |2
       <embed yaml>
   namespace: default
```

I've been trying out k0s with a project I'm working on and am not having luck getting the Helm chart quite right

1. I specify the namespace as something other than `default` but the resulting components end up in `default` namespace.
1. I specify values (that I would normally specify on the command line with --set), but the resulting deployment doesn't seem to reflect the values.

**Is there a way to see some debug output or a log of what Helm is doing?**  Something in a log file that is equivalent to passing the --debug argument to Helm would be **very** helpful.

**How To Reproduce**
Add this to k0s.yaml (I have tried with `values: |2`, `values: |`, and `values:` ... all yield the same results):

```yaml
extensions:
  helm:
    repositories:
    - name: akri
      url: https://deislabs.github.io/akri
    charts:
    - name: akri-dev
      chartname: akri/akri-dev
      version: "0.0.42"
      namespace: akri
      values: |2
        debugEcho.enabled: true
        debugEcho.shared: false
        agent.allowDebugEcho: true
```

I'm trying to do the equivalent of `helm install akri akri-helm-charts/akri-dev --version 0.0.42 --set debugEcho.enabled=true,debugEcho.shared=false,agent.allowDebugEcho=true`

**Expected behavior**
I expected to see the akri components show up in `akri` namespace rather than `default`  ( though, it is entirely possible that akri is messing up :) )

I also expected to see a few more components starting based on the values specified, which makes me wonder if the values are being communicated to Helm ( though, again, it is entirely possible that akri is messing up :) )

```sh
$ kubectl get pods -A
NAMESPACE     NAME                                          READY   STATUS             RESTARTS   AGE
default       akri-agent-daemonset-l84qp                    1/1     Running            10         2d20h
default       akri-controller-deployment-6949859f84-7gml9   1/1     Running            7          2d20h
```

**Screenshots & Logs**

**Additional context**
Following up on the idea that the values were not being applied to the Akri helm chart, I added a paramater that would result in a path being mounted in the resulting containers.

```yaml
extensions:
  helm:
    repositories:
    - name: akri
      url: https://deislabs.github.io/akri
    charts:
    - name: akri-dev
      chartname: akri/akri-dev
      version: "0.0.42"
      namespace: akri
      values: |2
        agent.host.dockerShimSock: /my/new/docker/shim/foo.sock
```

After deploying k0s, I can then test whether this new path was mounted as expected:
```sh
$ sudo crictl --runtime-endpoint unix:///run/k0s/containerd.sock inspect $(sudo crictl --runtime-endpoint unix:///run/k0s/containerd.sock ps | grep agent | awk '{print $1}') | grep hostPath | grep sock | grep /my/new/docker/shim/foo.sock && echo "Successfully passed Helm value" || echo "Failed to pass Helm value"
Failed to pass Helm value
```

It (from this contrived test) seems that the values are not being applied to the Helm chart.

jnummelin:

**Issue**
Fixes #447 

**What this PR Includes**
Intorduces a cleanup function to make all nested maps being converted into `map[string]interface{}` types. By default golang parses the nested maps into `map[interface{}]interface{}` types which makes Helm to actually ignore those keys when merging the chart default values into given values.

in the Helm code which merges the defaults in, there’s this:
```
if dest, ok := value.(map[string]interface{}); ok {
```
see: https://github.com/helm/helm/blob/6aefbdcf38e14998d0bcb24030061822b1e8888d/pkg/chartutil/coalesce.go#L156-L157

The cast is not successfull with `map[interface{}]interface{}` type and thus no merging actually happened

Adds also a verification into the related smoke case that we really see the given values being propagated into place.

### DIFF
--- a/pkg/apis/helm.k0sproject.io/v1beta1/generic_hash.go
+++ b/pkg/apis/helm.k0sproject.io/v1beta1/generic_hash.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2020 Mirantis, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package v1beta1
+
+import "fmt"
+
+// Cleans up a slice of interfaces into slice of actual values
+func cleanUpInterfaceArray(in []interface{}) []interface{} {
+	result := make([]interface{}, len(in))
+	for i, v := range in {
+		result[i] = cleanUpMapValue(v)
+	}
+	return result
+}
+
+// Cleans up the map keys to be strings
+func cleanUpInterfaceMap(in map[interface{}]interface{}) map[string]interface{} {
+	result := make(map[string]interface{})
+	for k, v := range in {
+		result[fmt.Sprintf("%v", k)] = cleanUpMapValue(v)
+	}
+	return result
+}
+
+// Cleans up the value in the map, recurses in case of arrays and maps
+func cleanUpMapValue(v interface{}) interface{} {
+	switch v := v.(type) {
+	case []interface{}:
+		return cleanUpInterfaceArray(v)
+	case map[interface{}]interface{}:
+		return cleanUpInterfaceMap(v)
+	case string:
+		return v
+	case int:
+		return v
+	case bool:
+		return v
+	case float64:
+		return v
+	default:
+		return fmt.Sprintf("%v", v)
+	}
+}
+
+// CleanUpGenericMap is a helper to "cleanup" generic yaml parsing where nested maps
+// are unmarshalled with type map[interface{}]interface{}
+func CleanUpGenericMap(in map[string]interface{}) map[string]interface{} {
+	result := make(map[string]interface{})
+	for k, v := range in {
+		result[fmt.Sprintf("%v", k)] = cleanUpMapValue(v)
+	}
+	return result
+}
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ bindata
 *.tfstate*
 aws_private.pem
 out.json
+.vscode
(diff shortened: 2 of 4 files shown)


jasmingacic (approved):

pkg/apis/helm.k0sproject.io/v1beta1/generic_hash.go
```diff
@@ -0,0 +1,51 @@
+package v1beta1
```
ncopa, line 1: New files need copyright boilerplate

ncopa (dismissed):
Congrats with a good catch!

pkg/apis/helm.k0sproject.io/v1beta1/generic_hash.go
jnummelin: @ncopa good catch. Added now, PTAL

jnummelin:
@ncopa @jasmingacic Had to force push and thus dismiss the reviews, could you PTAL again, thanks.

jasmingacic (approved):
I just tested this setup:
```    
charts:
    - name: akri-dev
      chartname: akri/akri-dev
      version: "0.0.42"
      namespace: akri
      values: |
        debugEcho:
          enabled: true
          shared: false
        agent:
          allowDebugEcho: true
          host:
            dockerShimSock: /my/new/docker/shim/foo.sock
```
Which resulted with:
```
# sudo crictl --runtime-endpoint unix:///var/lib/k0s/run/containerd.sock ps
CONTAINER           IMAGE               CREATED             STATE               NAME                         ATTEMPT             POD ID
2e50e5244dcc4       c0bd9c7f70f66       2 seconds ago       Running             akri-controller              2                   fe89ca98b8eba
ac6573b365fd9       37c2efaddb793       2 seconds ago       Running             akri-agent                   2   
```
And this:

```
# sudo crictl --runtime-endpoint unix:///var/lib/k0s/run/containerd.sock inspect ac6573b365fd9 | grep -i hostpath
         ...
        "hostPath": "/my/new/docker/shim/foo.sock",
```

Looking good

Issue #206, opened 2020-09-23
Add subcommand "k0s install <controller/worker>"


trawler:
we can add a helper script for deploying systemd startup scripts on the host os, to allow mke to start as a service and log to system journal.
(possibly also used as helper script for the upgrade process)
we can use https://godoc.org/github.com/ayufan/golang-kardianos-service for generating the service definitions (supports both windows & linux)

PR #478, opened 2020-12-01
Install cmd

trawler:
**Issue**
Part of #206

**What this PR Includes**
This PR adds the `install` command, that initializes the correct users for k0s server

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -1,0 +1,83 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"reflect"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/k0sproject/k0s/pkg/apis/v1beta1"
+	"github.com/k0sproject/k0s/pkg/install"
+)
+
+func init() {
+	installCmd.Flags().StringVar(&role, "role", "server", "node role (possible values: server or worker. In a single-node setup, a worker role should be used)")
+}
+
+var (
+	role string
+
+	installCmd = &cobra.Command{
+		Use:   "install",
+		Short: "Helper command for setting up k0s on a brand-new system. Must be run as root (or with sudo)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			switch role {
+			case "server", "worker":
+				return setup()
+			default:
+				logrus.Errorf("invalid value %s for install role", role)
+				return cmd.Help()
+			}
+		},
+	}
+)
+
+// the setup functions:
+// * Ensures that the proper users are created
+// * sets up startup and logging for k0s
+func setup() error {
+	if os.Geteuid() != 0 {
+		logrus.Fatal("this command must be run as root!")
+	}
+
+	if role == "server" {
+		if err := createServerUsers(); err != nil {
+			logrus.Errorf("failed to create server users: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func createServerUsers() error {
+	clusterConfig, err := ConfigFromYaml(cfgFile)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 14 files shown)


cmd/install.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return nil
+}
+
+func createServerUsers() error {
+	clusterConfig, err := ConfigFromYaml(cfgFile)
+	if err != nil {
+		return err
+	}
+	users := getUserList(*clusterConfig.Install.SystemUsers)
```
jasmingacic, line 60: If the config yaml file doesn't contain `installConfig:` the command will fail with:

```
# sudo ./k0s install --config cfg.yaml 
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1aef50d]

goroutine 1 [running]:
github.com/k0sproject/k0s/cmd.createServerUsers(0x6b, 0x0)
        /root/src/github.com/k0sproject/k0s/cmd/install.go:60 +0x6d
...
```
Maybe we should check if value of system users are present

cmd/user.go
```diff
@@ -49,10 +49,11 @@ users:
 
 	// userCmd creates new certs and kubeConfig for a user
 	userCmd = &cobra.Command{
-		Use:   "user",
+		Use:   "user [command]",
```
jasmingacic, line 52: I know it isn't related to this PR but the command name is a little misleading here. As the command is described as "Manage user access" but in fact it just creates a kubeconfig.

internal/util/users.go
```diff
@@ -37,3 +37,26 @@ func GetGID(name string) (int, error) {
...
+func GetLinuxDist() (string, error) {
+	if runtime.GOOS == "windows" {
+		return "", fmt.Errorf("unsupported OS")
+	}
+	cfg, err := ini.Load("/etc/os-release")
+	if err != nil {
+		fmt.Printf("failed to read file: %v", err)
+	}
+	return cfg.Section("").Key("ID").String(), nil
+}*/
```
jasmingacic, line 62: maybe we could remove this

pkg/install/users.go
```diff
@@ -0,0 +1,87 @@
...
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/k0sproject/k0s/internal/util"
+)
+
+// EnsureUser checks if a user exists, and creates it, if it doesn't
+// TODO: we should also consider modifying the user, if the user exists, but with wrong settings
+func EnsureUser(name string, homeDir string) error {
```
jasmingacic, line 17: The function does what it is supposed to do:
```
# cat /etc/passwd | grep k0s
kube-controller-manager:x:997:997::/var/lib/k0s:/usr/sbin/nologin
etcd:x:996:996::/var/lib/k0s:/usr/sbin/nologin
kube-apiserver:x:995:995::/var/lib/k0s:/usr/sbin/nologin
konnectivity-server:x:994:994::/var/lib/k0s:/usr/sbin/nologin
kube-scheduler:x:993:993::/var/lib/k0s:/usr/sbin/nologin
```

cmd/install.go
trawler: Resolved  with https://github.com/k0sproject/k0s/blob/a93e731e3652f53123cd224c01064dd78b7f811c/cmd/helpers.go#L47-L49

internal/util/users.go
trawler: This is a stub that I thought would be useful for future iterations, and I didn't want to remove it :)

cmd/install.go
jasmingacic: yes it is resolved

jasmingacic (approved):

cmd/user.go
trawler: I believe this stems from the original issue: #389 and perhaps should be better clarified in a future PR.

PR #495, opened 2020-12-05
Custom volume plugin dir

### LINKED ISSUE #494: k0s does not work on Fedora CoreOS
**Version**
```
v0.8.1-git
```
**Platform**
```
Fedora release 33 (Thirty Three)
```
**What happened?**
k0s fails because `/usr` is read only.

**How To Reproduce**
Run k0s on any system where `/usr` is read only.

**Expected behavior**
k0s should be configurable to avoid using `/usr` on systems where it cannot be read.

Currently I've found this to apply to Calico and it's `flexvol-driver-host` and the default `volumePluginDir`.

**Additional context**
This applies to all(?) systems using rpm-ostree for package managment. I've only been able to try it on Fedora IoT, Fedora Silverblue and Fedora CoreOS though, so it might not apply for all systems with rpm-ostree.

llndqvst:
---
name: Pull Request
about: Create a Pull Request
title: 'Custom volume plugin dir'
labels: ''
assignees: ''

---

**Issue**
Fixes #494 

**What this PR Includes**
This PR includes new configuration options to allow the user setting a custom volumePluginDir. Sample configs are also included for getting k0s to work on Fedora CoreOS.


docs/configuration.md
```diff
@@ -89,6 +89,7 @@ spec:
       vxlanVNI: 4096
       mtu: 1450
       wireguard: false
+      flexVolumeDriverPath: /usr/libexec/kubernetes/kubelet-plugins/volume/exec/nodeagent~uds
```
jnummelin, line 92: I think we should have a explanatory title/item below for this too.

pkg/component/worker/kubelet.go
```diff
@@ -82,7 +77,6 @@ func (k *Kubelet) Run() error {
 
 	args := []string{
 		fmt.Sprintf("--root-dir=%s", k.dataDir),
-		fmt.Sprintf("--volume-plugin-dir=%s", k.K0sVars.KubeletVolumePluginDir),
```
jnummelin, line 85: Unsetting this will now make the default location to change :/ Could we overcome this by setting the default worker profile to have the current default path?

jnummelin (changes requested):
The changing path to kubelet volume plugin dir needs IMO some attention

llndqvst:
I have applied the requested changes. Now the default location should be properly set inside the default kubelet configuration for workers.

However, I want to note that this change includes a slight regression from main. Currently k0s checks if the path supplied to `--volume-plugin-dir` is writeable on the worker nodes and exits if it's not. With my changes that is not possible, and instead k0s will fail silently. Only noticeable to the user as Calico getting stuck on init.

I see two possible fixes for this:

- Change the configuration file to include a specific field for cluster-wide kubelet configuration with manually mapped values. This could look something like this:

```
kubelet:
  volumePluginDir: /var/libexec...
```

- Parse the applied kubeletConfiguration for values that need to be tested by k0s before starting the kubelet.

jnummelin:
I think the regression from main is acceptable given that customizing the plugin dir should not be needed in most cases.

jnummelin (approved):

PR #500, opened 2020-12-10
install command: service setup

### LINKED ISSUE #206: Add subcommand "k0s install <controller/worker>"

### LINKED ISSUE #473: Automation-friendly cluster setup
As a cluster operator, I would like to create a cluster using infrastructure as code. The requirement to use the encoded kubeconfig join token makes this more challenging than I would like. My current solution works something like the following.

1. Create base VM image with k0s included.
2. Using terraform, create controller and worker instances.
3. Using ansible, set up and start a systemd service running k0s on the primary master.
3. Using ansible, create a controller and worker tokens on the primary controller.
4. Using ansible, copy the controller join token to each secondary controller, set up and start a systemd service running k0s.
5. Using ansible, copy the worker join token to each worker, set up and start a systemd service running k0s.

I would like to have a secondary join flow which allows me to join a worker or a controller to the primary controller by providing a pre-shared key together with the address of the primary controller. This would allow a solution something like the following.

1. Create base VM with k0s included.
2. Using terraform, create the primary controller; provide the k0s configuration (including pre-shared join token) and the systemd unit file as write_file directives in the cloud-init configuration.
3-4. Repeat 2, except for secondary controllers and workers and providing the IP address of the primary controller in addition to the join token.

Any join flow which allows cloud-init to be used rather than requiring direct provisioning steps will greatly simplify deployment on all modern cloud and virtualization platforms.

trawler:
**Issue**
Fixes #206
Fixes #473

**What this PR Includes**
This PR allows a user to run `cmd install` which automatically creates startup scripts on systemd and OpenRC systems, for k0s server or k0s worker.


go.mod
```diff
@@ -60,4 +62,5 @@ replace (
 	github.com/docker/distribution => github.com/docker/distribution v0.0.0-20191216044856-a8371794149d
 	github.com/docker/docker => github.com/moby/moby v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
 	github.com/docker/libnetwork => github.com/moby/libnetwork v0.8.0-dev.2.0.20201031180254-535ef365dc1d
+// github.com/k0sproject/kardianos-service => /home/ubuntu/go/src/github.com/k0sproject/kardianos-service
```
jnummelin, line 65: looks like a leftover from local testing

jnummelin (approved):
LGTM

@ncopa you're super familiar with init systems, maybe also take a look

ncopa (approved):
LGTM

jasmingacic (dismissed):
I tested it. It works as described.

PR #508, opened 2020-12-14
Windows support

### LINKED ISSUE #36: Support windows workers

soider:
---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #36 

**What this PR Includes**
Support for windows nodes is introduced here

To join windows node, the cluster must have at least single linux node (for example, server --enable-worker)

The command example 

`.\k0s.exe worker --api-server=https://172.31.2.164:9443  --cri-socket=docker:npipe:////./pipe/docker_engine  <token>`


pkg/constant/constant_windows.go
```diff
@@ -0,0 +1,120 @@
+// +build windows
```
soider, line 1: todo: clean up duplicates here

pkg/component/worker/kubelet.go
```diff
@@ -169,3 +200,22 @@ func splitRuntimeConfig(rtConfig string) (string, string, error) {
 
 	return runtimeType, runtimeSocket, nil
 }
+
+const awsMetaInformationURI = "http://169.254.169.254/latest/meta-data/local-hostname"
+
+func getNodeName() (string, error) {
+	req, err := http.NewRequest("GET", awsMetaInformationURI, nil)
```
jnummelin, line 207: This ties the implementation now 100% to AWS.

pkg/component/worker/kubeproxy_windows.go
```diff
@@ -0,0 +1,61 @@
...
+	supervisor supervisor.Supervisor
+}
+
+// Init
+func (k KubeProxy) Init() error {
+	return assets.Stage(k.K0sVars.BinDir, "kube-proxy.exe", constant.BinDirMode)
+}
+
+func (k KubeProxy) Run() error {
+	node, err := getNodeName()
```
jnummelin, line 24: Do we need to always override the hostname? If not, I'd propose to do e.g. an env var to override it and thus "externalize" the problem out of k0s

cmd/api.go
```diff
@@ -146,6 +153,70 @@ func etcdHandler() http.Handler {
...
+				continue
+			}
+			found = true
+			secretWithToken = secret
+			break
+		}
+		if !found {
+			sendError(fmt.Errorf("no calico-node-token secret found"), resp)
+			return
+		}
```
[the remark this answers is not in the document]
soider: We don't really have a proper hame, only prefix

pkg/component/worker/kubelet.go
soider: No, it has fallback to the actual os.Hostname.
The thing is that we need to have internal aws hostname if we run in ec2 and as far as I understand it is the only way to get it and check if we in the ec2.

cmd/worker.go
```diff
@@ -39,6 +40,7 @@ import (
 func init() {
 	workerCmd.Flags().StringVar(&workerProfile, "profile", "default", "worker profile to use on the node")
 	workerCmd.Flags().StringVar(&criSocket, "cri-socket", "", "contrainer runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]")
+	workerCmd.Flags().StringVar(&apiServer, "api-server", "", "HACK: api-server for the windows worker node")
```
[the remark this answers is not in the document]
soider: yeah, we have separate issue to get rid of this (and two other "HACK:" arguments) #516

pkg/component/worker/kubeproxy_windows.go
soider: We need to override it for EC2 instances and I think it's better to have in the code, because it could not that user friendly to understand which EC2 hostname to use, because calico docs states that we need to use internal hostname always, so that could led user to have some confusion.

jnummelin (approved):
LGTM 
As we have separate issues to track the known immediate next steps let's merge this 

🎉

BahramRadan (approved):

jnummelin (approved):
LGTM

This has now been proven to work on both AWS and Equinix Metal infra.

We've got many enhancements known already and tracked in https://github.com/k0sproject/k0s/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%2Fwindows

Slach:
@soider @jnummelin 
please like https://github.com/ScoopInstaller/Main/pull/1702
I just add k0s installation via `scoop`

PR #527, opened 2020-12-15
Config option for external api address

### LINKED ISSUE #455: Question: Is it possible to put the control plane behind a load-balancer?
I've only done limited testing, so please forgive me if I've overlooked something that should have been obvious.

In our current configuration (not k0s just upstream k8s), our k8s masters are behind a load-balancer. New nodes communicate with the load-balancer to join the cluster and masters can come and go. I'd like to have a similar configuration with k0s, but I've hit a problem.

It seems like the token generation code is tied to the value of the API address in the k0s configuration file. I'm running in AWS and tried to use the DNS name of an elastic load-balancer. However, it complained that it wasn't an IP. If I don't set it, it gets the IP of the host and new nodes must communicate with a specific master to join. Is that correct? If so, do you have suggestions on automating adding nodes that that doesn't assume the initial control plane node is always available? We rebuild all of our servers fairly frequently.

jnummelin:

**Issue**
Fixes #455 

**What this PR Includes**

This PR introduces new options to configure external API address. This is needed in cases where there's a LB in front of the controller nodes.

So the `api.externalAddress` value is being now used in few cases:
- the server address in join tokens
- the API address when configuring cluster components that need to talk with the API (konnectivity agents, kube-proxy etc.)
- new reconciler that resolves the `externalAddress` into a IP addresses and pushes them into the `kubernetes` service endpoints.

### Example use case

So say I configure my cluster with AWS ELB in front of controllers. The ELB has address `my-k0s-lb.us-west-2.elb.amazonaws.com` and it resolves to addresses:
- `1.2.3.4`
- `5.6.7.8`

This would essentially mean the following:
- Join token has server address `https://my-k0s-lb.us-west-2.elb.amazonaws.com:6443`
- All components connecting DIRECTLY to API will connect to the same address as above
- automatic `endpoints/kubernetes` addresses will be disabled and k0s controller will resolve it to have `1.2.3.4:6443` and `5.6.7.8:6443` addresses. This essentially means all in-cluster components connecting through the `kunernetes` service will also connect through the LB

## TODO

- [x] add the external address to SANs list automatically

### DIFF
--- a/pkg/component/server/apiendpointreconciler.go
+++ b/pkg/component/server/apiendpointreconciler.go
@@ -102,11 +102,13 @@ func (a *APIEndpointReconciler) Healthy() error { return nil }
 func (a *APIEndpointReconciler) reconcileEndpoints() error {
 	if !a.leaderElector.IsLeader() {
 		a.L.Debug("we're not the leader, not reconciling api endpoints")
+		return nil
 	}
 
 	ips, err := net.LookupIP(a.ClusterConfig.Spec.API.ExternalAddress)
 	if err != nil {
 		a.L.Errorf("cannot resolve api.externalAddress: %s", err.Error())
+		return err
 	}
 	// Sort the addresses so we can more easily tell if we need to update the endpoints or not
 	ipStrings := make([]string, len(ips))
--- a/pkg/component/server/apiendpointreconciler.go
+++ b/pkg/component/server/apiendpointreconciler.go
@@ -1,0 +1,221 @@
+/*
+Copyright 2020 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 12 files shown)


trawler (approved):

jasmingacic (approved):

jnummelin:
There's still one TODO item left in my list on this, let's not merge yet

jnummelin:
Added the automation to get  `api.externalAddress` into SANs. @trawler @jasmingacic  PTAL again

pkg/component/server/apiendpointreconciler.go
```diff
@@ -0,0 +1,221 @@
...
+	close(a.stopCh)
+	return nil
+}
+
+// Healthy dummy implementation
+func (a *APIEndpointReconciler) Healthy() error { return nil }
+
+func (a *APIEndpointReconciler) reconcileEndpoints() error {
+	if !a.leaderElector.IsLeader() {
+		a.L.Debug("we're not the leader, not reconciling api endpoints")
```
soider, line 104: I think return should be added here, to end function if we are not the leader?

soider (changes requested):

pkg/component/server/apiendpointreconciler.go
jnummelin: oh, very good catch 👍

jnummelin: fixed

jnummelin:
@soider Fixed the change request, PTAL again.

soider (approved):

PR #543, opened 2020-12-17
update k8s artifacts to 1.20.0

### LINKED ISSUE #542: Wrong version for kubelet configmap
**What happened?**
We updated to kube 1.20 in #498  but internal kubeconfig configmap still is named with 1.19 version

So the issue is that both controller and worker use "wrong" configmap. Although they are in sync so nothing is currently actually broken. :)

**Expected behavior**
k0s controller should create new 1.20 related configmap for kubelet.

trawler:
**Issue**
Fixes #542

**What this PR Includes**
Update a few more constants to k8s v1.20.0
This PR also:
* Updates sonobuoy to version 0.20
* Adds the `kube-conformance-image-version` flag to sonobuoy conformance and e2e testing
* Updates conformance test docs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM alpine:3.12
 
 RUN apk add --no-cache bash coreutils findutils iptables curl
 
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.19.0/bin/linux/amd64/kubectl \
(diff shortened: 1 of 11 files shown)


ncopa:
whoa... the string `1.19` was hardcoded in many places. would be nice to reduce that (but not in this PR)

ncopa (approved):

PR #551, opened 2020-12-17
[WIP] added more information in 'k0s install; command

jasmingacic:
This is just a placeholder for cmd install improvement. 

this is depending on https://github.com/kardianos/service/pull/255 getting merged
The PR is required so I can easily integrate k0s into LP. The vital piece of information is ability to edit k0s.service for worker so the join token can be injected.

### DIFF
--- a/pkg/install/service.go
+++ b/pkg/install/service.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	"github.com/kardianos/service"
-	"github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/install/service.go
```diff
@@ -65,7 +65,10 @@ func EnsureService(args []string) error {
 	svcConfig.Dependencies = deps
 	svcConfig.Arguments = args
 
-	logrus.Info("Installing k0s service")
+	fmt.Println("Installing k0s service")
+	fmt.Println("Service information:", s.Platform())
```
jasmingacic, line 69: I intentionally put fmt to print out these values since I was not able to remove time stamps from logrus.Infof

jasmingacic:
I will close this for now since the upstream PR (kardianos/service#255) hasn't been merged yet.

Issue #537, opened 2020-12-16
Pod cannot communicate between adm64 and arm64 workers


lucj:
**Version**
```
$ k0s version
v0.8.1
```
**Platform**
Which platform did you run k0s on?
- master runs on Ubuntu 20.04
- worker 1 runs on Ubuntu 18.04
- worker 2 runs on RPI 4

**What happened?**

When deploying several Pods in the cluster, it appears that Pods running on worker1 cannot communicate with Pods running on worker2. That is totally fine if both workers runs on Ubuntu though.

**Screenshots & Logs**

```
$ export KUBECONFIG=/var/lib/k0s/pki/admin.conf
$ kubectl get no
NAME      STATUS   ROLES    AGE    VERSION
worker-1  Ready    <none>   1h23m  v1.19.4       # -> Ubuntu
worker-2  Ready    <node>   4m     v1.19.4         # -> RPI 4
```

Pods running in the cluster:
```
$ kubectl get po -o custom-columns=NAME:.metadata.name,NODE:.spec.nodeName,IP:.status.podIP
NAME           NODE       IP
whoami-5b5lx   worker-1   10.244.0.75
whoami-8lxhp   worker-2   10.244.246.2
```

Running a test Pod
```
$ kubectl run test -ti --image=alpine -- sh
```

Communication from test Pod to Pod running on worker-1 is fine.

```
/ # curl 10.244.0.75:8080
I'm whoami-5b5lx running on linux/amd64
...
```

Communication from test Pod to Pod running on worker-2 does not work.

```
/ # curl 10.244.246.2:8080
... # Hanging
```

In this example, it happens that the Pod is running on worker-1. If it was running on worker-2 that would be the same thing, it could only communicate with the Pod running on the same machine it is running on.

jnummelin:
@lucj What OS and version is your RPI4 running?

> That is totally fine if both workers runs on Ubuntu though.

Based on that I'll assume the networking on RPI4 node is busted.

First thing to always verify is that there's no firewall blocking the traffic. By default calico is running in vxlan mode so check that there's nothing blocking the vxlan UDP traffic on port 4789. One way to test the connectivity is with nmap:
```
nmap 172.17.0.5 -p 4789 -sU
```
Where the `172.17.0.5` is the nodes IP.

Check the logs of calico pods if they have any errors or other clues.

[...]

PR #562, opened 2020-12-21
Always set calico to auto detect iptables backend mode

jnummelin:

**Issue**
Might be a partial fix for #537 but even if not, making Calico autodetect the used mode is a good idea.

**What this PR Includes**
This PR sets Calico to always use auto-detect mode for iptables backend.

✅  SIG-Network tests succesful at https://github.com/k0sproject/k0s/actions/runs/435568064


soider (approved):
lgtm

trawler (approved):

PR #567, opened 2020-12-21
Refactor makefile

ncopa:
**What this PR Includes**

Refactor Makefile:
- dedupe build rules for windows and linux
- fix race condition that caused arm64 build to fail
- add `make smoketests` target so we can run all tests in one go


soider (approved):
Good stuff

PR #575, opened 2020-12-22
Add KillMode=process to systemd unit

jnummelin:

**Issue**
The current default way for systemd is to kill ALL processes in the services group. In k0s case it means all the managed processes AND al the containers too.

**What this PR Includes**
This changes the systemd unit to send kill signals only to the main process (k0s itself) so it'll allow k0s to shutdown other components gracefully.


jasmingacic (approved):
This is what we talked about yesterday. Cool

trawler (approved):

PR #592, opened 2020-12-30
Accept cri socket param in k0s server subcommand

### LINKED ISSUE #579: Allow k0s server to run with a BYO CRI Engine
`k0s server` does not support the `--cri-socket` flag, and so cannot run with a runtime other than containerD

Add the above mentioned flag to server as well (for single-node setup)

**Additional context**
https://github.com/k0sproject/k0s/issues/523#issuecomment-750382756

binoychitale:
Signed-off-by: binoychitale <binoychitale@gmail.com>

---
name: Accept cri socket param in k0s server subcommand
about: implements requirement to support the `--cri-socket` flag for `k0s server` subcommand when running an all-in-one cluster
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #579

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -429,7 +429,7 @@ func enableServerWorker(clusterConfig *config.ClusterConfig, k0sVars constant.Cf
 		}
 		componentManager.Add(containerd)
 	}
-	
+
 	if err := kubelet.Init(); err != nil {
 		logrus.Errorf("failed to init kubelet: %s", err)
 	}
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -420,22 +420,24 @@ func enableServerWorker(clusterConfig *config.ClusterConfig, k0sVars constant.Cf
 		K0sVars:             k0sVars,
 	}
 
-	if err := containerd.Init(); err != nil {
-		logrus.Errorf("failed to init containerd: %s", err)
+	if criSocket == "" {
+		if err := containerd.Init(); err != nil {
+			logrus.Errorf("failed to init containerd: %s", err)
+		}
+		if err := containerd.Run(); err != nil {
+			logrus.Errorf("failed to run containerd: %s", err)
+		}
+		componentManager.Add(containerd)
 	}
+	
 	if err := kubelet.Init(); err != nil {
 		logrus.Errorf("failed to init kubelet: %s", err)
 	}
-	if err := containerd.Run(); err != nil {
-		logrus.Errorf("failed to run containerd: %s", err)
-	}
+
 	if err := kubelet.Run(); err != nil {
 		logrus.Errorf("failed to run kubelet: %s", err)
 	}
 
-	if criSocket == "" {
-		componentManager.Add(containerd)
-	}
 	componentManager.Add(kubelet)
 
 	return nil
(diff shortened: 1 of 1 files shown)


jasmingacic (commented):
Code changed don't seem to be extensive. But we will need to test this

usrbinkat:
Neat @binoychitale , I scraped a test together but cant root cause for another week or so. Here is the result of my quick test.

### Test:
Built binary off this PR and tested single node behavior with systemd exec string:
```
ExecStart=/usr/local/bin/k0s server -c /root/.k0s/k0s.yaml --enable-worker --cri-socket remote:unix:///run/crio/crio.sock
```

### Result: pod scheduling failed
  - Kube API Resources: http://sprunge.us/5ziJQb?bash
  - Events: http://sprunge.us/lCRLgI?bash
  - Logs: http://sprunge.us/RdN6yq?bash
 
### Observations:
Starting on line 922 of the [logs](http://sprunge.us/RdN6yq?bash) we see the binary still trying to startup containerd.
On line 946 see args confirm crio.sock is requested
Starting on line 1805 crio+cgroup manager (systemd) errors begin to log failure output resulting in pod startup failure eg:
```
RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = cri-o configured with systemd cgroup manager, but did not receive slice as parent
```
  
### Env:
  Ubuntu 20.10
  Kernel 5.8.0
  Golang 1.15.6

binoychitale:
Hi @usrbinkat 

> Starting on line 922 of the logs we see the binary still trying to startup containerd.

There was a bug in the PR which caused this. Have added a commit which should take care of this

> Starting on line 1805 crio+cgroup manager (systemd) errors begin to log failure output resulting in pod startup failure eg:

I don't think I can be of much help here since I'm new to the project, however I did observe one thing -
https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-control-plane-node
according to these docs, the `cgroupDriver` is set to `cgroupfs` by default, which would have to be changed to `systemd` if not already done.

Hope this helps. If not, the core team would be better placed to help debug this.

usrbinkat:
Okay, then to use k0s+crio, i need to make sure I understand correctly, should I be re-configuring the host crio to utilize `cgroupfs` instead? or should I be flipping k0s configuration somewhere so that k0s adopts systemd as the cgroupDriver?

binoychitale:
I don't think there's a "right" way, but I found the following in the [docs](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers)

>A single cgroup manager simplifies the view of what resources are being allocated and will by default have a more consistent view of the available and in-use resources. When there are two cgroup managers on a system, you end up with two views of those resources. In the field, people have reported cases where nodes that are configured to use cgroupfs for the kubelet and Docker, but systemd for the rest of the processes, become unstable under resource pressure.

This would mean that if other services running on the node are managed by systemd, using systemd throughout seems to be the way to go.

I did some digging, and it seems that to change the cgroup manager for kubelet in k0s, setting the `cgroupDriver` in the kubelet config file (defaults to `/var/lib/k0s/kubelet-config.yaml`) for k0s might do the trick.

Also tagging @jnummelin / @jasmingacic  from the core team. You guys can correct me if I'm wrong here.

jnummelin:
> Should I be re-configuring the host crio to utilize cgroupfs instead? or should I be flipping k0s configuration somewhere so that k0s adopts systemd as the cgroupDriver?

@usrbinkat you can do it either way. :) If you wish to do it k0s side, you can use custom worker profiles to configure non-default settings for kubelet: https://docs.k0sproject.io/main/configuration/#specworkerprofiles

So say you create a profile called `crio` you can start the server with worker with: `k0s server -c /root/.k0s/k0s.yaml --enable-worker --cri-socket remote:unix:///run/crio/crio.sock --profile crio`

jnummelin (approved):

Issue #593, opened 2020-12-31
Install command help has misleading text for --role


jakolehm:
**Version**
```
$ k0s version
v0.9.1
```
**What happened?**
`k0s install --help` says
> --role string   node role (possible values: server or worker. In a single-node setup, a worker role should be used) (default "server")

which is misleading (or wrong) because docs say:

> enable-worker (single-node) setup is not supported.

jnummelin:
IMO we need to come up a way to make the install command to take in any supported flags and args for the server/worker commands. This way we do not need to copy all flags to install cmd.

By making the install command take in anything supported we can make it easily "consumable" for the `get.k0s.sh` script too. So something like:
```
curl ... | K0S_INSTALL_EXEC="worker --token-file /my/token"
```
would run the installer script and part of that run the `k0s install worker --token-file /my/token` to setup the systemd/openrc unit

trawler:
Hey, jumping in a bit late, but if I recall correctly, we were discussing the whole purpose of the “install” command with @ncopa way back, and wanted to have the install command support users for setting up the service in the officially recommended way. I.e, not run all services as root, etc. 
Given this assumption, I supposed that the single node setup isn’t in fact a “recommended” method. By supporting it in the install command we are actually making it easier to run it that way. Is that a route we want to take?
I agree that the help output should be clarified, but I am not yet convinced we should modify the whole behavior of the install command.

jasmingacic:
By merging #594 I believe this shouldn't be an issue anymore.
@jakolehm can you confirm this so we can close this issue?

jasmingacic:
Fixed #594

PR #594, opened 2020-12-31
Install improvement

jasmingacic:
This improvement allows to pass all possible k0s parameters to install command. 
Syntax is like this:

```
# ./k0s install --role server --enable-worker
INFO[2020-12-31 13:12:53] Installing k0s service                       
# cat /etc/systemd/system/k0s.service 
[Unit]
Description=k0s - Zero Friction Kubernetes
ConditionFileIsExecutable=/root/go/src/github.com/k0sproject/k0s/k0s

After=network.target 

[Service]
StartLimitInterval=5
StartLimitBurst=10
ExecStart=/root/go/src/github.com/k0sproject/k0s/k0s "server" "--config" "/root/go/src/github.com/k0sproject/k0s/k0s.yaml" "--enable-worker"
...
```

I believe this will clear up the questions related to #593  or similar

p.s. as @jnummelin mentioned maybe it would be worth considering adding separate install commands for server and worker so it would be easier to understand.

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -21,44 +21,28 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
-
 	"github.com/k0sproject/k0s/pkg/apis/v1beta1"
 	"github.com/k0sproject/k0s/pkg/install"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 )
 
 func init() {
-	installCmd.Flags().StringVar(&role, "role", "server", "node role (possible values: server or worker. In a single-node setup, a worker role should be used)")
-
-	// shell completion options
-	_ = installCmd.RegisterFlagCompletionFunc("role", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		return []string{"server", "worker"}, cobra.ShellCompDirectiveDefault
-	})
+	installCmd.AddCommand(installServerCmd)
+	installCmd.AddCommand(installWorkerCmd)
 }
 
 var (
-	role string
-
 	installCmd = &cobra.Command{
 		Use:   "install",
 		Short: "Helper command for setting up k0s on a brand-new system. Must be run as root (or with sudo)",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			switch role {
-			case "server", "worker":
-				return setup()
-			default:
-				logrus.Errorf("invalid value %s for install role", role)
-				return cmd.Help()
-			}
-		},
 	}
 )
 
 // the setup functions:
 // * Ensures that the proper users are created
 // * sets up startup and logging for k0s
-func setup() error {
+func setup(role string, args []string) error {
 	if os.Geteuid() != 0 {
 		logrus.Fatal("this command must be run as root!")
 	}
@@ -68,9 +52,8 @@ func setup() error {
 			logrus.Errorf("failed to create server users: %v", err)
 		}
 	}
-	// set-up service and logging
-	serviceArgs := getCmdArgs(role)
-	err := install.EnsureService(serviceArgs)
+
+	err := install.EnsureService(args)
 	if err != nil {
 		logrus.Errorf("failed to install k0s service: %v", err)
 	}
@@ -82,6 +65,7 @@ func createServerUsers() error {
 	if err != nil {
 		return err
 	}
+
 	users := getUserList(*clusterConfig.Install.SystemUsers)
 
 	var messages []string
@@ -105,22 +89,4 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
 		values[i] = v.Field(i).String()
 	}
 	return values
-}
-
-func getCmdArgs(role string) []string {
-	var args []string
-
-	if role == "server" {
-		args = append(args, "server")
-		if cfgFile != "" {
-			args = append(args, "--config", cfgFile)
-		}
-	} else {
-		args = append(args, "worker", "--token-file", "REPLACEME")
-	}
-
-	if debug {
-		args = append(args, "--debug")
-	}
-	return args
 }
--- a/cmd/installWorker.go
+++ b/cmd/installWorker.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2021 Mirantis, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+var (
+	installWorkerCmd = &cobra.Command{
+		Use:   "worker",
+		Short: "Helper command for setting up k0s as a worker node on a brand-new system. Must be run as root (or with sudo)",
+		Example: `Worker subcommand allows you to pass in all available worker parameters. 
+All default values of worker command will be passed to the service stub unless overriden.
+To get information about default values run "k0s install worker --help"
+
+Windows flags like "--api-server", "--cidr-range" and "--cluster-dns" will be ignored since install command doesn't yet support Windows services`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			flagsAndVals := []string{"worker"}
+			cmd.Flags().Visit(func(f *pflag.Flag) {
+				flagsAndVals = append(flagsAndVals, fmt.Sprintf("--%s", f.Name), f.Value.String())
+			})
+
+			return setup("worker", flagsAndVals)
+		},
+	}
+)
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/avast/retry-go v2.6.0+incompatible
 	github.com/cloudflare/cfssl v1.4.1
 	github.com/containerd/containerd v1.4.1 // indirect
-	github.com/davecgh/go-spew v1.1.1
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/docker/libnetwork v0.5.6
 	github.com/evanphx/json-patch v4.9.0+incompatible
@@ -34,8 +33,11 @@ require (
 	github.com/segmentio/backo-go v0.0.0-20200129164019-23eae7c10bd3 // indirect
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/cobra v1.1.1
+	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.6.1
+	github.com/vishvananda/netlink v1.1.0 // indirect
+	github.com/vishvananda/netns v0.0.0-20210104183010-2eb08e3e575f // indirect
 	github.com/weaveworks/footloose v0.0.0-20200609124411-8f3df89ea188
 	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
 	go.etcd.io/etcd v0.5.0-alpha.5.0.20200819165624-17cef6e3e9d5
@@ -45,7 +47,6 @@ require (
 	golang.org/x/tools v0.0.0-20201013201025-64a9e34f3752 // indirect
 	gopkg.in/check.v1 v1.0.0-20200227125254-8fa46927fb4f // indirect
 	gopkg.in/fsnotify.v1 v1.4.7
-	gopkg.in/ini.v1 v1.51.0
 	gopkg.in/segmentio/analytics-go.v3 v3.1.0
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 // indirect
--- a/go.sum
+++ b/go.sum
@@ -207,8 +207,6 @@ github.com/docker/go-metrics v0.0.0-20180209012529-399ea8c73916/go.mod h1:/u0gXw
 github.com/docker/go-units v0.3.3/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/docker/go-units v0.4.0 h1:3uh0PgVws3nIA0Q+MwDC8yjEPf9zjRfZZWXZYDct3Tw=
 github.com/docker/go-units v0.4.0/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
-github.com/docker/libnetwork v0.5.6 h1:hnGiypBsZR6PW1I8lqaBHh06U6LCJbI3IhOvfsZiymY=
-github.com/docker/libnetwork v0.5.6/go.mod h1:93m0aTqz6z+g32wla4l4WxTrdtvBRmVzYRkYvasA5Z8=
 github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1 h1:ZClxb8laGDf5arXfYcAtECDFgAgHklGI8CxgjHnXKJ4=
 github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1/go.mod h1:cyGadeNEkKy96OOhEzfZl+yxihPEzKnqJwvfuSUqbZE=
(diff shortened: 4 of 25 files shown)


cmd/install.go
```diff
@@ -108,19 +108,57 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+		//server params
+		if serverWorkerProfile != "default" {
+			params["profile"] = serverWorkerProfile
+		}
+
+		if enableWorker {
+			params["enable-worker"] = ""
+		}
+
+		params["token-file"] = serverToken
```
kke, line 129: I don't see where this `serverToken` comes from

cmd/install.go
```diff
@@ -108,19 +108,57 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+
 		if cfgFile != "" {
 			args = append(args, "--config", cfgFile)
 		}
 	} else {
-		args = append(args, "worker", "--token-file", "REPLACEME")
+		args = append(args, "worker")
+
+		//worker params
+		if workerProfile != "default" {
```
kke, line 138: Same with this, where does `workerProfile` come from?

cmd/install.go
```diff
@@ -108,19 +108,57 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+			params["profile"] = workerProfile
```
kke, line 139: Same with this, where does `workerProfile` come from?

cmd/install.go
jasmingacic: it comes from https://github.com/k0sproject/k0s/blob/main/cmd/server.go#L59

cmd/install.go
```diff
@@ -108,19 +108,57 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+		}
+
+		if tokenFile != "" {
+			params["token-file"] = tokenFile
+		} else {
+			params["token-file"] = "REPLACEME"
+
+		}
+		params["cri-socket"] = criSocket
```
kke, line 148: Same with these, where are `workerProfile`, `tokenFile` and `criSocket` coming from?

cmd/install.go
kke: How does it appear in `cmd/install.go`'s `getCmdArgs` function scope?

cmd/install.go
jasmingacic: https://github.com/k0sproject/k0s/blob/main/cmd/worker.go#L42

cmd/install.go
kke: image

jasmingacic:
The most recent changes are:
```
# ./k0s install -h
Helper command for setting up k0s on a brand-new system. Must be run as root (or with sudo)
Usage:
  k0s install [command]
Available Commands:
  server      Helper command for setting up k0s as server node on a brand-new system. Must be run as root (or with sudo)
  worker      Helper command for setting up k0s as a worker node on a brand-new system. Must be run as root (or with sudo)
```
Server subcommand:
```
# ./k0s install server -h
Helper command for setting up k0s as server node on a brand-new system. Must be run as root (or with sudo)

Usage:
  k0s install server [flags]

Examples:
All default values of server command will be passed to the service stub unless overriden. 
To get information about default values run "k0s install server --help"

With server subcommand you can setup a single node cluster by running:

        k0s install server --enable-worker

```
Worker subcommand:
```
# ./k0s install worker -h
Helper command for setting up k0s as a worker node on a brand-new system. Must be run as root (or with sudo)

Usage:
  k0s install worker [flags]

Examples:
Worker subcommand allows you to pass in all available worker parameters. 
All default values of worker command will be passed to the service stub unless overriden.
To get information about default values run "k0s install server --help"

Note: if --token-file value is not passed service stub will add "REPLACEME" value instead of it.
Windows flags like "--api-server", "--cidr-range" and "--cluster-dns" will be ignores since install command doesn't yet support Windows services

Flags:
      --api-server string       HACK: api-server for the windows worker node
      --cidr-range string       HACK: cidr range for the windows worker node (default "10.96.0.0/12")
      --cluster-dns string      HACK: cluster dns for the windows worker node (default "10.96.0.10")
      --cri-socket string       contrainer runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]
      --enable-cloud-provider   Whether or not to enable cloud provider support in kubelet
  -h, --help                    help for worker
      --profile string          worker profile to use on the node (default "default")
      --token-file string       Path to the file containing token.

```
Install server with default values:
```
# ./k0s install server
INFO[2021-01-04 13:50:14] Installing k0s service                       
# cat /etc/systemd/system/k0sserver.service 
[Unit]
Description=k0s - Zero Friction Kubernetes
ConditionFileIsExecutable=/root/go/src/github.com/k0sproject/k0s/k0s
After=network.target 
[Service]
StartLimitInterval=5
StartLimitBurst=10
ExecStart=/root/go/src/github.com/k0sproject/k0s/k0s "server" "--config" "/root/go/src/github.com/k0sproject/k0s/k0s.yaml"
```
Install worker with default values:
```
# ./k0s install worker --token-file whatever
INFO[2021-01-04 13:51:25] Installing k0s service                       
root@k0s:~/go/src/github.com/k0sproject/k0s# cat /etc/systemd/system/k0sworker.service 
[Unit]
Description=k0s - Zero Friction Kubernetes
ConditionFileIsExecutable=/root/go/src/github.com/k0sproject/k0s/k0s
After=network.target 
[Service]
StartLimitInterval=5
StartLimitBurst=10
ExecStart=/root/go/src/github.com/k0sproject/k0s/k0s "worker" "--token-file whatever"
```

jasmingacic:
One major change is that service name will be set according to k0s role. So when installing k0s server the service file will be names k0sserver.service and k0sworker.service respectively.

pkg/install/service.go
```diff
@@ -30,14 +30,16 @@ func EnsureService(args []string) error {
 	var k0sDisplayName string
 
 	prg := &program{}
-
 	for _, v := range args {
 		if v == "server" {
 			k0sDisplayName = "k0s server"
+			k0sServiceName = k0sServiceName + v
+		} else if v == "worker" {
```
kke, line 37: Can this somehow go to the `else` branch? 

worker` doesn't seem to get a `k0sDisplayName` like before.

Looks to me like the whole conditional could be skipped and replaced with:
```go
for _, v := range args {
  k0sDisplayName = "k0s " + v
  k0sServiceName = "k0s" + v
}
```

I have no clue what it is doing, but looks weird.

kke: Or maybe inside `if v == "server" || v == "worker"`

jasmingacic: Display name and service name are not the same. 
Service name is the name of  the service stub like k0sserver.service or k0sworker.service where the display name is just how it is going to appear when listed with systemctl.
p.s. Args slice contains more things than just role name so it had to have the condition as above

go.mod
```diff
@@ -36,6 +35,8 @@ require (
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.6.1
+	github.com/vishvananda/netlink v1.1.0 // indirect
+	github.com/vishvananda/netns v0.0.0-20201230012202-c4f3ca719c73 // indirect
```
kke, line 39: Seem to be unrelated

go.sum
```diff
@@ -800,6 +794,11 @@ github.com/urfave/cli v1.22.1/go.mod h1:Gos4lmkARVdJ6EkW0WaNv/tZAAMe9V7XWyB60NtX
 github.com/valyala/bytebufferpool v1.0.0/go.mod h1:6bBcMArwyJ5K/AmCkWv1jt77kVWyCJ6HpOuEn7z0Csc=
 github.com/valyala/fasttemplate v1.0.1/go.mod h1:UQGH1tvbgY+Nz5t2n7tXsz52dQxojPUpymEIMZ47gx8=
 github.com/vektah/gqlparser v1.1.2/go.mod h1:1ycwN7Ij5njmMkPPAOaRFY4rET2Enx7IkVv3vaXspKw=
+github.com/vishvananda/netlink v1.1.0 h1:1iyaYNBLmP6L0220aDnYQpo1QEV4t4hJ+xEEhhJH8j0=
+github.com/vishvananda/netlink v1.1.0/go.mod h1:cTgwzPIzzgDAYoQrMm0EdrjRUBkTqKYppBueQtXaqoE=
+github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df/go.mod h1:JP3t17pCcGlemwknint6hfoeCVQrEMVwxRLRjXpq+BU=
+github.com/vishvananda/netns v0.0.0-20201230012202-c4f3ca719c73 h1:JGMFiSX7pNu3l0DRO8sjxgA7ybQNE+HeuKIG23PjsN4=
+github.com/vishvananda/netns v0.0.0-20201230012202-c4f3ca719c73/go.mod h1:DD4vA1DwXk04H54A1oHXtwZmA0grkVMdPxx/VGLCah0=
```
kke, line 801: Seem to be unrelated

jasmingacic: I did `go mod tidy` so I haven't touched it

go.mod
jasmingacic: I did `go mod tidy` so I haven't touched it

pkg/install/service.go
kke: Probably doesn't matter, but maybe it should break after the first match.

go.mod
jasmingacic: Weirdly `go mod tidy` actually adds these here

cmd/install.go
```diff
@@ -108,19 +94,58 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
 
 	if debug {
 		args = append(args, "--debug")
 	}
+
+	for k, v := range params {
+		if v != "" {
+			args = append(args, (fmt.Sprintf("--%s %s", k, v)))
+		}
+		if k == "enable-worker" || k == "enable-cloud-provider" {
```
kke, line 145: Couldn't this just be:

```suggestion
		} else {
```

jasmingacic: yes it could

pkg/install/service.go
kke: Could have been simply
```go
  for _, v := range args {
    if v == "server" || v == "worker" {
      k0sDisplayName = "k0s " + v
      k0sServiceName = k0sServiceName + v
      break
    }
  }
```

jasmingacic: Fixed it

cmd/install.go
kke, line 145: Appears `params['enable-cloud-provider']` will never be set.

kke (changes requested):

cmd/install.go
jasmingacic: It has default value set to false 
`workerCmd.Flags().BoolVar(&cloudProvider, "enable-cloud-provider", false, "Whether or not to enable cloud provider support in kubelet")`

Such flags like enable-worker or enable-cloud-provider are by default set to false but when they are passed like `k0s worker --enable-cloud-provider` then they are set to true.

jasmingacic: Which I never set in the block before. Fixed it now.

cmd/install.go
```diff
@@ -108,19 +94,61 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+		params["cri-socket"] = criSocket
 	}
 
 	if debug {
 		args = append(args, "--debug")
 	}
+
+	for k, v := range params {
+		if v != "" {
+			args = append(args, (fmt.Sprintf("--%s %s", k, v)))
```
kke, line 145: Extra parenthesis around `fmt.Sprintf`

cmd/install.go
```diff
@@ -108,19 +94,61 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+		}
+
+		if k == "enable-worker" || k == "enable-cloud-provider" {
+			args = append(args, (fmt.Sprintf("--%s", k)))
+
+		}
+	}
```
kke, line 152:
```suggestion
	for k, v := range params {
		arg := "--" + k
		if v != "" {
			arg += "=" + v
		}
		args = append(args, arg)
	}
```

I don't see a reason to identify the `enable-worker` and `enable-cloud-provider` keys here. It will require an extra change you have to make if you add any other flag without a value.

The intermediate `map[string]string` is a bit unnecessary too, the whole function could just simply be:

```go
func getCmdArgs(role string) args []string {
	args = append(args, role)
	args = append(args, "--data-dir", dataDir)
	if role == "server" {
		args = append(args, "--token-file", serverToken)

		if serverWorkerProfile != "default" {
			args = append(args, "--profile" , serverWorkerProfile)
		}

		if enableWorker {
			args = append(args, "--enable-worker")
		}

		if cfgFile != "" {
			args = append(args, "--config", cfgFile)
		}
	} else {
		args = append(args, "--cri-socket", criSocket)

		if workerProfile != "default" {
			args = append(args, "--profile", workerProfile)
		}

		if tokenFile != "" {
			args = append(args, "--token-file", tokenFile)
		} else {
			args = append(args, "--token-file", "REPLACEME")
		}

		if cloudProvider {
			args = append(args, "--enable-cloud-provider")
		}
	}

	if debug {
		args = append(args, "--debug")
	}

	return args
}
```

Shorter, simpler, less intermediate allocation, only uses anonymous string literals (=constants)

cmd/installWorker.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	installWorkerCmd = &cobra.Command{
+		Use:   "worker",
+		Short: "Helper command for setting up k0s as a worker node on a brand-new system. Must be run as root (or with sudo)",
+		Example: `Worker subcommand allows you to pass in all available worker parameters. 
+All default values of worker command will be passed to the service stub unless overriden.
+To get information about default values run "k0s install server --help"
```
kke, line 28:
```suggestion
To get information about default values run "k0s install worker --help"
```

cmd/installWorker.go
```diff
@@ -0,0 +1,36 @@
...
+
+Note: if --token-file value is not passed service stub will add "REPLACEME" value instead of it.
```
kke, line 30:
```suggestion
Note: if --token-file value is not passed, service stub will add "REPLACEME" value instead of it.
```

kke (dismissed):
I don't see why the temporary `params map[string]string` construct in `cmd/install.go` is needed, I wrote a full replacement for this function without that in a [review comment](https://github.com/k0sproject/k0s/pull/594#discussion_r553805473):

```go
func getCmdArgs(role string) args []string {
	args = append(args, role)
	args = append(args, "--data-dir", dataDir)
	if role == "server" {
		args = append(args, "--token-file", serverToken)

		if serverWorkerProfile != "default" {
			args = append(args, "--profile" , serverWorkerProfile)
		}

		if enableWorker {
			args = append(args, "--enable-worker")
		}

		if cfgFile != "" {
			args = append(args, "--config", cfgFile)
		}
	} else if role == "worker" {
		args = append(args, "--cri-socket", criSocket)

		if workerProfile != "default" {
			args = append(args, "--profile", workerProfile)
		}

		if tokenFile != "" {
			args = append(args, "--token-file", tokenFile)
		} else {
			args = append(args, "--token-file", "REPLACEME")
		}

		if cloudProvider {
			args = append(args, "--enable-cloud-provider")
		}
	}

	if debug {
		args = append(args, "--debug")
	}

	return
}
```

Other than that, LGTM.

cmd/install.go
jasmingacic: @kke can you please re-review this file. I added requested changes.

kke (dismissed):

cmd/install.go
```diff
@@ -107,20 +93,50 @@ func getUserList(sysUsers v1beta1.SystemUser) []string {
...
+	args = append(args, role)
+
+	if dataDir != "" {
+		args = append(args, "--data-dir", dataDir)
+	}
 
 	if role == "server" {
-		args = append(args, "server")
+		if serverToken != "" {
+			args = append(args, "--token-file", serverToken)
```
trawler, line 105: shouldn't `serverToken` actually be changed to `tokenFile`?

trawler (changes requested):
- please also update https://github.com/k0sproject/k0s/blob/main/docs/install.md
- You have some unrelated commits bundled in this PR. Please rebase and clean up your branch before merging. It makes it a bit hard to review and pinpoint the relevant changes.

cmd/install.go
jasmingacic: The variable is inherited from here https://github.com/k0sproject/k0s/blob/main/cmd/server.go#L59 so it is already parsed.

trawler: I believe this is the argument you are looking for: https://github.com/k0sproject/k0s/blob/main/cmd/server.go#L53

jasmingacic: Fixed it, can you please re-review it.

jasmingacic:
I wonder if we should change status of this command from alpha to beta? https://github.com/k0sproject/k0s/blob/main/docs/install.md

jasmingacic:
Rebased and addressed all the remarks

jnummelin (approved):
LGTM

soider (approved):
good stuff

PR #606, opened 2021-01-06
Support cross subnet overlay

### LINKED ISSUE #558: Non-overlay networking
It should be possible to use non-overlay networking for calico when operating inside single L2 subnet, as this option has the least overhead.

Currently the documentation states only IPIP and VXLAN are possible options.

binoychitale:
---
name: Support configurable overlay mode in calico
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #558 

**What this PR Includes**
Currently the overlay mode is hardcoded to `Always` when using calico. We want to keep this configurable so that it can be configured to use cross-subnet overlay mode.


binoychitale:
marked this as draft as I wanted some clarifications -

Since we have modified the static manifests in `static/manifests/calico/DaemonSet/calico-node.yaml`, it requires re-generation of `static/gen_manifests.go`, as these assets seem to be baked into the binary.

Do I need to manually check-in the `gen_manifests` file into version control? or is this done by some hooks

kat33rd (dismissed):

soider:
@binoychitale yes, plese run `make bindata-manifests` to regen it.

binoychitale:
Hi @soider , I generated the file using `make bindata-manifests`, however I am seeing a large diff, which also fails lint because the generated methods are `_` separated.
Not sure why this is.
```
$ go-bindata -version
go-bindata 3.1.0 (Go runtime go1.15.5).
Copyright (c) 2010-2013, Jim Teeuwen.
```

binoychitale:
Never mind, I was using an outdated version of go-bindata

jnummelin (approved):
LGTM

Thanks @binoychitale for the PR 🎉

PR #607, opened 2021-01-06
merge init/k0s.service into systemd template in pkg/install

### LINKED ISSUE #605: Decide on the the systemd example implentation
There currently are two systemd unit file templates in the repo:

A) how `k0s install` creates the unit: https://github.com/k0sproject/k0s/blob/main/pkg/install/service.go 
B) https://github.com/k0sproject/k0s/blob/main/init/k0s.service

It would be good if both implementations were to be combined for consistency.

trawler:

**Issue**
Fixes #605

**What this PR Includes**
Cleans up the systemd template the `k0s install` command and merges the settings previously existing in init/k0s.service file.


jasmingacic (dismissed):

jasmingacic (approved):
Tested it and it is working fine.

Go ahead merge it

PR #612, opened 2021-01-08
Add an example for deploying K0s to a Raspberry Pi 4 cluster

shaneutt:
**What this PR Includes**

This adds a new example in `/examples` that users can use or study to deploy K0s onto a cluster running ARM64, with a particular emphasis on using Raspberry Pi 4 systems.

### DIFF
--- a/examples/raspberry-pi4/README.md
+++ b/examples/raspberry-pi4/README.md
@@ -1,0 +1,272 @@
+# Raspberry Pi 4 K0s Kubernetes Cluster
+
+This is a highly opinionated example of deploying the [K0s](https://github.com/k0sproject/k0s) distribution of [Kubernetes](https://kubernetes.io) to a cluster comprised of [Raspberry Pi 4 Computers](https://www.raspberrypi.org/products/raspberry-pi-4-model-b/) with [Ubuntu 20.04 LTS](https://ubuntu.com) as the operating system.
+
+# Prerequisites
+
+The following tools should be installed on your local workstation to use this example:
+
+* [Kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) `v1.19.4`+
+* [Raspberry Pi Imager](https://github.com/raspberrypi/rpi-imager) `v1.5`+
+
+# Walkthrough
+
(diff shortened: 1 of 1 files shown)


jnummelin:
@shaneutt Thanks for the PR.

Just wondering if it would be possible / makes sense to re-use what has already been done for general Ansible at: https://github.com/movd/k0s-ansible

We already have a example also for it at: https://github.com/k0sproject/k0s/blob/main/docs/examples/ansible-playbook.md

shaneutt:
> @shaneutt Thanks for the PR.
> 
> Just wondering if it would be possible / makes sense to re-use what has already been done for general Ansible at: https://github.com/movd/k0s-ansible
> 
> We already have a example also for it at: https://github.com/k0sproject/k0s/blob/main/docs/examples/ansible-playbook.md

Hi @jnummelin thank you for your considerations :bow: 

The intention here was to focus on Raspberry Pis, and the use of Ansible was meant to be convenient, but teaching ansible wasn't the specific point, so I wonder: if I were to remove Ansible from the equation or otherwise change this example do you think there's some way we could make it align better with your expectations for examples in this repo?

Or perhaps more generally: could you please help me to better understand what you're looking for in examples submitted to this repository?

jnummelin:
I think we want to have examples that cover wide range of use cases and deployment architectures to guide the community.

The example itself is good and fit's perfectly the above goal, I'm just mainly thinking out loud if we could re-use the ansible bits so we do not end up having two examples with totally different ansible playbooks.

shaneutt:
To avoid having multiple split ansible playbooks like you're suggesting, I'm considering removing the ansible parts here and focusing instead on showing the commands needed for this specific kind of environment, making this lean a little bit more to being a documentation example over being more of a functional example.

My proposal for action items is:

1. remove the ansible
2. move the commands and context into the README.md
3. submit it again as a PR against `examples/`

What are your thoughts on that? Do you prefer a different approach?

jnummelin:
Sounds like a good plan 👍

shaneutt:
@jnummelin updated it according to our previous discussion, let me know what you think.

jnummelin (approved):
Good stuff. Thanks @shaneutt

PR #624, opened 2021-01-12
Use same kube clients everywhere

### LINKED ISSUE #610: Many instances of kube client on controller
Sub-ticket of #604 

**Version**
```
0.9.1
```
**Platform**
ALL

**What happened?**
We create many kube clients for different purposes. The issue this creates is the fact that each client caches lot of information (api resources, swagger, etc.) which uses quite a bit of memory.

**Expected behavior**
We can re-use the same kube client instance for most purposes as the client is documented to be "thread-safe".

jnummelin:

**Issue**
Fixes #610

**What this PR Includes**
This PR refactors all the components that needs kube client to use the one shared clients through `ClientFactory` insterface. The implementation is made as a "lazy-loading" factory so that we can create the factory itself before any components that uses it as at that point we do not yet have the capabilities (admin kubeconfig namely) to create the actual clients.

### DIFF
--- a/pkg/component/server/helmaddons.go
+++ b/pkg/component/server/helmaddons.go
@@ -29,28 +29,30 @@ import (
 
 // Helm watch for Chart crd
 type HelmAddons struct {
-	Client        clientset.ChartV1Beta1Interface
-	ClusterConfig *k0sv1beta1.ClusterConfig
-	saver         manifestsSaver
-	L             *logrus.Entry
-	stopCh        chan struct{}
-	informer      cache.SharedIndexInformer
-	helm          *helm.Commands
-	kubeConfig    string
+	Client            clientset.ChartV1Beta1Interface
+	ClusterConfig     *k0sv1beta1.ClusterConfig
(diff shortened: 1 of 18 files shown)


jasmingacic (dismissed):

pkg/component/server/helmaddons.go
```diff
@@ -62,13 +64,14 @@ const (
...
 
-// Init
+// Run runs the helm controller
 func (h *HelmAddons) Run() error {
 	h.L.Info("run begin")
 	if h.ClusterConfig.Extensions == nil || h.ClusterConfig.Extensions.Helm == nil {
 		h.L.Info("No helm addons specified, do not run HelmAddons reconciler")
 		return nil
 	}
+	// TODO Can we use the shared kube client factory to create the clientser for helm CRDs?
```
jasmingacic, line 74: clientser looks like a typo

jasmingacic (approved):

ncopa (approved):
LGTM

PR #633, opened 2021-01-16
Manifests at build time

### LINKED ISSUE #615: Generate manifests at build time
Currently, the static assets such as YAML configs are baked into the k0s binary using `go-bindata`. Whenever a change is made to these static assets, we need to run `make bindata-manifests` to generate a source file with generated assets.

Run the manifest generation at build time. i.e when doing a `make build`. 
This seems like a good option because -
1) It does not seem to be a costly operation 
2) Generated assets file need not be checked into version control (e.g `static/gen_manifests.go`)
3) Generation of these assets will be automated (dev does not need to manually add generated files into git)

**Additional context**

binoychitale:
---
name: Generate manifests at build time
about: 
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #615 

**What this PR Includes**
1. Run the manifest generation at build time. i.e when doing a make build.
2. Remove `static/gen_manifests.go` from version control


binoychitale:
The roadblock I'm facing here is that the build pipeline does not have `go-bindata` set up in the environment, and hence the pipeline fails. I did not find any existing github actions to do the same.

I have a feeling this might be the reason why this hadn't been picked up yet 🤔

binoychitale:
Ive made some change to the github workflow pipeline to pre-install go-bindata for manifest generation at build time

.github/workflows/go.yml
```diff
@@ -66,6 +66,10 @@ jobs:
           key: ${{ runner.os }}-embedded-bins-${{ hashFiles('**/embedded-bins/**/*') }}
           restore-keys: |
             ${{ runner.os }}-embedded-bins-${{ hashFiles('**/embedded-bins/**/*') }}
+      - name: Generate manifests
+        id: gen-manifests
+        run: |
+          sudo apt-get install go-bindata
```
soider, line 72: What do you think about having it as a dependency (the same way we do with go-lint) to have way to pin and update versions without relying on debian maintainers?

Makefile
```diff
@@ -81,20 +83,19 @@ $(smoketests): k0s
...
+	rm -f pkg/assets/zz_generated_offsets_*.go k0s k0s.exe .bins.*stamp bindata* static/gen_manifests.go
 	$(MAKE) -C embedded-bins clean
 
 .PHONY: manifests
 manifests:
 	controller-gen crd paths="./..." output:crd:artifacts:config=static/manifests/helm/CustomResourceDefinition object
 
-.PHONY: bindata-manifests
-bindata-manifests:
+static/gen_manifests.go:
```
soider, line 98: I think, this should have makefile level dependency on the static files, otherwise bindata wouldn't be regenerated in case if we change something in "static" directory

binoychitale: Makes sense. Will add this

.github/workflows/go.yml
soider: No, I don't

Let me guide you through what I meant.
In a Makefile we have target `lint` to run linter. It could be used either locally to run linter without even touching CI or through CI.
By having this `go get ...` code we pin the particular version of the golint to use.
I suggest do the same way for go-bindata because with go-bindata it could be flight critical to have particular version pinned.
```
golint := $(shell which golangci-lint)
ifeq ($(golint),)
golint := go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.31.0 && "${GOPATH}/bin/golangci-lint"
endif

.PHONY: lint
lint: pkg/assets/zz_generated_offsets_$(TARGET_OS).go
	$(golint) run ./...
```

binoychitale: Got it, thanks for the clarification. 
I've implemented it similar to the `golangci-lint` (i.e check if installed locally, or else get via `go get ...` and use that binary)

.gitignore
```diff
@@ -12,3 +12,4 @@ bindata*
 aws_private.pem
 out.json
 .vscode
+static/gen_manifests.go
```
soider, line 15: I would rather prefer to keep generated code in the repo, but it's absolutelly open for a discussion. @ncopa do you have an opinion about?

soider: I actually just cautght a thought that bindata output is non-human readable anyway, so resolving the conversation

ncopa: In my experience, checking in generated code/data into git creates more problems that it solves, and should be avoided.

.github/workflows/go.yml
```diff
@@ -307,9 +306,10 @@ jobs:
     if: github.ref != 'refs/heads/main'
     steps:
       - uses: actions/checkout@v2
-
       - name: codegen
-        run: make pkg/assets/zz_generated_offsets_linux.go
+        run: |
+          make pkg/assets/zz_generated_offsets_linux.go
+          make static/gen_manifests.go
```
ncopa, line 312:
```suggestion
          make -j$(nproc) pkg/assets/zz_generated_offsets_linux.go static/gen_manifests.go
```

They can be generated in parallel

ncopa (changes requested):
I think this is a good idea.

But we maybe don't need 13 commits for it? I think this could be done in a single commit.
https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md#squashing-commits

I also like when commit message has a short explanation of why the change was done.

.github/workflows/go.yml
binoychitale: Makes sense, I have made these changes. 
I could only test the CI pipeline by committing to the repo, hence the sea of commits :)
I've squashed them now

ncopa: yeah, its perfectly fine while getting feedback and working on the PR. But its nice that the git history of the final result is clean and tidy. Thanks!

ncopa (approved):

ncopa:
Also passed network test https://github.com/k0sproject/k0s/actions/runs/496485335

Issue #602, opened 2021-01-05
Determine version and role of currently running k0s instance


kke:
K0s needs something along the lines of:

```
$ k0s status
not running
```

```
$ k0s status
running: k0s worker version 0.9.1 
```

```
$ k0s status
running: k0s controller version 0.9.1
```

I guess it could be made simply with something like `/run/k0s/status` which would look something like:

```
pid=1234
role=worker
version=0.9.1
```

The `k0s status` command would read this status file, check the pid exists and looks like a k0s,  and then report the rest of it.

A regular pid file like `/run/k0s/k0s.pid` could also be handy, it could be checked when trying to launch a new instance to avoid multiple instances and maybe the PID file could be somehow utilized by systemd/openrc/whatever init system.

jnummelin:
I _think_ we can dig out the needed info from standard proc bits:
```
~# ps -ef | grep ‘k0s server’
root       91598   91571  0 15:30 pts/0    00:00:00 sudo k0s server
root       91600   91598  0 15:30 pts/0    00:00:00 k0s server
root       91926   91776  0 15:32 pts/1    00:00:00 grep --color=auto k0s server
~# /proc/91600/exe version
v0.8.1
~# cat /proc/91600/cmdline
k0s--debug--data-dir=/var/lib/k0s/custom-data-dirworker--token-file=/token
```

Of course it would be probably easier to have all the needed info in some more easily parseable format in some custom status file

jasmingacic:
#637 should handle this.

Closing the issue.

PR #637, opened 2021-01-19
k0s status

### LINKED ISSUE #544: Make "k0s install" a bit more verbose
When running `k0s install` it would be useful to have some more information:

* The type of service installed
* The location of the startup script
* An output of the startup command

jasmingacic:
---
name: Pull Request
about: Create a Pull Request
title: 'k0s status'
labels: 'enhancememt'
assignees: ''

---

**Issue**
Fixes #544 #602 

**What this PR Includes**
A simple command that prints out following:
```
Process ID: 1076667
Role: server+worker
Init System: linux-systemd
```
of if --out parameter is used it will print json/yaml output
```
{
   "Pid": 1076667,
   "PPid": 856280,
   "Role": "server+worker",
   "SysInit": "linux-systemd",
   "StubFile": ""
}
```
```
pid: 1076667
ppid: 856280
role: server+worker
sysinit: linux-systemd
stubfile: ""
```

This is just a draft as I want to hear more input on what we would/should surface with this command

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -26,7 +26,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/k0sproject/k0s/pkg/build"
 	"github.com/k0sproject/k0s/pkg/install"
 	ps "github.com/mitchellh/go-ps"
 	"github.com/spf13/cobra"
@@ -53,6 +52,12 @@ var (
 			}
 
 			if status.Pid != 0 {
+				if ver, err := getK0sVersion(status.Pid); err != nil {
+					return err
+				} else {
+					status.Version = ver
+				}
+
 				if user, err := getProcessOwner(status.Pid); err != nil {
 					return err
 				} else if !strings.Contains(user, "root") {
@@ -90,7 +95,6 @@ type K0sStatus struct {
 }
 
 func (s K0sStatus) String() {
-	s.Version = build.Version
 	switch s.output {
 	case "json":
 		jsn, _ := json.MarshalIndent(s, "", "   ")
@@ -99,10 +103,11 @@ func (s K0sStatus) String() {
 		ym, _ := yaml.Marshal(s)
 		fmt.Println(string(ym))
 	default:
-		fmt.Println("Version:", s.Version)
 		if s.Pid == 0 {
 			fmt.Println("K0s not running")
+			return
 		} else {
+			fmt.Println("Version:", s.Version)
 			fmt.Println("Process ID:", s.Pid)
 			fmt.Println("Parent Process ID:", s.PPid)
 			fmt.Println("Role:", s.Role)
@@ -194,4 +199,13 @@ func getProcessOwner(pid int) (string, error) {
 		return "", err
 	}
 	return string(stdout), nil
+}
+
+func getK0sVersion(pid int) (string, error) {
+	cmd := fmt.Sprintf("/proc/%d/exe", pid)
+	stdout, err := exec.Command(cmd, "version").Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSuffix(string(stdout), "\n"), nil
 }
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -52,11 +52,11 @@ var (
 			}
 
 			if status.Pid != 0 {
-				if ver, err := getK0sVersion(status.Pid); err != nil {
+				ver, err := getK0sVersion(status.Pid)
+				if err != nil {
 					return err
-				} else {
-					status.Version = ver
 				}
+				status.Version = ver
 
 				if user, err := getProcessOwner(status.Pid); err != nil {
 					return err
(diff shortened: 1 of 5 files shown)


jasmingacic:
I haven't tested this with worker nodes or any other OSs but linux. If anyone has k0s running as worker or on windows host please test it.

ncopa:
What happens if there are multiple instances of k0s running on the same machine?

I wonder if it would be better to query the running instance via a unix socket or similar, rather than do assumptions based on what is found by the process list.

jasmingacic:
@ncopa valid question.  Current setup wouldn't be able to detect all running instances. We could perhaps try looking if the binary is running as a service because we can't have two running services with the same name.

ncopa:
@jasmingacic I think it would make more sense to connect to the running instance via an API. Then we we can also show state of supervised process (running/restarting) and do a health check of the components that supports that. Currently only etcd I think.

jasmingacic:
@ncopa we briefly discussed the option you suggested [here](https://github.com/k0sproject/k0s/pull/637#issuecomment-762959720) and as @jnummelin suggested there might be some security concerns. Also currently it isn't not possible to run two k0s instances on the same node.
This command is essentially just convenience for some work that needs to be done on k0sctl.

cmd/status.go
```diff
@@ -0,0 +1,168 @@
...
+				PPid: p.PPid()}
+
+			return status, nil
+		}
+	}
+
+	return &K0sStatus{}, nil
+}
+
+func hasChildern(pid int, processes []ps.Process) bool {
```
jnummelin, line 161: `hasChildern ` -> `hasChildren `

cmd/status.go
```diff
@@ -0,0 +1,168 @@
...
+
+}
+func getRole(pid int) (role string, err error) {
+	if runtime.GOOS == "windows" {
+		return "worker", nil
+	}
+
+	var raw []byte
+	if raw, err = ioutil.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid)); err != nil {
+		return "", fmt.Errorf("K0s not running")
```
jnummelin, line 132: the most obvious reason for this to fails is not running the command as root, not k0s not running

cmd/status.go
```diff
@@ -0,0 +1,168 @@
...
+	output string
+
+	status    *K0sStatus
+	statusCmd = &cobra.Command{
+		Use:     "status",
+		Short:   "Helper command for get general information about k0s",
+		Example: `The command will return information about system init, PID, k0s role, kubeconfig and similar.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+
```
jnummelin, line 44: I think we should check if we're running this as root or not. we're looking for stuff like process list, `/proc/1234/cmdline` which will not be readable without root privileges

cmd/status.go
```diff
@@ -0,0 +1,168 @@
...
+	var raw []byte
+	if raw, err = ioutil.ReadFile(fmt.Sprintf("/proc/%d/cmdline", pid)); err != nil {
+		return "", fmt.Errorf("K0s not running")
+	}
+	cmdln := string(raw)
+	if strings.Contains(cmdln, "enable-worker") {
+		return "server+worker", nil
+	} else if strings.Contains(cmdln, "server") {
+		return "server", nil
+	}
```
jnummelin, line 139: Missing case to detect pure worker

cmd/status.go
```diff
@@ -0,0 +1,168 @@
...
+	return role, nil
```
jnummelin, line 140: I don't think it makes sense to return empty string as role as a "fallback"

jnummelin (changes requested):

jasmingacic:
```
# ./k0s status --out json
{
   "Version": "v0.9.0-41-g1e859ee",
   "Pid": 0,
   "PPid": 0,
   "Role": "",
   "SysInit": "",
   "StubFile": ""
}
```
```
# ./k0s status --out yaml
version: v0.9.0-41-g1e859ee
pid: 0
ppid: 0
role: ""
sysinit: ""
stubfile: ""
```
```
# ./k0s status 
Version: v0.9.0-41-g1e859ee
K0s not running
```

cmd/status.go
```diff
@@ -0,0 +1,197 @@
...
+	for _, p := range processes {
+		if p.PPid() == pid {
+			return true
+		}
+	}
+	return false
+}
+
+func getProcessOwner(pid int) (string, error) {
+	stdout, err := exec.Command("ps", "-o", "user=", "-p", strconv.Itoa(pid)).Output()
```
kke, line 192: I think you could just use the [process.Username](https://pkg.go.dev/github.com/shirou/gopsutil/process#Process.Username)

cmd/status.go
```diff
@@ -0,0 +1,197 @@
...
+	Pid      int
+	PPid     int
+	Role     string
+	SysInit  string
+	StubFile string
+	output   string
+}
+
+func (s K0sStatus) String() {
+	s.Version = build.Version
```
kke, line 93: This needs to be the version of the running daemon, not the version of current executable.

jasmingacic: @jnummelin do you think we can achieve this in any other way?

kke: 1) write stuff to a file under the run-dir
2) `/proc/$pid/exe version`

kke: method 1 would be nice, k0s could pretty much write a `K0sStatus` json to rundir `status.json`

jasmingacic:
Once approved I will squash the commits

kke (approved):
Ok for now.

The slightly hackish ways to determine information about the running process should probably be replaced with some sort of `status.json` in the run-dir in the future which would be written out when k0s starts, deleted when it ends and maybe kept up to date for some parts during k0s lifetime.

cmd/status.go
```diff
@@ -0,0 +1,211 @@
...
+	default:
+		if s.Pid == 0 {
+			fmt.Println("K0s not running")
+			return
+		}
+
+		fmt.Println("Version:", s.Version)
+		fmt.Println("Process ID:", s.Pid)
+		fmt.Println("Parent Process ID:", s.PPid)
+		fmt.Println("Role:", s.Role)
```
kke, line 114: (could just default to yaml output and drop these, yaml is human readable)

jnummelin (approved):

PR #666, opened 2021-01-28
Make supervisor exit on fatal errors

### LINKED ISSUE #225: k0s supervisor should exit on fatal errors
Example: when mke binaries do not exist. 
We should identify the fatal errors, and clean up all dependent processes before exiting.

ncopa:
Verify that executable actually starts and return an error if it does
not. This covers the case when binary does not exist or has wrong
permissions or exec(2) fails for any other reason.

fixes #225

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #225

**What this PR Includes**
Refactor supervisor to verify that executable actually starts and return failure if it doesn't so k0s can abort the startup.

Note that this does not make k0s shutdown if a process successfully starts up but immediately exits with error for some reason (cannot bind to port, disk full or similar).

### DIFF
--- a/pkg/supervisor/supervisor.go
+++ b/pkg/supervisor/supervisor.go
@@ -107,14 +107,9 @@ func (s *Supervisor) Supervise() error {
 		s.TimeoutRespawn = 5 * time.Second
 	}
 
-	s.quit = make(chan bool)
-	s.done = make(chan bool)
-
+	started := make(chan error)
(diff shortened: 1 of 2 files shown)


jnummelin (approved):

trawler (approved):
lgtm

PR #670, opened 2021-01-29
Use warning/info instead of error messages for default config

### LINKED ISSUE #668: drop error message that claims defaults may not work
k0s prints this warning when no config is specified:
```
ERRO[2021-01-29 11:13:40] Failed to read cluster config: failed to read config file at /root/k0s.yaml: open /root/k0s.yaml: no such file or directory
ERRO[2021-01-29 11:13:40] THINGS MIGHT NOT WORK PROPERLY AS WE'RE GONNA USE DEFAULTS 
```
Remove that error message or change it to "Failed to read config file. Using defaults".

It can be either "info"  or "warning", but IMHO it should not be an error.

**Additional context**
The default values should work just fine. If they don't we should fix it.

ncopa:
fixes #668

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #668

**What this PR Includes**
Dot show error message when default config is used. Display a warning + info instead

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -34,8 +34,8 @@ func ConfigFromYaml(cfgPath string) (clusterConfig *config.ClusterConfig, err er
 	}
 
 	if err != nil {
-		logrus.Errorf("Failed to read cluster config: %s", err.Error())
-		logrus.Error("THINGS MIGHT NOT WORK PROPERLY AS WE'RE GONNA USE DEFAULTS")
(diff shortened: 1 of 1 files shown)


ncopa:
I wonder if we should drop the warning as well, at least in the "no such file" case.

mviitane (approved):
Looks like a very good improvement! I wonder if "Failed to read..." is still too much like a failure for quite a normal use case. It could also be something like "Cannot find config file, using default".

PR #672, opened 2021-01-29
Dynamic HA configuration for Konnectivity

### LINKED ISSUE #595: Multi-node controller get stuck
Hi, I'm trying to setup a multi-node controller cluster and after a while (5 mins) the system turn inestable and get stuck two of three masters machines whit a konnectivity-server error.

**Version**
```
v0.9.1
```
**Platform**

```
OS:                 linux (amd64)
OS Image:           Fedora CoreOS 33.20201201.3.0
Kernel version:     5.9.11-200.fc33.x86_64
Container runtime:  containerd://1.4.3
Kubelet version:    v1.20.1-k0s1
```
**What happened?**

**Logs**

```
Jan 01 19:18:15 k0s-1 k0s[822]: time="2021-01-01 19:18:10" level=info msg="E0101 19:18:10.548256    2040 egress_selector.go:201] failed to create connection to uds name /run/k0s/konnectivity-server/konnectivity-server.sock, error: dial unix /run/k0s/konnectivity-server/konnectivity-server.sock: socket: too many open files" component=kube-apiserver
Jan 01 19:18:15 k0s-1 k0s[822]: time="2021-01-01 19:18:10" level=info msg="W0101 19:18:10.548442    2040 clientconn.go:1223] grpc: addrConn.createTransport failed to connect to {/run/k0s/konnectivity-server/konnectivity-server.sock  <nil> 0 <nil>}. Err :connection error: desc = \"transport: Error while dialing dial unix /run/k0s/konnectivity-server/konnectivity-server.sock: socket: too many open files\". Reconnecting..." component=kube-apiserver
```

### LINKED ISSUE #596: 2nd and 3rd node controllers are going to down after the worker node join.
**Version**
```
$ k0s version
v0.9.1
```
**Platform**
```
$ lsb_release -a
LSB Version:	core-11.1.0ubuntu2-noarch:security-11.1.0ubuntu2-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.1 LTS
Release:	20.04
Codename:	focal
```
```
$ uname -a
Linux host1 5.4.0-1032-aws #33-Ubuntu SMP Wed Dec 9 17:17:06 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
```

**What happened?**
When joining the worker nodes into Multi-node controller cluster, 2nd and 3rd controller nodes show increasing file descriptor of the nodes, eating the memory and it causes hung controller nodes.
This may be the same issue as #595. It seems to be coming from konnectivity and kube-apiserver.

**How To Reproduce**
1st controller: k0s -c k0s.yaml server, and then token create --role=controller, --role worker
2nd and 3rd controller: k0s server  -c k0s.yaml server <token>
worker node: k0s worker <token>

Right after joining worker nodes into the cluster, the following logs observed every 5secin 2nd and 3rd controller node.

```
INFO[2020-12-31 12:39:08] E1231 12:39:08.104653    1077 server.go:276] "Failed to get a backend" err="No backend available"  component=konnectivity
```

Eventually, 2nd nodes and 3rd nodes are going to down. kube-apiserver on 2nd and 3rd node eats huge memory and increasing file descriptor are observed unlimitedly as bellow.

```
root@host3:~#  while true; do lsof -p <kube-apiserver PID> |wc -l && date ; sleep 5; done
514
Sun Jan  3 01:38:36 JST 2021
520
Sun Jan  3 01:38:41 JST 2021
525
Sun Jan  3 01:38:46 JST 2021
530
Sun Jan  3 01:38:51 JST 2021

root@host3:~#  while true; do lsof -p <konnectivity-server PID> |wc -l && date ; sleep 3; done
461
Sun Jan  3 01:39:05 JST 2021
466
Sun Jan  3 01:39:08 JST 2021
471
Sun Jan  3 01:39:11 JST 2021
471
Sun Jan  3 01:39:14 JST 2021
477
Sun Jan  3 01:39:17 JST 2021
```

Guessing that Konnectivity agent settings may cause this.
Although the default setting of "--proxy-server-host=" in konnectivity-agent.yaml points 1st controller nodes as a default, in case the setting is changed to 2nd or 3rd controller, the increase of the file-descriptor is not observed in 2nd or 3rd controllers node. On the other hand, the increase of the file descriptor starts to be shown on the 1st node.

### LINKED ISSUE #600: konnectivity agents connecting to only one controller
**Version**
```
0.9.1
```
**Platform**
ALL

**What happened?**
Currently the konnectivity-agent configs are done so that the agent connects to single address. it's either the direct node address or the given `externalAddress` (LB). 

**How To Reproduce**
Just run k0s with the ddefault configs.

**Expected behavior**
The agent should connect to ALL controllers. For the agent to recognize this the konnectivity-server needs to be configured with:
- `--server-count`: How many unique connections each agent should make
- `--server-id`: unique id for the server (we can probably use the "machineID")

The problem is that how can we get a number of running (or expected to be running) controller nodes dynamically?

One idea is to have a `lease` per controller node and then on each controller we count all the nodes having a lease.

Why lease and not some other object?

Lease ensures that if a controller is lost for good, we also update the expected number of controllers as the lease expires.

**Additional context**
This is actually sort-of "identified" by issues #596 and #595. Leaking FDs is probably actually a bug in either or both konnectivity-server and kube api-server

trawler:
**Issue**
Fixes #600 
Fixes #596 and 
Fixes #595

**What this PR Includes**
This PR introduces a lease package, that allows a service to create a lease per controller. This lease then is checked on a regular basis to give back the number of "live" service backends.

Konnectivity uses this lease package to provide the server-count dynamically for a truly HA konnectivity setup. If that number changes (I.E, a server is added, or goes down), konnectivity restarts with the new settings.


trawler:
I separated APIAddress and APIAddressURL function in https://github.com/k0sproject/k0s/tree/e72d12c92ab7c9ed34103398a3ec00c324720c59

Once it's merged, I'll rebase this PR.

pkg/component/server/konnectivity.go
```diff
@@ -124,10 +154,105 @@ func (k *Konnectivity) writeKonnectivityAgent() error {
...
+			select {
+			case <-ticker.C:
+				observedLeaseHolders := k.leaseLock.CountValidLeaseHolders(ctx)
+				if observedLeaseHolders != k.serverCount {
+					logrus.Debugf("change in %v lease holders detected. refreshing service.", serviceName)
+					k.serverCount = observedLeaseHolders
+
+					// restarting service
+					k.Stop()
+					k.Run()
```
ncopa: As far I can see, `runLease()` will start two go routines that never returns. (both has a `for {}` loop without any return/break or condition to exit the loop). I can not see that `k.Stop()` shuts those go routines down, so they will remain running. The `k.Run()` here will recursively spawn new instances of those go-routines (`Run()` -> `runLease()` -> `Run()` ...).

Do we have a leak here?

pkg/component/server/konnectivity.go
```diff
@@ -188,28 +188,28 @@ func (k *Konnectivity) runLease() {
...
 
 					// restarting service
-					k.Stop()
-					k.Run()
+					if err := k.Stop(); err != nil {
+						logrus.Errorf("error stopping %v service: %v", serviceName, err)
+					}
+					if err := k.Run(); err != nil {
+						logrus.Errorf("error starting %v service: %v", serviceName, err)
+					}
```
ncopa, line 205: If `k.Stop()` here above fails (means that konnectivity is still running), should we really start another instance? Would it be better to have an `else` there, so we only do `k.Run()` if the `k.Stop()` was successful?

ncopa (commented):
It looks like we are introduce resource leak.

jnummelin (changes requested):

pkg/component/server/konnectivity.go
jnummelin: oh, right. so instead of dealing with the level of `k.Stop()` and `k.Run()` we should rather stop and start only the supervisor (konnectivity-server process only).

ncopa: either that or we implement the `runLease()` thingy as a separate Component, that is conditionally added from `cmd/server.go` and passes in the konnectivity component.

pkg/component/server/konnectivity.go
```diff
@@ -124,10 +154,91 @@ func (k *Konnectivity) writeKonnectivityAgent() error {
...
+		}
+		k.leaseLock = leaseLock
+		logrus.Debugf("found %v lease holders", k.serverCount)
+
+		ctx := context.Background()
+		defer func() {
+			k.done <- true
+		}()
+		for {
+			k.leaseLock.LeaseRunner(ctx)
```
jnummelin, line 177: We don't return/break out from the loop at all, ever?

pkg/component/server/konnectivity.go
```diff
@@ -124,7 +161,97 @@ func (k *Konnectivity) writeKonnectivityAgent() error {
...
+			default:
+				k.leaseLock.LeaseRunner(context.Background())
+			}
+		}
+	}(k.leaseCtx)
+
+	go func(ctx context.Context) {
+		logrus.Infof("watching %v lease holders", serviceName)
+		ticker := time.NewTicker(30 * time.Second)
+
```
jnummelin, line 190:
```suggestion
defer ticker.Close()
```
Just to be sure we do not leak tickers when restarting konnectivity

jnummelin (changes requested):

jnummelin (approved):

PR #673, opened 2021-01-29
Introduce configuration changes for the ipv6 support

### LINKED ISSUE #394: Support for dual-stack Kubernetes cluster
I'm trying to run k0s as a dual stack Kubernetes cluster but I'm unable to modify the calico ipam in order to enable the ipv6 support.

I would like to be able to turn on the IPv6 support in calico like described here: https://docs.projectcalico.org/networking/ipv6.

N/A

**Additional context**
N/A

soider:
Pass CIDR as separate ENV variables for calico-node
Change default node-cidr-mask-size-ipv6 to be a smaller size value
Enable CALICO_IPV6POOL_NAT_OUTGOING for dualstack setup
Change kube-proxy to ipvs in case of dual-stack setup

closes #394 

The changes here enable to create ipv6\ipv4 dualstack setup

Limitations: 
- Connections to the outer world work (e.g. `curl -6 ipv6.google.com` will work from any pod)
- It seems like currently k8s doesn't assign internal ipv6 address automatically on bare-metal, but could be done by workaround with external cloud provider and manual node status update as described in https://github.com/kubernetes/kubernetes/issues/42125
- each pod gets ipv6 address from the CNI
- cross pod connectivity exists across nodes

Additional links used during this PR:
https://www.projectcalico.org/dual-stack-operation-with-calico-on-kubernetes/
https://docs.projectcalico.org/networking/ipv6
https://kubernetes.io/docs/tasks/network/validate-dual-stack/ (this one mentions how to create dual\single stack services)
https://kubernetes.io/docs/concepts/services-networking/dual-stack/

In case of testing this on AWS the source\destination checks must be disabled on the interface.

k0s.yaml used:
```
spec:
  network:
    calico:
      mode: "bird"
    dualStack:
      enabled: true
      IPv6podCIDR: "fd00::/108"
      IPv6serviceCIDR: "fd01::/108"
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -300,6 +300,37 @@ jobs:
           path: |
             /tmp/*.log
 
+  smoketest-dualstack:
+    name: Smoke test for IPv6 dualstack
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get PR Reference and Set Cache Name
+        run: |
+          PR_NUMBER=$(echo ${GITHUB_REF} | cut -d / -f 3 )
+          echo "cachePrefix=k0s-${PR_NUMBER}-${{ github.sha }}" >> $GITHUB_ENV
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Cache compiled binary for smoke testing
+        uses: actions/cache@v2
+        id: restore-compiled-binary
+        with:
+          path: |
+            k0s
+          key: build-${{env.cachePrefix}}
+
+      - name: Run smoke test .
+        run: make -C inttest check-dualstack
+
+      - name: Collect test logs
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          path: |
+            /tmp/*.log
+
   lint:
     name: Lint
     runs-on: ubuntu-latest
--- a/pkg/apis/v1beta1/network.go
+++ b/pkg/apis/v1beta1/network.go
@@ -28,7 +28,7 @@ type Network struct {
 	ServiceCIDR string    `yaml:"serviceCIDR"`
 	Provider    string    `yaml:"provider"`
 	Calico      *Calico   `yaml:"calico"`
-	DualStack   DualStack `yaml:"dualStack"`
+	DualStack   DualStack `yaml:"dualStack,omitempty"`
 }
 
 // DefaultNetwork creates the Network config struct with sane default values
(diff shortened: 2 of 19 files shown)


jasmingacic (dismissed):

jnummelin (changes requested):
Few minor tweaks needed

.github/workflows/go.yml
```diff
@@ -300,6 +300,38 @@ jobs:
           path: |
             /tmp/*.log
 
+  smoketest-dualstack:
+    name: Smoke test for IPv6 dualstack
+    needs: build
+    runs-on: ubuntu-latest
+    if: false
```
jnummelin, line 307: Why skip on actions?

unixfox:
Could you allow dual stack mode for `custom` network provider? Currently, I can't do that due to this error: `network dual stack is supported only for calico mode `bird``.

My config:
```yml
network:
    provider: custom
    dualStack:
      enabled: true
      IPv6podCIDR: "fd00::/108"
      IPv6serviceCIDR: "fd01::/108"
```

unixfox (dismissed):
Works great!

I tested:
- [x] reaching the exposed port of a nodeport from a pod using the nginx image
- [x] pod interconnectivity with pods in other nodes
- [x] connectivity to the internet from a pod

unixfox (approved):

jnummelin (approved):

pkg/apis/v1beta1/network.go
```diff
@@ -46,6 +48,9 @@ func (n *Network) Validate() []error {
 	if n.Provider != "calico" && n.Provider != "custom" {
 		errors = append(errors, fmt.Errorf("unsupported network provider: %s", n.Provider))
 	}
+	if n.Provider == "custom" && n.DualStack.Enabled && n.Calico.Mode != "bird" {
```
jnummelin, line 51:
```suggestion
	if n.Provider == "calico" && n.DualStack.Enabled && n.Calico.Mode != "bird" {
```
I don't think we want to validate Calico mode when running custom CNI :)

jnummelin (changes requested):

.github/workflows/go.yml
soider: accident

soider:
> pported only for calico mode bird

@unixfox should be fixed in latest version

pkg/apis/v1beta1/network.go
soider: That cause to validate calico mode only. Like, if we have calico enabled we want to have dualstack enabled only with "bird" mode

soider:
@unixfox thank you for the help with testing things!

jnummelin (approved):

PR #678, opened 2021-02-03
Documentation page headers

mviitane:
Updated documentation page headers and aligning them with navigation.

Signed-off-by: mviitanen@mirantis.com


docs/raspberry-pi4.md
```diff
@@ -1,15 +1,15 @@
-# Raspberry Pi 4 K0s Kubernetes Cluster
+# Creating a cluster with Raspberry Pi 4
```
jasmingacic, line 1: I'd go with  `Running cluster on Raspberry Pi 4` rather than `Createing a cluster with...`

jasmingacic: Or maybe `Creating Raspberry Pi 4 k0s cluster`

docs/k0s-in-docker.md
```diff
@@ -19,7 +19,7 @@ token=$(docker exec -t -i k0s k0s token create --role=worker)
 
 Then join a new worker by running the container with:
 ```
-docker run -d --name k0s-worker1 --hostname k0s-worker1 --privileged -v /var/lib/k0s docker.pkg.github.com/k0sproject/k0s/k0s:<version> k0s worker $token
+docker run -d --name k0s-worker1 --hostname k0s-worker1 --privileged -v /var/lib/k0s docker.io/k0sproject/k0s:latest k0s worker $token
```
soider, line 22: I think, it's better to have both images as an example here, because as far as I remember docker.io registry could have limits now and if people would run this example while we hit limits there could be questions and incoming issues

soider (dismissed):
looking good, left one minor suggestion, but it's not a blocker

mkdocs.yml
```diff
@@ -10,26 +10,27 @@ copyright: Copyright &copy; 2021 <a href="https://mirantis.com/">Mirantis Inc.</
...
+      - Installing the Traefik Ingress Controller: examples/traefik-ingress.md
+      - Installing the Ambassador Gateway: examples/ambassador-ingress.md
+      - Creating a cluster with Raspberry Pi 4: raspberry-pi4.md
+      - Creating a cluster with Ansible Playbook: examples/ansible-playbook.md
   - CLI Commands:
-      - General:            cli/k0s.md
+      - k0s:                cli/k0s.md
       - k0s api:            cli/k0s_api.md
       - k0s completion:     cli/k0s_completion.md
       - k0s default-config: cli/k0s_default-config.md
```
jasmingacic, line 36: let move default-config to the second place on the list, then api and completion. Or even better point CLI command reference to cli/README.md since it is already linking to appropriate sections.

jasmingacic (changes requested):

docs/k0s-in-docker.md
mviitane: This change was about aligning github/docker hub links. Previously 2/3 links were already updated, but this one forgotten. However, I could add a separate note about this github link as an alternative to docker hub.

docs/raspberry-pi4.md
mviitane: I'll update accordingly.

mkdocs.yml
mviitane: Do we want to break the alphabetical order? Now I think the commands are easy to find because of it.
I'll anyway update a link to README.md.

jasmingacic: I was thinking from the perspective of a user and what they might be using.

mviitane: To be fixed with a separate PR.

ncopa (approved):
LGTM

docs/containerd_config.md
```diff
@@ -1,4 +1,4 @@
-# containerd configuration
+# Containerd configuration
```
jasmingacic, line 1: official name is with lower case c `https://containerd.io/docs/`

docs/install.md
```diff
@@ -1,26 +1,24 @@
...
 * Supported services: OpenRC & Systemd
 
-### Server setup
+## Server setup
 This is the default mode of operation. When a server role is picked, the installer will do the following:
 
 * Create user accounts for the different components (see https://github.com/k0sproject/k0s/blob/main/pkg/apis/v1beta1/system.go#L6)
 * Create a service file (OpenRC/Systemd) and redirects logging to `/var/log/k0s.log`.
 * If the `--debug` flag is used, it will also pass this flag along to the service file.
 * `enable-worker` (single-node) setup is not supported. If you would like to run your service in that way, a possible solution would be to run `cmd install ` as worker, and edit the startup command by hand.
```
jasmingacic, line 16: enable-worker is supported `k0s install server --enable-worker`

docs/install.md
```diff
@@ -1,26 +1,24 @@
...
-
-This is an **alpha state** feature.
+Sub-command `k0s install` allows users to easily install k0s as a service, and define its logging.
 
 ## Caveats
 * This command is strictly a helper command. It is not meant to provide a fully-automated solution, since you can run k0s in multiple, very different ways.
 * It configures your service set-up as either a worker or a server, and will have different tasks, depending on the role you pick.
 * Supported services: OpenRC & Systemd
 
-### Server setup
```
jasmingacic, line 12: Perhaps adding an example here how to setup server node with install command
```
k0s install server --config cfg.yaml
```
and similar more details can be found here https://github.com/k0sproject/k0s/pull/594#issuecomment-754003983

docs/install.md
```diff
@@ -1,26 +1,24 @@
...
 
 * Create user accounts for the different components (see https://github.com/k0sproject/k0s/blob/main/pkg/apis/v1beta1/system.go#L6)
 * Create a service file (OpenRC/Systemd) and redirects logging to `/var/log/k0s.log`.
 * If the `--debug` flag is used, it will also pass this flag along to the service file.
 * `enable-worker` (single-node) setup is not supported. If you would like to run your service in that way, a possible solution would be to run `cmd install ` as worker, and edit the startup command by hand.
 
-### Worker Setup
+## Worker Setup
 * A worker cannot run with any other user, other than `root`, so no special users will be created.
 * The service file will include the `--token-file` flag, with a value that needs to be manually changed.
```
jasmingacic, line 20: This is not done manually anymore. You can find details at the bottom of this comment here https://github.com/k0sproject/k0s/pull/594#issuecomment-754003983

docs/install.md
jasmingacic: also there no need to edit service file

jasmingacic (changes requested):

docs/install.md
mviitane: This could be done with a separate PR. Probably a similar example should be provided for the worker setup?

docs/install.md
mviitane: Should be done with a separate PR.

docs/install.md
jasmingacic: Ok then we should reference this in an issue so we don't miss it.

docs/install.md
mviitane: Added a separate chapter for Single-node setup with this example.

docs/containerd_config.md
mviitane: changed

jasmingacic (approved):

PR #685, opened 2021-02-05
k0s terminology: server -> controller

### LINKED ISSUE #603: Decide on terminology for server/controller
The seems to be a mix'n'match of the terms `controller` and `server` used to describe the same thing. The documentation for example is talking about adding controllers and there's for example `k0s token create --role=controller` which then generates a token to be used for `kos install server` (#594).

trawler:

---

**Issue**
Fixes #603

**What this PR Includes**
This PR standardizes the different terminology used throughout the code base and documentations whereby the k0s controller was sometimes named "server" and sometimes "controller". Per the discussion in the attached issue, we've decided "controller" would work best.

**Notes**
- For backward compatibility, the older `k0s server` command is left intact, with a warning.
- To test the above, the single node integration test still use the `k0s server` command. When the command will be rendered deprecated, this test should be changed as well.

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -127,8 +127,8 @@ func (s K0sStatus) String() {
 
 }
 func getSysInit(role string) (sysInit string, stubFile string, err error) {
-	if role == "server+worker" {
-		role = "server"
+	if role == "controller+worker" {
+		role = "controller"
 	}
 	if sysInit, err = install.GetSysInit(); err != nil {
 		return sysInit, stubFile, err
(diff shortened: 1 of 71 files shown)


kke (commented):
Nice 😁 Will break k0sctl and any other integrations though unless there's some code that will make it accept either variant for a few versions forward..

trawler:
> Nice 😁 Will break k0sctl and any other integrations though unless there's some code that will make it accept either variant for a few versions forward..

Fair enough :) I left a stub of the old command, with a warning: https://github.com/k0sproject/k0s/blob/cfe65d3f0cc81377d07e5e1d097f4778666cf98d/cmd/controller.go#L97-L119

cmd/status.go
```diff
@@ -160,9 +160,9 @@ func getRole(pid int) (role string, err error) {
 	}
 	cmdln := string(raw)
 	if strings.Contains(cmdln, "enable-worker") {
-		return "server+worker", nil
-	} else if strings.Contains(cmdln, "server") {
-		return "server", nil
+		return "controller+worker", nil
```
jasmingacic, line 163: Just leaving this note here so we don't forget to refactor k0sctl since it is reading this value as server+worker

jasmingacic (approved):

trawler:
merged after rebase

PR #694, opened 2021-02-10
Fix for broken worker join in case of spec.api.address set to be ipv6 address

### LINKED ISSUE #688: k0s worker can't join a cluster with an API server address set to an IPv6 address
**Version**
v0.10.0

**Platform**
Debian 10 amd64

**What happened?**
When I set an IPv6 address for the API server, no workers will be able to join the cluster.

**How To Reproduce**
Find your public IPv6 address, you may be able to reproduce it with a private IPv6 address that starts with `fe80::` but I haven't tested on that case.

Generate a new default configuration:
```
k0s default-config > ./k0s.yaml
```

Set spec.api.address to an IPv6 address and add it so the san like this example:
```yml
spec:
  api:
    address: 2a03:b0c0:2:d0::f60:3001
    sans:
    - 2a03:b0c0:2:d0::f60:3001
```

Launch k0s server with a worker: `k0s server --enable-worker`.

You should now see no worker appearing with `kubectl get nodes`, no `kubelet` running and a very high CPU usage from the apiserver.

**Expected behavior**
To have k0s working normally like it would with an IPv4 address set for the API server address.

**Screenshots & Logs**
N/A

**Additional context**
N/A

soider:

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #688

**What this PR Includes**
    Change the way spec.network.api build API urls (use brackets in case of ipv6 address).
    Change order in CIDR definitions in case of dualstack
    Add ipv6 address to the apiserver certificate, if dualstack is enabled

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	k8s.io/cli-runtime v0.20.2
 	k8s.io/client-go v0.20.2
 	k8s.io/kubectl v0.20.2
+	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
 )
 
 // We need to force to a git commit of 3.4.13 release, see https://github.com/etcd-io/etcd/issues/12109
--- a/pkg/apis/v1beta1/cluster.go
+++ b/pkg/apis/v1beta1/cluster.go
@@ -17,11 +17,11 @@ package v1beta1
 
 import (
 	"fmt"
-	"io"
-	"io/ioutil"
-
 	"github.com/pkg/errors"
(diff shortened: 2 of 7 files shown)


soider:
It works under my setup (explicitly set address to be node public ipv6 address) though I am stil ltrying to figure out, why we use kube-apiserver `advertise-address` argument equals spec.api.address so the PR is marked as draft

jasmingacic (approved):

jnummelin (approved):

PR #695, opened 2021-02-11
fix the path order for k0s components

trawler:
Signed-off-by: Karen Almog <kalmog@mirantis.com>

**What this PR Includes**
supervisor components path is currently passed to the end of the PATH variable, which might cause external components to be used instead of k0s internal components. This PR fixes this.

PR #696, opened 2021-02-11
Fix order of PATH for binaries

ncopa:
Search the k0s path first so those binaries are preferred over system
binaries.

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''


ncopa:
@trawler was faster than me :rocket:

#695

PR #709, opened 2021-02-15
Make config parsing fail when reading invalid or non-existing config

### LINKED ISSUE #705: Failure in parsing the config will result default config to be used
**Version**
```
0.10.0
```

**What happened?**
When using an invalid or non-existing config, k0s commands will "revert" to use the default config in which e.g. the api addresses are autodetected. See #704 for example behaviour in token commands.

**How To Reproduce**
```
/ # cat k0s.yaml 
kind: Cluster
metadata:
  name: k0s
spec:
  api:
    address: 10.0.20.11
gf.bv,;)

/ # ./k0s token create -c /non/existing/k0s.yaml --role=worker | base64 -d | gunzip - | grep "server:"
    server: https://172.17.0.2:6443
```

**Expected behavior**
The user should be given a clear and concise error status that the config is invalid.

jnummelin:

**Issue**
Fixes #705 

**What this PR Includes**

Currently when k0s commands are reading a non-existing config or invalid config it'll, sometimes even silently, fall back to use default config. This PR makes the config parsing actually fail the commands properly.

### DIFF
--- a/cmd/helpers.go
+++ b/cmd/helpers.go
@@ -27,16 +27,17 @@ import (
 
 // ConfigFromYaml returns given k0s config or default config
 func ConfigFromYaml(cfgPath string) (clusterConfig *config.ClusterConfig, err error) {
-	if isInputFromPipe() {
+	if cfgPath == "" {
+		logrus.Info("no config file given, using defaults")
+		clusterConfig = config.DefaultClusterConfig()
+	} else if isInputFromPipe() {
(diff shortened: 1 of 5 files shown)


jasmingacic (approved):

jasmingacic (approved):
Essentially you removed unused bits of the code.

jnummelin:
> Essentially you removed unused bits of the code.

Mostly yes. Fun when a bug fix actually mainly just removes code :D

PR #715, opened 2021-02-17
Disable leader elector on single controllers setup (#540)

### LINKED ISSUE #540: Cannot override "leader-elect" on scheduler
For non-ha purposes it would be great if user could disable leader-elect completely. It burns quite a bit CPU which is completely unnecessary if you have only one controller.

ncopa:
---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #540

**What this PR Includes**

- disable leader elector when there is no `ExternalAddress` (single controller setup)
- refactor helm addon to re-use the controller.LeaderElector
- re-use controller.LeaderElector for the stack applier.

### DIFF
--- a/pkg/component/controller/dummyleaderelector.go
+++ b/pkg/component/controller/dummyleaderelector.go
@@ -1,0 +1,26 @@
+/*
+Copyright 2021 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controller
+
+type DummyLeaderElector struct {
+	Leader bool
+}
+
+func (l *DummyLeaderElector) Init() error    { return nil }
+func (l *DummyLeaderElector) Run() error     { return nil }
+func (l *DummyLeaderElector) Stop() error    { return nil }
+func (l *DummyLeaderElector) IsLeader() bool { return l.Leader }
+func (l *DummyLeaderElector) Healthy() error { return nil }
--- a/pkg/component/controller/dummyleaderelector.go
+++ b/pkg/component/controller/dummyleaderelector.go
@@ -16,11 +16,29 @@ limitations under the License.
 package controller
 
 type DummyLeaderElector struct {
-	Leader bool
+	Leader    bool
+	callbacks []func()
 }
 
 func (l *DummyLeaderElector) Init() error    { return nil }
-func (l *DummyLeaderElector) Run() error     { return nil }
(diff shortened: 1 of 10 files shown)


pkg/component/controller/dummyleaderelector.go
```diff
@@ -0,0 +1,24 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controller
+
+type DummyLeaderElector struct{}
```
jasmingacic, line 18: We already have `fakeAlwaysLeaderElector` here https://github.com/k0sproject/k0s/blob/main/pkg/component/controller/apiendpointreconciler_test.go#L30-L40
Maybe we should replace it with DummyLeaderElector?

ncopa: I didn't notice the `fake*LeaderElector`s until after i wrote the `DummyLeaderElector`. The test has both `fakeAlways...` and `fakeNever...` so I thought it was cleaner to keep the test as is.

But I think you are right. We should clean this up. How about we do `DummyLeaderElector` with a boolean variable?
```diff
diff --git a/cmd/controller.go b/cmd/controller.go
index 1e528d8..552b3e8 100644
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -226,7 +226,7 @@ func startController(token string) error {
        if clusterConfig.Spec.API.ExternalAddress != "" {
                leaderElector = controller.NewLeaderElector(clusterConfig, adminClientFactory)
        } else {
-               leaderElector = &controller.DummyLeaderElector{}
+               leaderElector = &controller.DummyLeaderElector{Leader: true}
        }
        componentManager.Add(leaderElector)
 
diff --git a/pkg/component/controller/dummyleaderelector.go b/pkg/component/controller/dummyleaderelector.go
index f83e329..a02426f 100644
--- a/pkg/component/controller/dummyleaderelector.go
+++ b/pkg/component/controller/dummyleaderelector.go
@@ -15,10 +15,12 @@ limitations under the License.
 */
 package controller
 
-type DummyLeaderElector struct{}
+type DummyLeaderElector struct {
+       Leader bool
+}
 
 func (l *DummyLeaderElector) Init() error    { return nil }
 func (l *DummyLeaderElector) Run() error     { return nil }
 func (l *DummyLeaderElector) Stop() error    { return nil }
-func (l *DummyLeaderElector) IsLeader() bool { return true }
+func (l *DummyLeaderElector) IsLeader() bool { return l.Leader }
 func (l *DummyLeaderElector) Healthy() error { return nil }
```

And maybe rename it to `FakeLeaderElector`? I kinda liked that better than `DummyLeaderElector`.

Then let the test use that instead.

jasmingacic: DummyLeaderElector sounds better

jnummelin (commented):
Good stuff.

I think we should also selectively disable the k8s leader electors:
- controller-manager: `--leader-elect=false`
- scheduler: `--leader-elect=false`

jasmingacic (approved):

jasmingacic (approved):

PR #722, opened 2021-02-19
Fix status command to pick up controller name

jnummelin:

**What this PR Includes**
Allows the status command to determine controller role for < 0.11 `k0s server` processes

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 Mirantis, Inc.
+Copyright 2021 k0s authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


cmd/status.go
```diff
@@ -165,6 +165,8 @@ func getRole(pid int) (role string, err error) {
 		return "controller", nil
 	} else if strings.Contains(cmdln, "worker") {
 		return "worker", nil
+	} else if strings.Contains(cmdln, "server") {
```
jasmingacic, line 168: Perhaps adding OR to the line 164 would make it nicer
```
} else if strings.Contains(cmdln, "controller") || else if strings.Contains(cmdln, "server") {
...
```

jnummelin: IMO this is better as separate branch as we're not going to keep it for long. Makes it stand out better so we remember to clean it after some time.

soider (approved):

jasmingacic (approved):
Tested it. Works as expected.

PR #725, opened 2021-02-19
K0s reset command - remove containers, namespaces and mounts

### LINKED ISSUE #278: Add k0s reset command to clean up an k0s worker
When the k0s process is killed on a worker, containerd-shim processes are left on the worker and have to be stopped manually.
To allow a cleanup of a worker node completely, we should add an `k0s reset` command, that would be similar to kubeadm reset:

> kubeadm reset is responsible for cleaning up a node local file system from files that were created using the kubeadm init or kubeadm join commands. For control-plane nodes reset also removes the local stacked etcd member of this node from the etcd cluster and also removes this node's information from the kubeadm ClusterStatus object. ClusterStatus is a kubeadm managed Kubernetes API object that holds a list of kube-apiserver endpoints.
kubeadm reset phase can be used to execute the separate phases of the above workflow. To skip a list of phases you can use the --skip-phases flag, which works in a similar way to the kubeadm join and kubeadm init phase runners.

https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/cmd/reset.go

trawler:
**Issue**
Fixes #278

**What this PR Includes**
This PR adds the `k0s reset` command for removing containers, network namespaces and kubelet mounts for k0s reset.
It also removes system init file, if exists.

It also moves service & user operations to the install package, to allow it to be re-used by other commands.


jasmingacic (approved):
After rebuilding k0s the reset command worked as expected
```
root@k0s:~/go/src/github.com/k0sproject/k0s# ./k0s reset
INFO[2021-02-24 10:45:25] Uninstalling the k0s service                 
ERRO failed to uninstall k0s service: failed to remove the k0s service: exit status 1 
INFO[2021-02-24 10:45:25] no config file given, using defaults         
INFO[2021-02-24 10:45:25] deleting user: etcd                          
userdel: user 'etcd' does not exist
INFO[2021-02-24 10:45:25] deleting user: kube-apiserver                
userdel: user 'kube-apiserver' does not exist
INFO[2021-02-24 10:45:25] deleting user: konnectivity-server           
userdel: user 'konnectivity-server' does not exist
INFO[2021-02-24 10:45:25] deleting user: kube-scheduler                
userdel: user 'kube-scheduler' does not exist
INFO failed to delete controller users: failed to run command "userdel" "etcd": exit status 6
failed to run command "userdel" "kube-apiserver": exit status 6
failed to run command "userdel" "konnectivity-server": exit status 6
failed to run command "userdel" "kube-scheduler": exit status 6 
INFO[2021-02-24 10:45:25] starting containerd for cleanup operations... 
INFO[2021-02-24 10:45:25] containerd succesfully started               
INFO[2021-02-24 10:45:25] attempting to clean up kubelet volumes...    
INFO[2021-02-24 10:45:25] successfully removed kubelet mounts!         
INFO[2021-02-24 10:45:25] attempting to clean up network namespaces... 
INFO[2021-02-24 10:45:25] successfully removed network namespaces!     
INFO[2021-02-24 10:45:25] attempting to stop containers...             
INFO[2021-02-24 10:45:32] successfully removed k0s containers!         
INFO[2021-02-24 10:45:32] deleting k0s generated data-dir (/var/lib/k0s) and run-dir (/run/k0s) 
ERRO[2021-02-24 10:45:32] k0s cleanup operations done. To ensure a full reset, a node reboot is recommended. 
```

pkg/install/cleanup.go
```diff
@@ -0,0 +1,116 @@
...
+func (c *CleanUpConfig) WorkerCleanup() error {
+	var msg []string
+
+	if err := c.workerPreFlightChecks(); err != nil {
+		logrus.Fatalf("failed clean up pre-flight-checks: %v", err)
+	}
+	logrus.Info("starting containerd for cleanup operations...")
+
+	quit := make(chan bool)
+	c.quit = quit
```
ncopa, line 47:
```suggestion
	c.quit := make(chan bool)
```

pkg/install/cleanup_unix.go
```diff
@@ -0,0 +1,209 @@
...
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (c *CleanUpConfig) stopContainerd() {
+	logrus.Debug("attempting to stop containerd")
+	c.quit <- true
```
ncopa, line 156: I think we technically should wait for the containerd process to exit and collect the exit status, or we'll have a containerd zombie process. I don't know if it matters though, since the uninstall is a short lived process.

pkg/install/process.go
```diff
@@ -0,0 +1,82 @@
...
+			pid := p.Pid()
+			ppid := p.PPid()
+			return &pid, &ppid, nil
+		}
+	}
+	return nil, nil, nil
+}
+
+func GetProcessOwner(pid int) (string, error) {
+	stdout, err := exec.Command("ps", "-o", "user=", "-p", strconv.Itoa(pid)).Output()
```
ncopa, line 46: Future improvement may parse `/proc/[pid]/status` to get the process owner instead of relying of external `ps` utility.

ncopa (commented):
good stuff!

pkg/install/process.go
jasmingacic: Why don't we use `github.com/mitchellh/go-ps` to get PID?

ncopa: > Why don't we use `github.com/mitchellh/go-ps` to get PID?

Does not seem to do what we need. We could use https://pkg.go.dev/github.com/prometheus/procfs#ProcStatus but I think its simpler to just parse it our selves than rely on external dependency in this case.

pkg/install/cleanup.go
trawler: > Suggested change 
> 	quit := make(chan bool)
> 	c.quit = quit
> 	c.quit := make(chan bool)

This doesn't actually work, because then the compiler complains that `c.quit evaluated but not used`

pkg/install/process.go
ncopa:
```go
func GetProcessOwner(pid int) (string, error) {
    content, err := os.ReadFile(fmt.Sprintf("/proc/%d/status", pid))
    if err != nil {
        return "", err 
    }
    for _, line := range strings.Split(string(content), "\n") {
        kv := strings.SplitN(line, ":", 2) 
        if kv[0] == "Uid" { 
            return strings.Fields(kv[1])[1], nil
        }
    }
    return "", errors.New("Uid not found in procfs")
}
```

trawler: Turns out this function is not needed. it is used to find out if "k0s status" is run as root, but this was easily mitigated by using `os.Geteuid()`

pkg/install/cleanup_unix.go
trawler: The function now uses os.SIGINT (https://golang.org/src/os/exec.go?s=2142:2233#L59) to stop containerd. If the process is still up, it will wait 3 seconds and will send SIGKILL.

embedded-bins/Makefile
```diff
@@ -5,6 +5,7 @@ kubernetes_version = 1.20.4
 kine_version = 0.6.0
 etcd_version = 3.4.14
 konnectivity_version = 0.0.14
+crictl_version = 1.20.0
```
soider, line 8: that gives us about +40 megabytes to the binary size:\ is it somehow possible to go without that tool? for example we can do the same unix socket requests or even use the tool but by importing code as a library

cmd/reset.go
```diff
@@ -0,0 +1,81 @@
+package cmd
```
ncopa, line 1: I think we need a copyright and licence boilerplate

pkg/install/cleanup_unix.go
```diff
@@ -0,0 +1,209 @@
+package install
```
ncopa, line 1: copyright and license boilerplate

pkg/install/cleanup_windows.go
```diff
@@ -0,0 +1,5 @@
+package install
```
ncopa, line 1: copyright and license boilerplate might be a good idea even if file is short

ncopa (dismissed):
I like that you got rid of the GetProcessOwner :+1:

embedded-bins/Makefile
ncopa: I agree, but I think this is good enough for this release.

trawler: I agree with both: For future releases we should aim to remove the dependency in an additional binary.

PR #727, opened 2021-02-19
Add ability to pass extra args for kubelet process

### LINKED ISSUE #643: Add "extraArgs" for kubelet
On a single-node cluster, I would like to start everything on a different networking interface, say, 192.168.0.1. I can get everything listening on only 192.168.0.1 just fine, including the `address` config field for `kubelet-config.yaml` in the `workerProfiles` section of `k0s.yaml`. This successfully causes kubelet to listen only on 192.168.0.1, but the kubelet still telling the cluster that its `INTERNAL-IP` is that of the external networking interface. Without the correct `INTERNAL-IP`, one cannot get logs or interact with running pods.

This issue is typically solved with the kubelet arg `--node-ip=192.168.0.1`, but in k0s I don't believe there is a way to pass args down to the kubelet process.

An `extraArgs` field in `k0s.yaml` for kubelet similar to the one for the control plane processes.

Attempted to use the `address` field for `kubelet-config.yaml`, but this doesn't affect what the advertised `INTERNAL-IP` is.

### LINKED ISSUE #707: Add ability to modify the InternalIP of a node
Currently, except by defining the argument `--node-ip` in a worker profile for kubelet it's not possible to modify the internal IP of a node.

I would like to be able to set the internal IP using the argument `--node-ip`.

Defining the `--node-ip` argument in the worker profiles:
```yaml
workerProfiles:
    - name: custom-role
      values:
         node-ip: 9.9.9.9
```

And then launch the worker with this custom profile.

**Additional context**
Requested in https://github.com/k0sproject/k0s/issues/492#issuecomment-778758460

jnummelin:

**Issue**
Fixes #707
Fixes #643 

**What this PR Includes**
This PR allows one to pass any set of extra args to kubelet:
```
k0s worker --kubelet-extra-args="--some-arg=foo --bool-flag --some-other=bar"
```

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -270,7 +270,7 @@ func (s *FootlooseSuite) GetJoinToken(role string, dataDir string) (string, erro
 }
 
 // RunWorkers joins all the workers to the cluster
-func (s *FootlooseSuite) RunWorkers(dataDir string) error {
+func (s *FootlooseSuite) RunWorkers(dataDir string, args ...string) error {
 	ssh, err := s.SSH("controller0")
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


inttest/common/footloosesuite.go
```diff
@@ -286,9 +286,9 @@ func (s *FootlooseSuite) RunWorkers(dataDir string) error {
 
 	var workerCommand string
 	if dataDir != "" {
-		workerCommand = fmt.Sprintf(`nohup k0s --debug --data-dir=%s worker --labels="k0sproject.io/foo=bar" "%s" - >/tmp/k0s-worker.log 2>&1 &`, dataDir, token)
+		workerCommand = fmt.Sprintf(`nohup k0s --debug --data-dir=%s worker --labels="k0sproject.io/foo=bar" --kubelet-extra-args="--address=0.0.0.0" "%s" - >/tmp/k0s-worker.log 2>&1 &`, dataDir, token)
 	} else {
-		workerCommand = fmt.Sprintf(`nohup k0s --debug worker --labels="k0sproject.io/foo=bar" "%s" >/tmp/k0s-worker.log 2>&1 &`, token)
+		workerCommand = fmt.Sprintf(`nohup k0s --debug worker --labels="k0sproject.io/foo=bar" --kubelet-extra-args="--address=0.0.0.0" "%s" >/tmp/k0s-worker.log 2>&1 &`, token)
```
ncopa, line 291: I think we should refactor `RunWorkers` to take a list of args instead of adding those args all tests.

Similar to what we did in df25442e1479e02dc59bfce62ac84f10befd7294

Can/should be done as separate issue though.

jnummelin: Yup, shall be refactored. I'll actually do it in this PR, better to do it sooner rather than later.

Just wanted to push early for general "design level" discussions plus to get the smokes running early

jnummelin: @ncopa I refactored this to be more flexible, PTAL

ncopa (approved):

PR #737, opened 2021-02-25
Add option --single for single node

### LINKED ISSUE #720: add option for --single
There are some processes running that are not needed when running `k0s` as a mini cluster on local laptop, like konnectivity.

Would be nice with a `--single` option that tells `k0s` that we intend to not expand the cluster with more controllers and that we want run the worker locally.

config file option?

**Additional context**

I think `--single` could:
- enable the `--enable-worker`
- use kine as default backend as default instead of etcd
- disable konnectivity
- disable the leader election
- anything else?

ncopa:
disables konnectivity, and implies --enable-worker

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #720

### DIFF
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -92,6 +92,7 @@ var (
 			}
 			if singleNode {
 				enableWorker = true
+				k0sVars.DefaultStorageType = "kine"
 			}
 
 			return startController(controllerToken)
@@ -157,7 +158,7 @@ func startController(token string) error {
 	var storageBackend component.Component
 
 	switch clusterConfig.Spec.Storage.Type {
-	case v1beta1.KineStorageType, "":
+	case v1beta1.KineStorageType:
 		storageBackend = &controller.Kine{
 			Config:  clusterConfig.Spec.Storage.Kine,
 			K0sVars: k0sVars,
--- a/cmd/controller.go
+++ b/cmd/controller.go
@@ -224,10 +224,12 @@ func startController(token string) error {
 	componentManager.Add(leaderElector)
 
 	componentManager.Add(&applier.Manager{K0sVars: k0sVars, KubeClientFactory: adminClientFactory, LeaderElector: leaderElector})
-	componentManager.Add(&controller.K0SControlAPI{
-		ConfigPath: cfgFile,
-		K0sVars:    k0sVars,
-	})
+	if !singleNode {
+		componentManager.Add(&controller.K0SControlAPI{
+			ConfigPath: cfgFile,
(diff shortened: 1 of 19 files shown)


soider:
May be in a single node set up we could also automatically use sqlite storage instead of etcd, to save some local machine resources?

soider (approved):

jnummelin (changes requested):
If we do not start konnectivity, we also need to remove the konnectivity related configs from kube-apiserver

cmd/controller.go
```diff
@@ -51,6 +51,7 @@ import (
 func init() {
 	controllerCmd.Flags().StringVar(&controllerWorkerProfile, "profile", "default", "worker profile to use on the node")
 	controllerCmd.Flags().BoolVar(&enableWorker, "enable-worker", false, "enable worker (default false)")
+	controllerCmd.Flags().BoolVar(&single, "single", false, "enable single node (implies --enable-worker, default false)")
```
trawler, line 54: Shouldn't we then also remove the `--enable-worker` flag? What happens is `--enable-worker` is true, but `--single` is not?

I would also suggest to use `--enable-worker` as an alias to `--single`, to avoid breakage. See: https://github.com/k0sproject/k0s/blob/main/cmd/installController.go#L30

trawler (changes requested):

ncopa:
> May be in a single node set up we could also automatically use sqlite storage instead of etcd, to save some local machine resources?

yes, absolutely, but solving that is trickier that it may seem.

ncopa:
also join api can be disabled

cmd/controller.go
soider: I think it's not exactly the same. We can run few controllers with `enable-worker` to work the same way as vanilla k8s clusters, though I don't know if there are any practical reasons to do it

trawler: I agree they are the same, that's why I don't see the reason in keeping both flags.

ncopa: We cannot remove `--enable-worker` flag. Without this flag you cannot expand to multiples nodes which acts as both controllers and workers. (think 3 nodes which acts as both controllers *and* workers)

The `--single` flag disables the things that makes it possible to expand the control plane with more controllers.

jnummelin (approved):

jasmingacic (dismissed):

trawler (approved):
lgtm 👍 (but we should also update our single node guide appropriately)

trawler (approved):

PR #733, opened 2021-02-24
k0s docs update

jasmingacic:
k0s install instructions and brief example how to deploy k0s with k0sctl and how to upgrade the cluster

Closes #691

PR #745, opened 2021-03-02
Documentation improvements

### LINKED ISSUE #691: Document k0sctl in k0s docs
k0sctl is not documented in k0s documentation.

We need to document k0sctl and define its scope in https://docs.k0sproject.io
Describe k0sctl in the multi-node documentation part, explain its uses and why should a user elect to use it vs vanilla installation.
We should keep the documentation on setting up a multi node cluster manually, and also describe the reasons for doing so.

jasmingacic:
This is the same thing as #733 which I closed as it had almos 200+ which were hard track.

Closes #691

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,8 +1,30 @@
+- [Configuration options](#configuration-options)
+    - [Configuration Validation](#configuration-validation)
+  - [Configuration Keys](#configuration-keys)
+    - [`spec.api`](#specapi)
+    - [`spec.network`](#specnetwork)
+    - [`extensions.helm`](#extensionshelm)
+  - [Configuring an HA Control Plane](#configuring-an-ha-control-plane)
+    - [Requirements](#requirements)
+        - [Load Balancer](#load-balancer)
+        - [Cluster configuration](#cluster-configuration)
+  - [Full config reference](#full-config-reference)
+    - [`spec.api`](#specapi-1)
+    - [`spec.controllerManager`](#speccontrollermanager)
+    - [`spec.scheduler`](#specscheduler)
+    - [`spec.storage`](#specstorage)
+    - [`spec.network`](#specnetwork-1)
+      - [`spec.network.calico`](#specnetworkcalico)
+    - [`spec.podSecurityPolicy`](#specpodsecuritypolicy)
+    - [`spec.workerProfiles`](#specworkerprofiles)
+    - [`images`](#images)
+    - [Extensions](#extensions)
+    - [Telemetry](#telemetry)
+
+
 # Configuration options
 
-## Control plane
-
-k0s Control plane can be configured via a YAML config file. By default `k0s server` command reads a file called `k0s.yaml` but can be told to read any yaml file via `--config` option.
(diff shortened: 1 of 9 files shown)


docs/configuration.md
```diff
@@ -31,6 +53,24 @@ extensions:
...
+2. [SANs addresses](#specapi-1)
+3. [Network providers](#specnetwork-1)
+4. [Worker profiles](#specworkerprofiles) 
+
+
+## Configuration Keys
+
 ### `spec.api`
 
 - `address`: The local address to bind API on. Also used as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.
```
trawler, line 76: we still have two sections describing spec.api, spec.network and helm extensions. one under the "configuration keys" and also under "full configuration".
It's not clear why we have both, so I would merge the configuration keys and the full configuration sections (and rename "full configuration" to "configuration keys")

docs/k0s-single-node.md
```diff
@@ -1,53 +1,27 @@
...
 ```sh
-mkdir -p ${HOME}/.k0s
+k0s default-config > k0s.yaml
 ```
 
-#### 2. Generate a default cluster configuration
+#### 2. Start k0s
 ```sh
-k0s default-config | tee ${HOME}/.k0s/k0s.yaml
+$ sudo k0s install controller -c k0s.yaml --enable-worker
```
trawler, line 18: question: would the systemd service be able to resolve the full path of k0s.yaml file from this command?

trawler (changes requested):
Please run the mkdocs test process detailed in  https://github.com/k0sproject/k0s/blob/main/docs/internal/publishing_docs_using_mkdocs.md#testing-docs-locally to pick up on errors such as:
```
k0s-docs_1  | WARNING -  Documentation file 'README.md' contains a link to 'create-cluster.md' which is not found in the documentation files.
k0s-docs_1  | WARNING -  Documentation file 'k0s-multi-node.md' contains a link to 'k0s-install.md' which is not found in the documentation files.
k0s-docs_1  | WARNING -  Documentation file 'k0s-single-node.md' contains a link to 'k0s-install.md' which is not found in the documentation files.
```

docs/k0s-single-node.md
jasmingacic: No it won't 
```
[Service]
StartLimitInterval=5
StartLimitBurst=10
ExecStart=/root/go/src/github.com/k0sproject/k0s/k0s controller --config=val.yaml
```

good catch

docs/configuration.md
jasmingacic: I think it would be better to remove those examples and just add configurations specs instead of this

docs/docker-compose.yml
```diff
@@ -5,4 +5,4 @@ services:
     volumes:
       - ../:/docs
     ports:
-      - 80:8000
\ No newline at end of file
+      - 8081:8000
```
trawler, line 8: Please remove this change from the commit

trawler (changes requested):

trawler (approved):

PR #750, opened 2021-03-03
convert relative file path to absolute

### LINKED ISSUE #749: k0s install: relative paths from options should be converted to absolute
When path to files with any of install subcommands  doesn't file path doesn't get parsed in service stubs.

```
k0s install controller -c k0s.yaml
```

```
ExecStart=/root/go/src/github.com/k0sproject/k0s/k0s controller --config=val.yaml
```

Which when running will not get resolved.

Also there was a mention of the same issue here: `https://miracloud.slack.com/archives/C01CQPVUELA/p1614696996086800`

jasmingacic:
When CLI parameters are converted to `--key=value` format,  file paths need to be absolute. 
This PR converts relative file path for parameters `--data-dir` `--config` and `--token-file` 
Closes #749

### DIFF
--- a/cmd/install.go
+++ b/cmd/install.go
@@ -16,11 +16,13 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"github.com/k0sproject/k0s/internal/util"
 	"github.com/k0sproject/k0s/pkg/install"
 )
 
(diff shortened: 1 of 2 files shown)


soider (dismissed):

cmd/install.go
```diff
@@ -81,22 +101,64 @@ Windows flags like "--api-server", "--cidr-range" and "--cluster-dns" will be ig
...
+		}
+		if !util.FileExists(tokenFile) {
+			return fmt.Errorf("%s does not exist", tokenFile)
+		}
 	}
+
 	return nil
 }
+
+func isDirectory(path string) bool {
```
jnummelin: We've got this already available in utils, re-use from there.

cmd/install.go
```diff
@@ -81,22 +101,64 @@ Windows flags like "--api-server", "--cidr-range" and "--cluster-dns" will be ig
...
 
 	err := install.EnsureService(args)
 	if err != nil {
-		logrus.Errorf("failed to install k0s service: %v", err)
+		return fmt.Errorf("failed to install k0s service: %v", err)
+	}
+	return nil
+}
+
+func convertParamsToAbsolute() (err error) {
```
jnummelin, line 131:
```suggestion
func convertFileParamsToAbsolute() (err error) {
```

jnummelin (changes requested):

jnummelin (approved):

PR #784, opened 2021-03-18
Refactoring of cobra cmd commands

### LINKED ISSUE #775: k0s embedded 'kubectl' doesn't work for all subcommands
**Version**
```
$ k0s version
v0.11.0
```

**Platform**
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Release:	20.04
Codename:	focal
```

**What happened?**

Some of the subcommands are working fine with the embedded 'kubectl', but some are not.
```
$ sudo k0s kubectl get pods -A
... everything works fine
````

But then:
```
$ sudo k0s kubectl create namespace nginx3
ERRO[2021-03-11 11:20:47] cannot read admin kubeconfig at , is the server running? 
Error: open : no such file or directory
````

**How To Reproduce**
````
$ sudo k0s kubectl create namespace nginx3
````

**Expected behavior**
````
$ kubectl create namespace nginx3
namespace/nginx3 created
````

### LINKED ISSUE #791: k0s prints usage info on any error
We should only print the usage help text when there is a problem with the args, not on every error.

For example:
```
$ ./k0s etcd leave
INFO[2021-03-18 17:23:12] no config file given, using defaults         
Error: can't connect to the etcd: open /var/lib/k0s/pki/apiserver-etcd-client.key: permission denied
Usage:
  k0s etcd leave [flags]

Flags:
  -h, --help                  help for leave
      --peer-address string   etcd peer address

Global Flags:
      --data-dir string        Data Directory for k0s (default: /var/lib/k0s). DO NOT CHANGE for an existing setup, things will break!
      --debugListenOn string   Http listenOn for debug pprof handler (default ":6060")
```

In this case the command line is not wrong, but it fails due to a permission problem. We should not print the usage text here as it hides the real problem.

### PR DESCRIPTION
**Pull request by trawler**
---

**Issue**
* Fixes #775
* Fixes #791

**What this PR Includes**
This PR fixes a problem where the `kubectl` was unable to resolve the correct kubeconfig path for k0s
To resolve this issue, a refactoring of the cmd structure needed to be made.
Also in this PR, cmd.SilenceUsage set to True, so that we don't get a usage printout for every error.

### CODE REVIEW

**Comment by soider** [comment]
Files with name `main.go` confuse me quiet a lot, because `main.go` usually name for a separate binary.
What do you think about may be renaming them to `command.go` or anything else but not `main.go`?

**Review by jasmingacic — APPROVED** [review]

**Comment by trawler** [comment]
> Files with name `main.go` confuse me quiet a lot, because `main.go` usually name for a separate binary.
> What do you think about may be renaming them to `command.go` or anything else but not `main.go`?

I haven't found any supporting documentation that states that main.go should belong to a separate binary in any coding guidelines. Do you have any links that confirm this? I find that using "command.go" would be even more confusing to have, personally.

File: cmd/airgap/listImages.go [hunk 1]
```diff
@@ -13,35 +13,30 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
```
soider, line 15: I think it's better have underscore instead of camel case in filename. camelcase could led to problems on case insensitive fs (like macos)

trawler:
tested also with k0sctl ✅

PR #786, opened 2021-03-18
merge token cmd files to single file

trawler:

---

**Issue**
Placing subcommands of the same command under separate files makes it more annoying to debug. This PR is simply re-organizing the token subcommands files into a single token command file.


trawler:
closing, in order to address as part of #784

PR #793, opened 2021-03-18
Fix exitcode on failure (#790)

### LINKED ISSUE #790: k0s always exit status 0, even on failure
k0s should not return exit status 0 (aka success) on errors.

example when there are no running controller:
```
$ ./k0s token create --role=worker ; echo $?
Error: failed to load kubeconfig: error loading config file "/var/lib/k0s/pki/admin.conf": open /var/lib/k0s/pki/admin.conf: permission denied
....
0
```

Other example:
```
$ ./k0s etcd leave ; echo $?
INFO[2021-03-18 17:23:12] no config file given, using defaults         
Error: can't connect to the etcd: open /var/lib/k0s/pki/apiserver-etcd-client.key: permission denied
Usage:
  k0s etcd leave [flags]

Flags:
  -h, --help                  help for leave
      --peer-address string   etcd peer address

Global Flags:
      --data-dir string        Data Directory for k0s (default: /var/lib/k0s). DO NOT CHANGE for an existing setup, things will break!
      --debugListenOn string   Http listenOn for debug pprof handler (default ":6060")

0
```

In both cases it returns 0, which means success. We should return non-zero in those cases.

ncopa:
Make sure we exit with error when a command fails

Fix regression introduced with commit d801eb1b768f (convert relative
file path to absolute)

**Issue**
Fixes #790

**What this PR Includes**
I believe this should fix some if not all of the flakiness we have seen in the smoke tests recently

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -148,7 +148,8 @@ func addPersistentFlags(cmd *cobra.Command) {
 }
 
 func Execute() {
-	// just a hack to trick linter which requires to check for errors
-	// cobra itself already prints out all errors that happen in subcommands
-	_ = rootCmd.Execute()
+	err := rootCmd.Execute()
+	if err != nil {
+		log.Fatal(err)
+	}
 }


ncopa:
I added this to PR #788 instead

PR #796, opened 2021-03-19
Simplify etcd health check by trying to get a value from the kv store

jnummelin:

**Issue**
The current etcd health check mechanism is bit complex and had couple flaws:
- always got the peers and used the first peer for the health check
- ignored http response codes != 200

**What this PR Includes**
This PR simplifies the health check by trying to get a value from the KV store. This is also done in the official etcdctl tool with command `etcdctl endpoint health`, see https://github.com/etcd-io/etcd/blob/3ead91ca3edf66112d56c453169343515bba71c3/etcdctl/ctlv3/command/ep_command.go#L124-L135

### DIFF
--- a/cmd/etcd.go
+++ b/cmd/etcd.go
@@ -18,6 +18,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


ncopa (approved):
LGTM

jasmingacic (approved):

PR #797, opened 2021-03-19
Make etcd use jwt tokens

### LINKED ISSUE #621: Etcd "simple token is not cryptographically signed"
**Version**
```
$ k0s version
```

> v0.8.1
**Platform**
```
$ lsb_release -a
```

> Distributor ID:	Ubuntu
> Description:	Ubuntu 20.04.1 LTS
> Release:	20.04

**What happened?**

When booting a completely vanilla new cluster I see this line:

> INFO\[2020-12-11 21:15:02\] 2020-12-11 21:15:02.034769 W | auth: simple token is not cryptographically signed component=etcd

This seems to indicate potentially that we are using "simple" token authentication.  I believe the ETCD cluster does spin up a certificate authority and such and so has the intention of auth/authz.

https://etcd.io/docs/v3.2.17/learning/auth_design/

> Two types of tokens: simple and JWT
> There are two kinds of token types: simple and JWT. **The simple token isn’t designed for production use cases.** Its tokens aren’t cryptographically signed and servers must statefully track token-user correspondence; it is meant for development testing. JWT tokens should be used for production deployments since it is cryptographically signed and verified. From the implementation perspective, JWT is stateless. Its token can include metadata including username and revision, so servers don’t need to remember correspondence between tokens and the metadata.

**How To Reproduce**

> k0s server

**Expected behavior**
Use JWT?

**Additional context**
https://auth0.com/blog/json-web-token-signing-algorithms-overview/
https://github.com/etcd-io/etcd/issues/9376
https://etcd.io/docs/v3.2.17/learning/auth_design/

jnummelin:

**Issue**
Fixes #621

**What this PR Includes**
Etcd will now be configured with jwt auth. The JWT sign key pair is created only for fresh setup and thus existing setup will be un-touched.

Includes also some refactorings, mainly for some generic utility re-locations.

### DIFF
--- a/internal/util/file.go
+++ b/internal/util/file.go
@@ -60,3 +60,18 @@ func GetExecPath(fileName string) (*string, error) {
 
 	return &path, nil
 }
+
+func ChownFile(file, owner string, permissions os.FileMode) error {
+	// Chown the file properly for the owner
+	uid, _ := GetUID(owner)
+	err := os.Chown(file, uid, -1)
+	if err != nil && os.Geteuid() == 0 {
+		return err
+	}
+	err = os.Chmod(file, permissions)
+	if err != nil && os.Geteuid() == 0 {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -16,6 +16,10 @@ limitations under the License.
 package certificate
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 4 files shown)


internal/util/file.go
```diff
@@ -60,3 +60,18 @@ func GetExecPath(fileName string) (*string, error) {
 
 	return &path, nil
 }
+
+func ChownFile(file, owner string, permissions os.FileMode) error {
+	// Chown the file properly for the owner
+	uid, _ := GetUID(owner)
+	err := os.Chown(file, uid, -1)
+	if err != nil && os.Geteuid() == 0 {
```
s0j, line 68: On Windows `os.Geteuid()` will always return `-1`, and this function will pass.  Is this a problem?

jnummelin: > On Windows os.Geteuid() will always return -1, and this function will pass. Is this a problem?

_should_ not be as this is used mainly for setting cert file permissions which we do not do on windows

PR #800, opened 2021-03-23
Cleanup/refactor footloose inttests

ncopa:
**What this PR Includes**
Cleanup and refactor the footloose integration tests.

- de-duplicate the copy/paste code
- use variadic args instead of always pass a dataDir option (so it is optional)
- add a regression test for #790 
- disconnect ssh when it should to avoid dangling ssh connections
- dont use hardcoded node names


soider (approved):
good stuff

PR #804, opened 2021-03-25
Issue/703/advertise different ports for components

### LINKED ISSUE #703: Add ability to advertise different ports for kube-apiserver, konnectivity and join API
Currently, the ports for kube-apiserver, konnectivity and join API are "hardcoded" and it's not possible to change them.
This complicates the setup of the control plane on Linux servers or load balancers where the public port range is limited and don't allow 6443 as a public exposed port for example.
A real world example is the fly.io load balancer that only allow port ranges 80, 443, 5000, and ports 10000-10100: https://fly.io/docs/reference/configuration/#services-ports

I would like the ability to change the advertised ports for the components kube-apiserver, konnectivity and join API.
On k3s for example it's possible to change the advertised port for the kube-apiserver with `--advertise-port`: https://rancher.com/docs/k3s/latest/en/installation/install-options/server-config/

I tried to force kube-apiserver to another port with:
```yaml
spec:
  api:
    extraArgs:
      secure-port: 5000
```
But that broke the other components.

**Additional context**
N/A

soider:
**Issue**
Fixes #703

**What this PR Includes**
The PR includes changes in the cluster spec. With the changes introduced it's now possible to configure custom ports for kube-api, k0s-api, konnectivity-admin and konnectivity-agent.
New set of smoke tests are introduced with simple external load balancer to verify that custom ports mapping works with load-balancer.


pkg/apis/v1beta1/cluster_test.go
```diff
@@ -136,6 +136,10 @@ spec:
 	assert.Equal(t, "unsupported network provider: invalidProvider", errors[0].Error())
 }
 
+func TestApiAddress(t *testing.T) {
+
+}
```
jasmingacic, line 141: If it is empty, maybe remove it

jasmingacic (approved):

inttest/common/footloosesuite.go
```diff
@@ -67,23 +73,27 @@ func (s *FootlooseSuite) SetupSuite() {
...
+
 	cluster, err := cluster.New(s.footlooseConfig)
+
 	if err != nil {
 		s.T().Logf("ERROR: failed to load footloose config: %s", err.Error())
 		s.T().FailNow()
 		return
 	}
 
+	// XXX: why do we create, delete and create again?
```
jnummelin, line 85: If there happens to be some leftover container from previous runs of the same suite lingering around

pkg/apis/v1beta1/api.go
```diff
@@ -27,6 +27,8 @@ import (
 // APISpec ...
 type APISpec struct {
 	Address         string            `yaml:"address"`
+	Port            int               `yaml:"port"`
+	K0sAPIPort      int               `yaml:"k0s_api_port,omitempty"`
```
jnummelin, line 31: hmm, I'm not a huge fan of the underscored yaml key "scheme". Why not use camel casing as with all other multi-word keys?

inttest/common/footloosesuite.go
soider: I changed it to be a explanation why do we do it

pkg/apis/v1beta1/api.go
soider: changed

inttest/customports/customports_test.go
```diff
@@ -0,0 +1,132 @@
...
+  name: k0s
+spec:
+  api:
+    externalAddress: %s
+    port: %d
+    k0sApiPort: %d
+  konnectivity:
+    agent_port: %d
+    admin_port: %d
+`
```
jnummelin, line 48: Wouldn't it be easier / more manageable to have this as a true text template?

inttest/customports/customports_test.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (ds *Suite) TestControllerJoinsWithCustomPort() {
+
+	ipAddress := ds.getMainIPAddress()
+	ds.T().Logf("ip address: %s", ipAddress)
+
+	ds.putFile("controller0", "/tmp/k0s.yaml", fmt.Sprintf(configWithExternaladdress, ipAddress, APIPort, k0sAPIPort, agentPort, adminPort))
+	ds.putFile("controller1", "/tmp/k0s.yaml", fmt.Sprintf(configWithExternaladdress, ipAddress, APIPort, k0sAPIPort, agentPort, adminPort))
+	ds.putFile("controller2", "/tmp/k0s.yaml", fmt.Sprintf(configWithExternaladdress, ipAddress, APIPort, k0sAPIPort, agentPort, adminPort))
```
jnummelin, line 97: Maybe do the "templating" only once and re-use results for all controllers

inttest/customports/customports_test.go
```diff
@@ -0,0 +1,132 @@
...
+	ds.putFile("worker0", "/tmp/k0s.yaml", fmt.Sprintf(configWithExternaladdress, ipAddress, APIPort, k0sAPIPort, agentPort, adminPort))
```
jnummelin, line 98: Why do we push the config yaml into workers?

jnummelin (changes requested):
Pretty good stuff. 👍 

Added few minor remarks

jnummelin (approved):
LGTM

PR #826, opened 2021-04-06
Use fmt.Errorf everywhere for error handling

### LINKED ISSUE #227: Migrate from github.com/pkg/errors to the standard library error handling
We use github.com/pkg/errors as main method to wrap errors.
Given that the library is in the maintenance only mode and for >1.13 versions it is a wrapper around stdlib, we could switch to the native library from standard library.

ncopa:
**Issue**
Fixes #227

**What this PR Includes**
Replace use of github.com/pkg/errors with standard fmt.Errorf

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	"github.com/avast/retry-go"
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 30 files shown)


jasmingacic (approved):

cmd/controller/controller.go
```diff
@@ -170,7 +169,7 @@ func (c *CmdOpts) startController() error {
 			LogLevel:    c.Logging["etcd"],
 		}
 	default:
-		return errors.New(fmt.Sprintf("Invalid storage type: %s", c.ClusterConfig.Spec.Storage.Type))
+		return fmt.Errorf("Invalid storage type: %s", c.ClusterConfig.Spec.Storage.Type)
```
soider, line 172: I don't actually understand how the linter pass that, the errors should start with lowercase :thinking:

jnummelin (approved):

PR #832, opened 2021-04-09
Build airgap bundle during release

### LINKED ISSUE #785: Build default bundle for the airgap install
With #755 is landing to the main branch we'd have a way to install k0s cluster at the airgaped environment

To perform such installation end-user need to have OCI bundle of images. There is guide available in the documentation on creating such bundle, but would be beneficial to the UX to have such bundle available on the release page.

One way is to modify release workflow to export OCI images from the cluster we setup during the e2e tests in the release workflow.

We can also set up completely separate workflow which would build bundle by using `docker images save` command

soider:
**Issue**
Fixes #785

**What this PR Includes**
GHA release.yaml changes to build bundle for airgap install during the release process for all supported architectures.

Also, includes PR #831


jasmingacic (approved):

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/onsi/gomega v1.10.2 // indirect
 	github.com/opencontainers/selinux v1.8.0 // indirect
 	github.com/pkg/errors v0.9.1
+	github.com/pkg/sftp v1.13.0 // indirect
```
jnummelin, line 40: I don't think the smokes are using ftp anymore?

inttest/footloose-alpine/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM alpine:3.13
 
-RUN apk add openrc openssh-server bash busybox-initscripts coreutils findutils iptables curl haproxy
+RUN apk add openrc openssh-server bash busybox-initscripts coreutils findutils iptables curl haproxy openssh-client
```
jnummelin, line 3: Do we really need the ssh-client packages?

soider: not anymore, I used it before I add volumes

.github/workflows/go.yml
```diff
@@ -455,6 +455,59 @@ jobs:
...
+
+    steps:
+      - name: Get PR Reference and Set Cache Name
+        run: |
+          PR_NUMBER=$(echo ${GITHUB_REF} | cut -d / -f 3 )
+          echo "cachePrefix=k0s-${PR_NUMBER}-${{ github.sha }}" >> $GITHUB_ENV
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Bindata
```
jnummelin, line 471: Why do we need the embedded-bins cache for smoke? We should already have the full k0s bin pre-build and restored from cache, right?

jnummelin (approved):
One minor comment, otherwise LGTM

jnummelin (approved):

jnummelin (approved):

PR #834, opened 2021-04-12
validate config

### LINKED ISSUE #774: 'k0s validate config' could detect mistyped/unsupported configuration parameters

When I mistype a configuration parameter name or value (in k0s.yaml), I don't easily find the problem. For example:
```
installConfig:
  users:
    etcdUserWRONG: etcd
````
or
```
spec:
  api:
    address: 172.WRONG.x.x
    addressWRONG: 172.x.x.x
    sansWRONG:
    - 172.x.x.x
  storageWRONG:
    type: etcd
    etcd:
      peerAddress: 172.x.x.x
  network:
    podCIDR__WRONG: 10.x.0.0/16
````

'k0s validate config' could help detecting problems like this. Also when starting a controller, the faulty configuration could be detected before actually launching the controller.

### PR DESCRIPTION
**Pull request by jasmingacic**
Closes #774 

There is still the question of semantic validation for each spec property, at least this will enforce that each spec implements Validate function

### DIFF
--- a/pkg/apis/v1beta1/api.go
+++ b/pkg/apis/v1beta1/api.go
@@ -19,10 +19,11 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/k0sproject/k0s/internal/util"
-
 	"github.com/asaskevich/govalidator"
+	"github.com/k0sproject/k0s/internal/util"
 )
+
+var _ Validateable = (*APISpec)(nil)
 
 // APISpec ...
 type APISpec struct {
@@ -107,6 +108,10 @@ func (a *APISpec) Validate() []error {
 			continue
 		}
 		errors = append(errors, fmt.Errorf("%s is not a valid address for sans", a))
+	}
+
+	if !govalidator.IsIP(a.Address) {
+		errors = append(errors, fmt.Errorf("spec.api.address: %q is not IP address", a.Address))
 	}
 
 	return errors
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: pkg/apis/v1beta1/cluster.go [hunk 1]
```diff
@@ -74,24 +82,51 @@ func (s *SchedulerSpec) IsZero() bool {
...
 	SystemUsers *SystemUser `yaml:"users,omitempty"`
 }
 
+// Validate stub for Validator interface
+func (i *InstallSpec) Validate() []error {
+	return nil
+}
+
+// Validator interface to ensure that all config components implement Validate function
+type Validator interface {
```
jnummelin, line 102: hmm, I'd maybe call this something like `Validateable` as this ensures we can validate such an interface

pkg/apis/v1beta1/api.go
```diff
@@ -109,5 +111,9 @@ func (a *APISpec) Validate() []error {
 		errors = append(errors, fmt.Errorf("%s is not a valid address for sans", a))
 	}
 
+	if !govalidator.IsIP(a.Address) {
+		errors = append(errors, fmt.Errorf("%v is not IP address", a.Address))
```
jnummelin, line 115: Let the user also know which field it was that failed. Now they only see that some address was not in correct form but have (maybe) no idea which field it was.

jnummelin (approved):

PR #835, opened 2021-04-12
K0s backup

### LINKED ISSUE #204: Cluster backup
We need an easy way to create backups from the cluster and naturally a way to restore a backup.

trawler:
**Issue**
Fixes #204 

**What this PR Includes**
This PR adds a `k0s backup` command that:

1. Backs up etcd data ([etcd snapshots](https://etcd.io/docs/v3.4/op-guide/recovery/)) in etcd clusters.
2. Runs a hot backup of Sqlite Dbs  in Kine clusters, by using [rqlite/rqlite](https://github.com/rqlite/rqlite) and [go-sqlite3](https://github.com/mattn/go-sqlite3)
3. Add a restore command that untars a backup file and restores the cluster.

Task List:
- [x] Add etcd backup
- [x] Sqlite backup + certificates
- [x] Add a Dockerfile to compile k0s in a `golang:1.15-alpine` image to support cross-compliation of go-sqlite3 (see: https://github.com/mattn/go-sqlite3/issues/384)
- [x] Update smoke tests with the new build option.
- [x] Test and create a restore command.
- [x] Docs


docs/backup.md
```diff
@@ -0,0 +1,42 @@
+# Backup/Restore overview
```
jnummelin, line 1: we need to "hang this up" somewhere at the docs tree in `mkdocs.yaml` file, otherwise it's not visible by default at all.

docs/backup.md
```diff
@@ -0,0 +1,42 @@
...
+To create backup run the following command on the controller node:
+```
+# k0s backup --save-path=<directory>
+```
+The directory used for the `save-path` value must exist and be writable. The default value is the current working directory.
+The command provides backup archive using following naming convention:
+
+```k0s_backup_<ISODatetimeString>.tar.gz```
+
+Because of the DateTime usage, it is guaranteed that none of the previously created archives would be overwrote.
```
jnummelin, line 32:
```suggestion
Because of the DateTime usage, it is guaranteed that none of the previously created archives would be overwritten.
```

jnummelin (changes requested):
Some minor changes requested. Overall pretty good stuff 🙌

jnummelin:
Overall this is good stuff.

The only thing on "high-level" I'm bit disliking (and having done some of the work myself, I'm partially to blame too 😄 ) is the fact that we now have to do lot of file/dir copying during the backup and restore. I think it would make sense, at least to try, to rather do "in-place" operations for the various files. Currently this basically means double disk usage for a while for the backup. Might be problematic in resource constrained environments.

The above can be refactored, if it's even doable, in a separate PR though.

docs/backup.md
```diff
@@ -0,0 +1,42 @@
...
+The backups created by `k0s backup` command have following pieces of your cluster:
+- certificates (the content of the `<data-dir>/pki` directory)
+- etcd snapshot, if the etcd storage is used
+- k0s.yaml
+- any custom defined manifests under the `<data-dir>/manifests`
+- any image bundles located under the `<data-dir>/images`
+- any helm configuration
+
+Parts not covered by the backup utility:
+- content for any persistence volume
```
mviitane, line 15: persistence-> persistent

docs/backup.md
```diff
@@ -0,0 +1,42 @@
...
+- database content, in case if the `kine` is used as a storage driver
```
mviitane, line 16: kine and etcd datastores could be placed next to each other.

pkg/backup/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+func (c configurationStep) Restore(restoreFrom, restoreTo string) error {
+	if !util.FileExists(restoreFrom) {
+		logrus.Info("No k0s.yaml in the backup archive")
+		return nil
+	}
+	logrus.Info("Previously used k0s.yaml saved under the data directory")
+	objectPathInArchive := path.Join(restoreFrom, "k0s.yaml")
+	objectPathInRestored := path.Join(restoreTo, "k0s.yaml")
+	return util.FileCopy(objectPathInArchive, objectPathInRestored)
```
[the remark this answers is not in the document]
soider: so the general flow here is to first unpack archive to temporary directory and then cherrypick everything defined from that temporary directory. 

If the question is generic like "why do we want k0s.yaml in backup", the answer is because we have some helm related setting there that can affect the cluster state pretty significantly.

docs/backup.md
soider: They are mentioned in different sections?

pkg/backup/manager.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Backuper defines interface for backup-restore step
+type Backuper interface {
+	Name() string
+	Backup() (StepResult, error)
+	Restore(from, to string) error
+}
+
+// StepResult backup result for the particular step
+type StepResult struct {
```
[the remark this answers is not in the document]
soider: I'd rather make a fields public because now, with public access for Add method, and the interface, it's possible to extend the backup tool by user (like, you can import k0s/pkg/backup and make your own backuper in addition to those we have now)

pkg/backup/manager.go
```diff
@@ -0,0 +1,171 @@
...
+		bm.Add(NewFilesystemStep(path))
+	}
+	bm.Add(newConfigurationStep(cfgPath))
+}
+
+// Add adds backup step
+func (bm *Manager) Add(step Backuper) {
+	if bm.steps == nil {
+		bm.steps = []Backuper{step}
+		return
```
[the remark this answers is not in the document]
soider: the slice has step as a single element

jnummelin (approved):

jnummelin:
Separate issue to track the kine/sqlite backup capability at #863

docs/backup.md
mviitane: ok, I misunderstood.

jnummelin (approved):

PR #847, opened 2021-04-15
Fix missing import in certificates generation

jasmingacic:
Fixes main branch

### DIFF
--- a/pkg/component/controller/certificates.go
+++ b/pkg/component/controller/certificates.go
@@ -18,6 +18,7 @@ package controller
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"


ncopa:
this is probably good, but the commit message says nothing about why this is needed. Maybe add a short explanation on what happened in the commit message?

jnummelin:
Still unsure how this got broken. We know it's #797 PR that removed the `fmt` import, but that PR passed on all checks.

ncopa (dismissed):
Add some sort of explanation in the commit messsage. For example:

``` 
The `fmt` was removed in 1c1513c491aa (Make etcd use jwt tokens) but is
still needed due to 5de216dfeae7 (Use fmt.Errorf everywhere for error
handling).
```

soider (approved):

soider (approved):

PR #849, opened 2021-04-16
Fix GH actions cache usage for embedded-bins

jnummelin:

**Issue**
Caching of the embedded-bin in GH actions is somehow broken and each PR is building ALL the bins now. This makes PR builds super slow. The reason is that we cache the `.stamp` file but as the `Makefile.variables` coms from checkout the timestamps do not match. Hence `make` determines it needs to do a full rebuild of embedded bins:
```
2021-04-16T08:54:53.2946938Z  File 'build' does not exist.
2021-04-16T08:54:53.2952041Z    File 'k0s' does not exist.
2021-04-16T08:54:53.3324428Z        Prerequisite 'embedded-bins/Makefile.variables' is newer than target '.bins.linux.stamp'.
2021-04-16T08:54:53.3325606Z       Must remake target '.bins.linux.stamp'.
```

**What this PR Includes**
This PR makes the `Makefile.variables` to be cached too so the timestamps match with the .stamp file. When we ACTUALLY change the component versions in the `Makefile.variables`, the cache will NOT be used at all as that file is part of the cache hash and thus there's full re-build for embedded bins as expected.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -53,10 +53,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
-      - name: embedded-bins BEFORE cache
-        run: ls -lah embedded-bins/
-
-      - name: Bindata
+      - name: Bindata cache
         uses: actions/cache@v2
         id: generated-bindata
         with:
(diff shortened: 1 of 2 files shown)


jasmingacic (approved):

ncopa (approved):
Good catch!

PR #855, opened 2021-04-22
remove CNI leftovers

### LINKED ISSUE #823: CNI configs left after reset
**Version**
```
$ k0s version
v0.12.0-35-gc760252
```
Checked out #779

**What happened?**
I ran `k0s reset` and ran `k0s install controller --enable-worker` and started k0s
```
k0s kc get pods -A
NAMESPACE     NAME                              READY   STATUS              RESTARTS   AGE
kube-system   coredns-5c98d7d4d8-z2rsc          0/1     ContainerCreating   0          6m19s
kube-system   konnectivity-agent-xq9xz          0/1     ContainerCreating   0          6m6s
kube-system   kube-proxy-g2s9h                  1/1     Running             0          6m16s
kube-system   kube-router-zlkf8                 1/1     Running             0          6m16s
kube-system   metrics-server-6fbcd86f7b-bhjc9   0/1     ContainerCreating   0          6m6s
```

the three pods above got stuck in `ContrainerCreating` forever. 

Remedy was to remove:
```
/etc/cni/net.d/10-calico.conflist 
/etc/cni/net.d/calico-kubeconfig 
```

It seems that reset doesn't clean up everything.

jasmingacic:
Fixes #823

### DIFF
--- a/pkg/install/cleanup_unix.go
+++ b/pkg/install/cleanup_unix.go
@@ -198,7 +198,7 @@ func (c *CleanUpConfig) RemoveAllDirectories() error {
 		}
 	}
 
-	msg = append(msg, removeCNILeftovers())
+	removeCNILeftovers()
 
 	logrus.Infof("deleting k0s generated data-dir (%v) and run-dir (%v)", c.dataDir, c.runDir)
 	if err := os.RemoveAll(c.dataDir); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/install/cleanup_unix.go
```diff
@@ -209,5 +212,29 @@ func (c *CleanUpConfig) RemoveAllDirectories() error {
...
 }
+
+func removeCNILeftovers() string {
+	var msg []string
+
+	calico10Conflist := "/etc/cni/net.d/10-calico.conflist"
+	calicoKubeconfig := "/etc/cni/net.d/calico-kubeconfig"
+	kuberouter10Conflist := "/etc/cni/net.d/10-kuberouter.conflist"
+
+	if err := os.RemoveAll(calico10Conflist); err != nil {
```
soider, line 226: RemoveAll removes path and any children it contains. It removes everything it can but returns the first error it encounters. If the path does not exist, RemoveAll returns nil (no error). If there is an error, it will be of type *PathError. 


I guess, that means that we going to always have error here or on line 232, depending which CNI we used to have

soider (changes requested):
Minor changes requested: lets' first check if file exists (otherwise we always have error message)

pkg/install/cleanup_unix.go
jasmingacic: Chances are that these files won't be present, so we could print debug messages here.

soider (approved):

PR #857, opened 2021-04-26
Do not fail if stdin is open but data is broken

### LINKED ISSUE #850: Error installing 0.13.1 with Vagrant
**Version**
```
$ k0s version
0.13.1
```

**Platform**
```
$ lsb_release -a
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.3 LTS
Release:	18.04
Codename:	bionic
```

**What happened?**

I have no problem to create a single node cluster using the following Vagrantfile with k0s version 0.12.0:

````
Vagrant.configure("2") do |config|
  config.vm.box = "hashicorp/bionic64"
  config.vm.network "private_network", ip: "192.168.33.11"
  config.vm.provision "shell", inline: <<-SHELL
    export K0S_VERSION=v0.12.0
    curl -sSLf https://get.k0s.sh | sudo --preserve-env=K0S_VERSION sh
    sudo k0s install controller --single
    sudo systemctl start k0scontroller && sleep 10
    sudo cp /var/lib/k0s/pki/admin.conf /vagrant/kubeconfig
    sed -i "s/localhost/192.168.33.11/" /vagrant/kubeconfig
  SHELL
end
````

But, it fails with the following error when using Vagrantfile with k0s version 0.13.1:

````
Vagrant.configure("2") do |config|
  config.vm.box = "hashicorp/bionic64"
  config.vm.network "private_network", ip: "192.168.33.11"
  config.vm.provision "shell", inline: <<-SHELL
    curl -sSLf https://get.k0s.sh | sudo sh
    sudo k0s install controller --single
    sudo systemctl start k0scontroller && sleep 10
    sudo cp /var/lib/k0s/pki/admin.conf /vagrant/kubeconfig
    sed -i "s/localhost/192.168.33.11/" /vagrant/kubeconfig
  SHELL
end
````

````
$ vagrant up
...
    default: Downloading k0s from URL: https://github.com/k0sproject/k0s/releases/download/v0.13.1/k0s-v0.13.1-amd64
    default: k0s is now executable in /usr/local/bin
    default: Error: yaml: unmarshal errors:
    default:   line 1: cannot unmarshal !!str `exit` into v1beta1.yclusterconfig
    default: Usage:
    default:   k0s install controller [flags]
````
I cannot figure out what I'm missing here.

### PR DESCRIPTION
**Pull request by soider**
**Issue**
Fixes #850

**What this PR Includes**
Sometimes we have stdin open but with no data on it (remote execution by ssh, execution in the context of Vagrantfile, probably some other tools as well). This PR prevents us from failing in that case, but do warning and fallback to the default config instead.

### DIFF
--- a/pkg/apis/v1beta1/cluster.go
+++ b/pkg/apis/v1beta1/cluster.go
@@ -137,19 +137,11 @@ func ConfigFromFile(filename string, k0sVars constant.CfgVars) (*ClusterConfig, 
 	return configFromString(string(buf), k0sVars)
 }
 
-func isInputFromPipe() bool {
-	fi, _ := os.Stdin.Stat()
-	return fi.Mode()&os.ModeCharDevice == 0
-}
-
 // ConfigFromStdin tries to read k0s.yaml config from stdin
 func ConfigFromStdin(k0sVars constant.CfgVars) (*ClusterConfig, error) {
-	if !isInputFromPipe() {
-		return nil, fmt.Errorf("can't read configuration from stdin")
-	}
 	input, err := ioutil.ReadAll(os.Stdin)
 	if err != nil {
-		return nil, nil
+		return nil, fmt.Errorf("can't read configration from stdin: %v", err)
 	}
 	return configFromString(string(input), k0sVars)
 
--- a/pkg/apis/v1beta1/cluster.go
+++ b/pkg/apis/v1beta1/cluster.go
@@ -17,8 +17,8 @@ package v1beta1
 
 import (
 	"fmt"
-	"io"
 	"io/ioutil"
+	"os"
 
 	"github.com/k0sproject/k0s/pkg/constant"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by jasmingacic — DISMISSED** [review]

File: pkg/config/yaml.go [hunk 1]
```diff
@@ -39,14 +38,14 @@ func GetYamlFromFile(cfgPath string, k0sVars constant.CfgVars) (clusterConfig *v
 }
 
 func ValidateYaml(cfgPath string, k0sVars constant.CfgVars) (clusterConfig *v1beta1.ClusterConfig, err error) {
-	if cfgPath != "" {
-		clusterConfig, err = v1beta1.FromYamlFile(cfgPath, k0sVars)
-	} else if isInputFromPipe() {
-		clusterConfig, err = v1beta1.FromYamlPipe(os.Stdin, k0sVars)
-	} else {
+	switch cfgPath {
+	case "-":
```
jnummelin, line 42: I think this option deserves a small note in the flag definitions. Something like
```
flagset.StringVarP(&CfgFile, "config", "c", "", "config file (default: ./k0s.yaml). Use '-' to read the config from stdin")
```

jnummelin (approved):
Minor nit only, otherwise good stuff

pkg/apis/v1beta1/cluster.go
```diff
@@ -127,28 +127,35 @@ func validateSpecs(v Validateable) []error {
...
-func FromYamlPipe(r io.Reader, k0sVars constant.CfgVars) (*ClusterConfig, error) {
-	input, err := ioutil.ReadAll(r)
+func isInputFromPipe() bool {
+	fi, _ := os.Stdin.Stat()
+	return fi.Mode()&os.ModeCharDevice == 0
+}
+
+// ConfigFromStdin tries to read k0s.yaml config from stdin
+func ConfigFromStdin(k0sVars constant.CfgVars) (*ClusterConfig, error) {
+	if !isInputFromPipe() {
```
ncopa, line 147: Do we really need this test?

What happens if you do: `k0s ... --config - <./path/to/file`

soider: > k0s ... --config - <./path/to/file

this will read the config from file.

> Do we really need this test?
actually it would either way just fall on the next line, so not that much difference if we drop it (and fix return nil, nil on the next line, which is a bug).

pkg/apis/v1beta1/cluster.go
```diff
@@ -127,28 +127,29 @@ func validateSpecs(v Validateable) []error {
...
+// ConfigFromStdin tries to read k0s.yaml config from stdin
+func ConfigFromStdin(k0sVars constant.CfgVars) (*ClusterConfig, error) {
+	input, err := ioutil.ReadAll(os.Stdin)
 	if err != nil {
-		return nil, nil
+		return nil, fmt.Errorf("can't read configration from stdin: %v", err)
 	}
-	return FromYamlString(string(input), k0sVars)
+	var si sysinfo.SysInfo
+	si.GetSysInfo()
```
jnummelin, line 147: This seems not relevant for this PR at all? And this seems to break the build

jnummelin (approved):

PR #858, opened 2021-04-26
Fix custom datadir (#771)

### LINKED ISSUE #771: kubelet fails to start if data dir on worker is different than on controller
**Version**
```
$ k0s version 0.11
```
**Platform**
```
alpine linux 3.13, x86_64
```
**What happened?**
worker fails to start when data directory does not correspond with `--data-dir` on controller

**How To Reproduce**

* Start controller as non-root with data dir where it has write permissions: `k0s controller --data-dir /tmp/k0s`
* create a join token and join a worker (as root) without setting `--data-dir`.

kubelet fails to start

**Expected behavior**
k0s worker should not care about the `--data-dir` setting on the controller

**Screenshots & Logs**
```...
INFO[2021-03-11 16:08:41] F0311 16:08:41.808708    2250 server.go:257] unable to load client CA file /tmp/k0s/pki/ca.crt: open /tmp/k0s/pki/ca.crt: no such file or directory  component=kubelet
INFO[2021-03-11 16:08:41] goroutine 1 [running]:                        component=kubelet
INFO[2021-03-11 16:08:41] k8s.io/kubernetes/vendor/k8s.io/klog/v2.stacks(0xc000010001, 0xc000181b80, 0x94, 0x9e)  component=kubelet
...
WARN[2021-03-11 16:08:41] exit status 255                               component=kubelet
INFO[2021-03-11 16:08:41] respawning in 5s                              component=kubelet
```

**Additional context**

ncopa:
---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #771

**What this PR Includes**
- Adds tests to verify issue #771 
- Clean up Makefile a bit (fix a warning and respect GO env var)
- fix kubernetes 1.20 leftovers
- fix a filehandle leak spotted by @jnummelin
- save the kubelet config as a template in configmap, so we can replace the path on worker side
- use text/template instead of html/template in template writer so the special chars are not html-escaped

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ go_bindata := go get github.com/kevinburke/go-bindata/...@v3.22.0 && "${GOPATH}/
 endif
 
 GOLANG_IMAGE = golang:1.16-alpine
-GO = GOCACHE=/tmp/.cache docker run --rm -v "$(CURDIR)":/go/src/github.com/k0sproject/k0s \
+GO ?= GOCACHE=/tmp/.cache docker run --rm -v "$(CURDIR)":/go/src/github.com/k0sproject/k0s \
 	-w /go/src/github.com/k0sproject/k0s \
 	-e GOOS \
 	-e CGO_ENABLED \
--- a/Makefile
+++ b/Makefile
@@ -103,14 +103,13 @@ lint: pkg/assets/zz_generated_offsets_$(TARGET_OS).go
 	$(golint) run ./...
 
 .PHONY: $(smoketests)
+check-airgap: image-bundle/bundle.tar
 $(smoketests): k0s
 	$(MAKE) -C inttest $@
 
 .PHONY: smoketests
 smoketests:  $(smoketests)
 
-check-airgap: image-bundle/bundle.tar k0s
-	$(MAKE) -C inttest check-airgap
 
 .PHONY: check-unit
 check-unit: pkg/assets/zz_generated_offsets_$(TARGET_OS).go static/gen_manifests.go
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -66,6 +66,7 @@ require (
 	k8s.io/kubectl v0.20.2
 	k8s.io/mount-utils v0.20.4
 	k8s.io/utils v0.0.0-20201110183641-67b214c5f920
+	k8s.io/kubelet v0.20.5
```
jnummelin, line 69: I don't think we need this dependency now with the templating approach

pkg/component/worker/kubeletconfigclient.go
```diff
@@ -55,3 +57,17 @@ func (k *KubeletConfigClient) Get(profile string) (string, error) {
 	}
 	return config, nil
 }
+
+// Get reads the config from kube api
+func (k *KubeletConfigClient) GetAsStruct(profile string) (*kubeletconfig.KubeletConfiguration, error) {
```
jnummelin, line 62: This func is not really used

jnummelin (changes requested):
Some "dead code" found. In general looks good though

jnummelin:
@jasmingacic @soider PTAL also, I'm bit too biased to review as I helped to conceive the whole template-in-template approach... 😄

jasmingacic (dismissed):
Looks legit

go.mod
ncopa: :+1: cleaned up

pkg/component/worker/kubeletconfigclient.go
ncopa: removed.

jasmingacic (approved):

jnummelin (approved):

PR #860, opened 2021-04-26
Scale coreDNS by the amount of nodes in the cluster

jnummelin:

**Issue**
Currently CoreDNS deployment is hardcoded with `replicas: 1`, which obviously makes it not so tolerant.

**What this PR Includes**
This PR makes coreDNS scale with a factor of node count. Check the unit test for "specification" how the scaling works.


soider (approved):
looks good

jasmingacic (approved):

PR #862, opened 2021-04-27
Enrich telemetry data with host properties

kke:

* Adds host properties (cpu count & architecture, os name & version) to the telemetry context info
* Moves the k0s version from event properties into the intended place: analytics.Context.App.Version

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -3,6 +3,8 @@ package telemetry
 import (
 	"context"
 	"fmt"
+	"os"
+	"runtime"
 
 	analytics "github.com/segmentio/analytics-go"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


pkg/telemetry/telemetry.go
```diff
@@ -140,14 +141,37 @@ func (c Component) sendTelemetry() {
...
+	}
+
+	hostData.App.Version = c.Version
+	hostData.App.Name = "k0s"
+	hostData.App.Namespace = "k0s"
+	hostData.Extra["cpuArch"] = runtime.GOARCH
+
+	if runtime.GOOS == "linux" {
+		var si sysinfo.SysInfo
+		si.GetSysInfo()
```
kke, line 156: Not sure if this gets heavy, it does all kinds of data retrieval from who knows where.

jnummelin: the telemetry is collected anyway once per 10mins by default, so IMO not an issue

kke:
I guess this can be reviewed.

soider (approved):
looks good

PR #872, opened 2021-04-28
Fix release bundle steps

jnummelin:
**Issue**
The image bundle release process has currently two issues:
- release.yaml not valid for GH actions as the `id` field has same value for two steps within a job
- Image bundle build on armv7 fails as it tries to include calico images. Those are not available for armv7 and thus the bundling fails.

**What this PR Includes**
- Fix yaml validity
- Exclude calico images when image list executed on arm(v7)

**Note:** This currently has also #869 commits in so I was able to actually test the release. Will be rebased out once that on is merged.

**Note2:** 1.20.0-beta... releases are baked out of this branch, thus those "checks" also show up on the PR checklist.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,7 +77,7 @@ jobs:
           asset_content_type: application/octet-stream
 
       - name: Upload Release Assets - Bundle
-        id: upload-release-asset
+        id: upload-release-asset-images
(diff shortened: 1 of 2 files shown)


soider (approved):

jnummelin:
@soider Rebased so needs re-review

soider (approved):

PR #772, opened 2021-03-11
k0s reset: don't use crictl as a dependency

jewertow:
Signed-off-by: Jacek Ewertowski <jacek.ewertowski1@gmail.com>

**Issue**
Fixes #743

**What this PR Includes**
1. Custom implementation of `crictl` commands.
2. Removal of embedded `crictl`.

**How to test this change**
1. Run k0s:
```
k0s install controller --enable-worker
systemctl start k0scontroller
```
2. Wait for k0s pods to be ready. There should be 6 pods:
```
systemctl status k0scontroller
k0s kubectl get pods -n kube-system
```
3. Stop k0s and reset:
```
systemctl stop k0scontroller
k0s reset --debug
```
All containers should be stopped and removed successfully.

Issue #743, opened 2021-03-01
k0s reset: don't use crictl as a dependency


trawler:
**Is your feature request related to a problem? Please describe.**
In order to decrease the size footprint of the compiled k0s binary, we should remove crictl from the embedded bins.

**Describe the solution you'd like**
crictl stop and rm operations should be imported into k0s as a library.

jewertow:
@trawler could I work on this issue?

jasmingacic:
@jewertow go ahead give it a try

jewertow:
Ok, cool. I will try to implement it.
I have one question: if the main goal is to decrease the size footprint of the compiled k0s binary, maybe it would be even better to import `cri-api` instead of `crictl` and implement these operations?

trawler:
@jewertow since cri-tools already use cri-api, I don't see why not :) that sounds like a good plan. 👍

PR #881, opened 2021-05-03
[WIP] Add command k0s crictl...

### LINKED ISSUE #856: Ability to list images on the filesystem
I see from #743 that crictl was removed from the binary, however I would like to be able to list the images on the filesystem without having to get another binary (crictl) and pointing it at the k0s containerd location and namespace.

So the ask here is to be able to do something like
```bash
k0s image list
k0s image pull ...
k0s container list
```

I know this would be solved similar to how k0s now has the `kc` command for kubectl, but given `crictl` was removed in #743, I'm not sure that is an option, but `k0s crictl image ls` would have been great.

jewertow:
Signed-off-by: Jacek Ewertowski <jacek.ewertowski1@gmail.com>

**Issue**
Fixes #856 

**Changes**
1. Revert removal of embedded crictl (issue: #743, pr: #772).

**TODO**
1. Allow commands to be passed to crictl as `k0s crictl <command>`.
2. Optionally remove embedded crictl and use crictl as library to reduce size of k0s.

**Important**
Size of the last k0s binary version is 173MB. After this change, k0s binary with embedded crictl will be 193MB. I don't know how crictl as library would reduce the size.


jewertow:
@jnummelin what do you think about command `k0s crictl <options>` instead of `k0s ctr <options>`? I think `ctr` might not be intuitive.

jnummelin:
> what do you think about command k0s crictl <options> instead of k0s ctr <options>? 

The issue with crictl, although it is general purpose for any CRI, is that k0s cannot really configure it properly for all CRIs. k0s only understands the containerD it manages, all other CRIs would need config flags etc. to work properly. And IMO that kinda partially beats the purpose of having such tool available for users.

Hence I'm leaning towards `ctr` more.

Also, I don't think we need to embed it as a binary, but rather "only" as a k0s subcommand like we do for kubectl. So user should be able to run `k0s ctr xyz` out of box. 

oh, and 🙌  for taking up on this :)

jewertow:
Ok, I understand. Thanks for clarification.

PR #884, opened 2021-05-04
Controller join cleanup

ncopa:
**What this PR Includes**

Misc code clean up to simplify and make controller join code more readable. No (significant) functional changes.


jnummelin (approved):

pkg/component/controller/etcd.go
```diff
@@ -126,46 +163,12 @@ func (e *Etcd) Run() error {
 
 	if util.FileExists(filepath.Join(e.K0sVars.EtcdDataDir, "member", "snap", "db")) {
 		logrus.Warnf("etcd db file(s) already exist, not gonna run join process")
```
soider, line 165: I think instead of "else if" here we can just have explicit return if the guard check has not passed, to keep nesting level lower.

soider (approved):
one minor suggestion, otherwise looks good. Feel free to refuse suggested changes.

PR #888, opened 2021-05-05
Drop segment fork, use the Extra field in upstream for Direct

kke:

As in https://github.com/k0sproject/k0sctl/pull/139 - The `Extra` field can be used for setting the `Direct: true`.

According to analytics-go [code comments](https://github.com/segmentio/analytics-go/blob/v3.0/context.go#L28-L32):

```go
	// This map is used to allow extensions to the context specifications that
	// may not be documented or could be introduced in the future.
	// The fields of this map are inlined in the serialized context object,
	// there is no actual "extra" field in the JSON representation.
	Extra map[string]interface{} `json:"-"`
```

So, the analytics-go fork which adds the `Direct`-field is not necessary for that functionality.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/rogpeppe/go-internal v1.6.1 // indirect
 	github.com/segmentio/analytics-go v3.1.0+incompatible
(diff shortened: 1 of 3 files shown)


kke:
Tested working with the lower case "direct".

jnummelin (approved):

PR #890, opened 2021-05-05
Check for unused go mod deps during linter step

### LINKED ISSUE #818: Check in GHA that `go mod tidy` doesn't change anything
I'd like to engage all contributors to run `go mod tidy` before merging branches. By having that we would always ensure that we don't have any noisy and garbage deps in the go.mod (they could appear during development, from debugging or obsolete deps). Also that would led us to have dependencies across branches converge to the same state which also could be beneficial. 

Separate GHA action, the same way we do for checking linter results or checking for merge commit.
The steps in that action could be:
- Clone repository
- Run `go mod tidy`
- If `git diff` shows diff in go.mod/go.sum fail action with message "Please, run `go mod tidy`"
- Separate makefile target could be introduced for that

soider:

**Issue**
Fixes #818

**What this PR Includes**
Add separate makefile target for the linter step to check if we have unused deps in the go.mod go.sum


Makefile
```diff
@@ -99,7 +99,7 @@ k0s.exe k0s: $(GO_SRCS)
 	touch $@
 
 .PHONY: lint
-lint: pkg/assets/zz_generated_offsets_$(TARGET_OS).go
+lint: pkg/assets/zz_generated_offsets_$(TARGET_OS).go lint-gomod
```
s0j, line 102: Instead of the `GOMODTIDYLINT` variable, couldn't you just tie in the `SKIP_GOMOD_LINT` check here as a conditional dependency?

ie. `lint: pkg/assets/zz_generated_offsets_$(TARGET_OS).go $(if $(SKIP_GOMOD_LINT), , lint-gomod)`

This way, you can just leave the body of `lint-gomod` as unconditional.

soider: Done not exactly how you suggested, but very close to:
- only one definition for lint-gomod target
- list of deps for lint target is updated by ifneq check for the SKIP_GOMOD_LINT variable

jasmingacic (approved):

ncopa (approved):
LGTM

PR #896, opened 2021-05-07
Add k0s start and stop for controlling service state

kke:

Adds `k0s start` and `k0s stop` subcommands for starting / stopping the service if it has been installed.

### DIFF
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -36,10 +36,7 @@ func NewStartCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			status, err := svc.Status()
-			if err != nil {
-				return err
-			}
+			status, _ := svc.Status()
 			if status == service.StatusRunning {
 				cmd.SilenceUsage = true
 				return fmt.Errorf("already running")
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -41,6 +41,7 @@ func NewStartCmd() *cobra.Command {
 				return err
 			}
 			if status == service.StatusRunning {
+				cmd.SilenceUsage = true
 				return fmt.Errorf("already running")
 			}
 			return svc.Start()
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2021 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


jasmingacic (dismissed):
Looks good. Make sure to document it before it is merged

kke:
I added some documentation.

I haven't tried this at all yet, need to set up some kind of build/test env.

cmd/start/start.go
```diff
@@ -0,0 +1,50 @@
...
+				return err
+			}
+			status, err := svc.Status()
+			if err != nil {
+				return err
+			}
+			if status == service.StatusRunning {
+				return fmt.Errorf("already running")
+			}
+			return svc.Start()
```
jasmingacic, line 46: can you suppress cobra help output so it doesn't return this:
```
# ./k0s start 
Error: already running
Usage:
  k0s start [flags]

Flags:
  -h, --help   help for start

Global Flags:
      --data-dir string   Data Directory for k0s (default: /var/lib/k0s). DO NOT CHANGE for an existing setup, things will break!
      --debug             Debug logging (default: false)

2021-05-18 08:27:24.350451 I | already running
```

cmd/start/start.go
```diff
@@ -0,0 +1,50 @@
...
+	return &cobra.Command{
+		Use:   "start",
+		Short: "Start the k0s service configured on this host. Must be run as root (or with sudo)",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if os.Geteuid() != 0 {
+				return fmt.Errorf("this command must be run as root")
+			}
+			svc, err := install.InstalledService()
+			if err != nil {
+				return err
```
jasmingacic, line 37: It seem it doesn't return correct error:

``` 
# ./k0s install controller --single
```
```
# stat  /etc/systemd/system/k0scontroller.service
  File: /etc/systemd/system/k0scontroller.service
  Size: 521             Blocks: 8          IO Block: 4096   regular file
Device: 8e3h/2275d      Inode: 655836      Links: 1
Access: (0644/-rw-r--r--)  Uid: (    0/    root)   Gid: (    0/    root)
...
```
```
# ./k0s start 
Error: the service is not installed
...
```

jasmingacic (changes requested):

cmd/start/start.go
kke: So it doesn't work on single node clusters?

jasmingacic: I already had the service running and it was able to execute `k0s stop`  but when I reinstalled it, it wouldn't start. I'm not sure if it is related to single node clusters.

kke: Maybe now?

cmd/start/start.go
kke: Done

jasmingacic (dismissed):
Works as described

jasmingacic (approved):

PR #904, opened 2021-05-12
Fix sysinfo errors in windows build

kke:

Moves the sysinfo gathering into a separate function which live in build-tagged files. On linux we get the full sysinfo, on others just get the runtime.GOOS.

Also adds a windows build step into CI.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -125,9 +125,6 @@ jobs:
       - name: Build
         run: make k0s.exe
 
-      - name: Run unit tests
-        run: make check-unit
-
   smoketest:
     name: Smoke test
     needs: build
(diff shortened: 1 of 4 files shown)


.github/workflows/go.yml
```diff
@@ -83,6 +81,53 @@ jobs:
...
+
+          key: ${{ runner.os }}-embedded-bins-win-${{ hashFiles('**/embedded-bins/**/*') }}
+          restore-keys: |
+            ${{ runner.os }}-embedded-bins-win-${{ hashFiles('**/embedded-bins/**/*') }}
+
+      - name: Build
+        run: make k0s.exe
+
+      - name: Run unit tests
+        run: make check-unit
```
jnummelin, line 129: I don't think we need to run unit tests on win build job. those are anyway running with GOOS=linux, so there's no benefit AFAIK

jnummelin (approved):
LGTM

PR #912, opened 2021-05-20
Fix pod logs check in smokes by always getting a fresh list of pods to find ready pods from

jnummelin:

**Issue**
The pod logs check in smokes is bit flaky as the test in correctly uses "stale" list of pods to find running pods from.

**What this PR Includes**
This makes the pod logs check to always get fresh list of pods so we do not try to re-use stale information.

### DIFF
--- a/inttest/common/util.go
+++ b/inttest/common/util.go
@@ -2,7 +2,6 @@ package common
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


ncopa (approved):
LGTM

PR #913, opened 2021-05-20
Fix token commands role defaulting

jnummelin:

**Issue**
`k0s token create` incorrectly defaults to controller token

**What this PR Includes**
Makes token create and list to use separate variables for the role as they have different defaults. Currently the list flag makes also the create flag default to empty string 🤦


ncopa (approved):
LGTM

PR #914, opened 2021-05-20
making sure that token create and list are accepting only controller and worker role

jasmingacic:
Ensuring that token create/list `--role` parameter can only accept values `worker` or `controller`.

### DIFF
--- a/cmd/token/create.go
+++ b/cmd/token/create.go
@@ -85,7 +85,7 @@ k0s token create --role worker --expiry 10m  //sets expiration time to 10 minute
 }
 
 func checkCreateTokenRole(cmd *cobra.Command, args []string) error {
-	if !(createTokenRole == controllerRole || createTokenRole == workerRole) {
+	if createTokenRole != controllerRole && createTokenRole != workerRole {
 		cmd.SilenceUsage = true
 		return fmt.Errorf("unsupported role %q, supported roles are %q and %q", createTokenRole, controllerRole, workerRole)
 	}
--- a/cmd/token/create.go
+++ b/cmd/token/create.go
@@ -37,6 +37,7 @@ func tokenCreateCmd() *cobra.Command {
 		Example: `k0s token create --role worker --expiry 100h //sets expiration time to 100 hours
 k0s token create --role worker --expiry 10m  //sets expiration time to 10 minutes
 `,
+		PreRunE: checkCreateTokenRole,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Disable logrus for token commands
 			logrus.SetLevel(logrus.FatalLevel)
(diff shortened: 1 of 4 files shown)


soider (dismissed):

cmd/token/list.go
```diff
@@ -76,3 +77,14 @@ func tokenListCmd() *cobra.Command {
 	cmd.PersistentFlags().AddFlagSet(config.GetPersistentFlagSet())
 	return cmd
 }
+
+func checkListTokenRole(cmd *cobra.Command, args []string) error {
+	if listTokenRole == "" {
```
jnummelin, line 82: Empty role IS actually supported in list command. When no role is given it lists both controller and worker tokens.

cmd/token/create.go
```diff
@@ -82,3 +83,11 @@ k0s token create --role worker --expiry 10m  //sets expiration time to 10 minute
 
 	return cmd
 }
+
+func checkCreateTokenRole(cmd *cobra.Command, args []string) error {
+	if !(createTokenRole == controllerRole || createTokenRole == workerRole) {
```
jnummelin, line 88: IMO the negated or'ed conditions is harder to fit into ones head, at least into mine, than having plain:
```
if createTokenRole != controllerRole && createTokenRole != workerRole) {
```

pkg/token/kubeconfig.go
```diff
@@ -78,9 +78,11 @@ func CreateKubeletBootstrapConfig(clusterConfig *config.ClusterConfig, k0sVars c
 	if role == "worker" {
 		data.User = "kubelet-bootstrap"
 		data.JoinURL = clusterConfig.Spec.API.APIAddressURL()
-	} else {
+	} else if role == "controller" {
```
jnummelin, line 81: Use the added consts for this and above worker ase

pkg/token/kubeconfig.go
```diff
@@ -78,9 +78,11 @@ func CreateKubeletBootstrapConfig(clusterConfig *config.ClusterConfig, k0sVars c
...
 		data.User = "controller-bootstrap"
 		data.JoinURL = clusterConfig.Spec.API.K0sControlPlaneAPIAddress()
+	} else {
+		return "", fmt.Errorf("unsupported role %s only supporter roles are %q and %q", role, "controller", "worker")
```
jnummelin, line 85:
```suggestion
		return "", fmt.Errorf("unsupported role %s only supported roles are %q and %q", role, "controller", "worker")
```

jnummelin (changes requested):

cmd/token/create.go
jasmingacic: Linter doesn't like it:
```
suspect or: listTokenRole != controllerRole || listTokenRole != workerRole
```
but you are right i will change the "if"

pkg/token/kubeconfig.go
jasmingacic: it is a different package I will move the constants at the top of this file

jnummelin (approved):

PR #924, opened 2021-05-24
Add command 'k0s ctr [command] [flags] [args]'

### LINKED ISSUE #856: Ability to list images on the filesystem

jewertow:
**Issue**
Fixes #856 

**What this PR Includes**
This PR adds command `ctr`, which is containerd CLI.

### DIFF
--- a/inttest/ctr/singlenode_test.go
+++ b/inttest/ctr/singlenode_test.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2021 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package ctr
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/k0sproject/k0s/inttest/common"
+	"github.com/stretchr/testify/suite"
+)
+
+type CtrSuite struct {
+	common.FootlooseSuite
+}
+
+func (s *CtrSuite) TestK0sCtrCommand() {
+	ssh, err := s.SSH(s.ControllerNode(0))
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	_, err = ssh.ExecWithOutput("k0s install controller --enable-worker")
+	s.Require().NoError(err)
+
+	_, err = ssh.ExecWithOutput("rc-service k0scontroller start")
+	s.Require().NoError(err)
+
+	err = s.WaitForKubeAPI(s.ControllerNode(0))
+	s.Require().NoError(err)
+
+	kc, err := s.KubeClient(s.ControllerNode(0))
+	s.NoError(err)
+
+	err = s.WaitForNodeReady(s.ControllerNode(0), kc)
+	s.NoError(err)
+
+	output, err := ssh.ExecWithOutput("k0s ctr namespaces list")
+	s.Require().NoError(err)
+
+	flatOutput := removeRedundantSpaces(output)
+	errMsg := fmt.Sprintf("returned output of command 'k0s ctr namespaces list' is different than expected: %s", output)
+	s.Equal("NAME LABELS k8s.io", flatOutput, errMsg)
+}
+
+func TestCtrCommandSuite(t *testing.T) {
+	s := CtrSuite{
+		common.FootlooseSuite{
+			ControllerCount: 1,
+		},
+	}
+	suite.Run(t, &s)
+}
+
+func removeRedundantSpaces(output string) string {
+	pattern := regexp.MustCompile(`\s+`)
+	out := pattern.ReplaceAllString(output, " ")
+	out = strings.TrimSpace(out)
+	return out
+}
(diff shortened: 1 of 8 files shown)


jewertow:
If you want to verify if it works, execute the commands below to see that subcommands, flags and arguments work properly.
```
k0s ctr images list
k0s ctr image pull docker.io/calico/node:v3.11.2
k0s ctr images label docker.io/calico/node:v3.11.2 app=cni
k0s ctr images label --replace-all docker.io/calico/node:v3.11.2 cni=calico
```

jasmingacic:
Can you please run `go mod tidy` to make CI pass?

jewertow:
**Go build / Smoke test for BYO CRI feature** failed but it seems that it is not related to my change. Could you retry this stage?

jasmingacic:
I'm trying to pass this
```
# ./k0s ctr -a /run/containerd/containerd.sock i list 
Error: unknown shorthand flag: 'a' in -a
```
where as ctr is accepting -a --address flag 
```
# ctr  -a /run/containerd/containerd.sock i list
REF TYPE DIGEST SIZE PLATFORMS LABELS 
```

jewertow:
I'm sorry. I know what is the problem.

jasmingacic:
One more thing `ctr` should have default values for address something like `path.Join(k.K0sVars.RunDir, "containerd.sock")` and default namespace set to `k8s.io` so we would be able to do `k0s ctr c ls` out of the box without the need to do `k0s ctr --address /run/k0s/containerd.sock -n k8s.io c ls`. 
Also the users should be able to override these values by passing their desired values.

Does this make sense?

jewertow:
Ok, thanks for the feedback. I will fix it soon.

jasmingacic:
@jewertow also it would be good to have some sort of integration tests for such scenarios which would allow us to check if the tool has been included entirely.

jewertow:
@jasmingacic does it make sense to set default namespace to `k8s.io`? The default behaviour of containerd is:
1. Use value of `CONTAINERD_NAMESPACE` env var if it' set.
2. Use namespace `default`.
Maybe it would be better to not set namespace and relay on default behaviour of the `ctr`?

jasmingacic:
The ctr in this case will be used only for k0s purposes so defaulting to namespace k8s.io where all containers reside makes sense. So we should default to `k8s.io` basically we already do here https://github.com/k0sproject/k0s/blob/main/pkg/component/worker/ocibundle.go#L46

jewertow:
@jasmingacic I significantly simplified implementation. Now it fully relies on the original implementation.

jewertow:
After changing implementation entirely, do you still see the need to add integration tests?
I think the current implementation is trivial and I don't see corner cases.

kke:
> After changing implementation entirely, do you still see the need to add integration tests?
> I think the current implementation is trivial and I don't see corner cases.

I don't see anything worth unit testing in it now, there's hardly any "units" to test. LGTM.

A pretty simple "smoke test" case would be to just add some `k0s ctr` command to one of the integration tests (such as inttest/install/singlenode_test.go) that would just run `k0s ctr images list | grep -q something-its-supposed-to-include` and error out if it fails.

jasmingacic (changes requested):
Both stand alone `ctr --address /run/k0s/containerd.sock -n k8s.io c ls` and `k0s ctr c ls` produce the same output.

Can you please add integration test as @kke proposed [here](https://github.com/k0sproject/k0s/pull/924#issuecomment-847765685), so we can merge this.

jewertow:
Yes, I will add a test soon.

jewertow:
@jasmingacic this failed CI plan looks like random failure, because it's not related to my changes and the test I added works fine. Could you trigger failed tests or run CI plan once again?

jasmingacic:
@jewertow The CI issue is due to some changes can you sync your fork with the upstream.

jasmingacic (dismissed):
@jewertow this is looking good.

@k0sproject/k0s-core can someone else take a look too especially to the GH action changes.

inttest/ctr/singlenode_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (s *CtrSuite) TestK0sCtrCommand() {
+	ssh, err := s.SSH(s.ControllerNode(0))
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	_, err = ssh.ExecWithOutput("k0s install controller --enable-worker")
+	s.Require().NoError(err)
+
+	_, err = ssh.ExecWithOutput("rc-service k0scontroller start")
```
kke, line 40: If you change this to `k0s start` we can test the #896 at the same time :)

```suggestion
	_, err = ssh.ExecWithOutput("k0s start")
```

kke (approved):

inttest/ctr/singlenode_test.go
jewertow: Thanks. You're right, but the command `re-service k0scontroller start/stop` is also used in other tests, so we could change all of them to `k0s start/stop` in another pull request, because it's not related to the issue. What do you think?

jasmingacic: I'd rather that we don't touch other tests. Let's scope this PR to ctr only.

jewertow: Ok, done

jasmingacic (approved):
Great work @jewertow  thank you

PR #929, opened 2021-05-25
Drop telemetry interval

### LINKED ISSUE #920: Consider dropping user-configurable telemetry interval
The config has `spec.telemetry.interval`, I don't think it's necessary.

It could even be considered harmful. It's possible to set a super low value there and have k0s send a flood of telemetry and freeze itself. Setting a longer than default value will only be confusing when the telemetry data is analyzed.

kke:
Fixes #920 

Removes the user configurable telemetry interval and uses constant 10min, which is the current default. I don't know why anyone would need to adjust it and what good it would do either way.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -57,7 +57,6 @@ spec:
   podSecurityPolicy:
     defaultPolicy: 00-k0s-privileged
   telemetry:
-    interval: 10m0s
     enabled: true
   installConfig:
     users:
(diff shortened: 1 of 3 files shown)


jasmingacic (approved):
seems legit

ncopa (approved):
LGTM

kittiphophum18 (commented):
****

kittiphophum18 (commented):
****

PR #931, opened 2021-05-25
Ensure that containerd shims are statically built

ncopa:
Containerd and the shims need different CGO_ENABLED setting. By default
containerd needs CGO while the shims should have those disabled.

Fixes commit b98908b92908 (Make various k0s/embedded-bins build options
configurable.)


s0j (approved):

jasmingacic (approved):
Legit. The issue I was seeing earlier is not present anymore.

PR #934, opened 2021-05-26
fixes #899 and removed "server" aliases for controller command

### LINKED ISSUE #899: "Role" printout fault during installation
**Version**
```
v1.21.0+k0s.0
```
**Platform**
```
Centos7 VM
```
**What happened?**

When running `sudo k0s install controller` command with the `--enable-worker` option, the role shows correctly: **controller+worker**. But when using with `--single` option, the role shows only the **controller**

**The output when using `--enable-worker`:**
```
Version: v0.11.0
Process ID: 436
Parent Process ID: 1
Role: controller+worker
Init System: linux-systemd
```

**The output when using `--single`:**
```
Version: v1.21.0+k0s.0
Process ID: 1264
Parent Process ID: 1
Role: controller
Init System: linux-systemd
```

jasmingacic:
fixes #899 and 
removed obsolete "server" aliases for controller command

### DIFF
--- a/cmd/status/status.go
+++ b/cmd/status/status.go
@@ -31,11 +31,9 @@ type CmdOpts config.CLIOptions
 
 var (
 	output string
-	s      *install.K0sStatus
 )
 
 func NewStatusCmd() *cobra.Command {
-	s = &install.K0sStatus{}
 	cmd := &cobra.Command{
 		Use:     "status",
 		Short:   "Helper command for get general information about k0s",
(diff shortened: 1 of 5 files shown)


pkg/install/process.go
```diff
@@ -37,32 +37,32 @@ type K0sStatus struct {
...
-		return nil, nil, err
+		return nil, err
 	}
-
 	for _, p := range processList {
-		if p.Executable() == "k0s" && hasChildren(p.Pid(), processList) {
-			pid := p.Pid()
-			ppid := p.PPid()
-			return &pid, &ppid, nil
+		if strings.Contains(p.Executable(), "k0s") && hasChildren(p.Pid(), processList) {
```
kke, line 61: btw, doesn't this only work if the executable name is `k0s`.. 🤔 maybe it should use `filepath.Base(os.Executable())` instead of `"k0s"`.

jasmingacic: https://github.com/k0sproject/k0s/blob/main/cmd/status/status.go#L45 it won't get to here at least not for status. 
But reset and install might fail.

cmd/status/status.go
```diff
@@ -44,34 +42,44 @@ func NewStatusCmd() *cobra.Command {
...
 				return err
 			}
+			if len(statuses) != 0 {
+				for _, s := range statuses {
+					status := install.K0sStatus{
+						Pid:  s.Pid,
+						PPid: s.PPid,
+					}
+					if len(statuses) > 1 {
+						fmt.Println("=================")
```
jnummelin, line 57: looks like some leftover debugs

jnummelin (commented):
I fail to see how this fixes #899 🤔

jnummelin (commented):
I fail to see how this fixes #899 🤔

cmd/status/status.go
kke: It's a separator for multiple items it seems. Don't know what happens with json/yaml.

jasmingacic:
This PR might introduce more edge cases than it would fix.
`k0s status` will need further improvements

PR #936, opened 2021-05-26
Notify slack channel on job execution

trawler:

**What this PR Includes**
Send a notification on Slack on job execution.

### DIFF
--- a/.github/workflows/check-network.yaml
+++ b/.github/workflows/check-network.yaml
@@ -5,10 +5,8 @@ on:
       prNumber:
         description: PR against which to this workflow
         required: false
-
   schedule:
     - cron: "0 23 * * *"
-
 jobs:
   terraform:
     env:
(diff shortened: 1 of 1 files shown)


kittiphophum18 (dismissed):

kke:
The message it generates is rather lengthy.

trawler:
> The message it generates is rather lengthy.

Valid point... I'm still playing with the format though. But I also suppose that since this check is only triggered once a day, it should be too spammy.

trawler:
slack message now shortened to include only:
```
workflow,ref,commit,author,job,took
```

kke:
Didn't change noticeably.  In my opinion it could pretty much be an oneliner like:

```
New PR by Foo: Insert foo into foofoo
```

```
PR Approved by Bar: Insert foo into foofoo
```

```
New issue: Foo doesn't work
```

```
PR Merged: Insert foo into foofoo
```

```
Issue closed: Foo doesn't work
```


 Now it creates this huge hunk of text that covers the same area as some 20 lines of regular discussion.

trawler:
> Didn't change noticeably. In my opinion it could pretty much be an oneliner like:
**_Before_**
Screenshot 2021-05-27 at 15 00 44

**_After_**
Screenshot 2021-05-27 at 15 01 03

jnummelin (approved):
LGTM

We can always tweak the format if it proves to be too spammy

kittiphophum18 (approved):

kittiphophum18:
> Signed-off-by: Karen Almog <kalmog@mirantis.com>
> 
> 
> 
> **What this PR Includes**
> 
> Send a notification on Slack on job execution.

PR #954, opened 2021-06-03
Add man pages generation to k0s docs subcommand

kke:

Once there is a proper installer or packaging, there should also be man pages.

Changes `k0s docs` subcommand to require the format as an argument, ie. `k0s docs markdown` or `k0s docs man`.

The output for markdown is `./docs/cli` and the output for man is `./man`. Maybe it should be configurable.


kke:
It seems to result in having to use something like `man k0s_install` to get install docs.

jasmingacic (approved):

PR #955, opened 2021-06-03
k0s reset: support for docker

### LINKED ISSUE #742: k0s reset enhancement: support docker CRI
k0s reset supports only containerd, it should also support docker as a container runtime.

jewertow:
**Issue**
Fixes #742

**What this PR Includes**
The purpose of this PR is to support docker and CRI-compatible container runtimes by `k0s reset` command.

**Changes**
Command `reset` accepts flag to specify CRI socket:
```
k0s reset --cri-socket docker:unix:///var/run/docker.sock
```

**How to test**
```
k0s install controller --enable-worker --cri-socket docker:unix:///var/run/docker.sock
k0s start
k0s kc get pods -n kube-system
```
Wait for startup of pods in the `kube-system` namespace
```
k0s stop
k0s reset --debug --cri-socket docker:unix:///var/run/docker.sock
docker ps -aq
```
There should be no containers.

### DIFF
--- a/cmd/reset/reset.go
+++ b/cmd/reset/reset.go
@@ -66,13 +66,6 @@ func (c *CmdOpts) reset() error {
 		logger.Fatal("k0s seems to be running! please stop k0s before reset.")
 	}
 
-	role := install.GetRoleByStagedKubelet(c.K0sVars.BinDir)
-	logrus.Debugf("detected role for cleanup: %v", role)
-	err := install.UninstallService(role)
-	if err != nil {
-		// might be that k0s was not run as a part of a service. just notify on uninstall error
-		logger.Infof("failed to uninstall k0s service: %v", err)
-	}
 	// Get Cleanup Config
 	cfg, err := cleanup.NewConfig(c.K0sVars.DataDir, c.WorkerOptions.CriSocket)
 	if err != nil {
--- a/cmd/reset/reset.go
+++ b/cmd/reset/reset.go
@@ -67,14 +67,11 @@ func (c *CmdOpts) reset() error {
 	}
 
 	// Get Cleanup Config
-	cfg, err := cleanup.NewConfig(c.K0sVars.DataDir, c.WorkerOptions.CriSocket)
+	cfg, err := cleanup.NewConfig(c.K0sVars, c.CfgFile, c.WorkerOptions.CriSocket)
 	if err != nil {
 		logger.Fatalf("failed to configure cleanup: %v", err)
 		return err
 	}
-
-	cfg.CfgFile = c.CfgFile
-	cfg.K0sVars = c.K0sVars
 
 	err = cfg.Cleanup()
 
--- a/cmd/reset/reset.go
+++ b/cmd/reset/reset.go
@@ -66,9 +66,17 @@ func (c *CmdOpts) reset() error {
 		logger.Fatal("k0s seems to be running! please stop k0s before reset.")
 	}
 
+	role := install.GetRoleByStagedKubelet(c.K0sVars.BinDir)
+	logrus.Debugf("detected role for cleanup: %v", role)
+	err := install.UninstallService(role)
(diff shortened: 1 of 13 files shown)


cmd/reset/reset.go
```diff
@@ -73,7 +74,11 @@ func (c *CmdOpts) reset() error {
 		logger.Infof("failed to uninstall k0s service: %v", err)
 	}
 	// Get Cleanup Config
-	cfg := install.NewCleanUpConfig(c.K0sVars.DataDir)
+	cfg, err := install.NewCleanUpConfig(c.K0sVars.DataDir, c.WorkerOptions.CriSocket)
+	if err != nil {
+		logger.Fatalf("Failed to configure clenup: %v", err)
```
kke, line 79:
```suggestion
		logger.Fatalf("Failed to configure cleanup: %v", err)
```

pkg/config/cli.go
```diff
@@ -100,11 +100,16 @@ func GetKubeCtlFlagSet() *pflag.FlagSet {
 	return flagset
 }
 
+func GetCriSocketFlag() *pflag.FlagSet {
+	flagset := &pflag.FlagSet{}
+	flagset.StringVar(&workerOpts.CriSocket, "cri-socket", "", "contrainer runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]")
```
kke, line 105:
```suggestion
	flagset.StringVar(&workerOpts.CriSocket, "cri-socket", "", "container runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]")
```

cmd/reset/reset.go
jewertow: Fixed

pkg/config/cli.go
jewertow: Fixed

kke:
overlaps with #959 - @jasmingacic should this come before or after? :)

jasmingacic:
@kke I would suggest that we integrate this later since the logic has changed significantly.

jewertow:
Ok, so can I come back to this feature when your pull request will be merged, right?

jasmingacic:
@jewertow yes you can but implementation will differ significantly

jasmingacic:
@jewertow go ahead and have this refactored so it fits into new design.

jewertow:
@jasmingacic I rebased commits and refactored code. Please review this change.

jasmingacic (changes requested):
Passing "cri-socket" to `k0s reset` doesn't feel right.  Ideally the users should just run `k0s reset` to remove all k0s leftovers. 

I know that we discover containerd by looking for the containerd binary in dataDir/bin/containerd and also by checking if this statement is true `c.Config.containerd == nil` can we based on that figure out `--docker-endpoint` that's passed to the kubelet process. 

We know based on `k0s status` PID of k0s process and  get kubelet by PPID
```
# pstree $K0SPID
k0s─┬─etcd───96*[{etcd}]
   ...
    ├─kubelet───57*[{kubelet}]
    └─64*[{k0s}]
``` 

We already have something similar here -> `https://github.com/k0sproject/k0s/blob/main/pkg/install/process.go`

jewertow:
@jasmingacic `pstree $K0SPID` will not return `k0s` and `kubelet` processes after k0s has stoppped, so `reset` can't relay on it.

jewertow:
Maybe command `install` could store CRI socket path somewhere to make it possible for `reset` to use this path later?

jasmingacic (approved):
Since we don't have any better way to do it, this solution is good.

kke:
> Maybe command `install` could store CRI socket path somewhere to make it possible for `reset` to use this path later?

Yeah this should be in the status socket api that has been discussed a couple of times lately.

ncopa (approved):
LGTM

Issue #928, opened 2021-05-24
"k0s install" UX improvement


mviitane:
**Is your feature request related to a problem? Please describe.**

When installing k0s and forgetting to add the config file, it becomes difficult to re-install.

```
$ sudo k0s install controller --single
# forgot the config file, let's try again

$ sudo k0s install controller --single -c ~/k0s.yaml 
INFO[2021-05-12 12:34:19] Installing k0s service            
Error: failed to install k0s service: failed to install service: Init already exists: /etc/systemd/system/k0scontroller.service
2021-05-12 12:34:19.372931 I | failed to install k0s service: failed to install service: Init already exists: /etc/systemd/system/k0scontroller.service

$ sudo k0s reset
INFO[2021-05-12 12:34:37] Uninstalling the k0s service         
INFO failed to uninstall k0s service: failed to remove the k0s service: exit status 1 
FATA[2021-05-12 12:34:37] failed clean up pre-flight-checks: failed to find directory `/var/lib/k0s`, check if the node is provisioned
```

**Describe the solution you'd like**
1. I'd like to see that `k0s install` re-writes the k0s service with the updated arguments.
2. I'd like to see that `k0s reset` cleans up the existing init service even if `/var/lib/k0s` was not found.

kke:
1. Maybe require `--force` for overwriting and suggest using force when it already exists (i think if the suggestions in #770 would be implemented, this problem would no longer exist)
2. Yes, seems to be a bug to me

jasmingacic:
It is somewhat related to #878 and #879

trawler:
By design, k0s reset doesn't stop with failure even if one or more sub-tasks fail. maybe we can return a better return code, to alert users that some tasks failed during the "reset" command.

ncopa:
Dropping the `k0s install` would save us for all those issues, as @kke mentions in https://github.com/k0sproject/k0s/issues/770#issuecomment-848613480.

If we cannot drop the k0s installer, then I would prefer that the installer installs a default config under `/etc`. Configuring an app before installing it is unusual. Would be more intuitive to install it before configuring it.

mviitane:
Closing issue. The second part (reset improvement) was implemented and delivered. This makes it rather easy to also re-install k0s if/when needed (first run reset, then install).

PR #959, opened 2021-06-04
reset command improvements

jasmingacic:
Fixes part of the issue #928.
Reset command will remove service stubs regardless of the role installed. 
When running reset command without k0s binaries already being unpacked in the data directory reset command is not able to detect which role has already been installed. So reset will now try to remove both service stubs k0scontroller and k0sworker.

Signed-off-by: Jasmin Gacic <jgacic@mirantis.com>


kke:
The behavior is a bit sketchy.

First it gets the role:

```go
	role := install.GetRoleByStagedKubelet(c.K0sVars.BinDir)
```

This will return "worker" if no installation was detected.

Then it proceeds to uninstall both controller and worker:

```go
	err := install.UninstallService("controller")
	if err != nil {  ...	}
	err = install.UninstallService("worker")
	if err != nil {  ...	}
```

I assume one of them will always report the `logger.Infof("failed to uninstall k0s service: %v", err)` in the err check block, as both of them will always run.

Finally it will proceed to do cleanup based on the `install.GetRoleByStagedKubelet` from above, regardless of which Uninstall succeeded. In case of `controller+worker` both cleanups will run in series. 

I think the best would be to fix the  `GetRoleByStagedKubelet` to only return "worker" if it somehow can validate that a worker config exists and use `install.InstalledService` to figure out if a service was installed.

jasmingacic:
@kke `GetRoleByStagedKubelet` relies on existence of kubelet or api binary on the host in case that neither exist it will return empty role. So in case that the install failed due to existence of service file and no binaries have been before extracted reset won't be able to detect the role and therefore to remove service stubs which are named k0scontroller and k0sworker. 
So this is a crude solution for only one part of the issue we are seeing.

kke:
> @kke `GetRoleByStagedKubelet` relies on existence of kubelet or api binary on the host in case that neither exist it will return empty role. So in case that the install failed due to existence of service file and no binaries have been before extracted reset won't be able to detect the role and therefore to remove service stubs which are named k0scontroller and k0sworker.
> So this is a crude solution for only one part of the issue we are seeing.

https://github.com/k0sproject/k0s/blob/de25ffec200e772267b0426421135220cd178874/pkg/install/cleanup.go#L67-L79

seems to me, it will never return an empty role, but always fall back to "worker".

jasmingacic:
Yeah returning "worker" when binaries are not present caused reset to behave weird.

cmd/reset/reset.go
```diff
@@ -65,36 +65,15 @@ func (c *CmdOpts) reset() error {
...
+	cfg.K0sVars = c.K0sVars
 
-	if err := cfg.RemoveAllDirectories(); err != nil {
-		logger.Info(err.Error())
+	if err := cfg.Cleanup(); err != nil {
+		return err
 	}
-	logrus.Info("k0s cleanup operations done. To ensure a full reset, a node reboot is recommended.")
+
+	fmt.Println("k0s cleanup operations done. To ensure a full reset, a node reboot is recommended.")
```
jnummelin, line 76: This get not outputted in case any of the steps had errors

pkg/cleanup/cleanup_unix.go
```diff
@@ -0,0 +1,98 @@
...
+	if util.FileExists(apiBinary) && util.FileExists(kubeletBinary) {
+		return "controller+worker"
+	} else if util.FileExists(apiBinary) {
+		return "controller"
+	} else {
+		return "worker"
+	}
+}
+
+type Step interface {
```
jnummelin, line 94: Needs a comment as it's exported

pkg/cleanup/services.go
```diff
@@ -0,0 +1,62 @@
...
+	if len(msg) > 0 {
+		return fmt.Errorf("%v", strings.Join(msg, "\n"))
+	}
+	return nil
+}
+
+func uninstallService(role string) error {
+	prg := &install.Program{}
+
+	if role == "controller+worker" {
```
jnummelin, line 50: I don't think this check is really needed. This func get called only from the above `Run()` and thus should be only called for `controller` or `worker` roles.

jnummelin (changes requested):
Added some specific comments.

There's bunch of funcs that are exported but lacks proper comments.

jnummelin (commented):
Minor nit about the output:
```
/ # ./k0s reset
*  containerd steps
INFO[2021-06-11 07:15:58] successfully removed k0s containers!         
INFO[2021-06-11 07:15:58] no config file given, using defaults         
*  remove directories step
WARN[2021-06-11 07:15:58] failed to unmount /var/lib/k0s               
*  CNI leftovers cleanup step
k0s cleanup operations done. To ensure a full reset, a node reboot is recommended.
Error: errors received during clean-up: [failed to delete /var/lib/k0s. err: unlinkat /var/lib/k0s: device or resource busy]
2021-06-11 07:15:58.173736 I | errors received during clean-up: [failed to delete /var/lib/k0s. err: unlinkat /var/lib/k0s: device or resource busy]
```

The output is now coming from both `fmt.Println` and through logrus which makes it bit "messy" IMO. I'd rather have all output through logrus.

kke:
> The output is now coming from both `fmt.Println` and through logrus which makes it bit "messy" IMO. I'd rather have all output through logrus.

There's also direct output from the commands started via os.Exec as the stdout and stderr are piped directly.

jnummelin (approved):

PR #962, opened 2021-06-06
Retry controller join for transient failures

### LINKED ISSUE #866: CA syncking should be retried
**Version**
```
main
```

**What happened?**
With automation tools like k0sctl the cluster bootstrap is FAST. Every now and then the process get slightly slowed down with systemd unit restarts. Once the systemd unit is restarted, It'll pass as the first controller has gotten all the APIs up-and-running.

**How To Reproduce**
Use automation, human is not fast enough, to setup HA controlplane.

**Expected behavior**
CA, and other. sync between the controllers should probably use some retry functionality for scenarios like this.

**Screenshots & Logs**
Also add any output from kubectl if applicable:
```
Apr 28 11:30:18 k0s-controller-1 k0s[1168]: time="2021-04-28 11:30:18" level=info msg="initializing CASyncer\n"
Apr 28 11:30:18 k0s-controller-1 k0s[1168]: Error: failed to sync CA: Get "https://49.12.16.201:9443/v1beta1/ca": EOF
Apr 28 11:30:18 k0s-controller-1 k0s[1168]: failed to sync CA: Get "https://49.12.16.201:9443/v1beta1/ca": EOF
Apr 28 11:30:18 k0s-controller-1 systemd[1]: k0scontroller.service: Main process exited, code=exited, status=1/FAILURE
Apr 28 11:30:18 k0s-controller-1 systemd[1]: k0scontroller.service: Failed with result 'exit-code'.
```

jnummelin:

**Issue**
Fixes #866 

**What this PR Includes**
This PR makes controller join process to retry the CA sync. With automation tools like k0sctl the controllers are joined so fast that on the initial controller the sync API might not yet be up-and-running. That'll make the underlying init system (e.g. systemd) to restart the service. The job will get done for sure, but it'll take a bit long. Testing with this fix and with k0sctl shows this shaves ~1min off from the 3+3 setup.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -134,11 +134,17 @@ func joinController(tokenArg string, certRootDir string) (*token.JoinClient, err
 		return nil, fmt.Errorf("failed to create join client: %w", err)
 	}
 
-	caData, err := joinClient.GetCA()
+	var caData v1beta1.CaResponse
+	err = retry.Do(func() error {
+		caData, err = joinClient.GetCA()
+		if err != nil {
+			return fmt.Errorf("failed to sync CA: %w", err)
+		}
+		return nil
+	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to sync CA: %w", err)
+		return nil, err
 	}
-
 	return joinClient, writeCerts(caData, certRootDir)
 }
 


cmd/controller/controller.go
```diff
@@ -134,11 +134,17 @@ func joinController(tokenArg string, certRootDir string) (*token.JoinClient, err
...
 
-	caData, err := joinClient.GetCA()
+	var caData v1beta1.CaResponse
+	err = retry.Do(func() error {
+		caData, err = joinClient.GetCA()
+		if err != nil {
+			return fmt.Errorf("failed to sync CA: %w", err)
+		}
+		return nil
+	})
```
kke, line 144: FYI, total retry time is ~52 seconds, pause between two last attempts is 25 seconds.

Maybe could be improved by giving up early in case of some unrecoverable failures:

```go
		err = retry.Do(
			func() error {
				....
			},
			retry.RetryIf(
				func(err error) bool {
					return !strings.Contains(err.Error(), "something")
				},
			),
		)
```

kke (dismissed):

cmd/controller/controller.go
jnummelin: > Maybe could be improved by giving up early in case of some unrecoverable failures

I actually did think of this, but it's just super hard to make difference what is un-recoverable and what is not. E.g. with k0sctl (as it's super fast) the initial trys might give plain IO timeout (the join API is not yet up on the first controller), some RBAC related 40x (initial controller not yet pushed RBAC stuff for join tokens) or something else.

> FYI, total retry time is ~52 seconds, pause between two last attempts is 25 seconds.
I think this is OK for this case.

kke (approved):

PR #967, opened 2021-06-09
Change out put to output in k0s status

kke:

```
k0s status --help
Helper command for get general information about k0s

Usage:
  k0s status [flags]

Examples:
The command will return information about system init, PID, k0s role, kubeconfig and similar.

Flags:
  -h, --help         help for status
  -o, --out string   sets type of out put to json or yaml
```

Removes eyesore and saves a byte in size!


jasmingacic (approved):
:D

PR #978, opened 2021-06-15
Fix eula escaping

jnummelin:
**Issue**
Release building broken as it injects eula text that has spaces.


jnummelin:
Created out of wrong base branch 🤕 will re-open soon

PR #980, opened 2021-06-16
Fix prerequisites for generated go files

ncopa:
We use $(GO) for generating the zz_*.go files, so we need tell make that
we need create the docker image to be able to generate those files.


s0j (approved):

jnummelin (approved):
LGTM, assuming all the smokes agree with me 😃

PR #985, opened 2021-06-23
getting k0s status from k0s status socket

jasmingacic:
This PR introduces "status" sockets to transmit information about the running process. It will surface information as follows:
```
# k0s status -o json
[
   {
      "Version": "v1.21.0+k0s.0-164-g6050cc15",
      "Pid": 1577859,
      "PPid": 0,
      "Role": "controller",
      "SysInit": "",
      "StubFile": "",
      "Output": "",
      "Workload": true,
      "Args": [
         "./k0s",
         "controller",
         "--single"
      ],
      "ClusterConfig": {
         "APIVersion": "k0s.k0sproject.io/v1beta1",
         "Kind": "Cluster",
         "Metadata": {
            "Name": "k0s"
         },
         "Spec": {
            "API": {
               "Address": "147.75.33.177",
               "Port": 6443,
               "K0sAPIPort": 9443,
               "ExternalAddress": "",
               "SANs": [
                  "147.75.33.177",
                  "10.12.18.133",
                  "172.17.0.1",
                  "10.244.0.1"
               ],
               "ExtraArgs": {}
            },
            "ControllerManager": {
               "ExtraArgs": {}
            },
            "Scheduler": {
               "ExtraArgs": {}
            },
            "Storage": {
               "Type": "kine",
               "Kine": {
                  "DataSource": "sqlite:///var/lib/k0s/db/state.db?more=rwc\u0026_journal=WAL\u0026cache=shared"
               },
               "Etcd": null
            },
            "Network": {
               "PodCIDR": "10.244.0.0/16",
               "ServiceCIDR": "10.96.0.0/12",
               "Provider": "kuberouter",
               "Calico": null,
               "KubeRouter": {
                  "MTU": 0,
                  "PeerRouterIPs": "",
                  "PeerRouterASNs": "",
                  "AutoMTU": true
               },
               "DualStack": {
                  "Enabled": false,
                  "IPv6PodCIDR": "",
                  "IPv6ServiceCIDR": ""
               },
               "KubeProxy": {
                  "Disabled": false,
                  "Mode": "iptables"
               }
            },
            "PodSecurityPolicy": {
               "DefaultPolicy": "00-k0s-privileged"
            },
            "WorkerProfiles": null,
            "Telemetry": {
               "Enabled": true
            },
            "Install": {
               "SystemUsers": {
                  "Etcd": "etcd",
                  "Kine": "kube-apiserver",
                  "Konnectivity": "konnectivity-server",
                  "KubeAPIServer": "kube-apiserver",
                  "KubeScheduler": "kube-scheduler"
               }
            },
            "Images": {
               "Konnectivity": {
                  "Image": "us.gcr.io/k8s-artifacts-prod/kas-network-proxy/proxy-agent",
                  "Version": "v0.0.20"
               },
               "MetricsServer": {
                  "Image": "gcr.io/k8s-staging-metrics-server/metrics-server",
                  "Version": "v0.3.7"
               },
               "KubeProxy": {
                  "Image": "k8s.gcr.io/kube-proxy",
                  "Version": "v1.21.2"
               },
               "CoreDNS": {
                  "Image": "docker.io/coredns/coredns",
                  "Version": "1.7.0"
               },
               "Calico": {
                  "CNI": {
                     "Image": "docker.io/calico/cni",
                     "Version": "v3.18.1"
                  },
                  "Node": {
                     "Image": "docker.io/calico/node",
                     "Version": "v3.18.1"
                  },
                  "KubeControllers": {
                     "Image": "docker.io/calico/kube-controllers",
                     "Version": "v3.18.1"
                  }
               },
               "KubeRouter": {
                  "CNI": {
                     "Image": "docker.io/cloudnativelabs/kube-router",
                     "Version": "v1.2.1"
                  },
                  "CNIInstaller": {
                     "Image": "quay.io/k0sproject/cni-node",
                     "Version": "0.1.0"
                  }
               },
               "Repository": "",
               "DefaultPullPolicy": "IfNotPresent"
            },
            "Extensions": null,
            "Konnectivity": {
               "AgentPort": 8132,
               "AdminPort": 8133
            }
         }
      },
      "K0sVars": {
         "AdminKubeConfigPath": "/var/lib/k0s/pki/admin.conf",
         "BinDir": "/var/lib/k0s/bin",
         "CertRootDir": "/var/lib/k0s/pki",
         "WindowsCertRootDir": "C:\\var\\lib\\k0s\\pki",
         "DataDir": "/var/lib/k0s",
         "EtcdCertDir": "/var/lib/k0s/pki/etcd",
         "EtcdDataDir": "/var/lib/k0s/etcd",
         "KineSocketPath": "/run/k0s/kine/kine.sock:2379",
         "KonnectivitySocketDir": "/run/k0s/konnectivity-server",
         "KubeletAuthConfigPath": "/var/lib/k0s/kubelet.conf",
         "KubeletBootstrapConfigPath": "/var/lib/k0s/kubelet-bootstrap.conf",
         "KubeletVolumePluginDir": "/usr/libexec/k0s/kubelet-plugins/volume/exec",
         "ManifestsDir": "/var/lib/k0s/manifests",
         "RunDir": "/run/k0s",
         "KonnectivityKubeConfigPath": "/var/lib/k0s/pki/konnectivity.conf",
         "OCIBundleDir": "/var/lib/k0s/images",
         "DefaultStorageType": "kine",
         "HelmHome": "/var/lib/k0s/helmhome",
         "HelmRepositoryCache": "/var/lib/k0s/helmhome/cache",
         "HelmRepositoryConfig": "/var/lib/k0s/helmhome/repositories.yaml"
      }
   }
]
```

or if running without out parameter it will return:
```
# ./k0s status 
Version: v1.21.0+k0s.0-164-g6050cc15
Process ID: 1577859
Role: controller
Workloads: true
```

This PR also removes the role "controller+worker" and introduces parameters workloads so instead of having "controller+worker" it will return Role: controller and Workloads: true to indicate that the controller is started as single node.

Also what might be confusing is that json and yaml will be returned as an array. This is because we need to be able to poll status in scenarios where users are running two instances of k0s on the same host. To make this work when k0s is bootstrapped as controller it will create rundir + controller.sock and worker.sock for worker instance.

### DIFF
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -70,24 +70,27 @@ func (c *CmdOpts) backup() error {
 	}
 
 	if !util.DirExists(savePath) {
-		logger.Fatalf("the save-path directory (%v) does not exist.", savePath)
+		return fmt.Errorf("the save-path directory (%v) does not exist", savePath)
 	}
 
 	if !util.DirExists(c.K0sVars.DataDir) {
-		logger.Fatalf("cannot find data-dir (%v). check your environment and/or command input and try again.", c.K0sVars.DataDir)
+		return fmt.Errorf("cannot find data-dir (%v). check your environment and/or command input and try again", c.K0sVars.DataDir)
 	}
 
-	role := install.GetRoleByStagedKubelet(c.K0sVars.BinDir)
-	logrus.Debugf("detected role for backup operations: %v", role)
+	status, err := install.GetStatusInfo(config.StatusSocket)
+	if err != nil {
+		return fmt.Errorf("unable to detect cluster status %s", err)
+	}
+	logrus.Debugf("detected role for backup operations: %v", status.Role)
 
-	if strings.Contains(role, "controller") {
+	if strings.Contains(status.Role, "controller") {
 		mgr, err := backup.NewBackupManager()
 		if err != nil {
 			return err
 		}
 		return mgr.RunBackup(c.CfgFile, c.ClusterConfig.Spec, c.K0sVars, savePath)
 	}
-	return fmt.Errorf("backup command must be run on the controller node, have `%s`", role)
+	return fmt.Errorf("backup command must be run on the controller node, have `%s`", status.Role)
 }
 
 func preRunValidateConfig(cmd *cobra.Command, args []string) error {
--- a/pkg/component/status/status.go
+++ b/pkg/component/status/status.go
@@ -1,0 +1,79 @@
+package status
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+
+	"github.com/k0sproject/k0s/internal/util"
+	"github.com/k0sproject/k0s/pkg/install"
+	"github.com/sirupsen/logrus"
+)
+
+type Status struct {
+	StatusInformation install.K0sStatus
+	Socket            string
+	L                 *logrus.Entry
+	httpserver        http.Server
+	listener          net.Listener
+}
+
+// Healthy dummy implementation
+func (s *Status) Healthy() error { return nil }
+
+// Init initializes component
+func (s *Status) Init() error {
+	s.L = logrus.WithFields(logrus.Fields{"component": "status"})
+
+	var err error
+	s.httpserver = http.Server{
+		Handler: &statusHandler{Status: s},
+	}
+	err = util.InitDirectory(s.StatusInformation.K0sVars.RunDir, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create %s: %w", s.Socket, err)
+	}
+
+	s.listener, err = net.Listen("unix", s.Socket)
+	if err != nil {
+		s.L.Errorf("failed to create listener %s", err)
+		return err
+	}
+	s.L.Infof("Listening address %s", s.Socket)
+
+	return nil
+}
+
+// Run runs the component
+func (s *Status) Run() error {
+	go func() {
+		if err := s.httpserver.Serve(s.listener); err != nil {
+			s.L.Errorf("failed to start status server at %s: %s", s.Socket, err)
+		}
+	}()
+	return nil
+}
+
+// Stop stops status component and removes the unix socket
+func (s *Status) Stop() error {
+	defer os.Remove(s.Socket)
+	if err := s.httpserver.Shutdown(context.TODO()); err != nil {
+		return err
+	}
+	return nil
+}
+
+type statusHandler struct {
+	Status *Status
+}
+
+// ServerHTTP implementation of handler interface
+func (sh *statusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	if json.NewEncoder(w).Encode(sh.Status.StatusInformation) != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
--- a/pkg/config/cli.go
+++ b/pkg/config/cli.go
@@ -16,6 +16,7 @@ limitations under the License.
 package config
 
 import (
+	"path/filepath"
 	"time"
 
 	"github.com/spf13/pflag"
@@ -31,6 +32,7 @@ var (
 	DataDir        string
 	Debug          bool
 	DebugListenOn  string
+	StatusSocket   string
 	K0sVars        constant.CfgVars
 	workerOpts     WorkerOptions
 	controllerOpts ControllerOptions
@@ -94,6 +96,7 @@ func GetPersistentFlagSet() *pflag.FlagSet {
 	flagset.StringVarP(&CfgFile, "config", "c", "", "config file, use '-' to read the config from stdin")
 	flagset.BoolVarP(&Debug, "debug", "d", false, "Debug logging (default: false)")
 	flagset.StringVar(&DataDir, "data-dir", "", "Data Directory for k0s (default: /var/lib/k0s). DO NOT CHANGE for an existing setup, things will break!")
+	flagset.StringVar(&StatusSocket, "status-socket", filepath.Join(K0sVars.RunDir, "status.sock"), "Full file path to the socket file.")
 	flagset.StringVar(&DebugListenOn, "debugListenOn", ":6060", "Http listenOn for Debug pprof handler")
 	return flagset
 }
--- a/pkg/install/process.go
+++ b/pkg/install/process.go
@@ -16,135 +16,55 @@ limitations under the License.
 package install
 
 import (
+	"context"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
-	"os/exec"
-	"runtime"
-	"strings"
+	"net"
+	"net/http"
 
-	"github.com/k0sproject/k0s/internal/util"
-	"github.com/mitchellh/go-ps"
-	"gopkg.in/yaml.v2"
+	config "github.com/k0sproject/k0s/pkg/apis/v1beta1"
+	"github.com/k0sproject/k0s/pkg/constant"
 )
 
 type K0sStatus struct {
-	Version  string
-	Pid      int
-	PPid     int
-	Role     string
-	SysInit  string
-	StubFile string
-	Output   string
+	Version       string
+	Pid           int
+	PPid          int
+	Role          string
+	SysInit       string
+	StubFile      string
+	Output        string
+	Workloads     bool
+	Args          []string
+	ClusterConfig *config.ClusterConfig
+	K0sVars       constant.CfgVars
 }
 
-func GetPid() (status *K0sStatus, err error) {
-	pid, ppid, err := getProcessID()
-	if err == nil && pid != nil {
-		status = &K0sStatus{
-			Pid:  *pid,
-			PPid: *ppid,
-		}
-		return status, nil
+func GetStatusInfo(socketPath string) (status *K0sStatus, err error) {
+	status = &K0sStatus{}
+
+	httpc := http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", socketPath)
+			},
+		},
 	}
-	return &K0sStatus{}, nil
-}
 
-func getProcessID() (pid *int, ppid *int, err error) {
-	processList, err := ps.Processes()
+	response, err := httpc.Get("http://localhost")
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
+	defer response.Body.Close()
 
-	for _, p := range processList {
-		if p.Executable() == "k0s" && hasChildren(p.Pid(), processList) {
-			pid := p.Pid()
-			ppid := p.PPid()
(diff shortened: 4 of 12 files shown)


mviitane:
Based on this output it seems impossible to know if k0s was started with "controller --single" or "controller --enable-worker":
  Role: controller
  Workloads: true

Could those two use cases be separated in the printout in some way?

pkg/component/status/status.go
```diff
@@ -0,0 +1,89 @@
...
+// ServerHTTP implementation of handler interface
+func (sh *statusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	jsn, err := json.Marshal(sh.Status.StatusInformation)
+	if err != nil {
+		return
+	}
+	_, err = w.Write(jsn)
+	if err != nil {
+		return
+	}
```
kke, line 88:
```suggestion
	w.Header().Set("Content-Type", "application/json")
	if json.NewEncoder(w).Encode(sh.Status.StatusInformation) != nil {
		w.WriteHeader(http. StatusInternalServerError)
	}
```

pkg/install/process.go
```diff
@@ -16,135 +16,67 @@ limitations under the License.
...
-		return nil, nil, err
-	}
-
-	for _, p := range processList {
-		if p.Executable() == "k0s" && hasChildren(p.Pid(), processList) {
-			pid := p.Pid()
-			ppid := p.PPid()
-			return &pid, &ppid, nil
+	for _, role := range roles {
+		address := filepath.Join(runDir, fmt.Sprintf("%s.sock", role))
```
kke, line 51: There's probably something going wrong if a rundir has data for multiple instances. A program should usually fail to start if the rundir indicates there's an instance already active (a pidfile exists in the rundir and the process listed in it is alive). 

I feel like there's all this trickery going on just to avoid fixing the constant rundir (for root, for regular users it's under data-dir/run).

The rationale for creating a status socket was to confidently get data from a specific instance.

pkg/component/status/status.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Healthy dummy implementation
+func (s *Status) Healthy() error { return nil }
+
+// Init initializes component
+func (s *Status) Init() error {
+	s.L = logrus.WithFields(logrus.Fields{"component": "status"})
+
+	if len(s.address) == 0 {
+		s.address = path.Join(s.StatusInformation.K0sVars.RunDir, fmt.Sprintf("%s.sock", s.StatusInformation.Role))
```
kke, line 34: To determine the role of an instance via the socket, you first need to know the role of the instance so you know which socket to communicate with.

kke:
> Based on this output it seems impossible to know if k0s was started with "controller --single" or "controller --enable-worker":

It also seems if not impossible, at least difficult to find out what the difference between those two is.

From looking at the source, I think the only difference is that `--single` also sets the default storage to kine, otherwise it's the same?

```go
			if c.SingleNode {
				c.EnableWorker = true
				c.K0sVars.DefaultStorageType = "kine"
			}
```

pkg/component/status/status.go
jasmingacic: We query both known sockets and gather data if present.

pkg/install/process.go
jasmingacic: Let's open another PR for configurable runDir

jasmingacic: #991

mviitane:
> > Based on this output it seems impossible to know if k0s was started with "controller --single" or "controller --enable-worker":
> 
> It also seems if not impossible, at least difficult to find out what the difference between those two is.

'--single' option does the following:

- Kine/SQLite used as data store
- The following processes are disabled (needed for multi-node clusters):
Worker-controller konnectivity
Controller join api
Controller leader election

The differences should be documented somewhere.

s0j:
The output looks great if the component is running + socket available, but what about if it (worker or controller) fails to start for some reason?  Some (most) of the data provided in the output doesn't need a running controller, and would be helpful for diagnosing the startup failure.

jasmingacic:
With the last commit I introduced `status-socket` CLI parameter which defaults to `status.sock` that will be created in the run dir.  In case that the user decides to run two k0s instances on the same host it will result in following error:
```
2021-06-28 12:25:12.599405 I | /run/k0s/status.sock already exists, use 'status-socket' to specify different unix socket
```
The status socket will return a single status object.
To run multiple k0s instances on the same host do the following:

``` 
$ k0s controller
...
```
```
$k0s token create --role worker > tokenf
```
```
$ k0s worker --token-file tokenf --status-socket worker.sock
...
```
To poll for status of the controller instance just run
```
$ k0s status
```
and for the worker instance run:
```
$ k0s status --status-socket worker.sock
```

jasmingacic:
Fixes #899

pkg/component/status/status.go
```diff
@@ -0,0 +1,91 @@
...
+	}()
+	return nil
+}
+
+// Stop stops status component and removes the unix socket
+func (s *Status) Stop() error {
+	if err := s.httpserver.Shutdown(context.TODO()); err != nil {
+		return err
+	}
+	os.Remove(s.address)
```
s0j, line 77: This should be `defer'd, otherwise a failure in shutdown will leave this lingering.

pkg/install/process.go
```diff
@@ -16,135 +16,62 @@ limitations under the License.
...
+			},
+		},
 	}
-	return "", fmt.Errorf("k0s role is not found")
-}
 
-func (s K0sStatus) GetK0sVersion() (string, error) {
-	cmd := fmt.Sprintf("/proc/%d/exe", s.Pid)
-	stdout, err := exec.Command(cmd, "version").Output()
+	response, err := httpc.Get("http:/" + address)
```
s0j, line 61: Whats with the single `/`?

jasmingacic: I tried the address string is file path so it has a slash at the beginning

pkg/component/status/status.go
```diff
@@ -0,0 +1,91 @@
...
+	}()
+	return nil
+}
+
+// Stop stops status component and removes the unix socket
+func (s *Status) Stop() error {
+	if err := s.httpserver.Shutdown(context.TODO()); err != nil {
+		return err
+	}
+	defer os.Remove(s.address)
```
s0j, line 77: oops, the defer still won't be reached here.

pkg/install/process.go
s0j: The transport/connection is setup with `unix://<address>`, and your handler doesn't require any paths, so shouldn't `httpc.Get("/")` just work here?

jasmingacic: `httpc.Get("/")` won't work. I replaced it with `https.Get("http://localhost")` so we don't have to concatenate strings.

pkg/component/status/status.go
jasmingacic: I moved defer before shutdown to ensure the socket file is removed once the function returns.

jasmingacic:
> Based on this output it seems impossible to know if k0s was started with "controller --single" or "controller --enable-worker":

The status returns command line arguments which were used to bootstrap the cluster. Based on that information we can determine whether it was started with `--single` or `--enable-worker` flag.

cmd/backup/backup.go
```diff
@@ -77,17 +77,20 @@ func (c *CmdOpts) backup() error {
 		logger.Fatalf("cannot find data-dir (%v). check your environment and/or command input and try again.", c.K0sVars.DataDir)
 	}
 
-	role := install.GetRoleByStagedKubelet(c.K0sVars.BinDir)
-	logrus.Debugf("detected role for backup operations: %v", role)
+	status, err := install.GetStatusInfo(c.K0sVars.RunDir, config.StatusSocket)
+	if err != nil {
+		logger.Fatalf("unable to detect cluster status %s", err)
```
soider, line 82: let's keep it consistent with whole function (it always return error, instead of doing Fatalf and kill process here).
In general, logger.Fatalf is not the best practice, for example, it prevents "defers" from being called (not this particular function case, just generic observation)

jasmingacic: good point

pkg/component/status/status.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+	s.L.Infof("Listening address %s", s.address)
+
+	return nil
+}
+
+// Run runs the component
+func (s *Status) Run() error {
+	spew.Dump(s)
```
soider, line 63: debug leftover

soider (dismissed):
Left one minor comment about debug output, apart from that looks good

soider (approved):
reapprove

pkg/config/cli.go
```diff
@@ -94,6 +95,7 @@ func GetPersistentFlagSet() *pflag.FlagSet {
 	flagset.StringVarP(&CfgFile, "config", "c", "", "config file, use '-' to read the config from stdin")
 	flagset.BoolVarP(&Debug, "debug", "d", false, "Debug logging (default: false)")
 	flagset.StringVar(&DataDir, "data-dir", "", "Data Directory for k0s (default: /var/lib/k0s). DO NOT CHANGE for an existing setup, things will break!")
+	flagset.StringVar(&StatusSocket, "status-socket", "status.sock", "Name of the status socket file in the runDir.")
```
ncopa, line 98: I'd prefer this to be the full path, not the "file in the runDir".

pkg/install/process.go
```diff
@@ -16,135 +16,62 @@ limitations under the License.
...
-
-func hasChildren(pid int, processes []ps.Process) bool {
-	for _, p := range processes {
-		if p.PPid() == pid {
-			return true
-		}
-	}
-	return false
-}
+func GetStatusInfo(runDir string, socket string) (status *K0sStatus, err error) {
```
ncopa, line 46: I'd prefer this to take a single socket path (or URI) instead of two args, runDir + socket.
```suggestion
func GetStatusInfo(socketPath string) (status *K0sStatus, err error) {
```

pkg/install/process.go
```diff
@@ -16,135 +16,62 @@ limitations under the License.
...
+	status = &K0sStatus{}
 
-func GetRoleByPID(pid int) (role string, err error) {
-	if runtime.GOOS == "windows" {
-		return "worker", nil
+	address := filepath.Join(runDir, socket)
+	if _, err := os.Stat(address); err != nil {
+		return nil, fmt.Errorf("%s does not exist", address)
```
ncopa, line 51: There is no reason to check for the existence of the `address`.

ncopa (changes requested):
Good job!

I would prefer that `--status-socket` option should take a full path/URI, ie `--status-socket=/run/k0s/k0s.sock`.

pkg/config/cli.go
jasmingacic: That would mean that we will allow users to put the file anywhere. This is not very practical from the perspective of `reset`.

pkg/install/process.go
jasmingacic: Just checking if the file is present before trying to `Dial`

ncopa (approved):
LGTM

PR #991, opened 2021-06-24
Configurable --run-dir

kke:

A silly attempt at making the rundir settable via `--run-dir` like the data-dir.

Obviously this is still merely a torso.

- On start-up, k0s should refuse to start if a pidfile exists in the rundir and the process listed in it exists. If a pidfile that was created more then X seconds ago exists but the process does not, it should expect the previous session crashed and clean up the rundir, maybe warn about it and start normally.
- A user running a single instance should not care about rundir.
- A user running multiple instances, should be able to set a different run dir for each instance.
- Commands like `k0s status` should use the default run-dir unless `--run-dir` given.
- Pid file, socket file, etc should have a constant name under the run-dir, `k0s.pid` and `k0s-status.sock` for example. To utilize the rundir, the operator should not need to know the role of the instance beforehand. You should be able to figure out the role by doing something like `curl --unix-socket=/run/k0s.1/k0s-status.sock http://localhost/ | grep '"Role":`.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -214,7 +214,7 @@ func (c *CmdOpts) startController() error {
 	// common factory to get the admin kube client that's needed in many components
 	adminClientFactory := kubernetes.NewAdminClientFactory(c.K0sVars)
 
-	componentManager.Add(&controller.APIServer{
+	componentManager.Add(&controller.KubeAPIServer{
 		ClusterConfig:      c.ClusterConfig,
 		K0sVars:            c.K0sVars,
 		LogLevel:           c.Logging["kube-apiserver"],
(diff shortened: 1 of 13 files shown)


jasmingacic (approved):

ncopa:
I'd rather have an `--socket-path` option for the status socket than a `--run-dir` option.

kke:
Multiple instances will still need unique run dirs I think.

jasmingacic:
To ensure that runDir is unique for each instance means we'd have to know what is the run directory of the other instance. Also I think that just passing `--socket-path` is much cleaner solution

kke:
> To ensure that runDir is unique for each instance means we'd have to know what is the run directory of the other instance. Also I think that just passing `--socket-path` is much cleaner solution

You don't need to ensure that if another instance refuses to start if the rundir is already populated by an active instance.

jnummelin:
The PR is marked as stale since no activity has been recorded in 30 days

PR #992, opened 2021-06-24
Ensure `make clean` runs through all target commands

s0j:
For those people who choose to prune their images, they would encounter that `make clean` wouldn't run to completion (due to the missing images)


ncopa (approved):
LGTM

PR #998, opened 2021-06-30
Update kardianos/service to the latest version to fix Start/Stop for worker

### LINKED ISSUE #997: `k0s start` & `k0s stop` don't work for the worker node
**Version**
```
$ k0s version
v1.21.2+k0s.0
```
**Platform**
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Release:	20.04
Codename:	focal
```
**What happened?**
`k0s start` and `k0s stop` don't work for the worker node.

**How To Reproduce**
$ curl -sSLf https://get.k0s.sh | sudo sh
Downloading k0s from URL: https://github.com/k0sproject/k0s/releases/download/v1.21.2+k0s.0/k0s-v1.21.2+k0s.0-amd64
k0s is now executable in /usr/local/bin

$ sudo k0s install worker --token-file ~/token 
INFO[2021-06-30 07:08:30] Installing k0s service

$ sudo k0s start
Error: exit status 5
...

**Expected behavior**
`k0s start` and `k0s stop` should work for the worker nodes in the same way they do for the controllers.

**Workaround**
Use the init system commands instead. For example:
`sudo systemctl start k0sworker.service`
`sudo systemctl stop k0sworker.service`
They work just fine.

soider:
Update kardianos/service to the latest version to fix Start/Stop for worker

fixes #997

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/ishidawataru/sctp v0.0.0-20191218070446-00ab2ac2db07 // indirect
 	github.com/jmoiron/sqlx v1.2.1-0.20190826204134-d7d95172beb5 // indirect
 	github.com/k0sproject/dig v0.1.0
-	github.com/kardianos/service v1.2.1-0.20201211143537-ef35c563203c
+	github.com/kardianos/service v1.2.1-0.20210616011951-36c9bf8c36a2
 	github.com/kr/text v0.2.0 // indirect
 	github.com/magiconair/properties v1.8.1
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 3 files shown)


jasmingacic (approved):
Looks good, I gave it a try and it was working.
Also we have to keep in mind that start/stop won't work in case of two k0s instances on the host. It will only start the first instance it finds.

@mviitane Maybe it would be good if we document it

jasmingacic (dismissed):

trawler (dismissed):

mviitane:
> Looks good, I gave it a try and it was working.
> Also we have to keep in mind that start/stop won't work in case of two k0s instances on the host. It will only start the first instance it finds.
> 
> @mviitane Maybe it would be good if we document it

Good point. I think we need to clarify running of two instances in general and also check if other commands like status and reset work for two. AFAIK, we don't state in the docs that we even support two instances.

jasmingacic (approved):

PR #1009, opened 2021-07-07
added another reset step 'bridge'

jasmingacic:
Fixes https://github.com/k0sproject/k0sctl/issues/167
Signed-off-by: Jasmin Gacic <jasmin.gacic@gmail.com>

### DIFF
--- a/pkg/cleanup/bridge_linux.go
+++ b/pkg/cleanup/bridge_linux.go
@@ -1,0 +1,49 @@
+package cleanup
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/vishvananda/netlink"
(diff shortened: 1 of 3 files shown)


ncopa (dismissed):

ncopa:
Seems like it needs to be skipped on windows

ncopa (approved):

soider:
can it also be somehow skipped for macos build? (Basically it's just to invert logic, make default implementation stubbed and overrire it in _linux.go files)

 The thing is that on my local macos I sometime use plain "go build main.go" to ensure that there are no dramatical error + some local tooling also rely on proper compilation

PR #1023, opened 2021-07-22
Allow to disable controller-manager and scheduler

jakolehm:
**What this PR Includes**

Adds flags to `k0s controller` command to disable all scheduling/controller-manager functionality. Basically puts k0s controller to "api only" mode (which can be handy in some cases).

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -228,7 +228,7 @@ func (c *CmdOpts) startController() error {
 			KubeClientFactory: adminClientFactory,
 		})
 	}
-	if !c.SingleNode {
+	if !c.SingleNode && (!c.DisableControllerManager && !c.DisableScheduler) {
 		componentManager.Add(&controller.Konnectivity{
 			ClusterConfig:     c.ClusterConfig,
 			LogLevel:          c.Logging["konnectivity-server"],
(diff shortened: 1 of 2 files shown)


jasmingacic (dismissed):

soider:
While totally understanding the tech side of the PR, I don't understand the particular usecase. Is it possible to add some doc changes as well, with examples how and why to use this setup?

jakolehm:
> While totally understanding the tech side of the PR, I don't understand the particular usecase. Is it possible to add some doc changes as well, with examples how and why to use this setup?

With these kind of changes it would be possible to run k0s as a vcluster (https://github.com/loft-sh/vcluster) or similar.

But let's not merge this yet, I think we need to fine tune this PR a bit more.

soider:
@jakolehm ok. I convert it to the draft, to prevent accidental merge

jakolehm:
#1025 does overlap with this a bit.

cmd/controller/controller.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/avast/retry-go"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	funk "github.com/thoas/go-funk"
```
jnummelin, line 33: Do we really need this for anything else than checking if string slice contains some string? We already have that implemented internally: https://github.com/k0sproject/k0s/blob/64c1392c3914d908871a498e7a2c30b4b06a2aff/pkg/util/slice.go

cmd/controller/controller.go
```diff
@@ -230,24 +238,28 @@ func (c *CmdOpts) startController() error {
 			KubeClientFactory: adminClientFactory,
 		})
 	}
-	if !c.SingleNode {
+	if !c.SingleNode && !funk.Contains(c.DisableComponents, "konnectivity-server") {
```
jnummelin, line 241: Could we use constants for the component names?

pkg/config/cli.go
```diff
@@ -133,11 +135,30 @@ func GetWorkerFlags() *pflag.FlagSet {
 	return flagset
 }
 
+func AvailableComponents() []string {
+	return []string{
+		"konnectivity-server",
```
jnummelin, line 140: Maybe define these as consts?

jnummelin (changes requested):
AFAIK no need for go-funk as we've got the used functionality internally implemented already. Also I'd like to see the components names as constants to avoid typo errors

cmd/controller/controller.go
jakolehm: Indeed we should use our own ... I didn't see this because it's actually under `internal/` (not `pkg/`).

cmd/controller/controller.go
jakolehm: Probably yes, but looking at this specific file we haven't really used constants earlier. Will try to refactor.

pkg/config/cli.go
jakolehm: Done.

cmd/controller/controller.go
jakolehm: Refactored.

cmd/controller/controller.go
jakolehm: Done.

jnummelin (dismissed):

ncopa:
Only thing I want is smoke tests to make sure that we don't break this functionality.

mviitane:
The new feature "disable-components" should also be documented somewhere, for example:
https://docs.k0sproject.io/main/configuration/
https://docs.k0sproject.io/main/k0s-multi-node/

PR #1025, opened 2021-07-25
feat(konnectivity): add 'disabled' to allow disable konnectivity

mritd:
add 'disabled' to allow disable konnectivity

Signed-off-by: mritd <mritd@linux.com>

**What this PR Includes**

In some cases, we don't need the konnectivity service; 
this PR adds a 'disabled' option to disable the konnectivity service.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -219,7 +219,7 @@ func (c *CmdOpts) startController() error {
 		K0sVars:            c.K0sVars,
 		LogLevel:           c.Logging["kube-apiserver"],
 		Storage:            storageBackend,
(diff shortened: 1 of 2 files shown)


jasmingacic:
Thanks @mritd for the contribution
I'm marking this PR as draft until we decide if we are going to proceed with this.

mritd:
ok, thanks.

I found that once konnectivity is enabled in my environment, kubectl will respond slowly. 
I am still investigating the specific reasons, so I want to expect any third-party components to be turned off and enabled separately, such as konnectivity and metrics server.

ncopa:
#1023 Introduces `--disable-components=konnectivity-server` which should solve the same problem?

PR #1029, opened 2021-07-26
API Config as CustomResource

### LINKED ISSUE #552: Store current configuration to cluster
e.g. for cluster upgrades we need to know what was the previous version & config.

I'd like to be able to read version & config info to be able to plan and do automatic upgrades etc. operations. Maybe a custom global CRD to store all this info in a single place so it'll be readable for all needed tools.

trawler:
**Issue**
Fixes #552

**What this PR Includes**
This PR adds a CustomResourceDefinition to the cluster, called `ClusterConfig`. This custom resource accepts the same previously configured `k0s.yaml` with minor changes.

### To Do
- [x] Move from `gopkg.in/yaml.v2` to `sigs.k8s.io/yaml`
- [x] Create a client package for ClusterConfig object, with Get(), Create(), Delete() & List() methods, using `client-get`
- [x] Separate parsing of bootstrapping and non-bootstrapping (cluster) config
- [x] Reconciliation: if cluster config is provided, and CR does not exist, write the contents to the CR
- [x] Implement CR write only by a cluster leader
- [x] Add a reconcile method to all the components
- [x] Changes in CR triggers component reconciliation 
- [x] Reconfigure the cluster to use the CR and/or Node config when booting.
- [x] Separate cluster config reconciler to remove all dependencies with other components
- [x] Reconciliation: If cluster config is provided, but a CR exists, CR should override, unless a --force flag is provided.
- [ ] Adding an "enable" flag to the reconciler (disabled by default)
- [ ] Add CR validation directives for the different fields (optional)
- [ ] docs


s0j (approved):
SGTM

ncopa (approved):

PR #1031, opened 2021-07-27
Update containerd and appply go mod tidy

soider:
Some go.mod clean ups and updates to fix security alert


trawler (approved):

jasmingacic (approved):

go.mod
```diff
@@ -5,17 +5,23 @@ go 1.13
...
 	github.com/Masterminds/sprig v2.22.0+incompatible
-	github.com/Microsoft/hcsshim v0.8.7
+	github.com/Microsoft/go-winio v0.5.0 // indirect
+	github.com/Microsoft/hcsshim v0.8.20
+	github.com/Microsoft/hcsshim/test v0.0.0-20210726222751-a8f5f611a659 // indirect
 	github.com/asaskevich/govalidator v0.0.0-20200428143746-21a406dcc535
 	github.com/avast/retry-go v2.6.0+incompatible
 	github.com/cloudflare/cfssl v1.4.1
-	github.com/containerd/containerd v1.4.1
+	github.com/containerd/containerd v1.5.4
```
ncopa, line 14: I wonder if we should use containerd 1.4.9 instead

ncopa (changes requested):
I think we should use containerd 1.4.9

We should also update the containerd in the replace section.

go.mod
ncopa: i think it is helm 3.6.0 that pulls in containerd 1.5.4

the help update is also breaking the windows build

ncopa:
superseded by #1055

PR #1038, opened 2021-07-30
Add spec.api.bindAddress configuration

### LINKED ISSUE #957: Changing "bind-address" of kube-api
I have systems with multiple network interfaces and I would like to initialize the cluster on only one of the interfaces, say 192.168.1.1.

## TLDR
Adding `spec.api.extraArgs.bind-address = 192.168.1.1` (and adding to sans) in k0s config successfully causes the kube-api server to only bind on 192.168.1.1. However, when changing the `/var/lib/k0s/pki/admin.conf`s `clusters[0].cluster.server = https://192.168.1.1:6443`, `k0s token create --role=worker` results in a `TLS handshake timeout`. 

## The rest
This issue is more a question of how to configure k0s to achieve what I am looking to do. I don't want the kube-api server listening on every interface in this case, but changing the bind-address breaks being able to access the cluster. I know it is a certificate issue, I'm just not sure where.

Here is the relevant section in my k0s.yaml file:
```yaml
spec:
  api:
    address: 192.168.1.1
    port: 6443
    k0sApiPort: 9443
    externalAddress: 192.168.1.1
    extraArgs:
      bind-address: 192.168.1.1
    sans:
    - 192.168.1.1
```

For context, this is a single node control plane cluster that manages airgapped nodes that are behind a specific network interface.

gakio12:
Should a new `bindAddress` option be created and change documentation for the existing `address` option, or should the existing `address` option's behavior be changed to match the existing documentation?

The later I can see causing some issues with existing setups, `bindAddress` is probably the safest route. Currently the default for `bindAddress` is "0.0.0.0", which is the default kube-api-server uses if the `--bind-address` flag isn't used.

**Issue**
Fixes #957

**What this PR Includes**
Adds the ability to define a `bindAddress` for the kube-api-server.

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
@@ -56,6 +56,7 @@ func DefaultAPISpec() *APISpec {
 	return &APISpec{
 		Port:                   6443,
 		K0sAPIPort:             9443,
+		BindAddress:            "0.0.0.0",
 		SANs:                   addresses,
 		Address:                publicAddress,
 		ExtraArgs:              make(map[string]string),
@@ -98,6 +99,20 @@ func (a *APISpec) getExternalURIForPort(port int) string {
 	return fmt.Sprintf("https://%s:%d", addr, port)
 }
 
+// APIServerAddress returns the address the API is listening on
+func (a *APISpec) APIServerAddress() string {
+	return a.getAPIServerAddress(a.BindAddress)
+}
+
+func (a *APISpec) getAPIServerAddress(address string) string {
+	switch address {
+	case "0.0.0.0":
+		return "localhost"
+	default:
+		return address
+	}
+}
+
 // Sans return the given SANS plus all local adresses and externalAddress if given
 func (a *APISpec) Sans() []string {
 	sans, _ := iface.AllAddresses()
@@ -126,6 +141,10 @@ func (a *APISpec) Validate() []error {
 
 	if !govalidator.IsIP(a.Address) {
 		errors = append(errors, fmt.Errorf("spec.api.address: %q is not IP address", a.Address))
+	}
+
+	if !govalidator.IsIP(a.BindAddress) {
+		errors = append(errors, fmt.Errorf("spec.api.bindAddress: %q is not IP address", a.BindAddress))
 	}
 
 	return errors
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
@@ -32,6 +32,9 @@ type APISpec struct {
 	// Local address on which to bind an API
 	Address string `json:"address"`
 
+	// Local address on which to bind the kube-apiserver. Defaults to 0.0.0.0.
+	BindAddress string `json:"bindAddress,omitempty"`
+
 	// The loadbalancer address (for k0s controllers running behind a loadbalancer)
 	ExternalAddress string `json:"externalAddress,omitempty"`
 	// TunneledNetworkingMode indicates if we access to KAS through konnectivity tunnel
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/api.go
@@ -108,18 +108,14 @@ func (a *APISpec) APIServerAddress() string {
 }
 
 func (a *APISpec) getAPIServerAddress(address string) string {
-	switch address {
-	case "0.0.0.0":
-		return "localhost"
-	default:
-		return address
-	}
+	return address
 }
 
 // Sans return the given SANS plus all local adresses and externalAddress if given
 func (a *APISpec) Sans() []string {
 	sans, _ := iface.AllAddresses()
 	sans = append(sans, a.Address)
+	sans = append(sans, a.BindAddress)
 	sans = append(sans, a.SANs...)
 	if a.ExternalAddress != "" {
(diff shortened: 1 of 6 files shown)


jnummelin:
@gakio12 seems there's some unit test failures

trawler:
Hey @gakio12 please rebase with the current code updates. Thanks!

jnummelin:
The PR is marked as stale since no activity has been recorded in 30 days

gakio12:
@jnummelin Is there anything more I need to do on this PR? I want to make sure I am not missing anything.

pkg/apis/k0s.k0sproject.io/v1beta1/api.go
```diff
@@ -95,6 +99,20 @@ func (a *APISpec) getExternalURIForPort(port int) string {
...
 }
 
+// APIServerAddress returns the address the API is listening on
+func (a *APISpec) APIServerAddress() string {
+	return a.getAPIServerAddress(a.BindAddress)
+}
+
+func (a *APISpec) getAPIServerAddress(address string) string {
+	switch address {
+	case "0.0.0.0":
```
soider, line 109: this is logically wrong, "0.0.0.0" means we want to listen on all interfaces available, while "localhost" means listening only on loopback iface. What it the logic behind that decision?

soider (changes requested):

pkg/apis/k0s.k0sproject.io/v1beta1/api.go
gakio12: The reason it is "0.0.0.0" here is because without specifying an interface (as the current version of k0s does), the kube-apiserver listens on 0.0.0.0. So the default is "0.0.0.0" to allow the functionality to remain the same as previous versions.

The reason we return "localhost" in `getAPIServerAddress` follows the same reason as the above, k0s uses a hardcoded "localhost" in the kubeconfig and other certs for the kube-apiserver in the current released version.

With this change, it keeps the default functionality of k0s, while allowing someone to change the bind address and still have the certs and kubeconfig be correct (a kubeconfig pointing at "localhost" when the kube-apiserver is listening on 192.168.1.1 doesn't work).

jnummelin: > since you can't connect to 0.0.0.0 directly

well, it turns out that one actually can do that as in many (most?) languages connecting to `0.0.0.0` actually means connecting to localhost. Now that this address is used for kube components which are all Golang, I think it should work if they connect to `0.0.0.0`. 

For reference: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/ipsock_posix.go;l=137-143

With this in mind, everything _should_ work if we use the bind address as the address for other components to connect to the API, right?

gakio12: Huh, I was not aware of that. That is good to know! Since this string is used for health checks and a kubeconfig, an easy test is to take an existing cluster and change the kubeconfig to point to `0.0.0.0`. The certificate will have to be signed with `0.0.0.0` as well, since "localhost" resolves to `127.0.0.1`. Just doing that quick test I outlined, I got the expected certificate error.

I'll do a bit more testing on this and see if this logic can be removed while maintaining how k0s currently functions.

soider: @gakio12 hello! do you have any updates on the issue?

ocraviotto:
@gakio12 Just curious, are you still planning to work on this? I'd need this to make my setup work as well, so was simply wondering what your plans were to decide on whether to wait. (Among other things, this would then need an addition to k0sctl to allow setting the BindAddress per controller etc.)

pkg/apis/k0s.k0sproject.io/v1beta1/api.go
gakio12: @soider Apologies for the delay on this, I am expecting to be able to work on this in the next few weeks. I have the changes in place locally, but I have lost my test environment the last two months, so was unable to test (apart from time constraints).

gakio12:
@ocraviotto Yes, I am planning to commit the changes proposed in the next few weeks now that I have gained some free time (and got my environment back).

pkg/apis/k0s.k0sproject.io/v1beta1/api.go
ocraviotto: My 2 cents here: the actual meaning of "0.0.0.0" depends on how it is used. When used to bind, it means all (or any) interfaces, but when used as a client, it means "this" host (or even more strongly, an "[unspecified](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/ip.go;l=117-121;drc=refs%2Ftags%2Fgo1.17.3)" host, and as pointed above, is interpreted to mean "localhost" or the loopback interface, at least in various cases (including the [apiserver self client](https://github.com/kubernetes/apiserver/blob/7f74d873f30ca894d2caf8eaa3e691a26740a520/pkg/server/config_selfclient.go#L67-L97))
So IMHO, although the address does seem to work (albeit with a failure as it's not in the cert SAN), it seems better to me to keep the old behavior and use "localhost" in this case.

mtb-xt:
Any chance this can be completed and merged before 1.25? :eyes:

jnummelin:
> Any chance this can be completed and merged before 1.25? 👀

Unfortunately that did not happen as the upstream konnectivity did not have this in any released version at that timeframe.

Hopefully gets into 1.26

kke:
This has been open for quite a while.

pschichtel:
@kke I rebased this PR and submitted as #3824.

Issue #1016, opened 2021-07-16
k0s extract


ncopa:
**Is your feature request related to a problem? Please describe.**
Would be nice to be able to have k0s extract the embedded binaries without any running anything. Possibly also generate the certificates.

**Describe the solution you'd like**
either: `k0s extract` or `k0s install --extract`.

jnummelin:
The issue is marked as stale since no activity has been recorded in 30 days

PR #1046, opened 2021-08-09
Bundle statically built iptables

### LINKED ISSUE #271: Make it possible to run k0s with only the kernel as a dependency
As MKE should be able to run on any Linux kernel, we don't want any hard dependencies. Currently, the `kubelet` has some hard dependencies, as documented in https://github.com/Mirantis/mke/blob/main/docs/internal/host-dependencies.md.

We should aim to remove all non-kernel dependencies, either through bundling more things into mke or, preferably, contribute to upstream where possible to remove those dependencies.

Spawned from #176

ncopa:
**Issue**
Partially fixes #271

**What this PR Includes**
Bundles statically built iptables with needed symlinks. 

With this PR there will be no external dependencies with kubernetes 1.22.

### DIFF
--- a/embedded-bins/Makefile
+++ b/embedded-bins/Makefile
@@ -7,7 +7,7 @@ TARGET_OS ?= linux
 export TARGET_OS
 
 bindir = staging/${TARGET_OS}/bin
-posix_bins = runc kubelet containerd containerd-shim containerd-shim-runc-v1 containerd-shim-runc-v2 kube-apiserver kube-scheduler kube-controller-manager etcd kine konnectivity-server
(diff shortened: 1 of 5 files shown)


kke:
🤔 k0sctl needs iptables to resolve host addresses. Maybe it could use it through k0s in the future.

jasmingacic (approved):

ncopa:
> 🤔 k0sctl needs iptables to resolve host addresses. Maybe it could use it through k0s in the future.

Maybe via a `k0s extract`?  https://github.com/k0sproject/k0s/issues/1016

PR #1049, opened 2021-08-12
bump kubernetes version to 1.22.0

jasmingacic:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.22.md

Signed-off-by: Jasmin Gacic <jasmin.gacic@gmail.com>

**What this PR Includes**
Update kubernetes to 1.22.0


jnummelin:
[1.22 removes](https://kubernetes.io/blog/2021/08/04/kubernetes-1-22-release-announcement/#removal-of-several-deprecated-beta-apis) some long deprecated API groups which I think some of the "addons" are still using. @jasmingacic pls take a look all the manifests we have embedded and "migrate" to new versions

ncopa:
I wonder if we should upgrade containerd to 1.5.x and etcd to 3.5.x at the same go.

jnummelin:
> I wonder if we should upgrade containerd to 1.5.x and etcd to 3.5.x at the same go.

etcd to 3.5.0 for sure

ncopa (approved):

ncopa:
> > I wonder if we should upgrade containerd to 1.5.x and etcd to 3.5.x at the same go.
> 
> etcd to 3.5.0 for sure

Talked with @jasmingacic and we decided to do etcd 3.5.0 upgrade in separate PR.

I did some investigation and it appears that the backup/restore needs to be refactored, since the API for v3.5.0 changed dramatically.

PR #1051, opened 2021-08-12
Add documentation of configuring environment variables

### LINKED ISSUE #754: Setup cluster behind a proxy
I can't set http_proxy, https_proxy, no_proxy for containerd. I have solution: set Environment in k0scontroller.service systemd unit. But this could be done after installing cluster with k0s/k0sctl. I need to install this during cluster installation. Is such a solution possible now or will it be added in the next version?

jewertow:
**Issue**
Fixes #754 

**What this PR Includes**
Simple documentation how to configure environment variable for components used by k0s on systems with SystemD and OpenRC.

**How to test**
[Here](https://github.com/jewertow/k0s-sandbox/tree/master/cluster-behind-a-proxy) is a sandbox environment with k0s behind a forward proxy where containerd has configured `HTTP_PROXY` in accordance with this documentation.

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1,0 +1,24 @@
+# Environment variables
+
+`k0s install` does not support environment variables.
+Setting environment variable for a component used by k0s depends on used init system.
+
+## SystemD
+
+Create a drop-in directory and add config file with a desired environment variable:
+
+```shell
+mkdir -p /etc/systemd/system/k0scontroller.service.d
+tee -a /etc/systemd/system/k0scontroller.service.d/http-proxy.conf <<EOT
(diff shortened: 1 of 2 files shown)


kke:
if #1054 would be implemented, this wouldn't be needed.

ncopa (approved):

jewertow:
> if #1054 would be implemented, this wouldn't be needed.

Even if `k0s install` will support env variables, this documentation might be useful when a user wants to add some env variables after installation of a cluster.

jnummelin (approved):

jnummelin:
Thanks for the contribution @jewertow 🎉

PR #1055, opened 2021-08-18
Bump etcd to 3.5.0; kube 0.22.0

jnummelin:
**Issue**
Kube 1.22 suggests etcd 3.5.0 as the main version to be used. 

**What this PR Includes**
This PR bump etcd to 3.5.0 both at the embedded bin and also all the go modules we use for etcd.

As usual with go mods this meant also that we need to update several other packages like all the kube stuff for example.

## TODO

- [x] figure out proper fixes for the k0s-cloud-provider parts


ncopa:
i think we may need that upstream fixes https://github.com/helm/helm/pull/9970 for us to be able to use new helm version with containerd 1.5

ncopa (approved):

PR #1064, opened 2021-08-23
logrus: standardize logging throughout k0s packages

trawler:

**What this PR Includes**
Chore: This PR creates loggers for k0s log files and k0s cli tools to the util package, so that we have standard loggers defined for these two types (unlike the custom loggers we had for different packages).

### DIFF
--- a/cmd/airgap/listimages.go
+++ b/cmd/airgap/listimages.go
@@ -18,6 +18,7 @@ package airgap
 import (
 	"fmt"
 
+	"github.com/k0sproject/k0s/internal/util"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
@@ -33,10 +34,12 @@ func NewAirgapListImagesCmd() *cobra.Command {
 		Short:   "List image names and version needed for air-gap install",
 		Example: `k0s airgap list-images`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// we don't need warning messages in case of default config
-			logrus.SetLevel(logrus.ErrorLevel)
 			c := CmdOpts(config.GetCmdOpts())
-			cfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
+			// we don't need warning messages in case of default config
+			c.Logger = util.CLILogger()
+			c.Logger.SetLevel(logrus.ErrorLevel)
+
+			cfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars, c.Logger)
 			if err != nil {
 				return err
 			}
--- a/cmd/kubeconfig/admin.go
+++ b/cmd/kubeconfig/admin.go
@@ -16,12 +16,10 @@ limitations under the License.
 package kubeconfig
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
 
-	"github.com/cloudflare/cfssl/log"
 	"github.com/k0sproject/k0s/internal/util"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/spf13/cobra"
@@ -37,20 +35,21 @@ func kubeConfigAdminCmd() *cobra.Command {
 	$ kubectl get nodes`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := CmdOpts(config.GetCmdOpts())
+			c.Logger = util.CLILogger()
 			if util.FileExists(c.K0sVars.AdminKubeConfigPath) {
 				content, err := ioutil.ReadFile(c.K0sVars.AdminKubeConfigPath)
 				if err != nil {
-					log.Fatal(err)
+					c.Logger.Fatal(err)
 				}
 
 				clusterAPIURL, err := c.getAPIURL()
 				if err != nil {
-					return fmt.Errorf("failed to fetch cluster's API Address: %w", err)
+					c.Logger.Fatalf("failed to fetch cluster's API Address: %v", err)
 				}
 				newContent := strings.Replace(string(content), "https://localhost:6443", clusterAPIURL, -1)
 				os.Stdout.Write([]byte(newContent))
 			} else {
-				return fmt.Errorf("failed to read admin config, check if the control plane is initialized on this node")
+				c.Logger.Fatal("failed to read admin config, check if the control plane is initialized on this node")
 			}
 			return nil
 		},
--- a/cmd/validate/validate.go
+++ b/cmd/validate/validate.go
@@ -16,7 +16,7 @@ limitations under the License.
 package validate
 
 import (
-	"fmt"
+	"github.com/k0sproject/k0s/internal/util"
 
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/spf13/cobra"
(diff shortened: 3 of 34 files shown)


cmd/airgap/listimages.go
```diff
@@ -33,10 +34,12 @@ func NewAirgapListImagesCmd() *cobra.Command {
 		Short:   "List image names and version needed for air-gap install",
 		Example: `k0s airgap list-images`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// we don't need warning messages in case of default config
-			logrus.SetLevel(logrus.ErrorLevel)
 			c := CmdOpts(config.GetCmdOpts())
-			cfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
+			// we don't need warning messages in case of default config
+			c.Logger = util.CLILogger()
```
s0j, line 39: Setting `c.Logger = util.CLILogger()` in `RunE` seems inconsistent in these changes (intentionally?).  I see that `listimages.go` sets it, but others like `etcd.go` don't.  Is this intentional?

internal/util/logger.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/sirupsen/logrus"
+)
+
+// CLILogger returns a logger with timestamps disabled & colors
+// Suitable for log output for CLI Operations
+func CLILogger() *logrus.Logger {
+	logger := logrus.New()
+	textFormatter := new(logrus.TextFormatter)
+	textFormatter.ForceColors = true
```
s0j, line 29: Doesn't this force colors even on non-TTYs?  (ie. ANSI-codes in piped log output)

cmd/airgap/listimages.go
trawler: CLILogger is only used when the command is a distinctive CLI command (unlike controller, api and worker commands that will normally need a more detailed log output, for example).
As for this specific use case (etcd command, which is a CLI command) I updated the PR.

internal/util/logger.go
trawler: Apparently it does :). Fixed.

cmd/kubeconfig/admin.go
```diff
@@ -37,6 +37,7 @@ func kubeConfigAdminCmd() *cobra.Command {
 	$ kubectl get nodes`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := CmdOpts(config.GetCmdOpts())
+			c.Logger = util.CLILogger()
```
s0j, line 40: Shouldn't the "fatal" error handling a few lines below also use `c.Logger`?

cmd/validate/validate.go
```diff
@@ -42,7 +44,8 @@ func validateConfigCmd() *cobra.Command {
    k0s validate config --config path_to_config.yaml`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := CmdOpts(config.GetCmdOpts())
-			_, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
+			c.Logger = util.CLILogger()
+			_, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars, c.Logger)
 			if err != nil {
 				fmt.Println(err)
```
s0j, line 50: Replace the `fmt.Println(err)` with `logrus`?

cmd/kubeconfig/admin.go
s0j, line 40: Shouldn't the "fatal" error handling a few lines below also use c.Logger?

cmd/kubeconfig/admin.go
```diff
@@ -37,20 +35,21 @@ func kubeConfigAdminCmd() *cobra.Command {
...
 				content, err := ioutil.ReadFile(c.K0sVars.AdminKubeConfigPath)
 				if err != nil {
-					log.Fatal(err)
+					c.Logger.Fatal(err)
 				}
 
 				clusterAPIURL, err := c.getAPIURL()
 				if err != nil {
-					return fmt.Errorf("failed to fetch cluster's API Address: %w", err)
+					c.Logger.Fatalf("failed to fetch cluster's API Address: %v", err)
```
s0j, line 47: Are you intentionally relying on `logrus.Fatalf()` exiting the process instead of returning `error` to the caller?  There are a number of these changes in this latest diff.

trawler: I adopted this logic from https://github.com/k0sproject/k0s/blob/main/pkg/backup/manager.go#L117
I thought it makes the code clearer, and applied it to cases where we actually need the command to fail.

kke:
Random thoughts:

Having a "util" package is usually not a clean way. If everything that needs logging imports `util`, who knows what else comes in..

Passing a logger around seems a bit complex. You could have a package like `log` which would be a wrapper/abstraction around logrus and defines a global instance that can be configured upon cli/api initialization. You could then have something like:

```go
import "github.com/k0sproject/k0s/pkg/log"

func main() {
  log.Logrus.SetOutput(os.Stderr)
  log.Errorf("failed!")
}
```

The wrapper would look something like:

```go
import "github.com/something/logrus"

var Logrus = logrus.NewLogger()

func Debugf(msg, args ...interface{}) { Logrus.Debugf(msg, args...) }
func Infof(msg, args ...interface{}) { Logrus.Debugf(msg, args...) }
```

Some inspiration: 
* https://jhall.io/posts/6-logging-mistakes/ 
* https://medium.com/@jfeng45/go-microservice-with-clean-architecture-application-logging-b43dc5839bce
* https://gitlab.com/tight5/kit/blob/master/logger/logger.go
* https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

cmd/kubeconfig/admin.go
soider: I would rather keep returning error, because cobra handling it by it's own, by rendering help and etc.
Plus in general log.Fatal is dangerous practice (for example it doesn't respect defers).

trawler:
> Having a "util" package is usually not a clean way. If everything that needs logging imports `util`, who knows what else comes in..

I'm afraid that ship has already sailed :) Perhaps it's a topic for improvement in a later PR, but this PR is focused on the inconsistent logging issue.

> Passing a logger around seems a bit complex. You could have a package like `log` which would be a wrapper/abstraction around logrus and defines a global instance that can be configured upon cli/api initialization. 

Sometimes passing a logger is necessary. Consider the `GetYaml` function from the config package: This function is called all over the code, both by API and by CLI commands. The output logging for this function changes according to the calling tool. In which case, passing the calling logger is (IMO) necessary.
There is some truth to what you're saying, that passing the logger is not necessary in all cases, and initializing a local logger for many functions should suffice. In which case, I'll trim the PR to only pass a logger when it's necessary.

pkg/applier/stackapplier.go
```diff
@@ -47,14 +48,15 @@ func NewStackApplier(path string, kubeClientFactory kubernetes.ClientFactory) (*
 		return nil, err
 	}
 	applier := NewApplier(path, kubeClientFactory)
-	log := logrus.WithField("component", "applier-"+applier.Name)
-	log.WithField("path", path).Debug("created stack applier")
+	logger := util.K0sLogger()
+	logger.WithField("component", "applier-"+applier.Name)
```
s0j, line 52: This `component` field is going to be lost unless you either log or use the `Entry` that is returned from it.  You can use `WithFields()` for this.

Also, the `logger` you're passing to `StackApplier` won't have any of these fields.

See: https://github.com/sirupsen/logrus/issues/1039

trawler:
Since this issue requires a bit more planning, closing this PR in favour of issue  #1081.

PR #1069, opened 2021-08-24
Document for NGINX Ingress controller

mviitane:
**What this PR Includes**

- Created a new document for installing NGINX Ingress controller
- Few minor corrections

Signed-off-by: mviitanen <mviitanen@mirantis.com>


docs/examples/nginx-ingress.md
```diff
@@ -0,0 +1,229 @@
+# Installing NGINX Ingress Controller
```
s0j, line 1: I think we need to be overly explicit as to which ingress this is as there are two ingress products named with 'nginx' coming from two different entities.  This should probably be pointed out as well.

* https://kubernetes.github.io/ingress-nginx
  * this is the one you are using, which is named "NGINX Ingress Controller", maintained by kubernetes iirc.

* https://docs.nginx.com/nginx-ingress-controller/
  * this is also named "NGINX Ingress Controller", but maintained by nginx

Someone googling for additional examples and yaml could lead people in different configuration directions which would be bad.

mviitane: Yes, this is a bit messy. I'll try to clarify.

mviitane: Done. Updated line 3.

kke (approved):

PR #1072, opened 2021-08-25
Export workflows go version to common Repo var

trawler:

**What this PR Includes**
This PR export the go-version variable to a common repo variable, so that we have only one place to update when go-version needs to be changed.


trawler:
closing, as secrets cannot be fetched by remote branches...

PR #1073, opened 2021-08-25
Export workflows go version to common Repo var

trawler:

**What this PR Includes**
This PR export the GO_VERSION variable to an external file, so that we only need to change it in one location, when updating it.


jnummelin:
I really like the idea of having go version (and probably some other stuff to in future) defined only once.

The only "complaint" I have is the fact that with this specific approach we need to add a step to each job always to read the env.

What if we define an env variable at the workflow level and then use it in the steps, something like:
```
name: Go build
env:
  GO_VERSION: "^1.16"
  ...
  
  jobs:
  build:
    name: Build
    runs-on: ubuntu-latest
    steps:
    ...
      - name: Set up Go 1.x
        uses: actions/setup-go@v2
        with:
          go-version: ${{ env.GO_VERSION }}
        id: go
...
```

trawler:
@jnummelin I applied your suggestion.
I wanted to also test the option of using our own custom GitHub action to collect the joint vars we're using throughout the workflows, but the testing takes longer than expected, so that will have to wait to another PR :)

PR #1103, opened 2021-09-05
Fix k0s reset containerD startup

### LINKED ISSUE #1102: containerd upgrade broke k0s reset
The update to containerd in  #1031 now breaks k0s reset.
The new containerd version validates the config.toml that is being started and therefore the startup command with the flag `--config=/etc/k0s/containerd.toml` will now fail to start containerd if the file doesn't exist.
Failure in starting containerd for cleanup results in zombie containers being left on the system.

trawler:

**Issue**
Fixes #1102

**What this PR Includes**
ContainerD will not start when a config flag points to a file that does not exist.
This PR checks if the custom config file exists, and will not add the `--config` flag if it doesn't.
It also adds an extra test to k0s reset command to test also number of pods left after reset completes (under CLI Test).


jnummelin:
We need to figure out how to catch this type of issue in a CI smoke test. E.g. we already run `k0s reset ...` in `singlenode` smoke, it _should've_ caught this issue but apparently it did not 🤦

trawler:
> We need to figure out how to catch this type of issue in a CI smoke test. E.g. we already run `k0s reset ...` in `singlenode` smoke, it _should've_ caught this issue but apparently it did not 🤦

Apparently the singlenode install test was left out of the workflow file in favor of cli test (which did not include the reset command). I added it to the CLI test, and removed the unused one.

ncopa (approved):

Issue #1087, opened 2021-09-02
[Feature Request] Ability to disable PodSecurityPolicy admission plugin, or maybe disable it by default?


yuha0:
**Is your feature request related to a problem? Please describe.**

Congrats on the new `v1.22.1+k0s.0` release. Now with the ability to disable optional system components, I can disable k0s predefined PSPs, but I would like to further disable PSP admission plugin completely. It is deprecated in Kubernetes `1.21` already and will be removed in `1.25`, so I think it is reasonable to start thinking about the path forward.

I don't want to have PSP in my cluster, but if I add `--disable-components default-psp` to `k0s`, it gives me a crippled cluster where system components like CNI pods complains about lack of PSP and cannot construct network.

Is it possible to implement a new configuration field to give user ability to explicitly disable it?

**Describe the solution you'd like**
Keep PSP enabled by default for now, but give users ability to explicitly disable it. And maybe in a later version, disable it by default while allowing users to explicitly enable it, until `1.25` comes out.

jnummelin:
> It is deprecated in Kubernetes 1.21 already and will be removed in 1.25, so I think it is reasonable to start thinking about the path forward.

You are totally correct here, we (as in the community around k0s) need to think what we want to do with the topic. Personally I see three options:
- keep current setup as-is and wait till PodSecurity admission matures (in alpha currently IIRC)
- drop pre-setup PSP/PodSecurity setup alltogether from k0s core and let users use whatever means they want
- bundle in "external" policy controller such as OPA, Kyverno, jsPolicy, ...

yuha0:
@jnummelin thank you. I missed that part in the documentation.

yuha0:
> You should be able to disable PSP completely using something like this as the config

Actually, it doesn't seem to work. If I set `disable-admission-plugins: PodSecurityPolicy`, API server refuses to start:

```
Sep  3 03:05:24 pi0 k0s[5111]: time="2021-09-03 03:05:24" level=info msg="Error: [PodSecurityPolicy] in enable-admission-plugins and disable-admission-plugins overlapped" component=kube-apiserver
```

And I cannot override the `enable-admission-plugins` flag:

```
Sep  3 03:07:53 pi0 k0s[6877]: time="2021-09-03 03:07:53" level=error msg="failed to start controller components: cannot override apiserver flag: enable-admission-plugins"
```

jnummelin:
[...]

PR #1105, opened 2021-09-06
Allow user to overwrite ANY flags with 'extraArgs'

### LINKED ISSUE #1104: k0s should allow to override ANY flags for kube-apiserver (and other components)
Originally reported as #1087 

**Version**
```
1.22.1+k0s.0
```
**What happened?**

Not able to overwrite api-server flags which k0s sets:
```
Sep  3 03:07:53 pi0 k0s[6877]: time="2021-09-03 03:07:53" level=error msg="failed to start controller components: cannot override apiserver flag: enable-admission-plugins"

```

**How To Reproduce**
Spin up controller using config that sets `extraArgs` values for some flags k0s also sets. E.g.:

```
spec:
  api:
    extraArgs:
      disable-admission-plugins: PodSecurityPolicy
```

**Expected behavior**
I'm expecting to be allowed to "overwrite" any api-server flags with:
```
spec:
  api:
    extraArgs:
      disable-admission-plugins: PodSecurityPolicy
      foo: bar
```

jnummelin:

**Issue**
Fixes #1104 

**What this PR Includes**
Currently k0s will error out if user tries to overwrite any flags k0s itself sets for kube-apiserver, controller-manager or scheduler. This PR makes k0s to "just" warn about that and still allows user to do the expected.

The most obvious case #1087 is now also tested as smoke test.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -865,6 +865,43 @@ jobs:
           path: |
             /tmp/*.log
 
+  smoketest-extraargs:
+    name: Smoke test for extra args for kube components
+    needs: build
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
+        id: go
+      - name: Get PR Reference and Set Cache Name
+        run: |
+          PR_NUMBER=$(echo ${GITHUB_REF} | cut -d / -f 3 )
+          echo "cachePrefix=k0s-${PR_NUMBER}-${{ github.sha }}" >> $GITHUB_ENV
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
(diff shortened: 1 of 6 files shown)


.github/workflows/go.yml
```diff
@@ -865,6 +865,43 @@ jobs:
           path: |
             /tmp/*.log
 
+  smoketest-extraargs:
```
soider, line 868: I start thinking that might be we need to have codegen tool for adding new e2e test suites 🙂

soider (approved):

.github/workflows/go.yml
jnummelin: I'm thinking if we could have something that "triggers" each smoke as a separate workflow or something. So basically after the build artifact is store in cache a separate step could then kick off each smoke as a separate workflow.

kke (approved):

PR #1110, opened 2021-09-07
Document needed kernel configs

### LINKED ISSUE #1092: Document needed kernel modules
We've seen some cases where the host OS is ocnfigured with minimal available kernel config. This has made some of the system components fail to properly bootup.

We should document the needed kernel modules that the core k0s (and all the system components) need to be enabled.

We already have the `k0s sysinfo` subcommand that does check the modules, we should also document these.

jnummelin:

**Issue**
Fixes #1092 

**What this PR Includes**
Adds docs for needed kernel configs/modules

### DIFF
--- a/docs/os-deps.md
+++ b/docs/os-deps.md
@@ -1,0 +1,83 @@
+# OS dependencies
+
+k0s is packaged as a single binary, which includes all the needed components. All the binaries are statically linked which means in typical use cases there's no OS level dependencies.
+
+However some of the underlying components _may_ have dependencies to OS level tools and packages in certain circumstances. The known cases are documented in below.
+
+## Kernel configuration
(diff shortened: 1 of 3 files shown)


docs/os-deps.md
```diff
@@ -0,0 +1,83 @@
+# OS dependencies
+
+k0s is packaged as a fully statically linked binary which contains all other needed components also as statically linked binaries. This means that in typical use cases there's no OS level dependencies.
```
mviitane, line 3: The first sentence seems a bit complex. Could it be simplified in some way, for example:
"k0s is packaged as a single binary, which includes all the needed components as statically linked."

mviitane (approved):

PR #1111, opened 2021-09-07
Add empty config file for containerd by default

### LINKED ISSUE #1095: Having a docker-configured ContainerD config.toml on a system breaks kubelet startup
**Version**
```
$ v1.22.1+k0s and up
```
**What happened?**
When running k0s worker (or controller with a worker enabled set-up), kubelet fails to start with the following error:
```
Sep 02 12:56:02 controller-0 k0s[2007862]: time="2021-09-02 12:56:02" level=info msg="E0902 12:56:02.139045 2009500 server.go:294] \"Failed to run kubelet\" e
rr=\"failed to run Kubelet: failed to create kubelet: get remote runtime typed version failed: rpc error: code = Unimplemented desc = unknown service runtime.
v1alpha2.RuntimeService\"" component=kubelet
Sep 02 12:56:02 controller-0 k0s[2007862]: time="2021-09-02 12:56:02" level=warning msg="exit status 1" component=kubelet
```

**Root Cause**
This problem only happens in systems that have containerd instance that's originally installed by `docker` and contains the following line:
```
disabled_plugins = ["cri"]
```
Since docker doesn't need CRI, it was disabled in https://github.com/docker/engine/pull/29, but kubelet does.
When starting k0s, containerd picks up the default `config.toml` created by the docker installation, and kubelet startup breaks. By removing the docker-created config.toml, or by enabling the cri plugin, kubelet starts up as expected.

The reason this is happening now, is that kubernetes just now implemented the deprecation of docker-shim that was announced back in December: https://kubernetes.io/blog/2020/12/02/dockershim-faq/#when-will-dockershim-be-removed

**How To Reproduce**
Run k0s version 1.22 and up on a system that has docker installed.

**Workaround**

The temporary workaround is to create separate containerd config before starting k0s
```
mkdir /etc/k0s
touch /etc/k0s/containerd.toml 
```
This trigger k0s to use the found file as the config for containerd as described in https://docs.k0sproject.io/v1.22.1+k0s.0/runtime/#containerd-configuration

And as the file is empty it'll make containerd to use the default config as k0s would expect anyways.

jnummelin:

**Issue**
Fixes #1095

**What this PR Includes**
This makes containerd to use default configuration rather than trying to load the system-wide (possibly from docker setup) configuration.

### DIFF
--- a/pkg/component/worker/containerd.go
+++ b/pkg/component/worker/containerd.go
@@ -23,10 +23,20 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 
+	dirutil "github.com/k0sproject/k0s/internal/pkg/dir"
+	fileutil "github.com/k0sproject/k0s/internal/pkg/file"
 	"github.com/k0sproject/k0s/pkg/assets"
 	"github.com/k0sproject/k0s/pkg/constant"
 	"github.com/k0sproject/k0s/pkg/supervisor"
 )
+
+const confTmpl = `
+# This is a placeholder configuration for k0s managed containerD.
+# If you wish to customize the config replace this file with your custom configuration.
+# For reference see https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md
+version = 2
+`
+const confPath = "/etc/k0s/containerd.toml"
 
 // ContainerD implement the component interface to manage containerd as k0s component
 type ContainerD struct {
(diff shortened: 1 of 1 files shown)


s0j (approved):

ncopa (approved):
I don't know if we also should remove it with `k0s reset`?

jnummelin:
> I don't know if we also should remove it with k0s reset?

hmm, maybe not as it could be user managed and thus we might wipe out user managed stuff.

PR #1120, opened 2021-09-13
fix disabling konnectivity-server

xinfengliu:
**What this PR Includes**

- Disabling konnectivity-server also needs disabling `egress-selector-config-file` in `kube-apiserver` args.
- ~If the controller is not enabled workloads, then ignore disabling `konnectivity-server` and log a warning message.~

Signed-off-by: Xinfeng Liu <xinfeng.liu@gmail.com>

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -171,12 +171,13 @@ func (c *CmdOpts) startController() error {
 
 	// common factory to get the admin kube client that's needed in many components
 	adminClientFactory := kubernetes.NewAdminClientFactory(c.K0sVars)
+	enableKonnectivity := !c.SingleNode && !stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName)
 	c.NodeComponents.Add(&controller.APIServer{
 		ClusterConfig:      c.NodeConfig,
 		K0sVars:            c.K0sVars,
 		LogLevel:           c.Logging["kube-apiserver"],
 		Storage:            storageBackend,
-		EnableKonnectivity: !c.SingleNode,
+		EnableKonnectivity: enableKonnectivity,
 	})
 
 	if c.NodeConfig.Spec.API.ExternalAddress != "" {
@@ -294,7 +295,7 @@ func (c *CmdOpts) startController() error {
 		ch <- syscall.SIGTERM
 	}
 
-	if !c.SingleNode && !stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName) {
+	if enableKonnectivity {
 		c.ClusterComponents.Add(&controller.Konnectivity{
 			LogLevel:          c.Logging[constant.KonnectivityServerComponentName],
 			K0sVars:           c.K0sVars,


cmd/controller/controller.go
```diff
@@ -225,12 +225,17 @@ func (c *CmdOpts) startController() error {
 	// common factory to get the admin kube client that's needed in many components
 	adminClientFactory := kubernetes.NewAdminClientFactory(c.K0sVars)
 
+	if !c.EnableWorker && stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName) {
+		logrus.Warnf("ignoring disableComponents: %q, because kube workerloads are not enabled on this controller", constant.KonnectivityServerComponentName)
+	}
+	enableKonnectivity := !c.EnableWorker || (!c.SingleNode && !stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName))
```
jnummelin, line 231: is there any way to simplify this e.g. by changing the logic a bit. Having bunch of negated bools makes this super hard to read, at least for me. :)

xinfengliu: Sorry for that. How about using `disableKconnectivity` :
```go
disableKconnectivity := c.SingleNode || (c.EnableWorker && stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName))
...
if !disableKconnectivity  {
    ....
}
```

jnummelin: that's at least bit easier to read IMO :)

xinfengliu: I changed the logic a bit and pushed the code changes, hope it's easier to understand.

cmd/controller/controller.go
```diff
@@ -225,12 +225,23 @@ func (c *CmdOpts) startController() error {
 	// common factory to get the admin kube client that's needed in many components
 	adminClientFactory := kubernetes.NewAdminClientFactory(c.K0sVars)
 
+	enableKonnectivity := true
```
soider, line 228: inverted could be a little easier to read

like this (though I don't like a lot of inversions here)

```
var enableKonnectivity bool // false by default
if !c.SingleNode && !c.EnableWorker{
enableKonnectivity = !stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName)
}
```

xinfengliu: Hi @soider, in above code snippet, if  `--enable-worker` is true, `enableKonnectivity` will always be false, this is not expected. In addition, if `!c.SingleNode && !c.EnableWorker`,  `enableKonnectivity` should always be true, that's one of purposes of this PR.

cmd/controller/controller.go
xinfengliu: >I changed the logic a bit and pushed the code changes, hope it's easier to understand.

@jnummelin Is this code change OK for you or you want more edits? Please see https://github.com/k0sproject/k0s/commit/1fe7472907064461b2d37ad1e7be89478a05bdaf

cmd/controller/controller.go
```diff
@@ -225,12 +225,23 @@ func (c *CmdOpts) startController() error {
...
+	if c.SingleNode {
+		enableKonnectivity = false
+	} else if stringslice.Contains(c.DisableComponents, constant.KonnectivityServerComponentName) {
+		if c.EnableWorker {
+			enableKonnectivity = false
```
jnummelin, line 233: I think we should still just disable konnectivity is user has given the flag to disable it. There are some use cases where users want to run only the kube API and nothing else, in this case there's never a need for konnectivity at all.

xinfengliu: OK. Let's respect user's instructions.

xinfengliu:
Now the only code change for this PR is to make sure `konnectivity` related stuff is not set up for `kube-apiserver` when user disables `konnectivity`. Otherwise, kube-apiserver would fail to start.
https://github.com/k0sproject/k0s/blob/4d8535cb73fca2e47af93969b357b0ccea2b21c7/pkg/component/controller/apiserver.go#L119-L126

ncopa:
This needs to be re-based, sorry.

While at it, can you please also remove the two commits:
- fix disabling konnectivity-server
- Revert "fix disabling konnectivity-server"

In a PR it is better to remove a bad commit using `git rebase -i` rather than revert it.

xinfengliu:
>In a PR it is better to remove a bad commit using git rebase -i rather than revert it.

Sure. Done.

jnummelin (approved):

PR #1140, opened 2021-09-22
Cleanup refactor reset/install

ncopa:
**What this PR Includes**
Misc code cleanups and simplifications in reset and install commands.

In general, instead of test if something exist and then delete, try to delete immediately and handle the error.

No significant change in behavior.

### DIFF
--- a/pkg/cleanup/cni.go
+++ b/pkg/cleanup/cni.go
@@ -16,11 +16,6 @@ func (c *cni) Name() string {
 	return "CNI leftovers cleanup step"
 }
 
-// NeedsToRun checks if there are and CNI leftovers
-func (c *cni) NeedsToRun() bool {
-	return true
-}
-
 // Run removes found CNI leftovers
 func (c *cni) Run() error {
 	var msg []error
--- a/pkg/cleanup/cni.go
+++ b/pkg/cleanup/cni.go
@@ -1,17 +1,15 @@
 package cleanup
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 
-	"github.com/k0sproject/k0s/internal/pkg/file"
 	"github.com/sirupsen/logrus"
 )
 
-type cni struct {
-	Config   *Config
-	toRemove []string
-}
+type cni struct{}
 
 // Name returns the name of the step
 func (c *cni) Name() string {
@@ -20,34 +18,22 @@ func (c *cni) Name() string {
 
 // NeedsToRun checks if there are and CNI leftovers
 func (c *cni) NeedsToRun() bool {
+	return true
+}
+
+// Run removes found CNI leftovers
+func (c *cni) Run() error {
+	var msg []error
+
 	files := []string{
 		"/etc/cni/net.d/10-calico.conflist",
 		"/etc/cni/net.d/calico-kubeconfig",
 		"/etc/cni/net.d/10-kuberouter.conflist",
 	}
-
 	for _, f := range files {
-		if file.Exists(f) {
-			c.toRemove = append(c.toRemove, f)
-		}
-	}
-	return len(c.toRemove) > 0
-}
-
-// Run removes found CNI leftovers
-func (c *cni) Run() error {
-	return removeCNILeftovers(c.toRemove)
-}
-
-func removeCNILeftovers(files []string) error {
-	var msg []error
-
-	for _, f := range files {
-		if file.Exists(f) {
-			if err := os.Remove(f); err != nil {
-				logrus.Debug("failed to remove", f, err)
-				msg = append(msg, err)
-			}
+		if err := os.Remove(f); !errors.Is(err, fs.ErrNotExist) {
+			logrus.Debug("failed to remove", f, err)
+			msg = append(msg, err)
 		}
 	}
 	if len(msg) > 0 {
--- a/pkg/cleanup/containers.go
+++ b/pkg/cleanup/containers.go
@@ -26,11 +26,6 @@ func (c *containers) Name() string {
 	return "containers steps"
 }
 
-// NeedsToRun checks if custom CRI is used, otherwise checks if containerd is present on the host
-func (c *containers) NeedsToRun() bool {
-	return true
-}
-
 // Run removes all the pods and mounts and stops containers afterwards
 // Run starts containerd if custom CRI is not configured
 func (c *containers) Run() error {
--- a/pkg/cleanup/containers.go
+++ b/pkg/cleanup/containers.go
@@ -1,7 +1,9 @@
 package cleanup
 
 import (
+	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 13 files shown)


soider (dismissed):
great stuff

pkg/install/users.go
```diff
@@ -97,72 +98,21 @@ func EnsureUser(name string, homeDir string) error {
...
-	if len(cmd.Args) == 0 {
-		return fmt.Sprintf("%q", cmd.Path)
-	}
-
-	var q []string
-	for _, s := range cmd.Args {
-		q = append(q, fmt.Sprintf("%q", s))
+func deleteUser(userName string) error {
+	_, err := exec.Command("userdel", userName).Output()
+	if err != nil {
```
kke, line 112: Should this and adduser maybe do something like `if err != exec. ErrNotFound { return err }` to avoid running the other command if the failure in the first one was something else than NotFound?

pkg/cleanup/users.go
```diff
@@ -16,28 +16,12 @@ func (u *users) Name() string {
...
-	return false
-}
-
 // Run removes all controller users that are present on the host
 func (u *users) Run() error {
 	logger := logrus.New()
 	clusterConfig, err := config.GetYamlFromFile(u.Config.cfgFile, u.Config.k0sVars)
 	if err != nil {
-		logger.Errorf("failed to get cluster setup: %v", err)
+		return fmt.Errorf("failed to get cluster setup: %v", err)
```
kke, line 24: This changes the behavior, the function never returned an error even if some parts of it failed.

pkg/cleanup/containers.go
```diff
@@ -23,30 +24,19 @@ func (c *containers) Name() string {
...
 func (c *containers) Run() error {
 	if !c.isCustomCriUsed() {
 		if err := c.startContainerd(); err != nil {
-			logrus.Debugf("error starting containerd: %v", err)
-			return err
+			if os.IsNotExist(err) {
+				logrus.Debugf("containerd binary not found. Skipping container cleanup")
+				return nil
+			}
+			return fmt.Errorf("failed to start containerd: %v", err)
```
kke, line 36: I think these `return fmt.Errorf("something: %v", err")` are supposed to use `%w` instead of `%v` for some fancy error unwrapping stuff we don't probably use.

pkg/cleanup/cni.go
```diff
@@ -4,47 +4,28 @@ import (
...
-	var msg []error
-
 	for _, f := range files {
-		if file.Exists(f) {
-			if err := os.Remove(f); err != nil {
+		if err := os.Remove(f); err != nil {
+			if !os.IsNotExist(err) {
 				logrus.Debug("failed to remove", f, err)
 				msg = append(msg, err)
 			}
```
kke, line 31:
```suggestion
		if err := os.Remove(f); err != nil && !os.IsNotExist(err) {
			logrus.Debug("failed to remove", f, err)
			msg = append(msg, err)
```

pkg/install/users.go
ncopa: That would be great, but I don't know how to catch that error. The error does no match `os.IsNotExist(err)` nor `exec.ErrNotFound`. The error is `executable file not found in $PATH`, but I'm not sure we want compare against a text string.

pkg/cleanup/users.go
ncopa: The `clusterConfig` (`cfg` after recent rebase) value is undefined if there are errors and is likely nil, in which case it would segfault.

`install.DeleteControllerUsers(nil)` ends with a segfault.
```
INFO[2021-09-29 12:13:36] * containers steps                           
INFO[2021-09-29 12:13:36] * remove k0s users step:                     
INFO[2021-09-29 12:13:36] no config file given, using defaults         
[signal SIGSEGV: segmentation violation code=0x1 addr=0x128 pc=0x1b5e57a]

goroutine 1 [running]:
github.com/k0sproject/k0s/pkg/install.DeleteControllerUsers(0x0, 0x0, 0xc0003c7be0)
	/go/src/github.com/k0sproject/k0s/pkg/install/users.go:54 +0x3a
github.com/k0sproject/k0s/pkg/cleanup.(*users).Run(0xc000378870, 0x4, 0xc0005c3250)
	/go/src/github.com/k0sproject/k0s/pkg/cleanup/users.go:27 +0x20a
github.com/k0sproject/k0s/pkg/cleanup.(*Config).Cleanup(0xc000ebe4e0, 0x1b, 0xc0008771a0)
	/go/src/github.com/k0sproject/k0s/pkg/cleanup/cleanup_unix.go:89 +0x2f9
github.com/k0sproject/k0s/cmd/reset.(*CmdOpts).reset(0xc0005c3878, 0x0, 0x33b18a9)
	/go/src/github.com/k0sproject/k0s/cmd/reset/reset.go:74 +0x2ce
github.com/k0sproject/k0s/cmd/reset.NewResetCmd.func1(0xc000caaf00, 0x51d2df8, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/k0sproject/k0s/cmd/reset/reset.go:42 +0xc5
github.com/spf13/cobra.(*Command).execute(0xc000caaf00, 0x51d2df8, 0x0, 0x0, 0xc000caaf00, 0x51d2df8)
	/gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc000692c80, 0x407905, 0xc00010e058, 0x2b09990)
	/gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
	/gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:902
github.com/k0sproject/k0s/cmd.Execute()
	/go/src/github.com/k0sproject/k0s/cmd/root.go:206 +0x2a
main.main()
	/go/src/github.com/k0sproject/k0s/main.go:40 +0x25
```
So this is technically a bug fix.

I guess that should be mentioned in the commit message.

ncopa: Oh, now I see what you mean. It will result in `k0s reset` returns error. I guess if it fails to parse the config it should return failure, shouldn't it?

pkg/cleanup/containers.go
ncopa: fixed.

pkg/cleanup/cni.go
ncopa: Seems like we should use `errors.Is(fs.ErrNotExist)`

From https://pkg.go.dev/os#IsNotExist
> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist).

pkg/install/users.go
ncopa: Figured it out. `exec.ErrNotFound` is what we want.

pkg/cleanup/users.go
kke: I'm not sure. Maybe it should just skip trying to do whatever is based on configuration and just do what it can outside of it (remove files, uninstall services, whatever?)

kke (approved):

PR #1142, opened 2021-09-24
Cleanup file renamed 

makhov:
**What this PR Includes**

`cleanup_unix.go` file is renamed to `cleanup.go`. 
The naming convention is `_$GOOS.go` (also `_$GOARCH` and `_$GOOS_$GOARCH`, but it's not our case), and all supported values for GOOS is listed here:
https://github.com/golang/go/blob/master/src/go/build/syslist.go#L10

`_unix` is an invalid value (yet?) and the project doesn't compile on macOS (go1.16.3). And the file doesn't contain platform-specific code. 
Also, the bridge for other platforms was added to make the project compile and let the tests pass

**Note**: should be merged after #1140


xinfengliu:
Thanks for this PR. You are right, `_unix`.go is not a filter for go build, it is usually for group OSes which support common unix features. The content of `cleanup_unix.go` is for that purpose. A better alternative may be adding build tags (e.g. `env_unix.go` in go source) to `cleanup_unix.go` after all we don't want Windows version to use this file and there is a `cleanup_windows.go` there.

makhov:
> after all we don't want Windows version to use this file

But the Windows version does use it since `_unix.go` is not a filter. I'm not sure if `cleanup_windows.go` is really needed, since it's not used. Looks like, cleanup files are unified now and all the differences in `bridge_*.go` files.

xinfengliu:
It's not unified. It does not make sense to make windows use `unix://%s/containerd.sock` in `cleanup_unix.go`. I didn't test windows version of k0s and don't know what will happen to windows k0s in current k0s version.

makhov:
But it's already is used, since for Windows works completely the same code as for Unix, right?

xinfengliu:
>But it's already is used, since for Windows works completely the same code as for Unix, right?

Yes, it's already used by Windows. But I suspect there are problems with Windows version for cleaning-up.

makhov:
> But I suspect there are problems with Windows version for cleaning-up

Yeah, seems so. My original purpose was to fix builds on darwin, and unfortunately, I don't have any Windows machine to check how it works.

Let me think about how I can fix it in another way without touching `cleanup_unix.go`, but I'll leave a fix for Windows  out of the scope of the change.

ncopa:
> Yeah, seems so. My original purpose was to fix builds on darwin, and unfortunately, I don't have any Windows machine to check how it works.

How do you build it on darwin/macos?

Is the intention to run a controlplane on MacOS? Worker nodes will certainly not work on macos. Control plane might.

It builds on my macos laptop (crosscompiles TARGET_OS=linux).

You can also cross-compile for windows with: `make k0s.exe`

ncopa:
Also, what error do you get? (How do I reproduce it?)

makhov:
> Is the intention to run a controlplane on MacOS?

No, I just tried to run unit test with `make check-unit` command. `bridge` is not implemented for darwin, so it doesn't compile.
Also there are a couple of other errors that I was going to fix in later in separate PRs.

ncopa:
Please see https://github.com/k0sproject/k0s/pull/1154

Thanks!

PR #1144, opened 2021-09-27
Context propagation

### LINKED ISSUE #917: Use signal.NotifyContext for the process signal handling
Since 1.16 there is https://pkg.go.dev/os/signal#NotifyContext

Now we use mostly direct signal handling, attaching signaling channel by calling `signal.Notify(ch, <signals>)` and then operating with a `ch` to handle OS level signals. 

At the same time, we actively use `context.TODO` to create context in different parts of the code.

By using NotifyContext we could achieve two things:
- make signal management more clear 
- use a proper root context instead of `context.TODO` and getting native cancelation for the long-running jobs during the shutdown process.

makhov:
**Issue**
Fixes #917

**What this PR Includes**
- Context propagation instead of `context.TODO()` usage.
- `signal.NotifyContext ` is used instead of `signal.Notify` for signal handling.

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -21,7 +21,6 @@ import (
 	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
 )
 
-
 // Component defines the interface each managed component implements
 type Component interface {
 	Init() error
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -15,12 +15,17 @@ limitations under the License.
 */
 package component
 
-import "github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+import (
+	"context"
+
+	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+)
+
 
 // Component defines the interface each managed component implements
 type Component interface {
 	Init() error
-	Run() error
+	Run(context.Context) error
 	Stop() error
 	Healthy() error
 }
@@ -28,5 +33,5 @@ type Component interface {
 // ReconcilerComponent defines the component interface that is reconciled based
 // on changes on the global config CR object changes
 type ReconcilerComponent interface {
-	Reconcile(*v1beta1.ClusterConfig) error
+	Reconcile(context.Context, *v1beta1.ClusterConfig) error
 }
--- a/pkg/component/controller/coredns.go
+++ b/pkg/component/controller/coredns.go
@@ -250,7 +250,6 @@ type CoreDNS struct {
 	manifestDir            string
 	K0sVars                constant.CfgVars
 	previousConfig         coreDNSConfig
-	stopCtx                context.Context
 	stopFunc               context.CancelFunc
 	lastKnownClusterConfig *v1beta1.ClusterConfig
 }
@@ -287,7 +286,7 @@ func (c *CoreDNS) Init() error {
 
 // Run runs the CoreDNS reconciler component
 func (c *CoreDNS) Run(ctx context.Context) error {
-	c.stopCtx, c.stopFunc = context.WithCancel(ctx)
+	ctx, c.stopFunc = context.WithCancel(ctx)
 
 	go func() {
 		ticker := time.NewTicker(10 * time.Second)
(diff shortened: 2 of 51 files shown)


ncopa (commented):
This is good!

I think we should merge/rebase this after #1029

ncopa:
Could you please rebase this against current main? or let me know if you need help with rebasing it.

makhov:
@ncopa Done! Had to add context to `ReconcilerComponent` interface as well.

makhov:
It seems I missed a couple of places to propagate the context. I've fixed that and at least `make check-basic` passed successfully.

ncopa (dismissed):

pkg/component/component.go
```diff
@@ -15,18 +15,22 @@ limitations under the License.
...
+	"context"
+
+	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+)
 
 // Component defines the interface each managed component implements
 type Component interface {
 	Init() error
-	Run() error
+	Run(context.Context) error
```
soider, line 27: @jnummelin  @ncopa  Just thinking aloud outside of the PR context, I wonder do we want to have the same definition in Stop function to introduce context based timeouts for the graceful shutdown?

pkg/component/controller/coredns.go
```diff
@@ -286,8 +286,8 @@ func (c *CoreDNS) Init() error {
 }
 
 // Run runs the CoreDNS reconciler component
-func (c *CoreDNS) Run() error {
-	c.stopCtx, c.stopFunc = context.WithCancel(context.Background())
+func (c *CoreDNS) Run(ctx context.Context) error {
+	c.stopCtx, c.stopFunc = context.WithCancel(ctx)
```
soider, line 290: why is it a field, is it used outside of the Run function anywhere? if no, can be a local variable

pkg/component/controller/etcd.go
```diff
@@ -211,15 +211,15 @@ func (e *Etcd) Reconcile() error {
...
+func (e *Etcd) setupCerts(ctx context.Context) error {
 	etcdCaCert := filepath.Join(e.K0sVars.EtcdCertDir, "ca.crt")
 	etcdCaCertKey := filepath.Join(e.K0sVars.EtcdCertDir, "ca.key")
 
 	if err := e.CertManager.EnsureCA("etcd/ca", "etcd-ca"); err != nil {
 		return fmt.Errorf("failed to create etcd ca: %w", err)
 	}
 
-	eg, _ := errgroup.WithContext(context.Background())
+	eg, _ := errgroup.WithContext(ctx)
```
soider, line 222: do we even need to have WithContext here, instead of regular errgroup? the returned contexted is underscored and none of the goroutines use it

is it better to use `var eg errgroup.Group` here?

pkg/component/controller/k0sLease.go
```diff
@@ -30,8 +30,8 @@ func (l *K0sLease) Init() error {
 }
 
 // Run runs the leader elector to keep the lease object up-to-date.
-func (l *K0sLease) Run() error {
-	l.cancelCtx, l.cancelFunc = context.WithCancel(context.Background())
+func (l *K0sLease) Run(ctx context.Context) error {
+	l.cancelCtx, l.cancelFunc = context.WithCancel(ctx)
```
soider, line 34: Same here, ctx could be a local variable instead of a struct field

soider (changes requested):
I see couple of places in which code could be improved a little. None of them are critical, just few simplifications.

pkg/component/controller/coredns.go
makhov: `c.stopFunc` is called in the `Stop()` function:
https://github.com/k0sproject/k0s/blob/be438c564d58ed81431e8d11b5ff25605e06e196/pkg/component/controller/coredns.go#L353

pkg/component/controller/k0sLease.go
makhov: done

pkg/component/controller/etcd.go
makhov: I haven't found any reason to keep `errgroup.WithContext` here. Changed to `var eg errgroup.Group`

pkg/component/controller/coredns.go
makhov: Done

pkg/component/component.go
jnummelin: hmm, not sure. We're now going into a direction where we have a "global" context anyways and thus it feels like the general `Stop()` function might not even be needed. I mean that when we `Run()` as component with a ctx the component needs to get hold of the given context. Then the component can stop all-the-things based on `<-ctx.Done()`, right? The "global" context is the one we use to signal EVERYTHING to stop, right?

But I agree, these parts go bit outside of this PR

ncopa: yeah. lets do that outside this PR

soider (dismissed):
good stuff

jnummelin (dismissed):

soider (approved):

ncopa (approved):

PR #1146, opened 2021-09-28
More supervisor tests

ncopa:
Add unit tests for supervisor to test:
- that respawn actually works
- stop when process is gone and waiting for respawn

Make sure that the signaling channels are closed.


ncopa:
yes, this is only for refactoring the supervisor code and adding more tests (paying off tech debt).

The `defer close(waitresult)` may not be needed but I guess it is good practice to always close channels. If nothing else to prevent surprises if the logic is changed in the future.

pkg/supervisor/supervisor.go
```diff
@@ -134,10 +139,10 @@ func (s *Supervisor) Supervise() error {
 			} else {
 				if s.quit == nil {
 					s.log.Info("Started successfully, go nuts")
-					s.quit = make(chan bool)
-					s.done = make(chan bool)
+					s.quit = make(chan struct{})
```
soider, line 142: IMO, logic in that file should be refactored to use root context coming from cobra command (or any other caller) and use context.Stop instead of manually orchestrating channels, wdyt?

ncopa: I agree, but I think we can do that in a separate PR.

ncopa: I dropped the "Close channels used by supervisor" commit.

soider: sounds fair

PR #1165, opened 2021-10-03
Add support for external etcd cluster

### LINKED ISSUE #1010: Add external etcd cluster as an option

The solution would be to add an external etcd cluster as an option to the storage mode. 
For now the only way to use an external storage is to use kine( and it doesnt include etcd) .

jewertow:
Signed-off-by: Jacek Ewertowski <jacek.ewertowski1@gmail.com>

**Issue**
Fixes #1010

**What this PR Includes**
This PR aims to enable users to use external etcd cluster as a storage for k0s. To do that a user has to define the following configuration:
```
  storage:
    etcd:
      externalCluster:
        endpoints:
        - http://192.168.10.1:2379
        - http://192.168.10.2:2379
        - http://192.168.10.3:2379
        etcdPrefix: k0s-tenant
    type: etcd
```
Field `endpoints` contains list of URLs that listen on for client requests. `etcdPrefix` is used to enable multi-tenancy in etcd and specifies the name of the root path which by default is `/registry`; this value will be used to pass as `--etcd-prefix` argument in kube-apiserver.

**Changes**
- Extend `ClusterConfig` CRD with `etcd.externalCluster` option.

**TODO**:
- [x] Skip running internal etcd if `etcd.externalCluster` is defined.
- [x] Configure external endpoints in `EtcdClient`.
- [x] Run kube-apiserver with configured external endpoints and `--etcd-prefix` argument.
- [x] Validate external cluster configuration.
- [x] Add option to specify paths to TLS key and certificate.


kke:
Good stuff. I'm thinking if it would be neater if instead of the `externalCluster` you would have `storage.etcd.endpoints` and the `etcd.IsExternal()` would be `return len(e.Endpoints) > 0`. I think prefix and clientCert can be there for the internal etcd too.

jewertow:
Thanks for your feedback. I decided to wrap `endpoints` and `etcdPrefix` in `externalCluster` object to make the default config as simple as possible:
```
  storage:
    type: etcd
    etcd:
      externalCluster: null
      peerAddress: 192.168.68.104
```
Additionally, wrapper object `externalCluster` gives more context about the purpose of the fields inside.

On the other hand, why a user would want to set custom prefix or client cert for the internal cluster? I think that internal etcd should be managed only by k0s to keep it as simple as possible.

jnummelin:
> In my opinion the first option gives more context and is more intuitive, so I vote for the first option.

I agree, feels more intuitive IMO too.

> Can TLS be optional in case of external cluster or always required?

Theoretically one can configure etcd not to use any TLS, I don't think many people do it though. So I'd say they can be optional.

> I'm going to not support k0s etcd member-list and k0s etcd leave when external cluster is used, because I think that if a user uses an external cluster, we can assume that he knows how to manage this cluster and he has tools to list members. When it comes to leave command, k0s should not support this, because it simply doesn't make sense. Do you agree with me?

Yes, I agree.

> We have to decide if etcdPrefix should be available also for internal cluster. I don't see a need to use a prefix in this case, but maybe you know some use case?

I do not know any use case where this would be required. So IMO ok to have it only on the `externalCluster`.

> What do you think about the name of etcdPrefix property? I named it so, to point out it's used as --etcd-prefix, but maybe it could be just prefix or pathPrefix?

Both `etcdPrefix` and `pathPrefix` sound ok to me.

pkg/apis/k0s.k0sproject.io/v1beta1/storage_test.go
```diff
@@ -97,3 +112,199 @@ spec:
...
+			Etcd: &EtcdConfig{
+				ExternalCluster: &ExternalCluster{
+					Endpoints:  []string{"http://192.168.10.10"},
+					EtcdPrefix: "tenant-1",
+				},
+			},
+		}
+
+		s.Nil(spec.Validate())
+	})
```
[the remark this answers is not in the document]
jewertow: Done

soider:
@jewertow can I ask you to rebase over upstream? I am afraid GHA won't start on the outdated branch

jewertow:
Rebased

soider:
>This pull request does not include documentation and smoke tests, but I'm going to add both these things in subsequent pull requests.

Would be great to have e2e test suite as a part of this PR, it's usual practice in the project to cover new features with tests originally to not let techdebt growing.

jewertow:
Ok, I will add some e2e tests.

soider:
@jewertow hello! Do you have any updates on this?

inttest/Makefile
```diff
@@ -10,6 +10,12 @@ bins = bin/sonobuoy
 
 include ../embedded-bins/Makefile.variables
 
+ifeq ($(ARCH),amd64)
+etcd_arch = amd64
+else
+etcd_arch = arm64
+endif
```
ncopa, line 17: why not `etcd_arch = $(ARCH)`?

ncopa (approved):
Good job!

inttest/Makefile
jewertow: I didn't know that value of `$(ARCH)` on armv7 is "arm64". It seems that `etcd_arch` is not necessary, so I will remove this variable and use `$(ARCH)` as a build arg.

jewertow:
This pull request is still missing a test case for external etcd with enabled TLS. I tested such a case manually on VMs and it works fine, but unfortunately I have no time this week to implement this test. I can implement this test next week. What do you think? Is it acceptable for you?

inttest/Makefile
jewertow: I can see in logs that `$(ARCH)` on armv7 is "arch", so I have to revert my last commit.

jewertow: `etcd_arch` must be equal to "amd64" or "arm64", because etcd has no release for armv7 architecture.

soider (approved):

ncopa (approved):

PR #1167, opened 2021-10-04
Create separate abstraction for components that acts as source for the ClusterConfig changes

### LINKED ISSUE #1149: Config CRD: Add option to disable CR based reconcile
In some cases users might want to always have the yaml config file as the source-of-truth

I'd like to see a flag/option to make the yaml config as the source of truth for the cluster config. This would mean that instead of getting the changes from CR we reconcile based on the locally provided config.

jnummelin:

**Issue**
Fixes #1149 

**What this PR Includes**
Disables the config CRD functionality by default.
Created an abstraction for what is the source of truth to allow us to use either kube API, files or something else as the source of truth for the config.


s0j (approved):

ncopa (approved):
Would be nice with some sort of test(s) for this.

jnummelin:
> Would be nice with some sort of test(s) for this.

hmm, it is already tested with all previous smokes (the part that pushes out the "static" config to reconcilers) plus the new config CRD related smoke now enables the dynamic config from API. what other sorts of tests we could/should do?

PR #1173, opened 2021-10-05
use client-gen to auto-generate API Client code

trawler:

**What this PR Includes**
Add a makefile target to run client-get for auto-generating the API client code


Makefile
```diff
@@ -188,6 +188,12 @@ static/gen_manifests.go: $(shell find static/manifests -type f)
 .PHONY: generate-bindata
 generate-bindata: pkg/assets/zz_generated_offsets_$(TARGET_OS).go
 
+.PHONY: generate-APIClient
+
+# install with go get k8s.io/code-generator/cmd/client-gen@v0.22.2
```
s0j, line 193: Would it make sense to include this command as a part of the target to avoid having the caller setup their environment properly?

s0j (approved):

Makefile
s0j: You can use `go env` for that:
```
generate-APIClient:
    go get k8s.io/code-generator/cmd/client-gen@v0.22.2
    $(go env GOPATH)/bin/client-gen --etc --etc
```

PR #1180, opened 2021-10-07
Use k8s.gcr.io and quay.io instead of docker.io

champtar:
This allow to avoid docker.io rate limiting as much as possible
kuberouter is only available at docker.io


champtar:
Updated, need approval for CI again

champtar:
This time I should have fixed everything needed, sorry about that

PR #1245, opened 2021-11-04
Validate config apiVersion and kind

### LINKED ISSUE #1237: Configuration version and kind are not validated
The k0s.yaml "header" fields `apiVersion` and `kind` are not checked.

```shell
$ cat k0s.yaml
apiVersion: hessu
kind: hopo
spec:
  api:
    address: 10.0.0.1
$ k0s validate config -c k0s.yaml
# no error
$ k0s controller -c k0s.yaml
# no error
```

kke:

Fixes #1237 

Errors out if configuration header apiVersion and kind fields do not have the expected values `k0s.k0sproject.io/v1beta1` and `ClusterConfig`.


soider (approved):

jnummelin:
`Smoke test for CRD based config changes` appears to be broken in some way, does not seem to pass even with few restarts

soider:
looks like actions are stuck? might be rebasing over main will help, we once already had that behavior.

kke:
I'm not sure why the config change smoke is failing. Is it building a config that does not have the fields?

soider:
The controller0 logs have ```
time="2021-11-26 08:14:11" level=error msg="cluster-config reconcile failed: failed to validate config: [expected apiVersion: k0s.k0sproject.io/v1beta1 but found  expected kind: ClusterConfig but found ]" component=clusterConfig-reconciler
```

so looks like it generates wrong config

**Comment by kke** [comment]
Hmm, in pkg/component/controller/clusterConfig.go there was:

```go
resourceType = v1.TypeMeta{APIVersion: "k0s.k0sproject.io/v1beta1", Kind: "clusterconfigs"}
```

I wonder if that was intentional. I also wonder what happens if that changes when upgrading.

kke:
The client-gen thing does have some plural override flag but I'm not sure if it should be used.

kke:
Maybe there should be different config kind for the crd without the node level stuff.

PR #1246, opened 2021-11-04
Give validation errors on unknown config fields

kke:

As mentioned in https://github.com/k0sproject/k0s/issues/1237#issuecomment-958726392 - the unknown field validation has not been working as intended since the move to https://sigs.k8s.io/yaml

Seems what breaks it is the `UnmarshalJSON` function which runs  basic `json.Unmarshal`, which will lose the strictness. This PR makes it use the `DisallowUnknownFields` JSON decoder option.

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
@@ -16,6 +16,7 @@ limitations under the License.
 package v1beta1
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -183,7 +184,7 @@ func ConfigFromStdin(dataDir string) (*ClusterConfig, error) {
 
 func ConfigFromString(yml string, dataDir string) (*ClusterConfig, error) {
 	config := DefaultClusterConfig(dataDir)
-	err := strictyaml.YamlUnmarshalStrictIgnoringFields([]byte(yml), config, []string{"interval"})
+	err := strictyaml.YamlUnmarshalStrictIgnoringFields([]byte(yml), config, "interval")
 	if err != nil {
 		return config, err
 	}
@@ -218,10 +219,10 @@ func (c *ClusterConfig) UnmarshalJSON(data []byte) error {
 	type config ClusterConfig
 	jc := (*config)(c)
 
-	if err := json.Unmarshal(data, jc); err != nil {
-		return err
-	}
-	return nil
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.DisallowUnknownFields()
+
+	return decoder.Decode(jc)
 }
 
 // DefaultClusterSpec default settings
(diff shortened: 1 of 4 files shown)


kke:
There's also `decoder.DisallowUnknownFields()` for json 🤔

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -218,10 +219,10 @@ func (c *ClusterConfig) UnmarshalJSON(data []byte) error {
 	type config ClusterConfig
 	jc := (*config)(c)
 
-	if err := json.Unmarshal(data, jc); err != nil {
-		return err
-	}
-	return nil
+	decoder := json.NewDecoder(bytes.NewReader(data))
+	decoder.DisallowUnknownFields()
```
trawler, line 223: the yaml package already turns `DisallowUnknownFields` on. Why isn't it working then?
https://github.com/kubernetes-sigs/yaml/blob/39f74b9b8a135185d095704ab98c760ca031a9f8/yaml.go#L64

trawler:
> Seems what breaks it is the `UnmarshalJSON` function which runs basic `json.Unmarshal`, which will lose the strictness. This PR makes it use the `DisallowUnknownFields` JSON decoder option.

The yaml.Unmarshalstrict is already turning on the DisallowUnknownFields option for the json unmarshaller. If this is not working as intended, I think a better route would be to submit a fix upstream.

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
kke: Yes. The `json.Unmarshal` in the `UnmarshalJSON` function creates a new decoder that doesn't know anything about the `DisallowUnknownFields` set by the yaml thing.

kke:
> I think a better route would be to submit a fix upstream.

I don't think it's possible to fix in upstream. The `UnmarshalJSON` function does not work like `UnmarshalYAML`, it does not send a function to be called:

```go
func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
  	// do stuff
  	
  	if err := unmarshal(yc); err != nil { // this function can inherit options from the original call
		return err
	}
	
	return nil
}
```


```go
func (c *Config) UnmarshalJSON(data []byte) error {
  	// do stuff

	if err := json.Unmarshal(data, c); err != nil { // this can not
		return err
	}

	return nil
}
```

kke:
Btw, I think the `DisallowUnknownFields` only works case-insensitively. The json decoder, I think, will accept `"aPiVeRSiON": "k0s.k0sproject.io/v1beta1"` and put it in the `APIVersion` struct field without any errors.

trawler:
> Btw, I think the `DisallowUnknownFields` only works case-insensitively. The json decoder, I think, will accept `"aPiVeRSiON": "k0s.k0sproject.io/v1beta1"` and put it in the `APIVersion` struct field without any errors.

Yes, that is correct. There is an issue about this same problem already open upstream: https://github.com/kubernetes-sigs/yaml/issues/15

trawler (approved):

PR #1247, opened 2021-11-04
pkg/config: refactoring config handling

trawler:

---

**Issue**

**What this PR Includes**
This PR refactors the entire handling of k0s config. Among other changes it:

1. Removes `--config` as a global flag to all commands
2. Keeps `--config` flag only on the `k0s controller` and `k0s reset` commands (note on the reset command below).
3. Adds a default value to k0s config: `/etc/k0s/k0s.yaml`
3. Generates a runtime config whenever k0s controller goes up and keeps it under `/run/k0s/k0s.yaml`. This runtime config file is removed when the controller is stopped.
4. The runtime config is a merged config of custom values and default config values (#1261)

Note on k0s reset:
Since the runtime config only exists while the controller is up, k0s reset (that can only be performed while the controller is down) needs to still retail the `--config` flag. A workaround to that would be to use the global location `/etc/k0s/k0s.yaml` as a source of configuration.


cmd/airgap/listimages.go
```diff
@@ -34,19 +33,23 @@ func NewAirgapListImagesCmd() *cobra.Command {
 		Example: `k0s airgap list-images`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// we don't need warning messages in case of default config
-			logrus.SetLevel(logrus.ErrorLevel)
-			c := CmdOpts(config.GetCmdOpts())
-			cfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
+			// logrus.SetLevel(logrus.ErrorLevel)
```
kke, line 36: I think these type of cases the logrus output could be sent to stderr so that piping would not include the logs.

cmd/backup/backup.go
```diff
@@ -89,16 +94,44 @@ func (c *CmdOpts) backup() error {
...
+	fullcfg, err := loadingRulesFull.Load()
+	if err != nil {
+		logrus.Errorf("failed to fetch cluster-config for backup: %v", err)
+		return ""
+	}
+	conf, _ := yaml.Marshal(fullcfg)
+
+	cfgFilePath, err := file.WriteTmpFile(string(conf), "k0s-config")
+	if err != nil {
+		logrus.Errorf("Error creating tempfile: %v", err)
```
kke, line 133:
```suggestion
		logrus.Errorf("error creating tempfile: %v", err)
```

cmd/controller/controller.go
```diff
@@ -119,11 +120,28 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 		return err
 	}
 
-	nodeConfig, err := config.GetNodeConfig(c.CfgFile, c.K0sVars)
-	if err != nil {
+	// get bootstrapping config first
+	// we first need to make sure our run-dir exists
+	if err := dir.Init(c.K0sVars.RunDir, constant.RunDirMode); err != nil {
+		logrus.Fatalf("failed to initialize dir: %v", err)
```
kke, line 126: This will exit, the `return err` on next line won't be reached.

cmd/controller/controller.go
```diff
@@ -324,9 +340,11 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 			return err
 		}
 	} else {
-		fullCfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
+		loadingRules := config.ClientConfigLoadingRules{}
+
+		fullCfg, err := loadingRules.Load()
 		if err != nil {
-			return err
+			logrus.Fatalf("failed to fetch controller's config: %s", err.Error())
```
kke, line 347:
```suggestion
			return fmt.Errorf("failed to fetch controller's config: %w", err)
```

cmd/controller/controller.go
```diff
@@ -395,8 +413,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 		logrus.Errorf("error while stopping node component %s", stopErr)
 	}
 	logrus.Info("all node components stopped")
```
kke, line 415: Not changed, but this will confusingly say something like:
```
[ERRO] error while stopping node component xyz
[INFO] all node components stopped
```

cmd/restore/restore.go
```diff
@@ -77,8 +74,8 @@ func (c *CmdOpts) restore(path string) error {
 		logger.Fatal("k0s seems to be running! k0s must be down during the restore operation.")
 	}
 
-	if !file.Exists(path) {
-		return fmt.Errorf("given file %s does not exist", path)
+	if !file.Exists(archivePath) {
+		return fmt.Errorf("given file %s does not exist", archivePath)
```
kke, line 78: Just a general thought, the `mgr.RunRestore` will return `fmt.Errorf("failed to unpack backup archive `%s`: %v", archivePath, err)` from the first row if the file doesn't exist (it should probably use `%w` to wrap it so it passes as `os.ErrNotExist`). Checking for file existence is often unnecessary.

pkg/config/file_config.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return err
+	}
+
+	return rules.writeConfig(yamlData)
+}
+
+// generate default config and return the file path
+func (rules *ClientConfigLoadingRules) generateDefaults() (config *v1beta1.ClusterConfig) {
+	logrus.Info("no config file given, generating default config")
```
kke, line 120: maybe this could just be a `println` so you don't have to filter it out or fiddle with logrus output levels when you want to hide this from pipes.

or we could have separation for levels, warn...panic to stderr, trace..info to stdout, as documented here: https://github.com/sirupsen/logrus/tree/f104497f2b2129ab888fd274891f3a278756bcde/hooks/writer - then this could be `logrus.Warn`.

pkg/config/file_config.go
```diff
@@ -0,0 +1,143 @@
...
+	if rules.RuntimeConfigPath == "" {
+		rules.RuntimeConfigPath = runtimeConfigPathDefault
+	}
+
+	// don't create the config file, if it already exists
+	if file.Exists(rules.RuntimeConfigPath) {
+		logrus.Infof("runtime config found: using %s", rules.RuntimeConfigPath)
+	}
+
+	switch CfgFile {
```
kke, line 85: There could be something like:

```go
package config

var errNoConfig = errors.New("no configuration")

func IsErrNoConfig(err error) bool {
    return err == errNoConfig
}

func ConfigReader() (io.Reader, error) [
  switch CfgFile {
  case "-":
    return os.Stdin, nil
  case "", constant.K0sConfigPathDefault:
    f, err := os.Open(constant.K0sConfigPathDefault)
    if err == nil {
      return f, nil
    }
    if os.IsNotExist(err) {
        return nil, errNoConfig
    }
    return err
  default:
    return os.Open(CfgFile)
  }
}
```

Then you could do:

```go
r, err := ConfigReader()
if err != nil {
  if IsErrNoConfig(err) {
    return rules.generateDefaults(), nil
  }
  return nil, err
}

return v1beta.ConfigFromReader(r, K0sVars.DataDir)
```

Or maybe `CfgFile` should only be defaulted to `constant.K0sConfigPathDefault` when that file exists, otherwise it would be `""` if that is somehow easily doable in cobra 🤔

pkg/constant/constant_windows.go
```diff
@@ -36,6 +36,7 @@ const (
 	KineSocket                     = "kine\\kine.sock:2379"
 	KubePauseContainerImage        = "mcr.microsoft.com/oss/kubernetes/pause"
 	KubePauseContainerImageVersion = "1.4.1"
+	K0sConfigPathDefault           = "C:\\etc\\k0s\\k0s.yaml"
```
kke, line 39: There's probably a better location for the default config (and rundir) on windows. Something like `path.Join(os.UserConfigDir(), "k0s")`. (which becomes something like `%APPDATA%/k0s`, or `C:\Users\User\AppData\Roaming\K0s`).

I guess some masochistic users might not even have `C:` or maybe a read-only `C:`.

Makefile
```diff
@@ -199,7 +199,7 @@ generate-APIClient:
 	$(go_clientgen) --go-header-file hack/client-gen/boilerplate.go.txt --input="k0s.k0sproject.io/v1beta1" --input-base github.com/k0sproject/k0s/pkg/apis --clientset-name="clientset" -p github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/
 
 image-bundle/image.list: k0s
-	./k0s airgap list-images > image-bundle/image.list
+	./k0s airgap list-images | grep -v 'time=' > image-bundle/image.list
```
kke, line 202: This shouldn't be necessary for users using `k0s airgap list-images` in scripts.

pkg/constant/constant_windows.go
trawler: > There's probably a better location for the default config (and rundir) on windows. Something like `path.Join(os.UserConfigDir(), "k0s")`. (which becomes something like `%APPDATA%/k0s`, or `C:\Users\User\AppData\Roaming\K0s`).
> 
> I guess some masochistic users might not even have `C:` or maybe a read-only `C:`.

Actually, there is no need at all for this var on windows, since we don't support windows controllers.
That also relates to the entire `pkg/config` package which is not needed at all in the windows binary. Unfortunately, in the current architecture, excluding the package from the windows build turned out to be a much bigger change than anticipated, and maybe should be tackled in a separate PR.

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -194,6 +194,15 @@ func ConfigFromString(yml string, dataDir string) (*ClusterConfig, error) {
 	return config, nil
 }
 
+// ConfigFromReader reads the configuration from any reader (can be stdin, file reader, etc)
```
kke, line 197: 👍 

I think this makes ConfigFromStdin obsolete/dead.

cmd/restore/restore.go
```diff
@@ -55,12 +52,12 @@ func NewRestoreCmd() *cobra.Command {
 	}
 
 	cmd.SilenceUsage = true
-	cmd.Flags().StringVar(&restoredConfigPath, "config-out", "", "Specify desired name and full path for the restored k0s.yaml file (default: ${cwd}/k0s_<archive timestamp>.yaml)")
+	cmd.Flags().StringVar(&restoredConfigPath, "config-out", "/etc/k0s/k0s.yaml", "Specify desired name and full path for the restored k0s.yaml file (default: /etc/k0s/k0s.yaml)")
```
kke, line 55: Doesn't this default hint come automatically from cobra?

```suggestion
	cmd.Flags().StringVar(&restoredConfigPath, "config-out", "/etc/k0s/k0s.yaml", "Specify desired name and full path for the restored k0s.yaml file")
```

trawler: fixed

cmd/airgap/listimages.go
soider: commented code is leftover

cmd/api/api.go
```diff
@@ -74,6 +69,14 @@ func NewAPICmd() *cobra.Command {
 }
 
 func (c *CmdOpts) startAPI() error {
+	// get k0s config
+	loadingRules := config.ClientConfigLoadingRules{Nodeconfig: true}
```
soider, line 73: can this be part of the GetCmdOpts function in github.com/k0sproject/k0s/pkg/config package, which is used as contruction function for CmdOptions?
It will allow to drop repetative calls of loading.Rules.Load everywhere.

soider (changes requested):
generally looks good, though couple of minor things.

Also, side note. This massive adding of omitempty here and there basically makes k0s default-config useless, because you now need read documentation to change the generated default configuration.

kke:
> This massive adding of omitempty here and there basically makes k0s default-config useless, because you now need read documentation to change the generated default configuration.

Good point! We could even go as far as using the CRD or some custom struct tags to somehow generate the default config with comments. That's what you get in the default config of many tools. Something like:

```yaml
spec:
  # APISpec defines the settings for the K0s API
  api:
    # Local address on which to bind an API
    address: 10.0.0.1
    # Custom port for k0s-api server to listen on     
    k0sApiPort: 9443
    # Custom port for kube-api server to listen on    
    port: 6443
    # List of additional addresses to push to API servers serving the certificate
    sans:
    - 172.19.0.1
    - 172.18.0.1
    - 172.20.0.1
    - 172.17.0.1
```

(btw this mixing of k0s and kubeapi stuff in spec.api is confusing, maybe it should have `spec.api` and `spec.kubeapi`)

PR #1255, opened 2021-11-09
Add a CR validator for API config

trawler:

**What this PR Includes**
When the config reconciler creates the config custom resource, CRValidator will validate the the created object has the right fields the k0s expects for this object.

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
@@ -282,6 +282,16 @@ func (c *ClusterConfig) Validate() []error {
 	return errors
 }
 
+// CRValidator is used to make sure a config CR is created with correct values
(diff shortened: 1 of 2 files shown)


pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -282,6 +282,16 @@ func (c *ClusterConfig) Validate() []error {
 	return errors
 }
 
+// CRValidator CRValidator is used to make sure a config CR is created with correct values
+func (c *ClusterConfig) CRValidator() *ClusterConfig {
+	copy := c.DeepCopy()
+	copy.ClusterName = "k0s"
+	copy.ObjectMeta.ClusterName = "k0s"
```
jnummelin, line 289: This should be `copy.ObjectMeta.Name = "k0s"`

The `ClusterName` field in all kube objects is some weird historical left over AFAIK

s0j (approved):

PR #1259, opened 2021-11-11
Docs: Add emphasis on enabling the dynamic config on every controller

kke:

The wording in the dynamic config docs did not state it clearly enough that you should have either all of the controllers with dynamic config enabled or none of them.

### DIFF
--- a/docs/dynamic-configuration.md
+++ b/docs/dynamic-configuration.md
@@ -2,7 +2,7 @@
 
 k0s comes with the option to enable dynamic configuration for cluster level components. This covers all the components other than etcd (or sqlite) and the Kubernetes api-server. This option enables k0s configuration directly via Kubernetes API as opposed to using a configuration file for all cluster configuration.
 
-This feature has to be separately enabled for all controllers using `k0s controller --enable-dynamic-config`.
+This feature has to be enabled for every controller in the cluster using the `--enable-dynamic-config` flag in `k0s controller` or `k0s install controller` commands. Having both types of controllers in the same cluster will cause a conflict.
 
 ## Dynamic vs. static configuration
 


docs/dynamic-configuration.md
```diff
@@ -2,7 +2,7 @@
 
 k0s comes with the option to enable dynamic configuration for cluster level components. This covers all the components other than etcd (or sqlite) and the Kubernetes api-server. This option enables k0s configuration directly via Kubernetes API as opposed to using a configuration file for all cluster configuration.
 
-This feature has to be separately enabled for all controllers using `k0s controller --enable-dynamic-config`.
+This feature has to be enabled for every controller in the cluster using the `--enable-dynamic-config` flag in `k0s controller` or `k0s install controller` commands. Having both types of controllers in the same cluster will cause a conflict.
```
mviitane, line 6: After this sentence should we briefly mention that in case of k0sctl, use 
```
  - role: controller
    installFlags:
    - --enable-dynamic-config
```
for each controller?

mviitane: The use of installFlags is not very obvious. I think we don't currently mention it anywhere in k0s docs.

kke: I think k0sctl needs proper support for dynamic config (https://github.com/k0sproject/k0sctl/issues/270). Currently if you enable it, you lose any way to modify k0s configuration via k0sctl.

kke: In my opinion it would be better to have all the installFlags features mapped out as settings in k0sctl.yaml.

mviitane: Yes, although I'm not sure how this would work, but I agree that we have a problem with installFlags. They need to be copied to each node section separately. You easily forget some of the nodes. One configuration should be enough (to reach all the nodes).

kke: Something like:

```yaml
spec:
  k0s:
    dynamicConfiguration: true
```

installFlags need to be separate for each.

mviitane: ok, then k0sctl doesn't need to be mentioned in this change. However, we have a problem here. Dynamic configuration is most useful for multi-controller clusters. It should work with k0sctl.

mviitane (approved):
LGTM

docs/dynamic-configuration.md
kke: It is possible to use yaml anchors for repetitive stuff.

```yaml
spec:
  hosts:
    - role: controller
      installFlags: &controllerFlags
        - --kubelet-extra-args=--foofoo
    - role: controller
      installFlags: *controllerFlags
    - role: controller
      installFlags: *controllerFlags
```

mviitane: ok, but I'm afraid most users don't know how to use them.

PR #1265, opened 2021-11-18
Enable the use of kubectl plugins in "k0s kubectl" subcommand

### LINKED ISSUE #1264: built-in kubectl cannot recognize kubectl plugin
**Version**
```
$ k0s version
v1.22.3+k0s.0
```
**Platform**
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.6 LTS
Release:	18.04
Codename:	bionic
```
**What happened?**
k0s built-in kubectl cannot recognize kubectl plugin, but native kubectl can do so.

**How To Reproduce**
1. install k0s.
2. install [Example Plugin](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#example-plugin) (kubectl-foo).
3. execute `k0s kubectl foo`, but cannot recognize `kubectl-foo` plugin (just print kubectl usage).
4. instal native [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#install-kubectl-binary-with-curl-on-linux).
5. execute `kubectl foo` and can get `I am a plugin named kubectl-foo` message.

**Expected behavior**
`k0s kubectl foo` prints `I am a plugin named kubectl-foo` message.

**Screenshots & Logs**
```
$ sudo cp /var/lib/k0s/pki/admin.conf kubeconfig.cfg
$ export KUBECONFIG=~/kubeconfig.cfg
$ cat > ./kubectl-foo
#!/bin/bash

# optional argument handling
if [[ "$1" == "version" ]]
then
    echo "1.0.0"
    exit 0
fi

# optional argument handling
if [[ "$1" == "config" ]]
then
    echo "$KUBECONFIG"
    exit 0
fi

echo "I am a plugin named kubectl-foo"
$ chmod +x kubectl-foo
$ sudo mv kubectl-foo /usr/local/bin/
$ /usr/local/bin/kubectl-foo 
I am a plugin named kubectl-foo
$ k0s kubectl foo
kubectl controls the Kubernetes cluster manager.
(snip)
$ curl -LO "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl"
$ chmod +x ./kubectl
$ ./kubectl foo
I am a plugin named kubectl-foo
```

**Additional context**
none.

kke:

Fixes #1264 : The plugin handler was not being initialized in `NewKubectlCommand` - this has now been fixed.

With #1299 in place, this can now be actually usable as a kubectl drop-in replacement even when using plugins. It will warn if there's no `kubectl` in the `$PATH` and hints how to create a symlink to work around that.

### DIFF
--- a/cmd/kubectl/kubectl.go
+++ b/cmd/kubectl/kubectl.go
@@ -16,7 +16,11 @@ limitations under the License.
 package kubectl
 
 import (
+	"fmt"
 	"os"
+	"os/exec"
+	"runtime"
+	"syscall"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -28,6 +32,42 @@ import (
 
 type CmdOpts config.CLIOptions
 
+type kubectlPluginHandler struct{}
+
+func (h *kubectlPluginHandler) Lookup(filename string) (string, bool) {
+	path, err := exec.LookPath(fmt.Sprintf("kubectl-%s", filename))
+	if err != nil || path == "" {
+		return "", false
+	}
+	return path, true
+}
+
+// adapted from kubectl.DefaultPluginHandler
+func (h *kubectlPluginHandler) Execute(executablePath string, cmdArgs, environment []string) error {
+	if _, err := exec.LookPath("kubectl"); err != nil {
+		if exe, err := os.Executable(); err == nil {
+			logrus.Warnf("kubectl not found in $PATH. many kubectl plugins try to run 'kubectl'. you can use k0s as a replacement by creating a symlink, for example: `sudo ln -s \"%s\" /usr/local/bin/kubectl`", exe)
+		}
+	}
+
+	// Windows does not support exec syscall.
+	if runtime.GOOS == "windows" {
+		cmd := exec.Command(executablePath, cmdArgs...)
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		cmd.Stdin = os.Stdin
+		cmd.Env = environment
+		if err := cmd.Run(); err != nil {
+			return err
+		}
+		os.Exit(0)
+	}
+
+	// invoke cmd binary relaying the environment and args given
+	// append executablePath to cmdArgs, as execve will make first argument the "binary name".
+	return syscall.Exec(executablePath, append([]string{executablePath}, cmdArgs...), environment)
+}
+
 func NewK0sKubectlCmd() *cobra.Command {
 	args := kubectl.KubectlOptions{
 		IOStreams: genericclioptions.IOStreams{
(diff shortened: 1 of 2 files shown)


kke:
To me it seems this now does what it is supposed to. `k0s kc foo` will now run `kubectl-foo` on `$PATH`. I'm a bit skeptical about plugins actually working, as they will often call back to kubectl. To fix this, there aren't many options. I think the only viable one is to install a `kubectl` symlink and make `k0s` detect that it was executed as `kubectl` and go directly to the `kubectl` subcommand.

kke:
Hmm well, there's also this option:

In kubectl-subcommand's pre-run:

1. Create a tempdir
2. Make "kubectl" symlink that points to os.Args[0]
3. Prepend tempdir to $PATH
4. Run the command
5. Clean up

kke:
> I think this could be done in a custom kubectl-plugin-handler.

Went that route.

kke:
Ah. `Error: flags cannot be placed before plugin name: --debug`

inttest/multicontroller/multicontroller_test.go
```diff
@@ -77,6 +77,8 @@ func TestMultiControllerSuite(t *testing.T) {
 }
 
 const k0sConfigWithMultiController = `
+apiVersion: k0s.k0sproject.io/v1beta1
+kind: ClusterConfig
```
kke, line 81: I don't know why this PR doesn't pass without this fix.

cmd/kubectl/kubectl.go
```diff
@@ -16,18 +16,94 @@ limitations under the License.
...
+	"github.com/k0sproject/k0s/internal/pkg/stringslice"
 	"github.com/k0sproject/k0s/pkg/config"
 )
 
 type CmdOpts config.CLIOptions
 
+// creates a temporary directory with a kubectl-symlink and prepends it to path. This is done to
+// allow kubectl plugins to call back to kubectl inside k0s
+func createTemp() (string, error) {
+	tempdir, err := os.MkdirTemp("", "k0s-kubectl")
```
ncopa, line 41: It is not uncommon that `/tmp` is mounted noexec, in which case things will fail.

I wonder if we can make a permanent symlink under datadir/bin

kke: Yeah maybe that would be better

kke: The kubectl subcommand will then only works when there's a k0s running on the host though.

kke: This will also require root for running kubectl in case datadir is in /var/lib.

kke: The tempdir has also the problem that defer won't fire so the tmp will not be cleaned 🤔

kke: Maybe just warn if there's no `kubectl` in PATH when launching a plugin and hint that you can make a symlink.

kke: Removed all the symlink trickery and replaced with simply a WARN when there's no kubectl in path.

soider (approved):

PR #1271, opened 2021-11-19
Bump version for image spec

soider:
Resolve dependabot update


soider:
better to update upstream of containerd once they update the dep

PR #1275, opened 2021-11-22
Use GH artifact for caching smoke-test binary

kke:
Makes it possible to download the binary for manual testing.


soider (dismissed):

soider (approved):

kke:
Maybe caching is better done using the cache. I mostly wanted this because my local build env wasn't working, now it is.

PR #1276, opened 2021-11-23
Merge supplied --config with defaults in "k0s default-config"

### LINKED ISSUE #1261: show merged config
when I have partial config in `/app/config.yaml` I'd like to see the "merged config" eg. defaults plus my changes:

I'd assume from the docs that this would do it:
```
$ k0s default-config -c /app/config.yaml
```

but it only shows the default values

kke:

Fixes #1261

When using `k0s default-config` with `--config <path>`, the supplied configuration is loaded, defaults filled and the result is displayed. When using without `--config`, it prints out the default config as before.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -131,11 +131,25 @@ func newDefaultConfigCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "default-config",
 		Short: "Output the default k0s configuration yaml to stdout",
+		Long:  "If a config file is passed via [-c|--config], its contents will be merged with the default config",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := cliOpts(config.GetCmdOpts())
-			if err := c.buildConfig(); err != nil {
+			var conf *v1beta1.ClusterConfig
+			if c.CfgFile != "" {
+				nodeConfig, err := config.GetNodeConfig(c.CfgFile, c.K0sVars)
+				if err != nil {
+					return err
+				}
+				conf = nodeConfig
+			} else {
+				conf = v1beta1.DefaultClusterConfig(c.K0sVars.DataDir)
+			}
+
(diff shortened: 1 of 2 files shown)


trawler:
This change is no longer necessary.
In the new config refactoring, providing a --config to the controller command will result in a combined config + defaults to be written in `/run/k0s/k0s.yaml`. The user can then view the merged output there.
https://github.com/k0sproject/k0s/blob/fb046a17a0a7b0b7f67d737aaf837fe5ee73424c/internal/testutil/runtime_config.go#L98

kke:
> This change is no longer necessary. In the new config refactoring, providing a --config to the controller command will result in a combined config + defaults to be written in `/run/k0s/k0s.yaml`. The user can then view the merged output there.

Seems the `k0s default-config --config` flag still exists in the config refactoring draft PR but it is still not used for anything. 

Someone may want to see the resulting configuration before restarting / spinning up a controller.

trawler:
> > This change is no longer necessary. In the new config refactoring, providing a --config to the controller command will result in a combined config + defaults to be written in `/run/k0s/k0s.yaml`. The user can then view the merged output there.
> 
> Seems the `k0s default-config --config` flag still exists in the config refactoring draft PR but it is still not used for anything.
That PR is still a work in progress.
> Someone may want to see the resulting configuration before restarting / spinning up a controller.
IMO, a command called `default-config` should be what it is called, and print out default-config only. If a new functionality is needed, then it should be exported to a new command. The planned workaround, where a merged config is created by k0s during runtime, is, I think, a legitimate solution to this need.

kke:
It could also be `k0s config <subcommand>` and the `k0s validate config` would be moved under `k0s config validate` as there doesn't seem to be anything else to validate. `k0s config generate` would maybe be the present day `default-config`.

kke:
Closing in favor of #1328

PR #1279, opened 2021-11-23
Remove labels used by kubernetes addon-manager

uesyn:
Signed-off-by: uesyn <suemura@zlab.co.jp>

`addonmanager.kubernetes.io/mode` label is used by kubernetres official [addon-manager](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons/addon-manager).
If we deploy addon-manager in kubernetes cluster built by k0s, konnectivity and default psp will be pruned unintentionally, because these manifests have `addonmanager.kubernetes.io/mode=Reconcile` label.

This PR is needed to be deploy addon-manager safely in cluster built by k0s.
As far as I have checked, these labels are not used by k0s, so I think we can remove this label.

**Issue**
None

**What this PR Includes**
Remove labels used by kubernetes addon-manager from konnectivity and default psp, because of deploying kubernetes addon-manager safely.

### DIFF
--- a/pkg/component/controller/defaultpsp.go
+++ b/pkg/component/controller/defaultpsp.go
@@ -109,7 +109,6 @@ metadata:
     seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
   labels:
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: Reconcile
 spec:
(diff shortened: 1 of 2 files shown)


soider:
Based on the addonmanager documentation, it makes sense

soider (approved):

PR #1280, opened 2021-11-24
Give the CI error log artifacts names

kke:

### Before:

image

I think each failing step will overwrite the artifact so you will only get one.

### After:

Something like:
 `k0s-<pr-number>-<commit-sha>-logs-check-basic-<build-number>`(`.zip`)


kke:
I think the smoke-tests could be run as a matrix, they seem to be identical in every other way except the smoke-test name. The arm ones are special.

PR #1282, opened 2021-11-24
Remove EMBEDDED_BINS_BUILDMODE=fetch from README

kke:
The README.md still has the `EMBEDDED_BINS_BUILDMODE=fetch` as the first documented way to build k0s. It was removed in #1089.

PR #1285, opened 2021-11-25
Testing if CI builds work without the cache

kke:

As mentioned in #1282, I'm starting to suspect there's something wrong with the build scripts. 

Testing here the build on a clean machine (using `ubuntu-18.04` instead of `ubuntu-latest`(20.04) and without any build cache.


kke:
Apparently it does work.

PR #1287, opened 2021-11-25
Tunneled network from node to api server

soider:
---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**What this PR Includes**
- New spec.api configuration fields
- New reconciler for the default `kubernetes` service endpoints
- Use forked version of the apiserver-network-proxy which supports tunneling traffic from the worker node to control plane.
- Some minor refactoring.

The reconciler changes internal cluster policy to be internal for the
default kubernetes service.
It creates endpoints for default kubernetes service to make it go to the local instance of konnectivity.
Each endpoint is <NODE_IP>:<KONNECTIVITY_PORT>


.github/workflows/go.yml
```diff
@@ -159,7 +159,7 @@ jobs:
           - check-disabledcomponents
           - check-extraargs
           - check-configchange
-          - check-upgrade
+          - check-tunneledkas
```
jnummelin, line 162: Why not add the new smoke rather than replace the existing?

pkg/apis/k0s.k0sproject.io/v1beta1/api.go
```diff
@@ -52,11 +53,12 @@ func DefaultAPISpec() *APISpec {
...
-		K0sAPIPort: 9443,
-		SANs:       addresses,
-		Address:    publicAddress,
-		ExtraArgs:  make(map[string]string),
+		Port:                   6443,
+		K0sAPIPort:             9443,
+		SANs:                   addresses,
+		Address:                publicAddress,
+		ExtraArgs:              make(map[string]string),
+		TunneledNetworkingMode: true,
```
jnummelin, line 61: Making it default to `true` means that once user upgrades, cluster will switch to this mode. Maybe we should disable it by default?

pkg/component/controller/konnectivity.go
```diff
@@ -366,6 +380,25 @@ spec:
...
+  internalTrafficPolicy: Local
+  ports:
+  - name: agent
+    port: 6443
+    protocol: TCP
+    targetPort: 6443
+  selector:
+    k8s-app: konnectivity-agent
+  sessionAffinity: None
+  type: ClusterIP
```
jnummelin, line 401: why do we need this?

pkg/component/controller/tunneledapiendpointreconciller.go
```diff
@@ -0,0 +1,214 @@
...
+		return nil, fmt.Errorf("can't list nodes: %w", err)
+	}
+
+	addresses := make([]v1core.EndpointAddress, 0, len(nodes.Items))
+	var ipAddr string
+	for _, node := range nodes.Items {
+		node := node
+		ipAddr = ""
+		for _, addr := range node.Status.Addresses {
+			if addr.Type != v1core.NodeInternalIP {
```
jnummelin, line 143: We need to verify if there are cases when node might have ONLY public address set. Those cases might be the ones on a cloud where node has only public address and the cloud-controller sets that as public on the node objet.

jnummelin (commented):
Good stuff. Added few comments/questions

pkg/apis/k0s.k0sproject.io/v1beta1/dualstack_test.go
```diff
@@ -1,36 +0,0 @@
-package v1beta1
-
-import (
-	"github.com/stretchr/testify/require"
-	"testing"
-)
-
-func TestAddDualStackArguments(t *testing.T) {
```
soider, line 8: for the context: deleted because this test is now a part of the feature gate helper test

pkg/component/controller/konnectivity.go
soider: no need, leftover

pkg/component/controller/tunneledapiendpointreconciller.go
soider: I wonder what should we do in such case - use the public address, skip until internal address would be there, or just use external address from config (basically dropping the tunneling). Do you have any opinion?

jnummelin: if there's only public address set for node, I _think_ those are only in case the node actually has that address in its interface. So no NAT etc. Which makes me think we can use the address as-is.

So basically we'd take the `InternalAddress` and if node does not have one, then `ExternalAddress`

mviitane:
See that this chapter remains up to date:
https://docs.k0sproject.io/main/networking/#controller-worker-communication

soider:
@mviitane it is internal change, disabled by default, I don't really want to describe it yet in the docs

jnummelin (approved):
LGTM

We still need to fix the armv7 image build (https://github.com/k0sproject/apiserver-network-proxy/pull/4) for the forked konnectivity before shipping this.

embedded-bins/Makefile.variables
```diff
@@ -39,8 +39,8 @@ etcd_build_go_cgo_enabled = 0
 etcd_build_go_ldflags = "-w -s"
 #etcd_build_go_ldflags_extra =
 
-konnectivity_version = 0.0.25
 konnectivity_buildimage = golang:1.17-alpine
+konnectivity_version = 0.0.25-k0s2
```
jnummelin, line 43: There's now `0.0.25-k0s3` with armv7 support too

pkg/constant/constant_shared.go
```diff
@@ -75,8 +75,8 @@ const (
 	// DefaultPSP defines the system level default PSP to apply
 	DefaultPSP = "00-k0s-privileged"
 	// Image Constants
-	KonnectivityImage                  = "k8s.gcr.io/kas-network-proxy/proxy-agent"
-	KonnectivityImageVersion           = "v0.0.25"
+	KonnectivityImage                  = "quay.io/k0sproject/apiserver-network-proxy-agent"
+	KonnectivityImageVersion           = "0.0.25-k0s2"
```
jnummelin, line 79: `0.0.25-k0s3` with armv7 support is now available

jnummelin (approved):

PR #1291, opened 2021-11-28
Separate command example and output in install.md to ease copying

### LINKED ISSUE #1290: Separate shell command example and example output in docs
This issue is in reference to the bug that was pointed out in #1205 , related to separating the command example and output, for easy copying and more readability!

Referring to the conversation [here](https://github.com/k0sproject/k0s/pull/1220#issuecomment-953531362), as there are many pages in the documentation where this bug can be improved/updated, i feel creating separate issues for each & making them as `good-first-issues` would be a nice way for new folks in the community to contribute to!

Hoping, i can proceed with making separate issues like this, for each page of the documentation 🙂
Would love your thoughts & suggestions on this!!

verma-kunal:
Signed-off-by: Kunal Verma <vkunal321@gmail.com>

---
name: Pull Request
about: Create a Pull Request
title: ''
labels: ''
assignees: ''

---

**Issue**
Fixes #1290 

**What this PR Includes**
Fixes a small bug, that would allow to only copy the particular command from the documentation & not the generated output

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -48,6 +48,9 @@ Though the Quick Start material is written for Debian/Ubuntu, you can use it for
 
     ```shell
     $ sudo k0s status
+    ```
+
+    ```shell
     Version: v1.22.4+k0s.1
     Process ID: 436
     Role: controller
@@ -63,6 +66,9 @@ Though the Quick Start material is written for Debian/Ubuntu, you can use it for
 
     ```shell
     $ sudo k0s kubectl get nodes
+    ```
+
+    ```shell
     NAME   STATUS   ROLES    AGE    VERSION
     k0s    Ready    <none>   4m6s   v1.22.4-k0s1
     ```
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -135,6 +135,33 @@ jobs:
     name: Smoke test
     needs: build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        smoke-suite:
+          - check-basic
+          - check-addons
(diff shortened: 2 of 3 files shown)


docs/install.md
```diff
@@ -48,6 +48,9 @@ Though the Quick Start material is written for Debian/Ubuntu, you can use it for
 
     ```shell
     $ sudo k0s status
```
verma-kunal, line 50:
```suggestionsudo k0s status
```

verma-kunal, line 50:
```suggestion
    sudo k0s status
```

docs/install.md
```diff
@@ -63,6 +66,9 @@ Though the Quick Start material is written for Debian/Ubuntu, you can use it for
 
     ```shell
     $ sudo k0s kubectl get nodes
```
verma-kunal, line 68:
```suggestion
    sudo k0s kubectl get nodes
```

kke:
Needs a rebase.

> Fixes a small bug, that would allow to only copy the particular command from the documentation & not the generated output

It is intentional to split the command and the output so that you can use the generated copy-button to copy just the command for trying it yourself without getting the output to clipboard.

kke (changes requested):
Needs a rebase.

Also:

>Fixes a small bug, that would allow to only copy the particular command from the documentation & not the generated output

It is intentional to split the command and the output so that you can use the generated copy-button to copy just the command for trying it yourself without getting the output to clipboard.

soider:
@verma-kunal needs to be rebased over the main branch

Issue #773, opened 2021-03-12
ipv6 only clusters (and more straightforward dualstack config)


asteven:
**Is your feature request related to a problem? Please describe.**
I want to use k0s to run pure ipv6 clusters. This is currently not supported. It crashes when generating the DNSAddress.

**Describe the solution you'd like**
I'd like k0s to support ipv4 only, ipv6 only or dual stack configurations, same as kubeadm does.

**Additional context**
The existing dualstack configuration is somewhat conflicting (or at least redundant) to implement a ipv6 only network config.
I therefore propose to refactor the existing network configuration to work like explained below.

The following config results in a ipv4 only cluster.

```
apiVersion: k0s.k0sproject.io/v1beta1
kind: Cluster
metadata:
  name: k0s
spec:
  api:
    address: 192.168.68.104
  network:
    podCIDR: 10.244.0.0/16
    serviceCIDR: 10.96.0.0/12
```

The following config results in a ipv6 only cluster:

```
apiVersion: k0s.k0sproject.io/v1beta1
kind: Cluster
metadata:
  name: k0s
spec:
  api:
    address: fd00:101::100
  network:
    podCIDR: fd00:102::/64
    serviceCIDR: fd00:103::/108
```

And the following config results in a dual stack cluster:

```
apiVersion: k0s.k0sproject.io/v1beta1
kind: Cluster
metadata:
  name: k0s
spec:
  api:
    # Either should work
    address: 192.168.68.104
    #address: fd00:101::100
  network:
    podCIDR: 10.244.0.0/16,fd00:102::/64
    serviceCIDR: 10.96.0.0/12,fd00:103::/108
```

I believe this is more straight forward then the current dualstack configuration and also inline with the approach taken by kubeadm which is good for user understanding and adoption.

I've basically already put this into code and wanted to discuss/check if you would accept a PR that implements this.

soider:
@asteven 

I see that you mentioned you already have some code changes implemented in place?
how about if you open PR and we discuss the changes?

In general, we are open to have that functionality in the k0s

jnummelin:
The issue is marked as stale since no activity has been recorded in 30 days

PR #1292, opened 2021-11-28
Support IPv6 only mode

olljanat:
**Issue**
Parly solves  #773

**What this PR Includes**
This start solving IPv6 only support. On this PR focus is add just enough logic to support k0s on vcluster https://www.vcluster.com/docs/operator/other-distributions#k0s with IPv6 only mode added by https://github.com/loft-sh/vcluster/pull/209

Without these changes that scenario crashed to error:
```
time="2021-11-28 20:33:54" level=info msg="using sans: [10.6.86.119]"
panic: runtime error: index out of range [3] with length 0

goroutine 1 [running]:
github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1.(*Network).DNSAddress(0xc0003db110, 0x4, 0x33d43d9, 0xe, 0xc000b48e10)
        /go/src/github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1/network.go:104 +0x3b0
github.com/k0sproject/k0s/cmd/controller.(*CmdOpts).startController(0xc000b49868, 0x3a39310, 0xc0000d8000, 0x0, 0x0)
        /go/src/github.com/k0sproject/k0s/cmd/controller/controller.go:146 +0x5ea
github.com/k0sproject/k0s/cmd/controller.NewControllerCmd.func1(0xc0003ef180, 0xc0005e11d0, 0x0, 0x3, 0x0, 0x0)
        /go/src/github.com/k0sproject/k0s/cmd/controller/controller.go:97 +0x39a
github.com/spf13/cobra.(*Command).execute(0xc0003ef180, 0xc0005e11a0, 0x3, 0x3, 0xc0003ef180, 0xc0005e11a0)
        /gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003ee000, 0x407905, 0xc0000b6058, 0x2b1cd90)
        /gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        /gocache/mod/github.com/spf13/cobra@v1.2.1/command.go:902
github.com/k0sproject/k0s/cmd.Execute()
        /go/src/github.com/k0sproject/k0s/cmd/root.go:206 +0x2a
main.main()
        /go/src/github.com/k0sproject/k0s/main.go:40 +0x25
```

### DIFF
--- a/internal/pkg/iface/iface.go
+++ b/internal/pkg/iface/iface.go
@@ -34,7 +34,7 @@ func AllAddresses() ([]string, error) {
 	for _, a := range addrs {
 		// check the address type and skip if loopback
 		if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-			if ipnet.IP.To4() != nil {
+			if ipnet.IP.To4() != nil || ipnet.IP.To16() != nil {
 				addresses = append(addresses, ipnet.IP.String())
 			}
 		}
@@ -45,12 +45,14 @@ func AllAddresses() ([]string, error) {
 	return addresses, nil
 }
 
-// FirstPublicAddress return the first found non-local address that's not part of pod network
+// FirstPublicAddress return the first found non-local IPv4 address that's not part of pod network
+// if any interface does not have any IPv4 address then return the first found non-local IPv6 address
 func FirstPublicAddress() (string, error) {
 	ifs, err := net.Interfaces()
 	if err != nil {
 		return "127.0.0.1", fmt.Errorf("failed to list network interfaces: %w", err)
 	}
+	ipv6addr := ""
 	for _, i := range ifs {
 		if i.Name == "vxlan.calico" {
 			// Skip calico interface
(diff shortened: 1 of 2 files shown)


internal/pkg/iface/iface.go
```diff
@@ -69,6 +69,14 @@ func FirstPublicAddress() (string, error) {
 				}
 			}
 		}
+		for _, a := range addresses {
+			// check the address type and skip if loopback
+			if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+				if ipnet.IP.To16() != nil && !ipnet.IP.IsLoopback() {
```
kke, line 75: The `!ipnet.IP.IsLoopback()` conditional is redundant, it's not possible to have `IsLoopback() == true` inside the branch. Looks like the same is true for the branch above. I believe these can be combined into one.

kke: I wonder if it would be ok to just do:

```go
if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() && s := ipnet.IP.String(); s != "<nil>" {
  return s, nil
}
```

The documentation says `String()` will return `<nil>` if the length would be zero.

The `String()` seems to do the equivalent of `len(ipnet.IP) == 0` to make that determination.

olljanat: Hmm. I think that it should default to IPv4 as long there is any of those addresses on any interface. However I just noticed that also my implementation handles incorrectly case where first interface have IPv6 only but others have also IPv4. Will update...

olljanat:
Updated and now logic is on way that IPv6 address will be only used if there is not any IPv4 address on any interface so it does not change behavior on dual-stack clusters even if those contains IPv6 only interfaces.

kke:
The failing ARM builds are a known issue, the self-hosted runners are being worked on.

soider:
@olljanat can I ask you to rebase over the main branch to fix the issue with failing e2e tests on arm?

olljanat:
@soider done

soider (approved):

ncopa (approved):

PR #1296, opened 2021-11-30
Validate image architectures automatically

soider:
Follow up on the #1295, some CI automation to check if we have all architectures for the default images.


s0j (approved):
LGTM

PR #1302, opened 2021-12-01
Avoid CI rebuild for unit tests

### LINKED ISSUE #1297: Unit tests step in the build workflow takes more time than actual build (> 4 minutes on average)
There is a separate step in the build workflow, to run unit tests.

It seens it doesn't use already existed modules and downloads them again, making the step execution time more than 4 minutes.

kke:

Fixes #1297 by tweaking the module/bindata cache settings.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -79,6 +79,14 @@ jobs:
       - name: Docker Layer Caching
         uses: satackey/action-docker-layer-caching@v0.0.11
 
+      - name: Docker volumes cache
+        uses: actions/cache@v2
+        with:
+          path: /var/lib/docker/volumes/
+          key: ${{ runner.os }}-docker-volumes
+          restore-keys: |
+            ${{ runner.os }}-docker-volumes
+
       - name: Build
         run: make build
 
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -68,10 +72,9 @@ jobs:
             bindata_linux
             pkg/assets/zz_generated_offsets_linux.go
             embedded-bins/Makefile.variables
-
           key: ${{ runner.os }}-embedded-bins-${{ hashFiles('**/embedded-bins/**/*') }}
           restore-keys: |
-            ${{ runner.os }}-embedded-bins-${{ hashFiles('**/embedded-bins/**/*') }}
+            ${{ runner.os }}-embedded-bins-
```
kke, line 77: Removing the exact restore key will bring in the latest cache, `make` should have wits to see what needs to be rebuilt.

kke:
Looks like it still redownloads everything in `make check-unit` 🤔

kke:
Hmm doesn't building happen in a container and testing happens directly via `go test`...

kke:
And the `k0sbuil.gocache` docker volume doesn't seem to be cached either 🤔

kke:
Needs a different kind of approach.

PR #1313, opened 2021-12-06
footloose inttests: allow overriding k0s path

### LINKED ISSUE #1312: check upgrade smoke test doesn't always run the same k0s version binary
During the check-upgrade smoke test, the test suite downloads the "old" k0s binary to the `/usr/local/bin` path, while the "new" k0s version is located, as usual, in `/usr/bin/k0s`.
However, when setting up the environment, the footloose suite uses simply the `k0s` command, without stating a full path, and so many commands are actually executed with the new binary, on the old binary cluster. This can cause the test to break, when some commands are handled differently between versions.
We should allow an override of the full k0s binary path, in order to tell the footloose suite exactly which binary to run.

trawler:

**Issue**
Fixes #1312 

**What this PR Includes**
This PR allows to override the default running k0s path inside our footloose VMs for smoke tests. It is mainly needed for the check-upgrade smoke test, since in that test we have two k0s binaries on the same VM. The change here allows us to finely-configure which binary exactly we want to run at any given moment (instead of relying on the internal footloose $PATH var).

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -52,6 +52,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
+var defaultK0sBinPath = "/usr/bin/k0s"
+
 // FootlooseSuite defines all the common stuff we need to be able to run k0s testing on footloose
 type FootlooseSuite struct {
 	suite.Suite
@@ -88,9 +90,6 @@ func (s *FootlooseSuite) SetupSuite() {
 	}
 	if s.KonnectivityAgentPort == 0 {
 		s.KonnectivityAgentPort = 8132
-	}
-	if s.K0sFullPath == "" {
-		s.K0sFullPath = "/usr/bin/k0s"
 	}
 
 	dir, err := os.MkdirTemp("", "footloose-keys")
@@ -728,6 +727,10 @@ func (s *FootlooseSuite) createConfig() config.Config {
 		s.FailNow("K0S_PATH env needs to be set to k0s binary")
 	}
 
+	if s.K0sFullPath == "" {
+		s.K0sFullPath = defaultK0sBinPath
+	}
+
 	volumes := []config.Volume{
 		{
 			Type:        "bind",
@@ -737,7 +740,7 @@ func (s *FootlooseSuite) createConfig() config.Config {
 		{
 			Type:        "bind",
 			Source:      binPath,
-			Destination: "/usr/bin/k0s",
+			Destination: s.K0sFullPath,
 		},
 		{
 			Type:        "volume",
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -59,12 +59,13 @@ type FootlooseSuite struct {
 	Cluster *cluster.Cluster
 
 	ControllerCount       int
-	WorkerCount           int
-	KubeAPIExternalPort   int
 	K0sAPIExternalPort    int
-	KonnectivityAgentPort int
+	K0sFullPath           string
 	KonnectivityAdminPort int
+	KonnectivityAgentPort int
+	KubeAPIExternalPort   int
 	WithLB                bool
+	WorkerCount           int
 
 	ExtraVolumes  []config.Volume
 	tearDownTimer *time.Timer
@@ -88,6 +89,10 @@ func (s *FootlooseSuite) SetupSuite() {
 	if s.KonnectivityAgentPort == 0 {
 		s.KonnectivityAgentPort = 8132
 	}
+	if s.K0sFullPath == "" {
+		s.K0sFullPath = "/usr/bin/k0s"
+	}
+
 	dir, err := os.MkdirTemp("", "footloose-keys")
 	if err != nil {
 		s.T().Logf("ERROR: failed to load footloose config: %s", err.Error())
@@ -97,7 +102,6 @@ func (s *FootlooseSuite) SetupSuite() {
 	s.footlooseConfig = s.createConfig()
 
 	suiteCluster, err := cluster.New(s.footlooseConfig)
-
 	if err != nil {
 		s.T().Logf("ERROR: failed to load footloose config: %s", err.Error())
 		s.T().FailNow()
@@ -165,7 +169,6 @@ func (s *FootlooseSuite) waitForSSH() {
 	}
 
 	err := g.Wait()
-
 	if err != nil {
 		s.FailNowf("failed to ssh one or many nodes: %s", err.Error())
 		s.T().FailNow()
@@ -243,7 +246,6 @@ func (s *FootlooseSuite) TearDownSuite() {
 		}
 
 	}
-
 }
 
 const keepAfterTestsEnv = "K0S_KEEP_AFTER_TESTS"
@@ -272,7 +274,6 @@ func getDataDirOpt(args []string) string {
 }
 
 func (s *FootlooseSuite) startHAProxy() {
-
 	addresses := s.getControllersIPAddresses()
 	ssh, err := s.SSH("lb0")
 	s.Require().NoError(err)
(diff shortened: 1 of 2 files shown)


inttest/upgrade/upgrade_test.go
```diff
@@ -174,14 +173,14 @@ func (s *UpgradeSuite) TestK0sGetsUp() {
...
 	s.NoError(err)
-
 }
 
 func TestUpgradeSuite(t *testing.T) {
 	s := UpgradeSuite{
 		common.FootlooseSuite{
 			ControllerCount: 1,
 			WorkerCount:     2,
+			K0sFullPath:     "/usr/local/bin/k0s",
```
ncopa, line 183: Will this not be overwritten by this test?

soider:
Why not to use env variable K0S_PATH?

kke:
> Why not to use env variable K0S_PATH?

I guess the tests could also just override $PATH if they're running `k0s` without absolute path.

trawler:
> Why not to use env variable K0S_PATH?
`K0S_PATH` is used by footloose to also deploy the build binary, so in the check-upgrade test we need the flexibility to separate these two functions, so that we can deploy the build binary to `/usr/bin/k0s` as usual, but elect to run the old version from `/usr/local/bin/k0s` if we want to.
This PR also allows us to switch between the two versions if we want to, but setting the K0sFullPath var between runs.

inttest/common/footloosesuite.go
```diff
@@ -88,6 +89,10 @@ func (s *FootlooseSuite) SetupSuite() {
 	if s.KonnectivityAgentPort == 0 {
 		s.KonnectivityAgentPort = 8132
 	}
+	if s.K0sFullPath == "" {
+		s.K0sFullPath = "/usr/bin/k0s"
```
soider, line 93: I think it is safer to use K0S_PATH env (or whatever the proper name) instead of trying to use system wide binary. 
Otherwise I see how many tests on the dev machine would be green just because somebody runs thems with systemwide binary.

soider (changes requested):

inttest/upgrade/upgrade_test.go
ncopa: I misunderstood the code

inttest/common/footloosesuite.go
ncopa: `K0S_PATH`  is the path outside the footloose container, which gets [mounted](https://github.com/k0sproject/k0s/pull/1313/files#diff-85e4a1b54e18b85233536b2d526e695deea45e09bd65ecf7e4ed55a7c1441b4eR740) to `/usr/bin/k0s` inside the container.

The upgrade test needs to have both an old k0s binary (fetched from network) and a new (mounted into footloose container).

ncopa:
I am a bit skeptic to the hardcoded `/usr/bin/k0s` in two places. If any of those changes in the future things will break.

What do you think about something like this here below? The idea here is that `K0sFullPath` controls where the locally built `k0s` binary is mounted to. This way you can change `K0sFullPath` in any test without anything breaks.

```diff
diff --git a/inttest/common/footloosesuite.go b/inttest/common/footloosesuite.go
index 5e722f5e..7204efba 100644
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -89,10 +89,6 @@ func (s *FootlooseSuite) SetupSuite() {
        if s.KonnectivityAgentPort == 0 {
                s.KonnectivityAgentPort = 8132
        }
-       if s.K0sFullPath == "" {
-               s.K0sFullPath = "/usr/bin/k0s"
-       }
-
        dir, err := os.MkdirTemp("", "footloose-keys")
        if err != nil {
                s.T().Logf("ERROR: failed to load footloose config: %s", err.Error())
@@ -727,6 +723,9 @@ func (s *FootlooseSuite) createConfig() config.Config {
        if binPath == "" {
                s.FailNow("K0S_PATH env needs to be set to k0s binary")
        }
+       if s.K0sFullPath == "" {
+               s.K0sFullPath = "/usr/bin/k0s"
+       }
 
        volumes := []config.Volume{
                {
@@ -737,7 +736,7 @@ func (s *FootlooseSuite) createConfig() config.Config {
                {
                        Type:        "bind",
                        Source:      binPath,
-                       Destination: "/usr/bin/k0s",
+                       Destination: s.K0sFullPath,
                },
                {
                        Type:        "volume",
diff --git a/inttest/upgrade/upgrade_test.go b/inttest/upgrade/upgrade_test.go
index f06293b9..ba304e10 100644
--- a/inttest/upgrade/upgrade_test.go
+++ b/inttest/upgrade/upgrade_test.go
@@ -75,6 +75,7 @@ func (s *UpgradeSuite) TestK0sGetsUp() {
        }
 
        s.Require().NoError(g.Wait())
+       s.K0sFullPath = "/usr/local/bin/k0s"
 
        s.Require().NoError(s.WaitForKubeAPI(s.ControllerNode(0)))
        token, err := s.GetJoinToken("worker")
@@ -135,7 +136,7 @@ func (s *UpgradeSuite) TestK0sGetsUp() {
                                return err
                        }
                        defer ssh.Disconnect()
-                       _, err = ssh.ExecWithOutput("rm /usr/local/bin/k0s && cp /usr/bin/k0s /usr/local/bin/k0s")
+                       _, err = ssh.ExecWithOutput("rm /usr/local/bin/k0s && cp /usr/bin/k0s.new /usr/local/bin/k0s")
                        if err != nil {
                                return err
                        }
@@ -154,7 +155,7 @@ func (s *UpgradeSuite) TestK0sGetsUp() {
                                return err
                        }
                        defer ssh.Disconnect()
-                       _, err = ssh.ExecWithOutput("rm /usr/local/bin/k0s && cp /usr/bin/k0s /usr/local/bin/k0s")
+                       _, err = ssh.ExecWithOutput("rm /usr/local/bin/k0s && cp /usr/bin/k0s.new /usr/local/bin/k0s")
                        if err != nil {
                                return err
                        }
@@ -180,7 +181,7 @@ func TestUpgradeSuite(t *testing.T) {
                common.FootlooseSuite{
                        ControllerCount: 1,
                        WorkerCount:     2,
-                       K0sFullPath:     "/usr/local/bin/k0s",
+                       K0sFullPath:     "/usr/bin/k0s.new",
                },
        }
        suite.Run(t, &s)
```

trawler:
>I am a bit skeptic to the hardcoded `/usr/bin/k0s` in two places. If any of those changes in the future things will break.
> What do you think about something like this here below? The idea here is that `K0sFullPath` controls where the locally built `k0s` binary is mounted to. This way you can change `K0sFullPath` in any test without anything breaks.

Well, we will still need to set some kind of default value, at least here: 
```
	if s.K0sFullPath == "" {
		s.K0sFullPath = "/usr/bin/k0s"
	}
```

ncopa:
> > I am a bit skeptic to the hardcoded `/usr/bin/k0s` in two places. If any of those changes in the future things will break.
> > What do you think about something like this here below? The idea here is that `K0sFullPath` controls where the locally built `k0s` binary is mounted to. This way you can change `K0sFullPath` in any test without anything breaks.
> 
> Well, we will still need to set some kind of default value, at least here:
> 
> ```
> 	if s.K0sFullPath == "" {
> 		s.K0sFullPath = "/usr/bin/k0s"
> 	}
> ```

I think that is fine, since It is only a single place.

inttest/common/footloosesuite.go
```diff
@@ -737,7 +739,7 @@ func (s *FootlooseSuite) createConfig() config.Config {
 		{
 			Type:        "bind",
 			Source:      binPath,
-			Destination: "/usr/bin/k0s",
+			Destination: defaultK0sBinPath,
```
ncopa, line 742: I think we should use `s.K0sFullPath` here so we can override the default path to k0s in individual tests.

ncopa (approved):

PR #1315, opened 2021-12-07
Vendor dependencies

kke:
Should speed up builds / CI runs and also adds a layer of protection against any of the repositories getting deleted / going offline.


kke:
Apparently ineffective

PR #1321, opened 2021-12-08
Add ephemeral env variable for the konnectivity-agent

soider:
Add ephemeral env variable to have konnectivit-agent restarted on each change of the server count to avoid relying on the connection management in the konnectivity-agent code

Also some minor context related refactoring, to drop using context as field (argument context derived from the same parent context)

### DIFF
--- a/pkg/component/controller/konnectivity.go
+++ b/pkg/component/controller/konnectivity.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


jnummelin (dismissed):

pkg/component/controller/konnectivity.go
```diff
@@ -337,6 +341,11 @@ spec:
           imagePullPolicy: {{ .PullPolicy }}
           name: konnectivity-agent
           command: ["/proxy-agent"]
+          env:
+              # the variable is not in a use
+              # we need it to have agent restarted on server count change
+              - name: EPHEMERAL_SERVER_COUNT
```
makhov, line 347: We can also add some prefix (`K0S_`) to indicate that it's used by k0s and not related to the konnectivity options.

makhov (approved):

makhov (approved):

PR #1325, opened 2021-12-08
Send log Trace..Info to stdout and Warn..Fatal to stderr

kke:

Complementary to #1247 to solve a couple of my own comments.

Splits logging to output all the harmless messages to STDOUT and all the suspicious messages (warn, error, fatal, panic) to STDERR. This will make commands like `k0s airgap list-images` and `default-config` work nice with redirecting.

```
$ k0s airgap list-images > images.txt
[WARN] 01:02:03 no config file given, using defaults
$ cat images.txt
foofoo:123
foobar:345
```

### DIFF
--- a/cmd/airgap/listimages.go
+++ b/cmd/airgap/listimages.go
@@ -33,8 +33,7 @@ func NewAirgapListImagesCmd() *cobra.Command {
 		Short:   "List image names and version needed for air-gap install",
 		Example: `k0s airgap list-images`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// we don't need warning messages in case of default config
-			logrus.SetLevel(logrus.ErrorLevel)
+			logrus.SetLevel(logrus.WarnLevel)
 			c := CmdOpts(config.GetCmdOpts())
 			cfg, err := config.GetYamlFromFile(c.CfgFile, c.K0sVars)
 			if err != nil {
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -59,15 +59,10 @@ func NewBackupCmd() *cobra.Command {
 }
 
 func (c *CmdOpts) backup() error {
-	logger := logrus.New()
-	textFormatter := new(logrus.TextFormatter)
-	textFormatter.ForceColors = true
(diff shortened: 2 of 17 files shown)


kke:
Maybe in the `default-config` it shouldn't display the message as that is what you were trying to accomplish in the first place.

trawler (dismissed):
LGTM 👍

cmd/airgap/listimages.go
```diff
@@ -34,7 +34,7 @@ func NewAirgapListImagesCmd() *cobra.Command {
 		Example: `k0s airgap list-images`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// we don't need warning messages in case of default config
```
soider, line 36: looks like the comment is obsolete
```suggestion
```

soider (dismissed):

kke:
Maybe all of the logrus stuff should by default go to stderr, it's not output, it's log data 🤔

kke:
> > Maybe the default log level should actually be `logrus.WarnLevel` and output to stderr unless you give `-v|--verbose`, the `logrus.Info` is not real output, it's log data
> 
> Or maybe we simply `sed -i -e 's/logrus\.Infof/logrus.Debugf/g' ....`?

Errors still should always go to stderr. Not even logrus.Fatal is going to stderr currently.

kke:
There are also a couple of oddities like using logrus to output json that need to be addressed.

soider:
@kke I think json is a nice thing, well not json, but structured logs.

kke:
> @kke I think json is a nice thing, well not json, but structured logs.

Yes, but for `k0s etcd list-members` we output the list of members as JSON and do it via: `logrus.SetFormatter(&logrus.JSONFormatter{}); logrus.WithField("members", members).Info("done")`, which results to something weird with timestamps, `"level":"info"` and `"members":["foofoo","foobar"],"msg":"done"`.

soider:
yep, that's definitely worth switching to something less fancy (my fault, btw, if I recall correctly, that's me who introduced it)

kke:
Closing in favor of #1333 which I think is a better solution.

PR #1332, opened 2021-12-09
Labels for controller nodes

### LINKED ISSUE #1086: [Feature request] make master(s) visible on k0s and add taints to disable workloads
It could be frustrating not seeing cluster masters when you fire a: `kubectl get nodes -o wide`

It could be a better option to add taints to the master which would then disable workloads and make them "visible", than "hide" them

makhov:
**Issue**
Fixes #1086

**What this PR Includes**
In this PR we assign the role label (`node-role.kubernetes.io/control-plane=true`) and the taint (`node-role.kubernetes.io/master=:NoSchedule`) to the controller nodes.
We don't set the taint in case the controller running in a `--single` mode. 

As a side effect, the PR improves `kubectl get nodes` output.
**Before:** 
```
bash-5.1# kubectl get nodes
NAME          STATUS   ROLES    AGE    VERSION
controller0   Ready    <none>   101s   v1.22.4+k0s
worker0       Ready    <none>   112s   v1.22.4+k0s
```

**After:**
```
bash-5.1# kubectl get nodes
NAME          STATUS   ROLES           AGE   VERSION
controller0   Ready    control-plane   31s   v1.22.4+k0s
worker0       Ready    <none>          42s   v1.22.4+k0s
```

Passing `node-role.kubernetes.io` labels to the kubelet options are not allowed, so we add custom `node.k0sproject.io/role=<role>` label first and then exchange it to the proper `node-role.kubernetes.io` one.


inttest/Makefile
```diff
@@ -47,6 +47,9 @@ check-customports: TIMEOUT=6m
 # Config change smoke runs actually many cases hence a bit longer timeout
 check-configchange: TIMEOUT=8m
 
+# Node role check runs several cases
+check-noderole: TIMEOUT=6m
```
soider, line 51: why to have extended timeout?

inttest/noderole/noderole_single_no_taints_test.go
```diff
@@ -0,0 +1,57 @@
...
+	s.NotContains(n.Spec.Taints, corev1.Taint{Key: "node-role.kubernetes.io/master", Effect: "NoSchedule"})
+
+	err = s.WaitForNodeLabel(kc, s.ControllerNode(0), "node-role.kubernetes.io/control-plane", "true")
+	s.NoError(err)
+}
+
+func TestNodeRoleSingleNoTaintsSuite(t *testing.T) {
+	s := NodeRoleSingleNoTaintsSuite{
+		common.FootlooseSuite{
+			ControllerCount: 1,
```
soider, line 53: I'd add the case for 2 controllers to check if all the controller workers would have proper labels

pkg/component/controller/noderole.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (n *NodeRole) ensureNodeLabel(ctx context.Context, client kubernetes.Interface, node corev1.Node) error {
+	var labelToAdd string
+	for label, value := range node.Labels {
+		if strings.HasPrefix(label, constant.NodeRoleLabelNamespace) {
+			return nil
+		}
+
+		if label == constant.K0SNodeRoleLabel {
```
soider, line 98: side note: the original reasoning to not have self-signing kubelets (https://github.com/kubernetes/kubernetes/issues/85073) IMO has a lot of sense.
While at the moment I think it is impossible to have list of controllers to make address based matching to select master-role workers (if I understand correctly, this is reason to use custom label), it would be nice to have at some moment controller nodes register in the kube-api.

inttest/Makefile
makhov: There are 3 test cases and 4 minutes is not enough.

pkg/component/controller/noderole.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to set label '%s' to node %s: %w", labelToAdd, node.Name, err)
+	}
+	return nil
+}
+
+func (n *NodeRole) addNodeLabel(ctx context.Context, client kubernetes.Interface, node, key, value string) (*corev1.Node, error) {
+	keyPath := fmt.Sprintf("/metadata/labels/%s", jsonpointer.Escape(key))
+	patch := fmt.Sprintf(`[{"op":"add", "path":"%s", "value":"%s" }]`, keyPath, value)
+	return client.CoreV1().Nodes().Patch(ctx, node, types.JSONPatchType, []byte(patch), metav1.PatchOptions{})
```
soider, line 113: why not to use client.CoreV1().Nodes().Update and pass node instance there insterad of building raw json?

soider:
generally speaking looks good, left couple of questions

inttest/noderole/noderole_single_no_taints_test.go
makhov: There are 3 test cases:
- controller + worker
- single controller
- single controller with `--no-taints`

Do you suggest changing the first one to 2 controllers + worker or adding one more with just 2 controllers

pkg/component/controller/noderole.go
makhov: My reasoning was that it may cause a race condition: someone or something could change the node in between Get and Update calls.

Looks like it's a real case: https://github.com/kubernetes/client-go/issues/414

inttest/noderole/noderole_single_no_taints_test.go
soider: yeah, I think just expanding first one to 2+1 would work

pkg/component/controller/noderole.go
soider: yeah, it's common thing in go-client, I think it is even expected to retry on already modified objects

soider (approved):
sgtm

soider:
Would be great to have some docs on new flags introduced.

soider (approved):

xinity:
@makhov any chances you could provide examples and/or few docs on how to leverage this PR ? :)

Thanks a million

makhov:
@xinity 
There are some docs for labels and taints: https://docs.k0sproject.io/v1.23.1+k0s.0/worker-node-config/#node-labels
Is there anything else that needs to be better clarified?

PR #1336, opened 2021-12-10
Fix `APIEndpointReconciler.needsUpdate` boolean negation

jnummelin:
The `needsUpdate` uses reflect.DeepEqual to check if the endpoint addresses need updating or not.
But as the function usage is a "negate" pattern, we need to negate that result. :facepalm:

**Issue**
The `APIEndpointReconciler` was not poperly updating the `kubernetes` service endpoint if the DNS records change.

**What this PR Includes**
Fix the negation so the endpoints get properly updated.

### DIFF
--- a/pkg/component/controller/apiendpointreconciler_test.go
+++ b/pkg/component/controller/apiendpointreconciler_test.go
@@ -160,6 +160,49 @@ func TestReconcilerWithNoNeedForUpdate(t *testing.T) {
 	assert.Equal(t, "bar", e.ObjectMeta.Annotations["foo"])
 }
 
+func TestReconcilerWithNeedForUpdate(t *testing.T) {
+	fakeFactory := testutil.NewFakeClientFactory()
+	existingEp := corev1.Endpoints{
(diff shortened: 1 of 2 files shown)


pkg/component/controller/apiendpointreconciler_test.go
```diff
@@ -160,6 +160,49 @@ func TestReconcilerWithNoNeedForUpdate(t *testing.T) {
 	assert.Equal(t, "bar", e.ObjectMeta.Annotations["foo"])
 }
 
+func TestReconcilerWithNeedForUpdate(t *testing.T) {
```
soider, line 163: awesome that we introduced new test as soon with the fix.

soider (approved):

pkg/component/controller/apiendpointreconciler_test.go
jnummelin: Together with @ncopa and @makhov we first created the test to be able to debug the issue in general. 😄

ncopa (approved):

PR #1337, opened 2021-12-10
Version bumps

ncopa:
- kine 0.8.1
- runc 1.0.3
- konnectivity 0.0.27


jnummelin:
Seems we cannot update konnectivity as the upstream images are missing for arm platforms

soider (approved):

Issue #1319, opened 2021-12-08
Apiserver not configured with dual-stack service-cluster-ip-range when dual stack is enabled in k0s configuration


erdii:
**Version**
```
$ k0s version
v1.22.4+k0s.1
```
**Platform**
Which platform did you run k0s on?
Odroid-HC4 ARM64
```
$ cat /etc/os-release 
NAME="Arch Linux ARM"
PRETTY_NAME="Arch Linux ARM"
ID=archarm
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinuxarm.org/"
DOCUMENTATION_URL="https://archlinuxarm.org/wiki"
SUPPORT_URL="https://archlinuxarm.org/forum"
BUG_REPORT_URL="https://github.com/archlinuxarm/PKGBUILDs/issues"
LOGO=archlinux-logo
```
**What happened?**
Enabled dual-stack in k0s configuration but I was not able to configure Services with the v6 IP-family even though pod networking had working v4 and v6 networking.
I found out that the kubernetes apiserver was missing the ipv6 part of the `service-cluster-ip-range` even though it was correctly configured on the controller-manager.

This is how the flag looks on the
- apiserver: `--service-cluster-ip-range=10.96.0.0/12`
- controller-manager: `--service-cluster-ip-range=10.96.0.0/12,fd01::/108`

**How To Reproduce**
Enable dual-stack networking in k0s configuration. I used k0sctl with this configuration:

```yaml
apiVersion: k0sctl.k0sproject.io/v1beta1
kind: Cluster
metadata:
  name: homelab
spec:
  hosts:
  - ssh:
      address: odroid-hc4-one.home.erdii.net
      user: root
      port: 22
    role: controller+worker
  - ssh:
      address: odroid-hc4-two.home.erdii.net
      user: root
      port: 22
    role: controller+worker
  - ssh:
      address: odroid-hc4-three.home.erdii.net
      user: root
      port: 22
    role: controller+worker
  k0s:
    version: 1.22.4+k0s.1
    config:
      apiVersion: k0s.k0sproject.io/v1beta1
      kind: Cluster
      metadata:
        name: k0s
      spec:
        api:
          k0sApiPort: 9443
          port: 6443
          externalAddress: kubelab.home.erdii.net
          sans:
          - kubelab.home.erdii.net
        network:
          provider: calico
          kubeProxy:
            mode: ipvs
          calico:
[...]

PR #1339, opened 2021-12-13
Cherry picks for 1.22.4+k0s.2

ncopa:
**What this PR Includes**
- Fix services with dual-stack (#1319)
- Fix kubernetes service endpoint address update on DNS change (#1336)
- kine 0.8.1
- runc 1.0.3
- sync CI config with main

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -135,6 +135,33 @@ jobs:
     name: Smoke test
     needs: build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        smoke-suite:
+          - check-basic
+          - check-addons
+          - check-byocri
+          - check-calico
+          - check-cnichange
+          - check-ctr
+          - check-customports
+          - check-dualstack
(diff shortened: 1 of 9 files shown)


.github/workflows/go.yml
```diff
@@ -87,642 +90,81 @@ jobs:
...
-        if: failure()
-        uses: actions/upload-artifact@v2
-        with:
-          path: |
-            /tmp/*.log
-
-  smoketest-airgap:
-    name: Smoke test for airgap install
-    needs: build
+  build_windows:
```
soider, line 93: is it intention?

soider (approved):

.github/workflows/go.yml
ncopa: trying to figure out where that change comes from?

git blame says a6b3153dcf185d92be99438e908efb33305e4943

ncopa: I think this is a result of how `git diff` calculates the diff.

Issue #1360, opened 2021-12-16
k0s 1.22 and newer uses etcd instead of kine as default storage backend


ncopa:
**Version**
```
$ k0s version
```

k0s release-1.22 and current git main.

**Platform**
Which platform did you run k0s on?

x86_64 and arm64.

alpine linux edge

**What happened?**

Starting `k0s controller --single` with no configuration uses `etcd` as storage.

It appears that the intention is to use kine.
https://github.com/k0sproject/k0s/blob/f37fcb96042a3eb3c30f186da6513a082277d065/cmd/controller/controller.go#L93
However, it appears that etcd is started and used instead of kine.

**How To Reproduce**
run `k0s controller --single` in foreground and look for `Using storage backend ...`, or run `ps xa | grep etcd` to detect if etcd is running.

**Expected behavior**
I would expect `k0s controller --single` to use kine as storage backed by default.

**Screenshots & Logs**
....

**Additional context**

k0s 1.21.x seems to use kine as expected.

ncopa:
This was introduced in between 1.22.2 and 1.22.3.

ncopa:
I bisected it. It was introduced with commit 22ffa9fb80c765961281bf1f4240a02f83918d35.

1player:
Still broken, see #1707

PR #1361, opened 2021-12-16
inttest: Verify that we actually use kine when we should

ncopa:
check that kine is extracted and actually runs when it is expected.


s0j (approved):

trawler (approved):

ncopa:
the CI breaks because current code is broken.  #1360

PR #1363, opened 2021-12-17
Fix storage for --single regression

ncopa:
Kine should be used as default storage with --single option specified.
commit 22ffa9fb80c7 (ClusterConfig CRD: remove k0Vars in favor of
DataDir var) introduced a regression that caused etcd be used by default
with --single.

Fix this and refactor so that DefaultClusterConfig takes an
optional/variadic clusterConfig. This way we don't always need to
specify which default cluster spec is to be used. Then we set the kine
storage in `ValidateYaml` depending on the DefaultStorageType in
K0sVars. `ValidateYaml` needs data dir from k0s vars so it makes sense
to do this from here.

Fixes https://github.com/k0sproject/k0s/issues/1360

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
@@ -26,7 +26,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/k0sproject/k0s/internal/pkg/strictyaml"
-	"github.com/k0sproject/k0s/pkg/constant"
 )
 
 const (
@@ -71,9 +70,8 @@ type ClusterConfig struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	metav1.TypeMeta   `json:",omitempty,inline"`
 
-	DataDir string              `json:"dataDir,omitempty"`
-	Spec    *ClusterSpec        `json:"spec,omitempty"`
-	Status  ClusterConfigStatus `json:"status,omitempty"`
+	Spec   *ClusterSpec        `json:"spec,omitempty"`
+	Status ClusterConfigStatus `json:"status,omitempty"`
 }
 
 // StripDefaults returns a copy of the config where the default values a nilled out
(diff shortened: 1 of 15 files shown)


pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -71,9 +70,9 @@ type ClusterConfig struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 	metav1.TypeMeta   `json:",omitempty,inline"`
 
-	DataDir string              `json:"dataDir,omitempty"`
-	Spec    *ClusterSpec        `json:"spec,omitempty"`
-	Status  ClusterConfigStatus `json:"status,omitempty"`
+	//	DataDir string              `json:"dataDir,omitempty"`
```
trawler, line 73: I think we can remove this line altogether.

ncopa: oh.. leftover.

I wonder what happens if there is a JSON config as CRD already. Will it not break if we remove it?

trawler: good question. but actually adding the data dir var was a bad workaround that was never meant to be exported in the CR anyway. it is also undocumented and does not appear in default-config.

ncopa: ok lets remove it and hope it does not blow up

ncopa: I verified it. `kubectl describe   -n kube-system clusterconfig k0s` does not have anything `DataDir`.

trawler (dismissed):
Lgtm :-)

pkg/config/config.go
```diff
@@ -212,9 +225,9 @@ func GetNodeConfig(cfgPath string, k0sVars constant.CfgVars) (*v1beta1.ClusterCo
 	nodeConfig := &v1beta1.ClusterConfig{
 		ObjectMeta: cfg.ObjectMeta,
 		TypeMeta:   cfg.TypeMeta,
-		DataDir:    cfg.DataDir,
-		Spec:       clusterSpec,
-		Status:     cfg.Status,
+		//		DataDir:    cfg.DataDir,
```
ncopa, line 228: forgot this one.

pkg/config/config.go
```diff
@@ -182,9 +190,9 @@ func ClusterConfigMinusNodeConfig(config *v1beta1.ClusterConfig) *v1beta1.Cluste
 	return &v1beta1.ClusterConfig{
 		ObjectMeta: config.ObjectMeta,
 		TypeMeta:   config.TypeMeta,
-		DataDir:    config.DataDir,
-		Spec:       clusterSpec,
-		Status:     config.Status,
+		//		DataDir:    config.DataDir,
```
ncopa, line 193: this also

trawler (approved):

PR #1367, opened 2021-12-17
Added checking for a leftover socket from the previous runs

makhov:
**Issue**
When k0s crashes or gets SIGKILL, the status server socket is not unlinked. This can cause troubles on the next runs.
```
Dec 16 12:31:13 ip-172-31-28-129 k0s[4636]: Error: listen unix /run/k0s/status.sock: bind: address already in use
Dec 16 12:31:13 ip-172-31-28-129 k0s[4636]: 2021/12/16 12:31:13 listen unix /run/k0s/status.sock: bind: address already in use
Dec 16 12:31:13 ip-172-31-28-129 systemd[1]: k0scontroller.service: Main process exited, code=exited, status=1/FAILURE
Dec 16 12:31:13 ip-172-31-28-129 systemd[1]: k0scontroller.service: Failed with result 'exit-code'.
Dec 16 12:33:13 ip-172-31-28-129 systemd[1]: k0scontroller.service: Scheduled restart job, restart counter is at 4.
Dec 16 12:33:13 ip-172-31-28-129 systemd[1]: Stopped k0s - Zero Friction Kubernetes.
```

**What this PR Includes**
In the change, we check if there is a leftover socket and if something is listening on this socket. If not, we remove socket file.

### DIFF
--- a/pkg/component/status/status.go
+++ b/pkg/component/status/status.go
@@ -3,9 +3,12 @@ package status
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
+	"os"
+	"strings"
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/pkg/install"
@@ -37,6 +40,11 @@ func (s *Status) Init() error {
 		return fmt.Errorf("failed to create %s: %w", s.Socket, err)
 	}
 
+	err = checkLeftoverSocket(s.Socket)
+	if err != nil {
+		return fmt.Errorf("socket error: %w", err)
+	}
+
 	s.listener, err = net.Listen("unix", s.Socket)
 	if err != nil {
 		s.L.Errorf("failed to create listener %s", err)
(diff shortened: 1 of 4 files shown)


pkg/component/status/status.go
```diff
@@ -47,6 +55,21 @@ func (s *Status) Init() error {
 	return nil
 }
 
+// checkLeftoverSocket checks if there is a leftover socket from the previous runs. This usually happens on crash or SIGKILL
+func checkLeftoverSocket(socket string) error {
+	_, err := net.Dial("unix", "socket")
```
ncopa, line 60: Shouldn't this be the variable `socket` instead of string literal `"socket"`?

ncopa:
I wonder if we really need to test for `ErrNotExist` and "connection refused" errors?

Wouldn't it be enough to always try to delete it unless we get a successful connection? If there are other errors (like permission denied) we'd fail to delete it, right?

I also think the error message `bind: address already in use` explains better what is going on than a `"socket is busy"`, so I think we can just ignore successful connection and let it fail during bind.

I believe it could be simplified to something like:
```go
	// try remove leftover sockets that nothing is listening on
	_, err := net.Dial("unix", s.Socket)
	if err != nil {
		_ := os.Remove(s.Socket)
	}
```

pkg/component/status/status.go
makhov: Yes, of course! Thank you!

makhov:
@ncopa Yeah, makes a lot of sense. Simplified it as you proposed. Thanks!

ncopa (approved):

PR #1376, opened 2021-12-21
Autogenerate docs/cli during publish-docs CI flow

### LINKED ISSUE #1369: docs/cli/ out of date
There's a bunch of subcommands missing from the documentation in `docs/cli` directory.

- k0s airgap (+subcommands)
- k0s backup
- k0s restore
- k0s sysinfo
- k0s reset
- k0s token (+subcommands)
- maybe others

Also many of the existing ones have been manually edited both for extra content and to pass linter (#953) 

Probably best to regenerate all and ignore on linter.

kke:

Fixes #1369

* Everything in docs/cli is `git rm`'ed
* Updated the publish docs workflow to generate a fresh set of cli docs 
* The docs for kubectl subcommands (k0s_kubectl_*) are deleted and any links to them are deleted from k0s_kubectl.md

There doesn't seem to be any way to verify if it works except by making a release.


.github/workflows/publish-docs.yml
```diff
@@ -37,18 +41,50 @@ jobs:
...
+        run: make EMBEDDED_BINS_BUILDMODE=none
+
+      - name: Clean Docker
+        run: |
+          docker system prune --all --volumes --force
+
+      - name: Generate CLI docs
+        run: |
+          mkdir -p docs/cli
+          ./k0s docs markdown
```
trawler, line 68: might be faster to simply run the docs command without compiling the whole binary with `go run main.go docs markdown`

kke: Nice, didn't know `go run` works for some commands.

trawler (approved):

PR #1379, opened 2021-12-21
Modernize the github issue/PR templates

kke:

I noticed fancy forms on another project's issue creation and looks like you can create these nice forms nowadays. Instead of a prefilled textbox, you get a pretty form instead.

Before:

image

After:

Something like:

image

Reference: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema


jnummelin (approved):
Awesome

kke:
Apparently didn't work. I wonder if they're so silly that it needs to be `.yml` and not `.yaml`

PR #1388, opened 2021-12-23
Use github-actions user for marking issues stale

### LINKED ISSUE #1387: Make "stale" bot to use a non personal token
Currently the stale action runs with my token which results into this:
image

I'd rather like to see those labels coming from a bot rather than from myself. 😄

kke:

Fixes #1387 

The [default](https://github.com/actions/stale#repo-token) should be `github.token` which I believe is the `github-actions` user.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,8 +9,8 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          repo-token: ${{ secrets.GH_TOKEN }}
           stale-issue-message: 'The issue is marked as stale since no activity has been recorded in 30 days'
           stale-pr-message: 'The PR is marked as stale since no activity has been recorded in 30 days'
           days-before-stale: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     steps:
       - uses: actions/stale@v3
         with:
-          repo-token: ${{ secrets.GH_TOKEN }}
```
soider, line 12: hm but why? it is used by default?

soider: nevermind, read the issue

soider (approved):

soider:
Should not we instead change the secrets value?

kke:
> Should not we instead change the secrets value?

I think the `secrets.GH_TOKEN` that was there is Jussi's token

soider:
Yeah, I mean change it to some proper value (don't know which one is proper). Anyway, it is a side note.

Issue #1222, opened 2021-10-27
Do not accept -c or --config option for the worker subcommand


soider:
while we don't use config in the worker process we still accept it as a cli argument which makes CLI interface confusing

jnummelin:
Should be fixed already

trawler:
fixed with #1389

PR #1389, opened 2021-12-23
Remove config flag as a global Flag + setting config default location to /etc/k0s/k0s.yaml

### LINKED ISSUE #1369: docs/cli/ out of date

trawler:
Fixes #1369 #1222

This PR is a sub-section of the work entailed in the original config-refactoring PR (#1247), since a rebase was too complex.

It includes:
* removal of the config flag from unnecessary commands
* setting the default config location to /etc/k0s/k0s.yaml
* update of the docs/cli documentation
* unifying the configFromStdin and ConfigFromFile for clusterConfig to ConfigFromReader

### DIFF
--- a/cmd/token/token.go
+++ b/cmd/token/token.go
@@ -43,6 +43,5 @@ func NewTokenCmd() *cobra.Command {
 	cmd.AddCommand(tokenCreateCmd())
 	cmd.AddCommand(tokenListCmd())
 	cmd.AddCommand(tokenInvalidateCmd())
-	cmd.PersistentFlags().AddFlagSet(config.GetPersistentFlagSet())
 	return cmd
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,7 +17,10 @@ package config
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
+	"os"
 	"strings"
 	"time"
 
@@ -26,6 +29,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/k0sproject/k0s/internal/pkg/file"
 	cfgClient "github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/clientset"
 	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
 	"github.com/k0sproject/k0s/pkg/constant"
@@ -35,7 +39,12 @@ import (
 var (
 	resourceType = v1.TypeMeta{APIVersion: "k0s.k0sproject.io/v1beta1", Kind: "clusterconfigs"}
 	getOpts      = v1.GetOptions{TypeMeta: resourceType}
+	errNoConfig  = errors.New("no configuration found")
 )
+
+func IsErrNoConfig(err error) bool {
+	return err == errNoConfig
+}
 
 func getConfigFromAPI(kubeConfig string) (*v1beta1.ClusterConfig, error) {
 	timeout := time.After(120 * time.Second)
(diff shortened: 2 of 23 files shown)


kke (changes requested):
This includes all the kubectl docs in the cli/docs

With #1376 all the docs/cli stuff could be dropped.

kke:
Doesn't seem to include #1358

cmd/token/token.go
```diff
@@ -43,6 +43,7 @@ func NewTokenCmd() *cobra.Command {
 	cmd.AddCommand(tokenCreateCmd())
 	cmd.AddCommand(tokenListCmd())
 	cmd.AddCommand(tokenInvalidateCmd())
+
 	cmd.PersistentFlags().AddFlagSet(config.GetPersistentFlagSet())
```
kke, line 47: (i don't think the `k0s token` umbrella command needs any flags)

trawler:
> Doesn't seem to include #1358
I listed the wrong issue. description fixed.

ncopa:
Would love to have a comment in the commit message explaining *why* we remove --config as global flag and *why* we use `/etc/k0s/k0s.yaml` as default config location, so future devs don't need to guess.

pkg/config/config.go
```diff
@@ -117,82 +126,50 @@ func GetYamlFromFile(cfgPath string, k0sVars constant.CfgVars) (clusterConfig *v
...
-	switch cfgPath {
-	case "-":
-		clusterConfig, err = v1beta1.ConfigFromStdin(storage)
-	case "":
-		clusterConfig = v1beta1.DefaultClusterConfig(storage)
-	default:
-		clusterConfig, err = v1beta1.ConfigFromFile(cfgPath, storage)
+	if file.Exists(constant.K0sConfigPathDefault) {
+		logrus.Debugf("found config file in %s", constant.K0sConfigPathDefault)
 	}
```
ncopa, line 141: Can we avoid use `file.Exist` here? There is a theoretical possibility that file is created or deleted directly after the `file.Exist` call but before it is read (in which case the debug log message would be misleading). We can print the debug message after the file has been read.

```suggestion
```

pkg/config/config.go
```diff
@@ -117,82 +126,50 @@ func GetYamlFromFile(cfgPath string, k0sVars constant.CfgVars) (clusterConfig *v
...
+		if IsErrNoConfig(err) {
+			cfg = v1beta1.DefaultClusterConfig(storage)
+		} else {
+			return nil, err
+		}
+	} else {
+		cfg, err = v1beta1.ConfigFromReader(cfgReader, storage)
+		if err != nil {
+			return nil, err
+		}
```
ncopa, line 153:
```suggestion
		}
		logrus.Debugf("Read config from %s", cfgReader.Name())
```

pkg/config/config.go
```diff
@@ -117,82 +126,50 @@ func GetYamlFromFile(cfgPath string, k0sVars constant.CfgVars) (clusterConfig *v
...
-	case "-":
-		clusterConfig, err = v1beta1.ConfigFromStdin(storage)
-	case "":
-		clusterConfig = v1beta1.DefaultClusterConfig(storage)
-	default:
-		clusterConfig, err = v1beta1.ConfigFromFile(cfgPath, storage)
+	if file.Exists(constant.K0sConfigPathDefault) {
+		logrus.Debugf("found config file in %s", constant.K0sConfigPathDefault)
 	}
+	cfgReader, err := getConfigReader()
```
ncopa, line 142: the `cfgReader` may be an open file, but is never closed?

We may need move the logic of `getConfigReader` to here, and add a `defer cfgReader.Close()`. Note that we should not close `stdin`.

pkg/config/config.go
```diff
@@ -201,31 +178,31 @@ func GetNodeConfig(cfgPath string, k0sVars constant.CfgVars) (*v1beta1.ClusterCo
...
-	nodeConfig := &v1beta1.ClusterConfig{
-		ObjectMeta: cfg.ObjectMeta,
-		TypeMeta:   cfg.TypeMeta,
-		Spec:       clusterSpec,
-		Status:     cfg.Status,
+func getConfigReader() (io.Reader, error) {
+	switch CfgFile {
+	case "-":
+		return os.Stdin, nil
+	case "", constant.K0sConfigPathDefault:
```
ncopa, line 196: If I read this correctly, this means that if `--config $configpath` is used, but `$configpath` does not exist, we will get an error - *unless*  `$configpath` happens to be identical with `constant.K0sConfigPathDefault`.

So a missing `--config /etc/k0s/k0s.conf` will lead to error, while a missing `--config /etc/k0s/k0s.yaml` will not error but silently use the defaults.

Wouldn't it be better to return error if `--config /etc/k0s/k0s.yaml` is used but the file is missing?

```suggestion
	case "":
```

ncopa (changes requested):
We need to make sure the config file is closed after it is read (unless we read from stdin)

pkg/config/config.go
trawler: > Wouldn't it be better to return error if `--config /etc/k0s/k0s.yaml` is used but the file is missing?

In my opinion, no - because the flag configuration for `--config` always defaults now to `/etc/k0s/k0s.yaml`, and that means that users who have not used any `--config` flag input (as was used before) will always fail. And that will hurt backwards compatibility.

ncopa: > that means that users who have not used any --config flag input (as was used before) will always fail.

What I mean is that when `--config` is used, it should always fail if the specified file cannot be read, regardless if it is the default config file or not.

When no `--config` is specified, it should try the default config path, but not fail if it does not exist. It should still fail if it fails to read for any other reason (permission denied for example).

ncopa (approved):

PR #1395, opened 2021-12-28
Fix CI CLI markdown docs generation and add manual dispatch

kke:
Signed-off-by: Kimmo Lehto <klehto@mirantis.com>

`go run` needs the `static/gen_manifests.go`.

Moved logic into Makefile.

PR #1401, opened 2021-12-29
Fix manual docs update workflow dispatch

kke:

Attempts to fix the manual dispatch of docs workflow introduced in #1395 

Moves the doc-generation into `docs/Makefile` so it can be easily included when backporting typofixes for example.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -31,24 +31,38 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install mkdocs
-          pip install mkdocs-exclude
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-docs.yml
```diff
@@ -76,22 +90,22 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
 
       - name: mike deploy main
-        if: "!github.event.workflow_dispatch && contains(github.ref, 'refs/heads/main')"
+        if: contains(github.ref, 'refs/heads/main')
```
kke, line 93: I'm not sure about this.. maybe it should be only done when it's not a manual dispatch

kke:
Maybe it would be cleaner to just split it into two separate files.

PR #1404, opened 2021-12-31
Refactor and clean up context propagation and component manager

ncopa:
### Refactor component manager.

- Simplify code a bit by removing AddSync. It was only used for certificate creation which could not happen in parallel with the other initialization.
- Only stop components that was actually started.
- Stop the started component before exiting with error on startup.

### Propagate context whenever possible

- pass on the context and avoid use context.Background()/TODO() where possible. this should improve the shutdown, specially if it happens during the init/startup phase.
- Use context with errgroups and retry.Do loops so they can be cancelled.

### Misc cleanups
- remove some dead code. (those commits may be excluded in this PR if wanted)
- remove status socket on shutdown


s0j (approved):
SGTM

PR #1407, opened 2022-01-03
Update copyright boilerplate year and add CI check

kke:

Updates the copyright year in `hack/client-gen/boilerplate.go.txt` and adds a linter check to see that the copyright header exists in all go files.

According to internet wisdom, it's not necessary to regex-replace copyright headerrs in every file, but only update it in files that are changed during that year.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -46,13 +46,10 @@ jobs:
         env:
           EMBEDDED_BINS_BUILDMODE: none
 
-
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.42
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yaml
```diff
@@ -46,13 +46,10 @@ jobs:
...
 
-
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.42
           skip-pkg-cache: true
           only-new-issues: true
           skip-build-cache: true
-          config: .golangci.yml
```
kke, line 57: this setting does not exist, there was a warning about unknown key.

ncopa (approved):

PR #1409, opened 2022-01-03
config refactoring: use runtime config

### LINKED ISSUE #1202: `k0s kubeconfig` ignores config value for the external address
**Version**
```
$ k0s version
```

Current main head

**Platform**
```
$ lsb_release -a
Distributor ID:	Ubuntu
Release:	20.04
Codename:	focal
```
**What happened?**
I use following cluster configuration with set up external address:
```
k0s kc -n kube-system get ClusterConfig -oyaml
apiVersion: v1
items:
- apiVersion: k0s.k0sproject.io/v1beta1
  kind: ClusterConfig
  metadata:
    creationTimestamp: "2021-10-20T10:52:18Z"
    generation: 1
    name: k0s
    namespace: kube-system
    resourceVersion: "223"
    uid: 185d9534-264f-4084-b29d-3e7d9d2c63cf
  spec:
    api:
      address: 10.0.0.24
      externalAddress: ec2-13-49-159-141.eu-north-1.compute.amazonaws.com
      port: 6443
    network:
      dualStack: {}
      kubeProxy:
        mode: iptables
      kuberouter:
        autoMTU: true
        mtu: 0
        peerRouterASNs: ""
        peerRouterIPs: ""
      podCIDR: 10.244.0.0/16
      provider: kuberouter
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Whenever I use `k0s kubeconfig` command to have kubeclient config, either for admin or by using `k0s kubeconfig create username`, I expect to see external address in the returned kubeconfig, but I get the internal address (10.0.0.24).

```
k0s kubeconfig create user

apiVersion: v1
clusters:
- cluster:
    server: https://10.0.0.24:6443
    certificate-authority-data: ...
  name: k0s
contexts:
- context:
    cluster: k0s
    user: user
  name: k0s
current-context: k0s
kind: Config
preferences: {}
users:
- name: user
  user:
    client-certificate-data: ...
    client-key-data: ...```

**How To Reproduce**
Use the steps from the description

**Expected behavior**
I expect kubeconfig to have external address as a value for the `cluster.server`

**Additional context**
Add any other context about the problem here.

### PR DESCRIPTION
**Pull request by trawler**

**Issue**
Fixes #1202 #1261
**What this PR Includes**
This PR refactors the entire handling of k0s config. Among other changes it:

1. Removes `--config` as a global flag to all commands
2. Keeps `--config` flag only on the `k0s controller` and `k0s reset` commands (note on the reset command below).
3. Adds a default value to k0s config: `/etc/k0s/k0s.yaml`
3. Generates a runtime config whenever k0s controller goes up and keeps it under `/run/k0s/k0s.yaml`. This runtime config file is removed when the controller is stopped.
4. The runtime config is a merged config of custom values and default config values (#1261)

Note on k0s reset:
Since the runtime config only exists while the controller is up, k0s reset (that can only be performed while the controller is down) needs to still retail the `--config` flag. A workaround to that would be to use the global location `/etc/k0s/k0s.yaml` as a source of configuration.

**To Do:**
- [X] Resolve correct runtime config file for single nodes
- [x] #1354
- [x] Fix prerun validate logic

### CODE REVIEW

File: cmd/controller/controller.go [hunk 1]
```diff
@@ -124,15 +120,25 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 	if err := dir.Init(c.K0sVars.CertRootDir, constant.CertRootDirMode); err != nil {
 		return err
 	}
+	// let's make sure run-dir exists
+	if err := dir.Init(c.K0sVars.RunDir, constant.RunDirMode); err != nil {
+		logrus.Fatalf("failed to initialize dir: %v", err)
```
kke, line 125: These could just return the (decorated?) error instead of relying on the exit functionality of `logrus.Fatalf`.

pkg/config/api_config.go
```diff
@@ -0,0 +1,99 @@
...
+	getOpts      = v1.GetOptions{TypeMeta: resourceType}
+)
+
+// run a config-request from the API and wait until the API is up
+func (rules *ClientConfigLoadingRules) getConfigFromAPI(client k0sv1beta1.K0sV1beta1Interface) (*v1beta1.ClusterConfig, error) {
+	timeout := time.After(120 * time.Second)
+	ticker := time.NewTicker(3 * time.Second)
+	defer ticker.Stop()
+	// Keep trying until we're timed out or got a result or got an error
+	for {
```
soider, line 43: let's use avast/retry-go as we do in other places?

cmd/reset/reset.go
```diff
@@ -75,9 +75,11 @@ func (c *CmdOpts) reset() error {
 
 func preRunValidateConfig(_ *cobra.Command, _ []string) error {
```
jnummelin, line 76: Seems this `preRunValidateConfig` func is duplicated in many places

cmd/restore/restore.go
```diff
@@ -52,12 +52,17 @@ func NewRestoreCmd() *cobra.Command {
...
-		PreRunE: preRunValidateConfig,
 	}
 
 	cmd.SilenceUsage = true
-	cmd.Flags().StringVar(&restoredConfigPath, "config-out", "", "Specify desired name and full path for the restored k0s.yaml file (default: ${cwd}/k0s_<archive timestamp>.yaml)")
-	cmd.Flags().AddFlagSet(config.FileInputFlag())
+
+	cwd, err := os.Getwd()
+	if err != nil {
+		return nil
```
jnummelin, line 61: Returning `nil` in case `os.Getwd()` return error seems bit off

pkg/config/cli.go
```diff
@@ -195,11 +197,26 @@ func FileInputFlag() *pflag.FlagSet {
 func GetCmdOpts() CLIOptions {
 	K0sVars = constant.GetConfig(DataDir)
 
+	if controllerOpts.SingleNode {
+		controllerOpts.EnableWorker = true
+		K0sVars.DefaultStorageType = "kine"
+	}
+
+	// when a custom config file is used, verify that it can be opened
+	if CfgFile != constant.K0sConfigPathDefault {
```
jnummelin, line 206: What happens now if I give `--config /etc/k0s/k0s.yaml` and it does not exist? I'd expect this to error out, but reading the code I'd assume it'll go and use defaults?

jnummelin (commented):
Left some questions and general comments. The direction is good in general.

I'm slightly worried that dropping the `-config` option in many commands will break backwards compatibility e.g. with k0sctl. Maybe rebase with the #1450 to verify this.

pkg/config/cli.go
trawler: technically you're right, but it would be a strange behaviour, since `--config` defaults to `/etc/k0s/k0s.yaml` anyway. If we error this out, then running k0s without any config would fail. The idea behind this was to enable backwards compatibility but also allow users to place their config in a default location without having to specify it explicitly.

kke:
> Left some questions and general comments. The direction is good in general.
> 
> I'm slightly worried that dropping the `-config` option in many commands will break backwards compatibility e.g. with k0sctl. Maybe rebase with the #1450 to verify this.

You could leave a nonfunctional hidden `--config` flag, maybe it could warn but ignore.

pkg/config/cli.go
kke: For these you would need some mechanism to detect if `--config` was given or defaulted.

cmd/restore/restore.go
kke: I don't see why it needs it in the first place. It could just `return fmt.Sprintf("k0s_%s.yaml", fName)`, the working directory is assumed when paths are relative.

cmd/api/api.go
```diff
@@ -63,16 +63,10 @@ func NewAPICmd() *cobra.Command {
 			logrus.SetOutput(os.Stdout)
 
 			c := CmdOpts(config.GetCmdOpts())
-			cfg, err := config.GetNodeConfig(c.CfgFile, c.K0sVars)
-			if err != nil {
-				return err
-			}
-			c.NodeConfig = cfg
 			return c.startAPI()
```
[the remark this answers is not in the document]
trawler: it gets it from the c.NodeConfig var: https://github.com/k0sproject/k0s/blob/d0d440c0feba328ee65869a212902a66f75acdf9/cmd/api/api.go#L104

trawler:
> Left some questions and general comments. The direction is good in general.
> 
> I'm slightly worried that dropping the `-config` option in many commands will break backwards compatibility e.g. with k0sctl. Maybe rebase with the #1450 to verify this.

k0sctl requires a very minor change for token creation. after that, `make check-k0sctl` passes: https://github.com/k0sproject/k0sctl/compare/main...trawler:k0s_config_flag_1409?expand=1

pkg/config/cli.go
trawler: Fixed in the latest commit

trawler:
Waiting for next k0sctl that includes https://github.com/k0sproject/k0sctl/pull/303 to run the `make check-k0sctl` test again

kke (approved):
It looks ok to me but it's fairly big and touches everything so I think the best would be to get it in as soon as possible before a release that includes it so that there is time to find any problems.

soider (approved):

PR #1411, opened 2022-01-03
Support for user lookup via NSS

### LINKED ISSUE #1402: Installation fails with nss-altfiles

### Version

v1.23.1+k0s.0

### Platform

```shell
Fedora CoreOS 32.20201104.3.0
```

### What happened?

As was foreshadowed in #1079
> forcing file based lookup will probably just break people's setups in a different way >:

the fix made there is breaking on rpm-ostree base distros (as far as I can tell that would be CoreOS/Flatcar/Silverblue) but the real issue is with [altfiles](https://github.com/aperezdc/nss-altfiles).

### Steps to reproduce

1. Install and configure nss-altfiles (or as in my case use Fedora Coreos)
2. Have one more of the installConfig.users defined in [/lib/passwd](https://github.com/coreos/fedora-coreos-config/blob/d1743185ffd47dce8fa5e885894d3a6a068b5dee/manifests/passwd#L16)
3. `k0s install controller`

### Expected behavior

exit 0, nor error

### Actual behavior

Errors with failure to create the user from step 2 (on a default fedora coreos this happens to etcd).
The user indeed exists so the error _to create_ is expected but this comes from the osusergo implementation not being able to find the user via nsss in the first place.

### Screenshots and logs

_No response_

### Additional context

The obvious workaround (once all the above troubleshooting was figured out) is to change the etcduser and create a new one
```
  installConfig:
    users:
      etcdUser: k0s-etcd
```
but it would be nice for it to work out of the box (other projects have resorted to having separate musl/glibc builds).

ncopa:
Add support to lookup users via NSS by calling external command  `id`.

Fixes #1402

### DIFF
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -16,26 +16,25 @@ limitations under the License.
 package users
 
 import (
+	"errors"
+	"os/exec"
 	"os/user"
 	"strconv"
+	"strings"
 )
 
 // GetUID returns uid of given username and logs a warning if its missing
 func GetUID(name string) (int, error) {
 	entry, err := user.Lookup(name)
-	if err != nil {
-		return 0, err
+	if err == nil {
+		return strconv.Atoi(entry.Uid)
 	}
-	return strconv.Atoi(entry.Uid)
-}
-
-func CheckIfUserExists(name string) (bool, error) {
-	_, err := user.Lookup(name)
-	if _, ok := err.(user.UnknownUserError); ok {
-		return false, nil
+	if errors.Is(err, user.UnknownUserError(name)) {
+		// fallback to call external `id` in case NSS is used
+		out, err := exec.Command("/usr/bin/id", "-u", name).CombinedOutput()
+		if err == nil {
+			return strconv.Atoi(strings.TrimSuffix(string(out), "\n"))
+		}
 	}
-	if err != nil {
-		return false, err
-	}
-	return true, nil
+	return 0, err
 }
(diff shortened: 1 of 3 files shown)


internal/pkg/users/users.go
```diff
@@ -16,26 +16,24 @@ limitations under the License.
...
-		return 0, err
+	if err == nil {
+		return strconv.Atoi(entry.Uid)
 	}
-	return strconv.Atoi(entry.Uid)
-}
-
-func CheckIfUserExists(name string) (bool, error) {
-	_, err := user.Lookup(name)
 	if _, ok := err.(user.UnknownUserError); ok {
```
s0j, line 31: This can probably be `errors.Is(..)`

internal/pkg/users/users.go
```diff
@@ -16,26 +16,24 @@ limitations under the License.
...
-		return false, nil
-	}
-	if err != nil {
-		return false, err
+		// fallback to call external `id` in case NSS is used
+		out, err := exec.Command("id", "-u", name).CombinedOutput()
```
s0j, line 33: Can this be the full path to `id` instead?

ncopa: It can.

internal/pkg/users/users.go
ncopa: good idea

pkg/install/users.go
```diff
@@ -73,22 +73,11 @@ func DeleteControllerUsers(clusterConfig *v1beta1.ClusterConfig) error {
...
 func EnsureUser(name string, homeDir string) error {
-	exists, err := users.CheckIfUserExists(name)
-	if err != nil {
-		return err
-	}
-
-	if exists {
-		return nil
+	_, err := users.GetUID(name)
+	if _, ok := err.(user.UnknownUserError); ok {
```
s0j, line 77: I missed this originally -- this could be `errors.Is()` as well. Not holding this up though.

s0j (dismissed):

pkg/install/users.go
ncopa: doh... I think we should fix this too.

s0j (approved):

PR #1413, opened 2022-01-04
Mass-replace copyright boilerplate year

kke:
Replaces the year in the copyright notice in all source files.

Not so nice diff but would remove the need to change it in every file touched by any PR during the following year.

* Fixes the CI linter header check
* Includes some other linter fixes + excludes
* Disables the `only-new-issues`

There were plenty of files with no header.


jnummelin (dismissed):

kke:
as a sidenote, we should maybe enable some of the other linters included in golangci-lint.

kke:
If we have `only-new-issues: false` it means we only get issues if the diff reaches the boilerplate, so practically only on new files. Changing code won't usually cause the diff to include the boilerplate and the linter won't complain. 

Of course we could just `grep -q "Copyright $(date +%Y)" $(git diff-tree --no-commit-id --name-only -r ${{ github.sha }})` instead of using the `goheader` linter.

ncopa:
I think it would be nice if the CI fixes were in separate commit so it easily can be cherry picked to stable branches

kke:
> I think it would be nice if the CI fixes were in separate commit so it easily can be cherry picked to stable branches

Unfortunately it's impossible as explained above. Keeping `only-new-issues: true` as is, makes the check worthless as the diff only hits the boilerplate on new files, changing it to `false` will make every PR fail for every file in the repo.

jnummelin (approved):

PR #1414, opened 2022-01-04
Add filename usage hint to "k0s restore"

kke:

Before:

```
$ ./k0s restore --help
restore k0s state from given backup archive. Must be run as root (or with sudo)

Usage:
  k0s restore [flags]
  ...
$ ./k0s restore
Error: path to backup archive expected
```

After:

```
$ ./k0s restore --help
restore k0s state from given backup archive. Must be run as root (or with sudo)

Usage:
  k0s restore filename [flags]
```

### DIFF
--- a/cmd/restore/restore.go
+++ b/cmd/restore/restore.go
@@ -2,7 +2,7 @@
 // +build !windows
 
 /*
-Copyright 2021 k0s authors
+Copyright 2022 k0s authors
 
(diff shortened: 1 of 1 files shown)


kke:
Btw this shouldn't have passed if #1407 actually worked.

makhov (approved):

PR #1429, opened 2022-01-10
Use path "-" in backup and restore to write to stdout or read from stdin

### LINKED ISSUE #1410: Encrypt backups

I would like to encrypt my backups to meet security requirements from my company.

Backup encryption happens automatically with keys I supply to the software.

Backups could be encrypted separately, but that leaves a window where a bad agent still has access to the unencrypted backup, thus it would be better if it happened as the backup is being made.

### Additional context

_No response_

kke:

Fixes #1410 

* `k0s backup --save-path -` will output the archive to stdout
* `k0s restore -` will read the archive from stdin

You can use these to get encrypted backups via something like:

```shell
k0s backup --save-path - | gpg -c > backup.tar.gz.gpg
```

```shell
gpg -d backup.tar.gz.gpg | k0s restore -
```

Or remote backups via something like:

```shell
ssh host.example.com k0s backup --save-path - > backup.tar.gz
```

```shell
ssh host.example.com k0s restore - < backup.tar.gz
```

Note that the unencrypted backups will still briefly exist on the controller filesystem while the archive is being built.

### DIFF
--- a/cmd/backup/backup.go
+++ b/cmd/backup/backup.go
@@ -55,7 +55,7 @@ func NewBackupCmd() *cobra.Command {
 		},
 		PreRunE: preRunValidateConfig,
 	}
-	cmd.Flags().StringVar(&savePath, "save-path", "", "destination directory path for backup assets")
+	cmd.Flags().StringVar(&savePath, "save-path", "", "destination directory path for backup assets, use '-' for stdout")
 	cmd.SilenceUsage = true
 	cmd.Flags().AddFlagSet(config.FileInputFlag())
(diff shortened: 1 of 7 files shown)


s0j (approved):

wkonitzer:
@kke Thanks for this, but how does this work with "k0sctl backup"? Please can we also update the docs with a tested encrypted backup example?

kke:
> Thanks for this, but how does this work with "k0sctl backup"? Please can we also update the docs with a tested encrypted backup example?

K0sctl implementation needs to be done separately in k0sctl.

PR #1433, opened 2022-01-11
Test that kubeconfig returns externalAddress

ncopa:
Add a test to verify that the server address is set to externalAddress
when it is set, and that the custom ports are included in the
kubeconfig.

This is a test for issue https://github.com/k0sproject/k0s/issues/1202

### DIFF
--- a/inttest/customports/customports_test.go
+++ b/inttest/customports/customports_test.go
@@ -138,13 +138,13 @@ func (ds *Suite) TestControllerJoinsWithCustomPort() {
 	ds.Require().NoError(common.WaitForPodLogs(kc, "kube-system"))
 
 	// https://github.com/k0sproject/k0s/issues/1202
-	ds.T().Log("test that kubeconfig includes externalAddress")
-	ssh, err := ds.SSH(ds.ControllerNode(0))
-	ds.Require().NoError(err)
-	defer ssh.Disconnect()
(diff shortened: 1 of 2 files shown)


trawler (dismissed):

trawler (approved):

inttest/customports/customports_test.go
```diff
@@ -135,4 +136,15 @@ func (ds *Suite) TestControllerJoinsWithCustomPort() {
 
 	ds.T().Log("waiting to get logs from pods")
 	ds.Require().NoError(common.WaitForPodLogs(kc, "kube-system"))
+
+	// https://github.com/k0sproject/k0s/issues/1202
+	ds.T().Log("test that kubeconfig includes externalAddress")
```
twz123, line 141: You could also wrap the following lines into their own sub-test via `ds.T().Run("kubeconfigIncludesExternalAddress", ...)` instead of logging it.

twz123 (approved):

inttest/customports/customports_test.go
ncopa: I guess we should do so with the other `Log` calls too in that case? for consistency

ncopa: I see what you mean now. Yeah, that makes sense.

twz123 (approved):

PR #1439, opened 2022-01-13
Fix stdout output of config when using k0s restore --config-out -

kke:

`io.Copy` args were reversed. How it passed tests is beyond my understanding. Why `os.Stdout` passes as `io.Reader`?


kke:
hmm some kind of confusion here

trawler (approved):

PR #1442, opened 2022-01-14
Building multiarch k0s image on release

makhov:

Currently, we are building only amd64 docker images: https://hub.docker.com/r/k0sproject/k0s/tags
Screenshot 2022-01-14 at 14 33 00

With this change, we will build a multiarch image with armv7/arm64 support:

Screenshot 2022-01-14 at 14 33 57

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
     outputs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
       tag_name: ${{ steps.branch_name.outputs.TAG_NAME }}
+      image_tag: ${{ steps.image_tag.outputs.IMAGE_TAGS }}
     steps:
       # Ugly hack to get the tag name
       # github.ref gives the full reference like refs.tags.v0.0.1-beta1
@@ -33,7 +34,13 @@ jobs:
           release_name: ${{ github.ref }}
           draft: true # So we can manually edit before publishing
           prerelease: ${{ contains(github.ref, '-') }} # v0.1.2-beta1, 1.2.3-rc1
-
+      - name: Prepare image tags
+        id: image_tag
+        # Basically just replace the '+' with '-' as '+' is not allowed in tags
+        run: |
+          TAGS=${{ steps.branch_name.outputs.TAG_NAME }}
+          TAGS=${TAGS//+/-}
+          echo ::set-output name=IMAGE_TAGS::${TAGS}
   x64:
     needs: release
     runs-on: [self-hosted,linux,x64]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,14 +1,15 @@
-FROM alpine:3.13
+ARG ARCH
+FROM ${ARCH}alpine:3.13
+
+ARG K0S_VERSION
 
 RUN apk add --no-cache bash coreutils findutils iptables curl tini
```
ncopa, line 6: I don't think we need iptables any more.

Dockerfile
```diff
@@ -1,14 +1,15 @@
-FROM alpine:3.13
+ARG ARCH
+FROM ${ARCH}alpine:3.13
```
ncopa, line 2: Maybe use `alpine:3.15`?

Dockerfile
```diff
@@ -1,14 +1,15 @@
...
+
+ARG K0S_VERSION
 
 RUN apk add --no-cache bash coreutils findutils iptables curl tini
 
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.20.1/bin/linux/amd64/kubectl \
-       && chmod +x ./kubectl \
-       && mv ./kubectl /usr/local/bin/kubectl
+RUN curl -sSLf https://get.k0s.sh | K0S_VERSION=${K0S_VERSION} sh
```
jnummelin, line 8: I don't think the binaries are downloadable at this point when we build the image 🤔

makhov: I moved image building to the very last step of the GHA workflow, so I think it should work.

Dockerfile
makhov: Removed iptables

Dockerfile
makhov: Updated the version

Dockerfile
makhov: Ok, now I see. I switched back to passing a binary via `ADD` in the Dockerfile

.github/workflows/release.yml
```diff
@@ -290,6 +272,63 @@ jobs:
...
+      # Need to remove this from maintenance branches
+      # Done as separate step for better control when we push latest
+      - name: Build image and push to Docker hub
+        if: "!contains(github.ref, '-')"
+        uses: docker/build-push-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
+          tags: docker.io/k0sproject/k0s:latest
+          push: true
+
```
jnummelin, line 331: I'm thinking if we should just stop pushing `latest` in general. It's a bad advise in general to use latest tags and for us it's bit problematic to manage which ones are actually latest tags.

.github/workflows/release.yml
```diff
@@ -290,6 +272,63 @@ jobs:
...
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Fetch k0s amd64
+        uses: actions/download-artifact@v2
+        with:
+          name: k0s-amd64
```
jnummelin, line 303: I don't get how `name: k0s-amd64` maps into the correct binary as the artifacts have name like `k0s-v1.23.1-rc.2+k0s.1-amd64`?

jnummelin (commented):
Good stuff, left one clarification question,

.github/workflows/release.yml
makhov: It's just build artifacts (not those that will be posted to the release), so they have [simple names](https://github.com/k0sproject/k0s/pull/1442/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R199) like `k0s-arm64`

.github/workflows/release.yml
makhov: Yeah, it could be useful only for some testing or development purposes. Should I remove it or do we need some discussion?

soider: I'd say we can safely drop it.

makhov: Deleted it

jnummelin (approved):

odidev:
Hi Team,  

I have seen that this PR is merged, and the changes are available in the master branch as well. However, multi-arch k0s image is still not available at [docker hub](https://hub.docker.com/r/k0sproject/k0s/tags). Docker hub only has Linux/AMD64 image available in the latest tag. 

May I know, are you planning to roll out k0s multi-arch docker images soon?

odidev:
Hi Team, may I know is there any ETA for the official Linux/ARM64 k0s docker image release?

jnummelin:
This did not get "backported" into existing release branches. This will mean that we'll probably ship the upcoming 1.24 (expected in mid April) with multiarch images.

k0s-in-container has not been super high in our priority as there's been very little usage for it. I guess that's in general coming from the fact that it'll mean running containers-in-containers which is not the most stable setup. 😄 

What's your use case for running k0s in a container?

odidev:
> What's your use case for running k0s in a container?

Thank you for the response @jnummelin.
I am just trying to install k0s using the [docker method](https://docs.k0sproject.io/v1.23.3+k0s.1/k0s-in-docker/) on my Linux/ARM64 machine, for that, I needed Linux/ARM64 k0s docker image.

PR #1408, opened 2022-01-03
Use k8s.io/base-component/cli for k0s kubectl command

soider:
Signed-off-by: Mikhail Sakhnov <msakhnov@mirantis.com>
addresses #1405

Issue #1405, opened 2022-01-02
k0s kubectl does not respect -v flag


adamlamar:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Version

v1.23.1+k0s.0

### Platform

```shell
# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 21.10
Release:	21.10
Codename:	impish
```

### What happened?

In `1.22.5+k0s.0`, running `k0s kubectl -v=8 get nodes` produces the expected node list and debug output from kubectl.

In `1.23.1+k0s.0`, the global `-v` flag to kubectl is unrecognized:

```
# k0s kubectl -v=8 get nodes
Error: unknown shorthand flag: 'v' in -v=8
See 'k0s kubectl get --help' for usage.
```

### Steps to reproduce

1. Run `k0s kubectl -v=8 get nodes`

### Expected behavior

The command should be run with verbose output.

### Actual behavior

The command fails to run.

soider:
Looks like it was introduced in the 2ecb0c52f16bfb4f29e5081605b1712f814b5b49

PR #1444, opened 2022-01-14
Add arguments from upstream kubectl; add e2e tests for common cases

soider:
Addressed #1405 

Ports e2e tests from #1408 but uses more simpler approach which doesn't fails on arm.

The approach used in #1408 is broken on arm due to different logic for argument handling: attaching KubectlCmd to the RootCmd overrides some arguments (but only on arm).

### DIFF
--- a/inttest/kubectl/kubectl_test.go
+++ b/inttest/kubectl/kubectl_test.go
@@ -1,0 +1,130 @@
+package kubectl
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/k0sproject/k0s/inttest/common"
+	"github.com/stretchr/testify/suite"
+)
+
+type KubectlSuite struct {
+	common.FootlooseSuite
+}
+
+const pluginContent = `#!/bin/bash
+
+echo "foo-plugin"
+`
+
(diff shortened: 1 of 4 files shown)


inttest/kubectl/kubectl_test.go
```diff
@@ -0,0 +1,130 @@
...
+	common.FootlooseSuite
+}
+
+const pluginContent = `#!/bin/bash
+
+echo "foo-plugin"
+`
+
+func (s *KubectlSuite) TestEmbeddedKubectl() {
+	s.Require().NoError(s.InitController(0, "--enable-k0s-cloud-provider", "--k0s-cloud-provider-update-frequency=5s"))
```
s0j, line 21: Do you really need `k0s-cloud-provider`?

s0j (dismissed):
LGTM

makhov (dismissed):

inttest/kubectl/kubectl_test.go
soider: not at all, some copy-paste left-overs from check-basic

makhov (approved):

PR #1457, opened 2022-01-19
Add landing page to docs. Amend navigation layout.

stevejr:
- Add new landing page
- Add new options to mkdocs.yaml to enable the landing page and the revised nav layout. Note that there now needs to be an index.md file in the docs root directory which will use the customer_dir/overrides/main.html. See https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/?h=navigation#section-index-pages
- Move analytics from override to mkdocs config - see https://squidfunk.github.io/mkdocs-material/setup/setting-up-site-analytics/#configuration

Signed-off-by: Steve Richards <srichards@mirantis.com>


stevejr:
Miska - Do we want an announcement banner at the top of the landing page? The Materials site has a banner to advertise following them on twitter. If we do not want a banner then the styling needs to be changed as currently this PR still includes the squid funk banner

miskun:
We can have there banner to encourage people to follow us on twitter as well.

stevejr:
@miskun - The nav and banner are now how you want them but I cannot get the image on the front page to dynamically change size as per the mkdocs-material site (https://squidfunk.github.io/mkdocs-material/). Maybe someone with a bit more HTML/CSS knowledge can assist?

PR #1459, opened 2022-01-20
Instead of compiling binary use latest release for the manual publish workflow

soider:
Use latest public release for publishing docs, instead of compiling by `go run main.go`

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -15,6 +15,8 @@ RUN \
   apk add --no-cache \
     git \
     git-fast-import \
+    curl \
+    jq \
   && apk add --no-cache --virtual .build gcc musl-dev \
   && python -m pip install --upgrade pip \
   && pip install -r requirements.txt \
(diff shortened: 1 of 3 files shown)


docs/Dockerfile
```diff
@@ -15,6 +15,8 @@ RUN \
   apk add --no-cache \
     git \
     git-fast-import \
+    curl \
+    jq \
```
soider, line 19: I don't know when this Dockerfile is used, but if it is used as runner for the publish workflow it is easier to pur k0s binary download process here

docs/Makefile
```diff
@@ -21,13 +21,19 @@ endif
 docs: cli
 	cd .. && mkdocs build --strict
 
+LATEST_RELEASE_META_URL=https://api.github.com/repos/k0sproject/k0s/releases/latest
```
jnummelin, line 24: Instead of "latest" shouldn't we use the version given as input for the workflow?

soider: yeah, makes sense

kke: yes

jnummelin (approved):

PR #1462, opened 2022-01-25
Download controller-gen tool through hack/ci-deps

soider:
Download controller-gen tool through hack/ci-deps
refresh manifest for the dynamic congifuration


Makefile
```diff
@@ -56,7 +56,12 @@ endif
 
 go_clientgen := $(shell which client-gen)
 ifeq ($(go_clientgen),)
-go_clientgen := cd hack/ci-deps && go install k8s.io/code-generator/cmd/client-gen@v0.22.2 && cd ../.. && test -x "${GOPATH}/bin/client-gen"
+go_clientgen := cd hack/ci-deps && go install k8s.io/code-generator/cmd/client-gen@v0.22.2 && cd ../.. && "${GOPATH}/bin/client-gen"
+endif
+
+go_controllergen := $(shell which controller-gen)
+ifeq ($(go_controllergen),)
+go_controllergen := cd hack/ci-deps && go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.8.0 && cd ../.. && "${GOPATH}/bin/controller-gen"
```
kke, line 64: `&& "${GOPATH}/bin/controller-gen"` should probably be echo or something?

soider: I guess no, it should just call this binary, otherwise it doesn't work.

They way it is used in further code is ```
$(go_controlergen) <controller-gen-args>
```

Comment (kke) [reply Makefile:64]: Hmm right, the whole thing is a string.

This is how it's done in k0sctl:

```makefile
golint := $(shell which golangci-lint)
ifeq ($(golint),)
golint := $(shell go env GOPATH)/bin/golangci-lint
endif

$(golint):
	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest
```

makhov (approved):

PR #1470, opened 2022-01-26
Fix k0s kubectl config flag

kke:

There was a strange `logs.AddFlags` call in `cmd/kubectl/kubectl.go`. Likely caused by some editor auto-importing magic.

Not entirely sure how it affected functionality.


kke:
Hmm, actually was it there to fix the log flush option?

kke:
Yes it was.

PR #1471, opened 2022-01-26
Config edit and status subcommands

### LINKED ISSUE #1453: UX: `k0s config edit` command
The command would allow editing the dynamic config on the API. Basically an "alias" for `kubectl edit k0sconfig k0s ...`

kke:

Fixes #1453 

Adds `k0s config edit` which is a shortcut to `k0s kubectl -n kube-system edit clusterconfig k0s`.
and `k0s config status` which is a shortcut to `k0s kubectl -n kube-system get event --field-selector involvedObject.name=k0s`

### DIFF
--- a/cmd/config/edit.go
+++ b/cmd/config/edit.go
@@ -16,21 +16,22 @@ limitations under the License.
 package config
 
 import (
+	"os"
+
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/spf13/cobra"
 )
 
-type CmdOpts config.CLIOptions
-
-func NewConfigCmd() *cobra.Command {
+func NewEditCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "config",
-		Short: "Configuration related sub-commands",
+		Use:   "edit",
+		Short: "Launch the editor configured in your shell to edit k0s configuration",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			c := CmdOpts(config.GetCmdOpts())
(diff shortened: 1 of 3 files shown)


kke:
Maybe it could be smoketested by creating a shell script and setting it as EDITOR and it would change something in the file and then check if that change was applied 🤔

cmd/config/edit.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/spf13/cobra"
+)
+
+func NewEditCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "edit",
+		Short: "Launch the editor configured in your shell to edit k0s configuration",
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			c := CmdOpts(config.GetCmdOpts())
+			os.Args = []string{"k0s", "kubectl", "--data-dir", c.K0sVars.DataDir, "-n", "kube-system", "edit", "clusterconfig", "k0s"}
```
jnummelin, line 31: Shouldn't this take the path to k0s from `os.Executable()`?

kke: It doesn't exec the binary, just fakes the os.Args for cmd.Execute (which only uses Args[1:])

But yes maybe it could reuse Args[0]

kke: `cmd.Execute()` calls `c.Parent().Root()`, that's how it ends up in the kubectl command.

jnummelin: oh, right. somehow my brain interpreted the `cmd.Execute` as `os.Execute`... 🤦

jnummelin (dismissed):

mviitane (approved):
LGTM

Issue #1420, opened 2022-01-06
plans for podSecurityPolicy given deprecated status


cortopy:
### Is your feature request related to a problem? Please describe.

Pod security polices are deprecated since 1.21. However, k0s still allows configuring a default PodSecurityPolicy and the security model seems to depend on it

### Describe the solution you would like

I've been searching on the repo for any discussion about this and couldn't find anything, so this issue is just to know what will happen next.

[PodSecurity](https://kubernetes.io/docs/concepts/security/pod-security-admission/) is confirmed to be the native successor and is already in beta and enabled by default in the latest k0s

There are other options too like Open Policy Agent (OPA). They may be more sophisticated but would bring in a new dependency to manage

### Describe alternatives you've considered

There is no other alternative, since PSP are already deprecated

makhov:
Technically, Pod Security Standards could be used with the current k0s version with some manual actions. 
Some automation could be done for this, but it requires proper feature design. 

**How to enable PodSecurity with k0s**

PodSecurity can be statically configured in the [Admission Controller configuration](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-admission-controller/#configure-the-admission-controller). 

For example:
1. Create admission control config file:
```
apiVersion: apiserver.config.k8s.io/v1
kind: AdmissionConfiguration
plugins:
- name: PodSecurity
  configuration:
    apiVersion: pod-security.admission.config.k8s.io/v1beta1
    kind: PodSecurityConfiguration
    # Defaults applied when a mode label is not set.
    defaults:
      enforce: "privileged"
      enforce-version: "latest"
    exemptions:
      # Don't forget to exempt namespaces or users that are responsible for deploying
      # cluster components, because they need to run privileged containers
      usernames: ["admin"] 
      namespaces: ["kube-system"] 
```
2. Add kube-apiserver arguments to the k0s configuration

```
apiVersion: k0s.k0sproject.io/v1beta1
kind: ClusterConfig
spec:
  api:
    extraArgs:
      disable-admission-plugins: PodSecurityPolicy # if you want to disable PodSecurityPolicy admission controller, not required
      enable-admission-plugins: PodSecurity        # only for Kubernetes 1.22, since 1.23 it's enabled by default
      admission-control-config-file: /path/to/admission/control/config.yaml
    
```
[...]

PR #1475, opened 2022-01-27
Pod Security Standards support

makhov:

Fixes: #1420


jnummelin:
We (the core maintainers) decided not to implement this at k0s level and allow users themselves to take full control of the PodSecurity Standards.

That being said it's probably worth adding some notes on this on the docs, #1628 for tracking that effort

PR #1486, opened 2022-02-01
Add configuration ability to collect  Calico PrometheusMetrics

alexapps:
This PR includes the changes that add ability to manage Calico Prometheus Metrics collection that by default is switched off

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
@@ -48,6 +48,9 @@ type Calico struct {
 
 	// Windows Nodes (default: false)
 	WithWindowsNodes bool `json:"withWindowsNodes"`
+
+	// PrometheusMetricsEnabled enables the Prometheus metrics server in Felix if set to true
+	PrometheusMetricsEnabled bool `json:"prometheusMetricsEnabled"`
 }
 
 // DefaultCalico returns sane defaults for calico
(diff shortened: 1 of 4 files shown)


static/manifests/calico/DaemonSet/calico-node.yaml
```diff
@@ -171,6 +171,11 @@ spec:
                 configMapKeyRef:
                   name: calico-config
                   key: veth_mtu
+            # Enable the Prometheus metrics server in Felix
+            {{- if .PrometheusMetricsEnabled }}
+            - name: FELIX_PROMETHEUSMETRICSENABLED
+              value: "true"
+            {{- end }}
```
jnummelin, line 178: If metrics is enabled, should we also put in the standard scrape labels?

jnummelin:
Is there a general drawback if we'd enable the metrics always?

pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
```diff
@@ -48,6 +48,9 @@ type Calico struct {
 
 	// Windows Nodes (default: false)
 	WithWindowsNodes bool `json:"withWindowsNodes"`
+
+	// PrometheusMetricsEnabled enables the Prometheus metrics server in Felix if set to true
+	PrometheusMetricsEnabled bool `json:"prometheusMetricsEnabled"`
```
soider, line 53: I think, the manifests are required to be regenerated by calling `make manifests`

static/manifests/calico/DaemonSet/calico-node.yaml
alexapps: Probably yes, I could not cover all implementation details

alexapps:
Implemented in the https://github.com/k0sproject/k0s/pull/1489

PR #1497, opened 2022-02-07
Ensure staged embedded-binaries + `k0s` have matching timestamps

s0j:
This provides enough information at startup to determine if the embedded-bin should be updated (sourced from a different `k0s` executable).

Previously, the current timestamp was used on staged embedded-bins which could open a window of time where embedded-bins won't get updated.

### DIFF
--- a/pkg/assets/stage.go
+++ b/pkg/assets/stage.go
@@ -27,21 +27,19 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// ExecutableIsOlder return true if currently running executable is older than given filepath
-func ExecutableIsOlder(filepath string) bool {
-	ex, err := os.Executable()
-	if err != nil {
-		return false
+// EmbeddedBinaryNeedsUpdate returns true if the provided embedded binary file should
+// be updated. This determination is based on the modification times of both the provided
+// executable, and embedded binary executable. It is expected that the embedded binary
+// modification times should match the main `k0s` executable.
+func EmbeddedBinaryNeedsUpdate(exinfo os.FileInfo, embeddedBinaryPath string) bool {
+	if pathinfo, err := os.Stat(embeddedBinaryPath); err == nil {
+		return !exinfo.ModTime().Equal(pathinfo.ModTime())
 	}
-	exinfo, err := os.Stat(ex)
-	if err != nil {
-		return false
-	}
-	pathinfo, err := os.Stat(filepath)
-	if err != nil {
-		return false
-	}
-	return exinfo.ModTime().Unix() < pathinfo.ModTime().Unix()
+
+	// If the stat fails, the file is either missing or permissions are missing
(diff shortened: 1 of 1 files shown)


pkg/assets/stage.go
```diff
@@ -73,7 +71,19 @@ func Stage(dataDir string, name string, filemode os.FileMode) error {
 		return fmt.Errorf("failed to create dir %s: %w", filepath.Dir(p), err)
 	}
 
-	if ExecutableIsOlder(p) {
+	selfexe, err := os.Executable()
+	if err != nil {
+		logrus.Warn(err)
```
twz123, line 76: If the error is returned anyways, wouldn't it be the caller's decision if and how this should be logged?

pkg/assets/stage.go
```diff
@@ -115,6 +120,12 @@ func Stage(dataDir string, name string, filemode os.FileMode) error {
 	if err := os.Chmod(p, 0550); err != nil {
 		return fmt.Errorf("failed to chmod %s: %w", name, err)
 	}
+
+	// It is expected that the staged embedded binary has the same modification
+	// time as the `k0s` executable.
+	if err := os.Chtimes(p, exinfo.ModTime(), exinfo.ModTime()); err != nil {
+		return fmt.Errorf("failed to chtimes %s: %w", name, err)
```
twz123, line 127:
```suggestion
		return fmt.Errorf("failed to set file modification time of %s: %w", p, err)
```

pkg/assets/stage.go
```diff
@@ -115,6 +120,12 @@ func Stage(dataDir string, name string, filemode os.FileMode) error {
 	if err := os.Chmod(p, 0550); err != nil {
 		return fmt.Errorf("failed to chmod %s: %w", name, err)
 	}
+
+	// It is expected that the staged embedded binary has the same modification
```
twz123, line 124: Maybe add a note that it's used for the update check.

twz123 (dismissed):

pkg/assets/stage.go
```diff
@@ -73,7 +71,19 @@ func Stage(dataDir string, name string, filemode os.FileMode) error {
 		return fmt.Errorf("failed to create dir %s: %w", filepath.Dir(p), err)
 	}
 
-	if ExecutableIsOlder(p) {
+	selfexe, err := os.Executable()
+	if err != nil {
+		logrus.Warn(err)
+		return err
```
soider, line 77: I would prefer to have here (and in other return err branches) fmt.Errorf, because it is often hard to find the exact codeline when we have some generic error message that's coming from the stdlibrary.

s0j: Thanks. I cleaned up all of the returns here and added some consistency to the names of things logged (now all full-paths vs name-only)

soider (approved):

twz123 (approved):

PR #1503, opened 2022-02-10
[release-1.22] Ensure staged embedded-binaries + `k0s` have matching timestamps

s0j:
This provides enough information at startup to determine if the embedded-bin
should be updated (sourced from a different `k0s` executable).

Previously, the current timestamp was used on staged embedded-bins which could
open a window of time where embedded-bins won't get updated.

(cherry picked from commit 7bbcf2cd349fa1ab32a478cc410658e185c7d766)

### DIFF
--- a/pkg/assets/stage.go
+++ b/pkg/assets/stage.go
@@ -27,21 +27,19 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// ExecutableIsOlder return true if currently running executable is older than given filepath
-func ExecutableIsOlder(filepath string) bool {
-	ex, err := os.Executable()
-	if err != nil {
-		return false
+// EmbeddedBinaryNeedsUpdate returns true if the provided embedded binary file should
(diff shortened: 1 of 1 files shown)


ncopa:
> The arm6l failure looks to be a disconnect between what `https://golang.org/VERSION?m=text` reports (go1.17.7), and what releases are available for download (current max is go1.17.6). I expect that this will sort itself out.

Makes me wonder if it is a good idea to use that URL for latest go version. Maybe we should git checkout the code first and then get the go version from `embedded-bins/Makefile.variables` with `make -s goversion`:
```diff
diff --git a/Makefile b/Makefile
index 2ede82c1..4c2321ce 100644
--- a/Makefile
+++ b/Makefile
@@ -215,3 +215,7 @@ image-bundle/bundle.tar: image-bundle/image.list
 .PHONY: docs
 docs:
        $(MAKE) -C docs
+
+.PHONY: goversion
+goversion:
+       @echo $(go_version)
```

PR #1512, opened 2022-02-13
Add support for external registry integration

### LINKED ISSUE #1424: support adding private registry through k0s configuration file

As a user I want to be able to pull images from my own hosted registry secured by auth.

To support it currently user needs to manually follow containerd documentation and add certificates.

https://github.com/containerd/containerd/blob/main/docs/cri/config.md#registry-configuration

Would be great if we introduce k0s config section related to the cri settings which will allow to easy setup certificates for private registry

Manual setup.

### Additional context

Request for corresponding guide appeared in the k0s slack channel couple of times.

martencassel:
## Issue
Fixes #1424  

## What this PR includes
This PR enables users to configure k0s and containerd to access a private registry that is secured by certificated based authentication. 
The k0s configuration file has been extended with an optional runtime section with a optional list of registry configurations for containerd.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added
 
## Checklist:


martencassel:
Adding this as a draft for now.
Where should i  populate ClusterConfig into the containerd component  ? 
@soider

jnummelin:
> Where in the code should i populate ClusterConfig so that the containerd component can read the configuration ?

hmm, that's the key problem here we need to figure out. Currently the worker nodes do NOT read that info at all, directly at least. What comes closest is the way we "push" custom configs for kubelets via [`workerProfiles`](https://docs.k0sproject.io/v1.23.3+k0s.0/configuration/#specworkerprofiles). Essentially what that does is it:
- on the controllers it breaks each profile into a `ConfigMap` with names like `kubelet-config-<profile_name>-1.23`
- when worker starts, it looks for the correct configmap based on the given profile `k0s worker --profile foobar`
- when the correct CM is found, it drops the data into disk and makes kubelet to use that config file

I'm thinking if we could/should use something similar for this use case too?

Some open questions for discussion:
- If we're essentially configuring `containerd` here, should we allow users to push full containerd configs rather than only registries?
- The "vehicle" for storing this stuff should probably be `Secret` as these might contain actual secrets? (like client certs etc.)
- Does allowing the node (with role `system:node`) to read secrets in `kube-system` open too many doors?
    - Or maybe more in general, how do we let worker nodes pull this information?

martencassel:
I don't see the point in this PR atm. 
Wouldn't it be easier just to push containerd configuration file to hosts using scp or ansible or manually ? 
What i know k3s has the feature to push a go template of containerd file to hosts. and render it.... 
Perhaps this is a solution to explore ?

jnummelin:
> I don't see the point in this PR atm.

Fair enough. I think this is like 50/50 whether this is really a k0s problem to solve or not. If you're not planning to work on this further, maybe just close the PR? GH still remembers it, so if we need to revisit this topic we'll find it still.

> As a user, it be easier just to push containerd configuration file to hosts using scp or ansible or manually ?

That's always an option, for sure. k0sctl also supports [pushing out config etc. files](https://github.com/k0sproject/k0sctl/#spechostsfiles-sequence-optional) to hosts during the cluster bootstrap

> Another solution is to offer a blog post about this setup.

That, or maybe short section in docs how to to this.

> Do people care about self-hosting docker registry ?

I think many organizations do it, but they usually have it properly setup with TLS

> the little i know of k3s is that they have a method to push and render containerd toml files to hosts

hmm, the rendering part I know but I didn't find any "centralized" way to push containerd configs to hosts.

jnummelin:
@martencassel if you ever decide to restart working on this, pls reopen the PR

PR #1514, opened 2022-02-14
[release-1.22] Add explicit setup-go action for lint GHA

s0j:
Linting is sensitive to the version of Go being used, so be explicit here.

(cherry picked from commit 6e891a4669b16ef0d503a51fbbce55ad2a68e702)


s0j:
I don't understand why `go1.17` is being used when I asked for `go1.16`.  Turning to draft for now.

https://github.com/k0sproject/k0s/runs/5177425768?check_suite_focus=true
```
Run actions/setup-go@v2
Setup go stable version spec ^1.16
Found in cache @ /opt/hostedtoolcache/go/1.17.6/x64
Added go to the path
Successfully setup go version ^1.16
go version go1.17.6 linux/amd64
```

twz123:
I think it's the caret, which allows higher minor versions?

s0j:
> I think it's the caret, which allows higher minor versions?

Yep, thats it, thanks. We want a tilde instead, otherwise we're going to be picking up all minor version bumps.

twz123 (approved):

PR #1515, opened 2022-02-14
Controller components metrics scraper

makhov:

## Description

Added a component for scraping metrics from controller components (kube-scheduler, kube-controller-manager) that are not accessible from the cluster. The new component deploys a dedicated [pushgateway with TTL](https://github.com/Eidison186/prometheus-pushgateway-with-ttl), scrapes metrics from the system components and pushes them to the gateway. TTL is used to make it possible to detect outages (missing Prometheus jobs).

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/metrics.go
+++ b/pkg/component/controller/metrics.go
@@ -1,0 +1,359 @@
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package controller
+
+import (
+	"bytes"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"k8s.io/client-go/rest"
+	"net/http"
+	"os"
+	"path"
+	"time"
+
+	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
+	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+	"github.com/k0sproject/k0s/pkg/component"
+	"github.com/k0sproject/k0s/pkg/constant"
+	"github.com/k0sproject/k0s/pkg/kubernetes"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	namespace       = "k0s-system"
+	pushGatewayName = "k0s-pushgateway"
+)
+
+// Metrics is the reconciler implementation for metrics server
+type Metrics struct {
+	hostname      string
+	log           *logrus.Entry
+	clusterConfig *v1beta1.ClusterConfig
+	tickerDone    context.CancelFunc
+	K0sVars       constant.CfgVars
+	jobs          []*job
+	saver         manifestsSaver
+	restClient    rest.Interface
+}
+
+var _ component.Component = &Metrics{}
+var _ component.ReconcilerComponent = &Metrics{}
+
+// NewMetrics creates new Metrics reconciler
+func NewMetrics(k0sVars constant.CfgVars, saver manifestsSaver, clientCF kubernetes.ClientFactoryInterface) (*Metrics, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return nil, err
+	}
+	log := logrus.WithFields(logrus.Fields{"component": "Metrics"})
+
+	restClient, err := clientCF.GetRESTClient()
+	if err != nil {
+		return nil, fmt.Errorf("error getting REST client for metrics: %w", err)
+	}
+
+	return &Metrics{
+		hostname:   hostname,
+		log:        log,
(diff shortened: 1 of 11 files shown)


pkg/component/controller/metrics.go
```diff
@@ -0,0 +1,359 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    component: "pushgateway"
+    app: k0s-observability
+  name: {{ .Name }}
+rules:
+  []
```
jnummelin, line 268: Why empty set of rules?

pkg/component/controller/metrics.go
```diff
@@ -0,0 +1,359 @@
...
+	if err != nil {
+		return err
+	}
+	m.jobs = append(m.jobs, j)
+
+	j, err = m.newJob("kube-controller-manager", "https://localhost:10257/metrics")
+	if err != nil {
+		return err
+	}
+	m.jobs = append(m.jobs, j)
```
jnummelin, line 110: I think we should also have a job for api-server. Or should we rely that users prom will pull the metrics via `kubernetes` services endpoints?

pkg/component/controller/metrics.go
```diff
@@ -0,0 +1,359 @@
...
+      serviceAccountName: {{ .Name }}
+      tolerations:
+        - key: "CriticalAddonsOnly"
+          operator: "Exists"
+        - key: "node-role.kubernetes.io/master"
+          operator: "Exists"
+          effect: "NoSchedule"
+      containers:
+        - name: prometheus-pushgateway
+          image: "dmathai/prom-pushgateway-ttl:1.4.0-ttl"
```
jnummelin, line 336: Should we handle this image as we do others (have it configurable)? And should we include this into airgap bundle too?

One downside on this is that is only available for amd64:
```
$ manifest-tool inspect dmathai/prom-pushgateway-ttl:1.4.0-ttl
dmathai/prom-pushgateway-ttl:1.4.0-ttl: manifest type: application/vnd.docker.distribution.manifest.v2+json
      Digest: sha256:a815e2c8aac8336beaf32f854009b33fe54d4fe2dc7fb30b2701a72c35b9a431
Architecture: amd64
          OS: linux
    # Layers: 4
      layer 1: digest = sha256:e5d9363303ddee1686b203170d78283404e46a742d4c62ac251aae5acbda8df8
      layer 2: digest = sha256:3430c2c42129e987db5d5c5434ea73f0cbb44b931719cb2072203fcc05617d54
      layer 3: digest = sha256:a09edd3728450ef9d785d66237dfa943554b8f518d025fd463094d165c3c8774
      layer 4: digest = sha256:547263457d796dc5c03b1846252566f39426aff85d3c5caed3405f5800bc64a4
```

pkg/component/controller/metrics.go
```diff
@@ -0,0 +1,359 @@
...
+		log:          m.log,
+		scrapeClient: httpClient,
+		restClient:   m.restClient,
+	}, nil
+}
+
+func (j *job) Run(ctx context.Context) {
+	j.log.Debugf("Running %s job", j.name)
+
+	t := time.NewTicker(time.Second * 30)
```
jnummelin, line 181: Is 30sec bit too often?

jnummelin (commented):
Mostly LGTM. Added some questions for clarifications.

pkg/component/controller/metrics.go
makhov: Yeah, I just copied a manifest generated by helm. 
I think pushgateway SA should use our default PodSecurityPolicy. 
Maybe we can remove this separate ClusterRole and just assign the `k0s:podsecuritypolicy:default` ClusterRole to the SA. WDYT?

pkg/component/controller/metrics.go
makhov: It depends on how we see this feature. If it just exposes cluster components' metrics that can't be scraped in another way, I don't think we need to add the api-server, since end-users can do it themselves.

pkg/component/controller/metrics.go
makhov: Yeah, fair enough. We also need to push the image to our own registry, I think.

pkg/component/controller/metrics.go
makhov: Yeah, maybe. I put 30s here because it's the default prometheus scrape_interval. What value are you suggesting?

pkg/component/controller/metrics.go
jnummelin: Fair enough

pkg/component/controller/metrics.go
jnummelin: Why do we need SA for it in the first place?

makhov: Yeah, we can just remove a separate SA usage.

makhov: Removed it

pkg/component/controller/metrics.go
makhov: Make it configurable and it uses our image by default now

pkg/component/controller/metrics.go
jnummelin: oh, somehow I remembered prom scrape default was 60s. I think 30s is fine

jnummelin (approved):
LGTM in general. I still think we'd benefit from having some sort of architectural description for this in the docs but can be done in separate PR too.

PR #1526, opened 2022-02-15
Add pre-flight checks to subcommands that run workloads

### LINKED ISSUE #1289: Pre-flight check for system level config
Every now and then we face issues where k0s and kube bits do not boot up properly. Often the case is about missing system requirements, mainly kernel modules, cgroup or other stuff like that.

The main k0s process could do an early "pre-flight" check and exit early with clear error what is missing. We already have a `k0s sysinfo` command which can dig the info out. Of course some flag like `--skip-system-check` would be in order to make it skip the check.

twz123:
## Description

Based on the sysinfo subcommand. Users can choose to ignore the results of the pre-flight checks by adding the `--ignore-pre-flight-checks` command line argument.

Fixes #1289.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


cmd/sysinfo/sysinfo.go
```diff
@@ -16,107 +16,17 @@ limitations under the License.
...
-	system "k8s.io/system-validators/validators"
 )
 
 func NewSysinfoCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "sysinfo",
 		Short: "Display system information",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runSysinfo()
+			sysinfo.ReportSysinfo()
```
jnummelin, line 28: IMO we could handle the error here too by logging it. In case the command fails to read kernel mod info, the output will be only partial and user does not know if everything is ok or not.

twz123: This needs some consideration. The system-validators library will return unfulfilled non-optional requirements both via the `Reporter` interface (marked as "bad") and as error, whereas "real" errors are solely reported as errors. As far as I can tell, there's no sustainable way to distinguishing those error types. When logging the errors here, we'd have all the failed requirements doubled.

That's probably the reason as to why the previous implementation just disregarded any errors (returning `nil` in any case).

inttest/common/footloosesuite.go
```diff
@@ -837,24 +838,55 @@ func (s *FootlooseSuite) initializeFootlooseClusterInDir(dir string) error {
...
+	// See https://github.com/kubernetes/system-validators/blob/v1.6.0/validators/kernel_validator.go#L180-L190
+
+	bindPaths := []string{
+		"/usr/src/linux/.config",
+		"/usr/lib/modules",
+		"/lib/modules",
+	}
+
+	if kernelVersion, err := (&system.KernelValidatorHelperImpl{}).GetKernelReleaseVersion(); err == nil {
+		bindPaths = append(bindPaths, []string{
```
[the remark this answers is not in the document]
twz123: `/proc` is treated specifically by Docker as far as I can tell. If `/proc/config.gz` is present on the host, it will be present inside the container, as well. At least that's the behavior on my box.

pkg/sysinfo/validation_linux.go
```diff
@@ -0,0 +1,69 @@
...
+	case unix.CGROUP2_SUPER_MAGIC:
+		return cgroupV2, nil
+	case unix.TMPFS_MAGIC:
+		return cgroupV1, nil
+	default:
+		return cgroupVersionUnknown, fmt.Errorf("unexpected file system type of %q: 0x%x", cgroupMountpoint, st.Type)
+	}
+}
+
+func (s *K0sSpec) validateCgroupVersion(reporter system.Reporter) error {
```
ncopa, line 48: I wonder if we could use https://github.com/kubernetes/system-validators/blob/master/validators/cgroup_validator_linux.go here?

twz123: Do you have something specific in mind? I didn't find a good trick to actually get the info whether it's v1 or v2 out of it, since the validator doesn't really care and kinda normalizes the differences away.

inttest/cli/cli_test.go
```diff
@@ -54,70 +56,96 @@ func (s *CliSuite) TestK0sCliCommandNegative() {
...
 
-	s.T().Log("running k0s install command")
-	_, err = ssh.ExecWithOutput("k0s install controller --enable-worker --disable-components konnectivity-server,metrics-server")
-	s.Require().NoError(err)
+	s.T().Run("preflightChecks", func(t *testing.T) {
+		out, err := ssh.ExecWithOutput(fmt.Sprintf(`
+			set -eu
+			fakeDir="$(mktemp -d)"
+			trap "rm -rf -- '$fakeDir'" INT EXIT
+			printf '#!/bin/sh\necho spoof' >"$fakeDir/uname"
```
ncopa: If this works it means that we depend on an external `uname` binary instead of using the [uname(2) syscall](https://man7.org/linux/man-pages/man2/uname.2.html), and I don't think we want depend on any external binaries.

Yup, k8s/system-validators execs external `uname`:
https://github.com/kubernetes/system-validators/blob/73d9d3d9f43503b1a64ec0506080252c683f8408/validators/os_validator.go#L40

inttest/common/footloosesuite.go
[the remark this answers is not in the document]
ncopa: ah... of course...

pkg/sysinfo/validation_linux.go
ncopa: that's fine. I was just wondering if it was considered.

twz123:
Turning this into a draft again, since there's more to be considered here.

twz123:
Will open up another PR that'tll supersede this one.

PR #1527, opened 2022-02-15
Sync release-1.23 branch with main for release 1.23.3+k0s.1

ncopa:
Merge main branch into release-1.23 so they are identical for release 1.23.3+k0s.1


ncopa:
I ended up pushing it directly to the release-1.23 branch to avoid the mergeable check

PR #1554, opened 2022-02-22
Restructure GitHub Actions for Golang

twz123:
## Description

* Remove GO_VERSION and GO_VERSION_WIN from build  
  Rely on the go version declared in the Makefiles instead. Use the shared script `.github/workflows/prepare-build-env.sh` to setup the environment for the different jobs in a unified way.

* Unify linux and windows builds into a single matrix job
  Split off the unit tests into their own job that's triggered after the matrix build. They're only running for Linux based builds, which saves conditions and also allows for the other dependent jobs to start earlier.

* Unify the smoke tests for ARM into a single matrix job
  This is more DRY, and saves some boilerplate. The differences in the jobs are handled via conditions on the steps and can possibly be unified in the future, once the setup-go action gets proper support for ARMv7.

* Rework the caching behavior
  Use regular directories for go caches and bind mount them into the Docker based builds. This replaces the volume based approach, which was not easily cacheable across builds. Use separate caches for GOCACHE and GOMODCACHE based on their usage pattern.

* Use upload/download artifacts to share the built binaries with downstream jobs
  Those artifacts are unique to each workflow run, which makes them better suited than caching, which is also shared across builds. As a plus, the binaries are downloadable by humans afterwards.

* Cache the airgap image bundle
  The image bundle is only dependent of the image list provides by k0s. Exploit this to add a special cache for the tar file, which is quite stable across builds. This speeds up the airgap smoke test significantly.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


s0j (dismissed):
LGTM, but needs more eyes.

makhov (approved):
LGTM. Since all builds pass, even if there are any issues, they are minor and we could fix them later.

jnummelin (approved):

s0j (approved):

PR #1556, opened 2022-02-22
docs: add file-structure

### LINKED ISSUE #951: Document k0s directories
Documentation should include a list of directories and their purposes, what goes in there and expected permissions etc.

- data-dir (/var/lib/k0s)
- run-dir (/var/run/k0s)
- bin-dir (/var/lib/k0s/bin?)

wenerme:
## Description

Fixes #951

describe the directory and file used by k0s

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/file-structure.md
+++ b/docs/file-structure.md
@@ -1,0 +1,76 @@
+# File Structure
+
+- /etc/k0s/k0s.yaml - default k0s.yaml
+- /etc/k0s/containerd.toml - bundled containerd config
+  - generate default by `/var/lib/k0s/bin/containerd config default > containerd.toml` 
+  - [containerd-config.toml.5](https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md)
+- /var/log/pods/ - logs of pod
+- /var/lib/cni/ - cni working directory
+- /var/openebs/ - OpenEBS default local pv path
+- /etc/machine-id - identity of the node, content is random 32 char - **REQUIRED**
+  - you can generate by `uuidgen | tr -d - > /etc/machine-id`
+- /var/log/k0s.log
+- /var/log/k0s.err - error log
+
+## Data directory
+
+- /var/lib/k0s/ - data directory - defined by `--data-dir`
+  - kubelet.conf
+  - kubelet-config.yaml
+  - kubelet-bootstrap.conf
+    - kubelet bootstrap config
+    - include worker token
+  - konnectivity.conf
+    - EgressSelectorConfiguration
+    - reference /run/k0s/konnectivity-server/konnectivity-server.sock
+- /var/lib/k0s/bin/ - extracted when starting k0s
+  - xtables-legacy-multi - bundled iptable
+    - target of symbol links: ip6tables, ip6tables-restore, ip6tables-save, iptables-restore, iptables-save
+  - containerd, containerd-shim, containerd-shim-runc-v1, containerd-shim-runc-v2
+  - runc
+  - etcd
+  - kine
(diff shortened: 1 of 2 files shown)


jnummelin:
@wenerme linter is not happy, check the `Files Changed` tab for its complaints.

kke:
Thinking if it would be feasible to change all file/path references in k0s code to some kind of `NamedPath` objects that could be used to generate docs (or code would be generated from the doc) 🤔

wenerme:
@kke windows support is experimental, may change later, and even use windows, there must have a controller+worker run in linux

kke:
> windows support is experimental, may change later, and even use windows, there must have a controller+worker run in linux

Yes, but we do have several `FooFooDefaultWindows` paths in the constants. This could be something like `namedpath.Get("rundir").Default()` and it would return the correct one depending on `GOOS`.

> A mapping of path to explain is good idea, but not easy to maintain

Could be easier than maintaining separately in docs.

wenerme:
I haven't deploy k0s on windows before, I can update the doc after I have some experience about windows k0s deployment.

trawler:
I am not convinced that a full listing of all files used by k0s, add meaningful information to the docs. 
Many of the listed files are auto-generated, and manipulating them will likely break k0s functionality if modified directly.
A concise description of the directory structure, directory contents and its uses and important configuration file paths would be more useful, in my opinion.

soider:
>I am not convinced that a full listing of all files used by k0s, add meaningful information to the docs.

I'd love to agree. The thing with listing all the files in the docs is that:
- it is now a contract to maintain that port of the documentation
- a lot of those files are actually used internally

trawler:
@wenerme: I think that such a detailed file and directory list defeats the purpose, and doesn't add as much value as expected. 
It includes files and directories that are auto-generated and may confuse users into trying to touch/edit the files, and this PR doesn't contain an explanation on the source and purpose of these files.
I would instead love to see the info in this PR added to https://docs.k0sproject.io/latest/architecture/
as a new segment: File Structure, with the general information of each directory, its' uses and the main configuration files (user configuration files, not system configuration files).
For example, list /var/lib/k0s: the main data directory for k0s and all it’s components, customizable with --data-dir flag and /var/lib/k0s/pki: CAs, keys and certificates managed by k0s for all kubernetes components, etc.
For other files: if it's necessary to list files like kubelet.conf / kubelet-config.yaml / kubelet-bootstrap.conf, etc, it's worth mentioning that these files should not be edited by hand, but instead are generated by k0s.

trawler (changes requested):
@wenerme: I think that such a detailed file and directory list defeats the purpose, and doesn't add as much value as expected.
It includes files and directories that are auto-generated and may confuse users into trying to touch/edit the files, and this PR doesn't contain an explanation on the source and purpose of these files.
I would instead love to see the info in this PR added to https://docs.k0sproject.io/latest/architecture/
as a new segment: File Structure, with the general information of each directory, its' uses and the main configuration files (user configuration files, not system configuration files).
For example, list /var/lib/k0s: the main data directory for k0s and all it’s components, customizable with --data-dir flag and /var/lib/k0s/pki: CAs, keys and certificates managed by k0s for all kubernetes components, etc.
For other files: if it's necessary to list files like kubelet.conf / kubelet-config.yaml / kubelet-bootstrap.conf, etc, it's worth mentioning that these files should not be edited by hand, but instead are generated by k0s.

wenerme:
@trawler Ok, I will split the content, only keep the k0s related file.

till:
Would be great to have a list of directories that I **can** adjust.

PR #1579, opened 2022-03-11
Update docs version management logic

kke:

## Description

- Uses https://github.com/k0sproject/version for sorting
- Every time a release is made or something is merged to main, the latest and stable version aliases will be updated.
- The "stable" will always be set as default.
- Files `latest.txt` and `stable.txt` are updated to include the plaintext version strings, the files are committed and pushed to the gh-pages branch.

## Type of change

- [X] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -22,19 +22,6 @@ jobs:
         with:
           python-version: ${{ env.PYTHON_VERSION }}
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install mkdocs
-          pip install git+https://${{ secrets.GH_TOKEN }}@github.com/lensapp/mkdocs-material-insiders.git
-          pip install mdx-truly-sane-lists
-          pip install mike
-
-      - name: Checkout k0s release
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
       - name: Set up Go ${{ env.GO_VERSION }}
         uses: actions/setup-go@v2
         with:
@@ -49,6 +36,20 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mkdocs
+          pip install git+https://${{ secrets.GH_TOKEN }}@github.com/lensapp/mkdocs-material-insiders.git
+          pip install mdx-truly-sane-lists
+          pip install mike
+          go install github.com/k0sproject/version/cmd/k0s_sort@v0.2.2
+
+      - name: checkout k0s release
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
       - name: Generate docs
         run: make -C docs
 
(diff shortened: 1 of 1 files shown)


kke:
We do setup go in the workflow anyway, could just use https://github.com/k0sproject/version to sort, just make a tiny cmd/main.go there.

kke:
I'm not 100% sure if `gh release list` will show the current release as draft or has it turned to published already at this point 🤔

.github/workflows/publish-docs.yml
```diff
@@ -49,6 +36,20 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
 
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mkdocs
+          pip install git+https://${{ secrets.gh_token }}@github.com/lensapp/mkdocs-material-insiders.git
```
twz123, line 43: Are you sure that `gh_token` can be lower case here?

kke: It has been like that since forever, but I'll change it anyway.

kke: Whoops, looks like I somehow managed to downcase it.

.github/workflows/publish-docs.yml
```diff
@@ -68,13 +69,37 @@ jobs:
...
+        run: |
+          LATEST=$(echo ${{ steps.tags.outputs.SORTED_TAGS }} | tail -1)
+          echo "${LATEST}" > site/latest
+          mike alias -u "${LATEST}" latest
+
+      - name: Set latest stable
+        id: latest_stable
+        run: |
+          STABLE=$(echo ${{ steps.tags.outputs.SORTED_TAGS }} | grep -v -- "-" | tail -1)
+          echo "${STABLE}" > site/stable
```
kke, line 93: These wont work, it would need some kind of deploy / publish / whatever

.github/workflows/publish-docs.yml
```diff
@@ -68,13 +69,37 @@ jobs:
...
       - name: Get the release version
-        if: contains(github.ref, 'refs/tags/v') && !github.event.release.prerelease # (don't generate for pre-releases)
+        if: contains(github.ref, 'refs/tags/v')
         id: get_version
         run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
 
-      - name: mike deploy new release
-        if: contains(github.ref, 'refs/tags/v') && !github.event.release.prerelease # (don't generate for pre-releases)
+      - name: Collect tags
+        id: tags
```
twz123, line 77: The GitHub token needs to be passed explicitly.

```suggestion
        id: tags
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/publish-docs.yml
```diff
@@ -62,19 +63,56 @@ jobs:
...
+      # Lists tags using the "gh" cli
+      # Gets rid of old versions (ones without +k0s) and drafts using grep
+      # The list is sorted using k0s_sort installed "install dependencies"
+      - name: Collect tags
+        id: tags
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAGS=$(gh release list -L 1000 -R k0sproject/k0s | grep "+k0s." | grep -v Draft | cut -f 1 | k0s_sort)
+          echo ::set-output name=SORTED_TAGS::${TAGS}
```
twz123, line 83: I think there will be line break issues when trying to pass on the tags without qutes. What about extracting latest and stable directly here, and use them later on via an `env:` block?

Moreover, `k0s_sort` will strip the leading "v", so we could add it back here, as well.

```suggestion
          LATEST=$(echo "$TAGS" | tail -1)
          STABLE=$(echo "$TAGS" | grep -v -- "-" | tail -1)
          echo ::set-output name=LATEST::"v${LATEST}"
          echo ::set-output name=STABLE::"v${STABLE}"
```

.github/workflows/publish-docs.yml
```diff
@@ -62,19 +63,56 @@ jobs:
...
+      # Ensures the current branch is gh-pages,
+      # Creates / updates the "stable" and "latest" plain text files with the corresponding versions
+      # Commits if the files were changed
+      # Finally pushes if there are unpushed commits
+      - name: Create version files
+        run: |
+          LATEST=${{ steps.set_versions.outputs.LATEST }}
+          STABLE=${{ steps.set_versions.outputs.STABLE }}
+          (git branch --show-current | grep -q gh-pages) || git checkout gh-pages
+          echo "${STABLE}" > stable
```
twz123, line 112: This will collide with the latest and stable alias directories created by mike in the previous step. What about adding a ".version" suffix to the files, or sth. like that?

.github/workflows/publish-docs.yml
```diff
@@ -62,19 +63,56 @@ jobs:
...
+
+      # Ensures the current branch is gh-pages,
+      # Creates / updates the "stable" and "latest" plain text files with the corresponding versions
+      # Commits if the files were changed
+      # Finally pushes if there are unpushed commits
+      - name: Create version files
+        run: |
+          LATEST=${{ steps.set_versions.outputs.LATEST }}
+          STABLE=${{ steps.set_versions.outputs.STABLE }}
+          (git branch --show-current | grep -q gh-pages) || git checkout gh-pages
```
twz123, line 111: I think we can unconditionally checkout gh-pages here.

.github/workflows/publish-docs.yml
```diff
@@ -62,19 +63,56 @@ jobs:
...
+          echo "${STABLE}" > stable
+          git add stable && git update-index --refresh
+          git diff-index --quiet HEAD -- || git commit -m "Set stable to ${STABLE}"
+          echo "${LATEST}" > latest
+          git add latest && git update-index --refresh
+          git diff-index --quiet HEAD -- || git commit -m "Set latest to ${LATEST}"
+          (git log origin/gh-pages..gh-pages | grep -q HEAD) && git push origin gh-pages
```
twz123, line 118: Also here, I think we can just push, without any conditions, since the commits are only happening if the files changed. When we push without any new commits, git won't do anything.

twz123 (changes requested):

.github/workflows/publish-docs.yml
kke: Hmm, k0s_sort needs to not strip it, I'll fix that there.

.github/workflows/publish-docs.yml
kke: Good catch, .txt is probably fine.

.github/workflows/publish-docs.yml
kke: Combined the steps so the list doesn't need to go to an output

twz123 (approved):

kke:
Let's see what happens on merge :)

kke:
Still no version selector :(

PR #1580, opened 2022-03-11
Makefile tweaks

twz123:
## Description

* Fix `BUILD_DATE` for GNU date  
  Whenever `SOURCE_DATE_EPOCH` was set, `BUILD_DATE` would be empty when using GNU date, failing with an "extra operand" error. Fix that by using the approach described at reproducible-builds.org. It will work with either GNU or BSD date, and fallback to ignore `SOURCE_DATE_EPOCH` if both fails. If `SOURCE_DATE_EPOCH` is not specified, try to use the commit date of the current git commit. If that fails, too, fallback to the current date.

* Exclude ./build/cache/ from `GO_SRCS`  
  `GOMODCACHE` is placed there. If all the Go files from there are included as dependencies, `make build` will take really long to figure out if the binary needs to be rebuilt.

* Introduce `GO_DIRS` to enumerate all the directories to be scanned by `go list`. Same reason here: if `go list` descends into `GOMODCACHE`, this will cause trouble.

* Add a `go mod tidy` target for `go.sum` to automatically update it whenever `go.mod` changed.

* Add a dependency to `go.sum` to all the targets that compile Go code, so, so that those get built whenever go.mod/go.sum have changed.

Fixes #1554.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 include embedded-bins/Makefile.variables
 include inttest/Makefile.variables
 
-GO_SRCS := $(shell find . -type f -name '*.go' -a ! -name 'zz_generated*')
+GO_SRCS := $(shell find . -type f -name '*.go' -not -path './build/cache/*' -not -name 'zz_generated*')
+GO_DIRS := . ./cmd/... ./pkg/... ./internal/... ./static/... ./hack/...
 
 # EMBEDDED_BINS_BUILDMODE can be either:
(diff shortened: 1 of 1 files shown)


kke (approved):
LGTM

jnummelin (approved):

PR #1581, opened 2022-03-11
openrc: increase default ulimit

wenerme:
## Description

```bash
 sudo head -1000 /var/log/k0s.log | grep 'too many open files'| wc -l
```

```
252
```

fix konnectivity-server.sock: socket: too many open files

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/install/linux_openrc.go
+++ b/pkg/install/linux_openrc.go
@@ -28,6 +28,8 @@ command_args="{{range .Arguments}}{{.}} {{end}}"
 name=$(basename $(readlink -f $command))
 supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"
 
+: "${rc_ulimit=-n 1048576 -u unlimited}"
+
 {{- if .Dependencies }}
 depend() {
 {{- range $i, $dep := .Dependencies}} 


wenerme:
@twz123 alpine is default to 1024/4096, this ulimit is copy from k3s https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/k3s/k3s.initd#L20

jnummelin:
@ncopa as Alpine wizard, PTAL 😄

pkg/install/linux_openrc.go
```diff
@@ -28,6 +28,8 @@ command_args="{{range .Arguments}}{{.}} {{end}}"
 name=$(basename $(readlink -f $command))
 supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"
 
+rc_ulimit="${K0S_ULIMIT:--c unlimited -n 1048576 -u unlimited}"
```
ncopa, line 31:
```suggestion
: "${rc_ulimit=-c unlimited -n 1048576 -u unlimited}"
```

This way user can override the setting in `/etc/conf.d/k0sworker` using `rc_ulimit`. It is also backwards compatible for users who already may have set `rc_ulimit` in their `/etc/conf.d/k0sworker`

ncopa (changes requested):
would be nice to be backwards compatible for users who may have already set `rc_ulimit` in their conf.d file.

pkg/install/linux_openrc.go
ncopa: actually, I don't think we should set `-c unlimited` by default for production.

wenerme: core file size default to zero, but leave -c untouched is ok for me.

ncopa (approved):

PR #1588, opened 2022-03-16
Fix docs traefik example link

kke:

## Description

Link pointed to `/main`, which got renamed to `/head`, but `/stable` is probably better here.

## Type of change

- [X] Documentation update

### DIFF
--- a/docs/examples/nginx-ingress.md
+++ b/docs/examples/nginx-ingress.md
@@ -49,7 +49,7 @@ Installing NGINX using NodePort is the most simple example for Ingress Controlle
     NAME    CONTROLLER             PARAMETERS   AGE
     nginx   k8s.io/ingress-nginx   <none>       162m
     ```
-  
+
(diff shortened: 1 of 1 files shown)


docs/examples/nginx-ingress.md
```diff
@@ -239,4 +239,4 @@ For more information about NGINX Ingress Controller installation, take a look at
 
 ## Alternative examples for Ingress Controllers on k0s
 
-[Traefik Ingress](https://docs.k0sproject.io/main/examples/traefik-ingress/)
+[Traefik Ingress](https://docs.k0sproject.io/stable/examples/traefik-ingress/)
```
jnummelin, line 242: I'd rather make this "relative" and not pin it to specific version of docs

kke: I think that will require using `../traefik-ingress`

kke: https://docs.k0sproject.io/v1.23.3+k0s.1/examples/nginx-ingress/../traefik-ingress seems to take the browser to the right location.

jnummelin (approved):

PR #1591, opened 2022-03-16
Use single node mode flag to decide if we want to run leader-election

### LINKED ISSUE #1585: Leader election disabled even in cases where external LB is not 100% required

### Version

1.23.3+k0s.1

### Platform

```shell
All affected
```

### What happened?

If I run multiple controllers (even with `--enable-worker`) and have NOT set `spec.api.externalAddress` in the config all kube controlplane components gets configured with `leader-elect=false`. This is because k0s deternimes that one cannot/shouldn't run HA controlplane without `spec.api.externalAddress`. But especially in cases when running controllers with `--enable.worker` one does not always need LB thus we should disable leader-election ONLY when running controller with `--single`.

### Steps to reproduce

1.
2.
3.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Screenshots and logs

_No response_

### Additional context

https://k8slens.slack.com/archives/C01FPC0N07P/p1647254187048809

soider:
Use single node mode flag to decide if we want to run leader-election

## Description

This PR introduces changes to disable leader-election only if we explicitly run cluster in the single node mode
The previously used behavior used to rely on the config ExternalAddress.

Fixes #1585

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [X] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -188,7 +188,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 	})
 
 	if c.NodeConfig.Spec.API.ExternalAddress != "" {
-		c.NodeComponents.Add(ctx, &controller.K0sLease{
+		c.NodeComponents.Add(ctx, &controller.K0sControllersLeaseCounter{
 			ClusterConfig:     c.NodeConfig,
 			KubeClientFactory: adminClientFactory,
 		})
@@ -197,7 +197,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 	var leaderElector controller.LeaderElector
 
 	// One leader elector per controller
-	if c.NodeConfig.Spec.API.ExternalAddress != "" {
+	if !c.SingleNode {
 		leaderElector = controller.NewLeaderElector(adminClientFactory)
 	} else {
 		leaderElector = &controller.DummyLeaderElector{Leader: true}
@@ -309,14 +309,16 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 	}
 	if !stringslice.Contains(c.DisableComponents, constant.KubeSchedulerComponentName) {
 		c.ClusterComponents.Add(ctx, &controller.Scheduler{
-			LogLevel: c.Logging[constant.KubeSchedulerComponentName],
-			K0sVars:  c.K0sVars,
+			LogLevel:   c.Logging[constant.KubeSchedulerComponentName],
+			K0sVars:    c.K0sVars,
+			SingleNode: c.SingleNode,
 		})
 	}
 	if !stringslice.Contains(c.DisableComponents, constant.KubeControllerManagerComponentName) {
 		c.ClusterComponents.Add(ctx, &controller.Manager{
-			LogLevel: c.Logging[constant.KubeControllerManagerComponentName],
-			K0sVars:  c.K0sVars,
+			LogLevel:   c.Logging[constant.KubeControllerManagerComponentName],
+			K0sVars:    c.K0sVars,
+			SingleNode: c.SingleNode,
 		})
 
 	}
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -187,7 +187,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 		EnableKonnectivity: enableKonnectivity,
 	})
 
-	if c.NodeConfig.Spec.API.ExternalAddress != "" {
+	if !c.SingleNode {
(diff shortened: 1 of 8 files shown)


cmd/controller/controller.go
```diff
@@ -188,7 +188,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 	})
 
 	if c.NodeConfig.Spec.API.ExternalAddress != "" {
```
jnummelin, line 190: I think this should be also guarded by `if singleNode` check?

jnummelin (changes requested):
There's also the k0s internal leader election which is currently guarded by checking the existence of `externalAddress`. See https://github.com/k0sproject/k0s/blob/main/cmd/controller/controller.go#L199-L204

cmd/controller/controller.go
jnummelin: @soider you marked this resolved but it's still using `externalAddress` check? Did I miss something? :)

ncopa (approved):

cmd/controller/controller.go
ncopa, line 190:
```suggestion
	if !c.SingleNode {
```

ncopa:
Maybe we also need replace it in `konnectivity.go`?
```diff
diff --git a/pkg/component/controller/konnectivity.go b/pkg/component/controller/konnectivity.go
index 851bf8e4..200044e4 100644
--- a/pkg/component/controller/konnectivity.go
+++ b/pkg/component/controller/konnectivity.go
@@ -100,7 +100,7 @@ func (k *Konnectivity) Run(ctx context.Context) error {
 // Reconcile detects changes in configuration and applies them to the component
 func (k *Konnectivity) Reconcile(ctx context.Context, clusterCfg *v1beta1.ClusterConfig) error {
        k.clusterConfig = clusterCfg
-       if k.NodeConfig.Spec.API.ExternalAddress != "" {
+       if k.SingleNode {
                go k.runLeaseCounter(ctx)
        } else {
                // It's a buffered channel so once we start the runServer routine it'll pick this up and just sees it never changing
```

ncopa:
I think this also:
```diff
diff --git a/pkg/component/controller/konnectivity.go b/pkg/component/controller/konnectivity.go
index 851bf8e4..200044e4 100644
--- a/pkg/component/controller/konnectivity.go
+++ b/pkg/component/controller/konnectivity.go
@@ -100,7 +100,7 @@ func (k *Konnectivity) Run(ctx context.Context) error {
 // Reconcile detects changes in configuration and applies them to the component
 func (k *Konnectivity) Reconcile(ctx context.Context, clusterCfg *v1beta1.ClusterConfig) error {
        k.clusterConfig = clusterCfg
-       if k.NodeConfig.Spec.API.ExternalAddress != "" {
+       if k.SingleNode {
                go k.runLeaseCounter(ctx)
        } else {
                // It's a buffered channel so once we start the runServer routine it'll pick this up and just sees it never changing
```

cmd/controller/controller.go
```diff
@@ -222,7 +222,7 @@ func (c *CmdOpts) startController(ctx context.Context) error {
 			adminClientFactory))
 	}
 
-	if c.NodeConfig.Spec.API.ExternalAddress != "" && !c.NodeConfig.Spec.API.TunneledNetworkingMode {
+	if !c.SingleNode && !c.NodeConfig.Spec.API.TunneledNetworkingMode {
```
jnummelin, line 225: hmm, I think this on was correct before. The k0s custom `EndpointReconciler` creates the `Endpoints` for kube api svc based on `externalAddress` via DNS resolution. Thus I don't think we should enable it unless there actually is `externalAddress`. Or am I missing something?

soider: Hm, I am trying to figure out what is the correct use case when we want to be:
- HA
- no external address
- tunneled network

at the same time.

I am even thinking, may be the logic should be totally different: use this reconciler if TunneledNetworkingMode, and add some validation somewhere, that if TunneledNetworkingMode is true but ExternalAddress is empty - fail the bootstrap. Wdyt?

jnummelin: There's many variables in this game for sure. :)

>  if TunneledNetworkingMode is true but ExternalAddress is empty - fail the bootstrap

That sounds about right and good thing to do in general, fails the bootstrap for config combos that are not gonna work properly. (dunno if it starts to creep a bit out of this PRs scope though)

soider: I think it might be better to get it out of that PR, so I just reverted the line you have commneted

jnummelin (approved):

PR #1618, opened 2022-03-24
Make kernel config smoketest more lenient

twz123:
## Description

It'll fail on systems which don't expose their kernel config at runtime, hence it may produce some unnecessary noise in those cases.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/sysinfo/probes/linux/kernel_test.go
+++ b/internal/pkg/sysinfo/probes/linux/kernel_test.go
@@ -118,7 +118,11 @@ func TestKConfigProber(t *testing.T) {
 
 		option, err := probeKConfig(ensureKConfig("I_CERTAINLY_DONT_EXIST"))
 		assert.Equal(t, option, kConfigUnknown)
-		assert.NoError(t, err)
+		if _, ok := err.(*noKConfigsFound); ok {
+			t.Log("system doesn't expose its kernel config")
+		} else {
+			assert.NoError(t, err)
+		}
 	})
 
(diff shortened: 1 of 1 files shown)


internal/pkg/sysinfo/probes/linux/kernel_test.go
```diff
@@ -118,7 +118,14 @@ func TestKConfigProber(t *testing.T) {
 
 		option, err := probeKConfig(ensureKConfig("I_CERTAINLY_DONT_EXIST"))
 		assert.Equal(t, option, kConfigUnknown)
-		assert.NoError(t, err)
+		switch err := err.(type) {
```
s0j, line 121: Can you use `errors.Is()` here? Looks like you'd be able to remove a case with it.

s0j: Oops, I meant `errors.As()`.  Something like this ..

```
var nce *noKConfigsFound
if errors.As(err, &nce) {
  t.Log("system doesn't ....")
} else {
  assert.NoError(t, err)
}
```

s0j (approved):
LGTM which ever way you choose, not blocking this.

internal/pkg/sysinfo/probes/linux/kernel_test.go
twz123: I've chosen a third option :-)

PR #1624, opened 2022-03-28
ARM64 support for etcd is stable now

twz123:
## Description

* Remove all mentions of ETCD_UNSUPPORTED_ARCH=arm64 or replace them
  just with arm.
* Update the unsupported arch check for etcd accordingly.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


pkg/component/controller/etcd.go
```diff
@@ -291,7 +291,8 @@ func (e *Etcd) Healthy() error {
 }
 
 func detectUnsupportedEtcdArch() error {
-	if strings.Contains(runtime.GOARCH, "arm") {
+	// https://github.com/etcd-io/etcd/blob/v3.5.2/server/etcdmain/etcd.go#L467-L472
+	if runtime.GOARCH != "amd64" || runtime.GOARCH == "arm64" {
```
ypnos, line 295: Is this logic correct? I would expect:
```
if runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64" {
```

Alternative:
```
if runtime.GOARCH == "amd64" || runtime.GOARCH == "arm64" {
    return nil
}
```

twz123: 🤦‍♂️ if you decide to do it the other way round after having done the first half already. Thanks for spotting!

twz123: Fixed and thanks again!

soider (approved):

PR #1640, opened 2022-04-07
Refactor sysinfo probes

twz123:
## Description

* Use probes.Probes directly in composite probe types
  So that the manual path prefixing is no longer necessary.
* Provide a simple default implementation for probes.ProbeDesc
  This allows for removing some private structs that were mostly only necessary to implement the probes.ProbeDesc interface.
* Remove the bogus "kernelConfig" intermediary probe layer
  It wasn't really reporting anything, and "just" provided a level of indentation. While this was somehow useful for the CLI output, it was a weird thing to do in the first place.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/sysinfo/probes/probes.go
+++ b/internal/pkg/sysinfo/probes/probes.go
@@ -97,8 +97,13 @@ type Probes interface {
 	Probe
 }
 
-// NewProbes returns a new, empty composite probe at the given path.
-func NewProbes(path ProbePath) Probes {
+// NewRootProbes returns a new, empty composite probe without a path.
+func NewRootProbes() Probes {
+	return &probes{nil, nil}
+}
+
+// NewProbesAtPath returns a new, empty composite probe at the given path.
+func NewProbesAtPath(path ProbePath) Probes {
 	return &probes{path, nil}
 }
 
--- a/internal/pkg/sysinfo/probes/probes.go
+++ b/internal/pkg/sysinfo/probes/probes.go
@@ -58,6 +58,11 @@ type ProbeDesc interface {
 	DisplayName() string
 }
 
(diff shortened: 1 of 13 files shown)


internal/pkg/sysinfo/sysinfo.go
```diff
@@ -52,7 +52,7 @@ func (s *K0sSysinfoSpec) RunPreFlightChecks(lenient bool) error {
 }
 
 func (s *K0sSysinfoSpec) NewSysinfoProbes() probes.Probes {
-	p := probes.NewProbes()
+	p := probes.NewProbes(nil)
```
makhov, line 55: Looks a bit fishy, is it on purpose?

makhov: IMHO, both will be better than `nil`, it's up to you what to choose :)

twz123: Done.

internal/pkg/sysinfo/probes/probes.go
```diff
@@ -92,11 +97,24 @@ type Probes interface {
 	Probe
 }
 
-// NewProbes returns a new, empty composite probe.
-func NewProbes() Probes {
-	return &probes{}
+// NewRootProbes returns a new, empty composite probe without a path.
+func NewRootProbes() Probes {
+	return &probes{nil, nil}
```
trawler, line 102: if path and probes values are left empty (`return &probes{}`) do they not equal nil by default? (meaning, do we need to set the values in the struct explicitly?)

makhov (approved):

PR #1333, opened 2021-12-09
Output logs to stderr, default log level to Warn, add global verbose flag

kke:
Signed-off-by: Kimmo Lehto <klehto@mirantis.com>

Closes #1325 (alternative)

* Changes default output level to WARN
* All logging goes to stderr
* Adds `--verbose|-v` for displaying progress information (aka, change loglevel to INFO)
* Changes the log level of several messages to make them more or less visible
* Changes the way `k0s etcd member-list` outputs the json.
* Output of `k0s controller`, `k0s worker` and `k0s api` is by default INFO level and goes to stdout (because they stay in the foreground)
* The stderr of some smoke-test commands is redirected to /dev/null because `ExecWithOutput` includes output from both streams.

Rationale:

* Almost all of our output is actually not "output"/"payload", but progress information and it should not be included with the output when redirecting or piping.
* None of our INFO level messages seem to contain important information / "payloads", those are always displayed via `fmt.Print`. 
* Warnings and errors belong to stderr, not stdout.

PR #1659, opened 2022-04-25
Restore --debug flag functionality

twz123:
## Description

The debug flag was not functional for the controller, worker and api subcommands since #1333.

Also:

* Remove the explicit dependency to viper.
* Remove ineffective duplicate definition of the debug flag

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/token/create.go
+++ b/cmd/token/create.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
@@ -41,8 +40,6 @@ k0s token create --role worker --expiry 10m  //sets expiration time to 10 minute
 `,
 		PreRunE: checkCreateTokenRole,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// Disable logrus for token commands
-			logrus.SetLevel(logrus.WarnLevel)
 			c := CmdOpts(config.GetCmdOpts())
 			expiry, err := time.ParseDuration(tokenExpiry)
 			if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/token/create.go
```diff
@@ -41,8 +40,6 @@ k0s token create --role worker --expiry 10m  //sets expiration time to 10 minute
 `,
 		PreRunE: checkCreateTokenRole,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			// Disable logrus for token commands
-			logrus.SetLevel(logrus.WarnLevel)
```
kke, line 45: The token commands could also do `logrus.SetOutput(os.Stderr)`

cmd/worker/worker.go
```diff
@@ -53,10 +53,13 @@ func NewWorkerCmd() *cobra.Command {
 	$ k0s worker --token-file [path_to_file]
 	Note: Token can be passed either as a CLI argument or as a flag`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			logrus.SetLevel(logrus.InfoLevel)
+			c := CmdOpts(config.GetCmdOpts())
```
kke, line 56: There could be a `c.LogLevel` 🤔

kke (approved):

cmd/token/create.go
twz123: stderr/warn is the k0s default, so I think it would be redundant here?

twz123:
@kke would you mind to have another look?

cmd/token/create.go
kke: Then it shouldn't matter if it outputs warnings as long as the token goes to stdout and everything else to stderr.

kke (approved):

PR #1665, opened 2022-04-28
Remove quoting when parsing flags for service unit files

jnummelin:
The way we quoted the flag values broke down in older systemd (219 at least) setups.

## Description

The current way we parse and handle the flags from install --> service unit files adds quotes aounrd the values IF they are string arrays and slices. That effectively results into following execs:

```
ExecStart=/usr/local/bin/k0s controller --config=/etc/k0s/k0s.yaml --disable-components="konnectivity-server,default-psp" --enable-worker=true --no-taints=true
```

Now this quoting breaks older systemd (219 at least):
```
Error: unknown component konnectivity-server,default-psp
```

This works perfectly fine on newer systemd (tested on 237 and 245)

So the fix is to remove quoting and rely on kardianos/service for encoding the possible spaces properly.

One possible caveat is that IF there are any kubelet flags passed via `--kubelet-extra-args` that might need spaces those will get probably broken. But reading through the kubelet flag helps does not imply any of those would have values with spaces in them.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/install/linux_openrc.go
+++ b/pkg/install/linux_openrc.go
@@ -23,7 +23,7 @@ name="{{.DisplayName}}"
 description="{{.Description}}"
 command={{.Path|cmdEscape}}
 {{- if .Arguments }}
-command_args="{{range .Arguments}}{{.}} {{end}}"
+command_args="{{range .Arguments}}'{{.}}' {{end}}"
 {{- end }}
 name=$(basename $(readlink -f $command))
 supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"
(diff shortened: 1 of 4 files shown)


makhov (dismissed):

kke (approved):

ncopa (approved):
LGTM

pkg/install/linux_openrc.go
```diff
@@ -23,7 +23,7 @@ name="{{.DisplayName}}"
 description="{{.Description}}"
 command={{.Path|cmdEscape}}
 {{- if .Arguments }}
-command_args="{{range .Arguments}}{{.}} {{end}}"
+command_args="{{range .Arguments}}'{{.}}' {{end}}"
```
kke, line 26: `--foo=bar baz,--bar=5` becomes `'--foo=bar baz' '--bar=5'`, looks weird but should work

Issue #1663, opened 2022-04-28
Bump k8s v1.24.0 / containerd 1.6.4


twz123:
This is a follow-up to #1634 with the next version bumps:

* [x] [Kubernetes v1.24.0](https://github.com/kubernetes/kubernetes/releases/tag/v1.24.0)
* [x] [containerd 1.6.4](https://github.com/containerd/containerd/releases/tag/v1.6.4)

twz123:
Turns out that k8s v1.24.0-rc.1 now requires go 1.18.1 ...

twz123:
We will skip containerd 1.6.3 and wait for containerd 1.6.4: https://github.com/containerd/containerd/issues/6876

iusetabs:
@twz123 Just running some tests using the latest version of k0s and containerd v1.6.3. My test fails and the pod is stuck in ContainerCreating with "failed to setup network for sandbox". There is no issue with containerd v1.6.4. Is 1.6.3 going to be skipped or? 😄

twz123:
@iusetabs Yep, k0s's integration tests were failing with 1.6.3... Skipped 1.6.3 (https://github.com/k0sproject/k0s/issues/1663#issuecomment-1113524232), already on 1.6.4 (#1690) on main.

iusetabs:
@twz123 awesome thx very much :-D

PR #1666, opened 2022-04-28
Bump go 1.18.1

twz123:
## Description

Kubernetes 1.24 switched to go 1.18.1 as well.

* Remove as many hard-coded go versions as possible
* Also regenerate go.mod/go.sum for CI dependencies
* Add empty go.mod to ignore build directory from go tools
  See https://github.com/golang/go/issues/30058#issuecomment-543815369
* Add missing dependencies to Makefile targets
* Set git config advice.detachedHead=false for embedded binaries
  So that there's less log spam
* Change go get to go install for Konnectivity

See #1663

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


twz123:
Corresponding Kubernetes PR: https://github.com/kubernetes/kubernetes/pull/109461

twz123:
And its minimum go version requirement: https://github.com/kubernetes/kubernetes/pull/109484

hack/ci-deps/go.mod
```diff
@@ -1,21 +1,233 @@
-module cideps
+module ci-deps
```
twz123, line 1: If we go for #1670, then those file will become supefluous.

.github/workflows/check-network.yaml
```diff
@@ -46,7 +42,10 @@ jobs:
         with:
           ref: ${{ env.github_ref }}
 
-      - name: Set up Go 1.x
+      - name: Prepare build environment
+        run: .github/workflows/prepare-build-env.sh
```
jnummelin, line 46: hmm, not sure if I'm getting blind but I do not see this file being part of the PR?

jnummelin: oh, never mind, it came already with some earlier PR and now you're wiring it up to be used everywhere 👍

twz123: This is part of the "Remove as many hard-coded go versions as possible" effort. It has already been done for the main build workflow. I thought it would be better to propagate that technique to the remaining workflows, instead of updating the hard-coded go versions there.

jnummelin (approved):

PR #1670, opened 2022-04-29
Use pure go install for CI binaries

twz123:
## Description

This makes it easier to maintain, as there's no go.mod/go.sum to be maintained anymore. Moreover, the build gets more consistent, because binaries in the PATH outside of the build environment are ignored.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


soider:
I don't really have super strong opinion here, but I like current hack/ci-deps/go.mod thing since it allows to pin versions the same way as any other deps.

Bummer that even with the new workspace feature the go doesn't allow to install binaries project wide.

twz123:
I had the feeling that it was a little bit difficult to properly update those files when trying out the go 1.18 bump. That's why I came up with this proposal. Having a single module to manage build dependencies for several independent go binaries artificially forces their dependencies to the exact same versions, even if the binaries are compiled separately and can happily use different dependency versions. All of the used binaries are go modules themselves, and have pinned down their dependency versions, except for go-bindada, which doesn't have any dependencies at all. Two of the four binaries already have their version specified in the Makefile, so their versions already cannot be updated via a single `go get`.

The only thing that we're gaining with those go.mod/go.sum files is a proof that the git tags of the binaries didn't change.

twz123:
There's already another solution on the way which makes this one obsolete. See #1731.

PR #1681, opened 2022-05-02
Remove unused Reconcile() methods from components

twz123:
## Description

This is probably a leftover from a previous refactoring. Those methods are not matching the signatures as required by ReconcilerComponent, and weren't doing anything useful in the first place. Also add a constraint to all the structs to actually adhere to the component interface.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -24,6 +24,7 @@ import (
 	"gopkg.in/fsnotify.v1"
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
+	"github.com/k0sproject/k0s/pkg/component"
 	"github.com/k0sproject/k0s/pkg/component/controller"
 	"github.com/k0sproject/k0s/pkg/constant"
 	kubeutil "github.com/k0sproject/k0s/pkg/kubernetes"
@@ -43,6 +44,8 @@ type Manager struct {
 
 	LeaderElector controller.LeaderElector
 }
+
+var _ component.Component = (*Manager)(nil)
 
 // Init initializes the Manager
 func (m *Manager) Init(ctx context.Context) error {
@@ -82,12 +85,6 @@ func (m *Manager) Stop() error {
 	if m.cancelWatcher != nil {
 		m.cancelWatcher()
 	}
-	return nil
-}
-
-// Reconcile reconciles the Manager
-func (m *Manager) Reconcile() error {
-	logrus.Debug("reconcile method called for: Manager")
 	return nil
 }
 
--- a/pkg/component/controller/csrapprover.go
+++ b/pkg/component/controller/csrapprover.go
@@ -33,6 +33,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 
 	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+	"github.com/k0sproject/k0s/pkg/component"
 	k8sutil "github.com/k0sproject/k0s/pkg/kubernetes"
 	kubeutil "github.com/k0sproject/k0s/pkg/kubernetes"
 )
@@ -59,6 +60,8 @@ type CSRApprover struct {
 	clientset         clientset.Interface
 }
 
+var _ component.Component = (*CSRApprover)(nil)
+
 // NewCSRApprover creates the CSRApprover component
 func NewCSRApprover(c *v1beta1.ClusterConfig, leaderElector LeaderElector, kubeClientFactory k8sutil.ClientFactoryInterface) *CSRApprover {
 	d := atomic.Value{}
@@ -76,12 +79,6 @@ func (a *CSRApprover) Healthy() error { return nil }
 // Stop stops the CSRApprover
 func (a *CSRApprover) Stop() error {
 	a.stop()
-	return nil
-}
-
-// Reconcile detects changes in configuration and applies them to the component
-func (a *CSRApprover) Reconcile() error {
-	logrus.Debug("reconcile method called for: CSRApprover")
 	return nil
 }
 
--- a/pkg/component/controller/leaderelector.go
+++ b/pkg/component/controller/leaderelector.go
@@ -31,11 +31,10 @@ type LeaderElector interface {
 	IsLeader() bool
 	AddAcquiredLeaseCallback(fn func())
 	AddLostLeaseCallback(fn func())
-	component.Component
 }
 
-type leaderElector struct {
-	L *logrus.Entry
+type LeasePoolLeaderElector struct {
+	log *logrus.Entry
 
 	stopCh            chan struct{}
 	leaderStatus      atomic.Value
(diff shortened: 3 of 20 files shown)


pkg/applier/manager.go
```diff
@@ -44,6 +45,8 @@ type Manager struct {
 	LeaderElector controller.LeaderElector
 }
 
+var _ component.Component = (*Manager)(nil)
```
makhov, line 48: Almost everywhere we do it like `var _ component.Component = &Manager{}`. I'd prefer to keep it consistent.

pkg/component/controller/csrapprover.go
```diff
@@ -60,7 +61,7 @@ type CSRApprover struct {
 }
 
 // NewCSRApprover creates the CSRApprover component
-func NewCSRApprover(c *v1beta1.ClusterConfig, leaderElector LeaderElector, kubeClientFactory k8sutil.ClientFactoryInterface) *CSRApprover {
+func NewCSRApprover(c *v1beta1.ClusterConfig, leaderElector LeaderElector, kubeClientFactory k8sutil.ClientFactoryInterface) component.Component {
```
makhov, line 64: What Is the reason for the change?

pkg/component/controller/leaderelector.go
```diff
@@ -47,7 +46,10 @@ type leaderElector struct {
 }
 
 // NewLeaderElector creates new leader elector
-func NewLeaderElector(kubeClientFactory kubeutil.ClientFactoryInterface) LeaderElector {
+func NewLeaderElector(kubeClientFactory kubeutil.ClientFactoryInterface) interface {
+	LeaderElector
+	component.Component
```
makhov, line 51: What is the idea here?

pkg/component/controller/csrapprover.go
twz123: The real interface that should be visible to the outside is just `component.Component`. The actual struct is just an implementation detail and can be hidden.

pkg/component/controller/leaderelector.go
twz123: Separation of concerns. We have two use cases for the return value: Once, for the component manager (which doesn't care about the component being a leader elector), and once for the rest of the code, which doesn't care that the leader elector is a component, and shouldn't call any of its lifecycle methods.

pkg/applier/manager.go
makhov: I'm ok with both, as I said we just need to be consistent here.

pkg/component/controller/csrapprover.go
makhov: It violates the "accept interfaces, return structs" rule. I'd prefer to keep it as is.

pkg/component/controller/leaderelector.go
makhov: Yeah, maybe we can do it this way:

```
type LeaderElectorComponent interface {
	LeaderElector
	component.Component
}

func NewLeaderElector(kubeClientFactory kubeutil.ClientFactoryInterface) LeaderElectorComponent 
```

twz123: Ah okay, so the answer here was "laziness" ;) I thought we could go without the extra type. Would you prefer this?

pkg/component/controller/csrapprover.go
twz123: Do you have a reference to this, so that I can read something about that?

I would argue that it's a common pattern to just expose interfaces (or traits, depending on the language) from factory methods. It makes it much clearer to understand what the return value should actually be used for. This is also not uncommon in Go, I'd say. All the standard library methods return plain errors, instead of e.g. *PathError or similar.

pkg/applier/manager.go
twz123: Opened #1689 for that.

pkg/component/controller/csrapprover.go
makhov: This is a very common mojo of Go. I'm not sure who was the first one who said it. Probably, it [was](https://www.reddit.com/r/golang/comments/dfe1qr/who_first_said_accept_interfaces_and_return/) Dave Cheney's tweet, but he has deleted all his tweets. 

Google [gives](https://www.google.com/search?q=accept+interfaces%2C+return+structs&newwindow=1&sxsrf=ALiCzsYhz81ejp3U1mbmzgFhCe9ghRgaqQ%3A1651656940650&source=hp&ei=7EhyYpWyJPGSxc8P79-v4A0&iflsig=AJiK0e8AAAAAYnJW_Id2hpxIC9xCnq-NHHG77fYXSznn&ved=0ahUKEwiV2PDJxcX3AhVxSfEDHe_vC9wQ4dUDCAY&uact=5&oq=accept+interfaces%2C+return+structs&gs_lcp=Cgdnd3Mtd2l6EAMyBQgAEIAEUABYAGDqA2gAcAB4AIABOIgBOJIBATGYAQCgAQKgAQE&sclient=gws-wiz) a lot of articles. 

> It makes it much clearer to understand what the return value should actually be used for.

Actually, it is the idea. Don't decide how to use it in the producer, let the consumer do it. The standard library has some exceptions for different reasons, but they are still exceptions.

pkg/component/controller/leaderelector.go
makhov: I don't know the reasons why we don't return just struct here, but yeah, if we decide to return an interface, I think having something like `LeaderElectorComponent` is a bit clearer.

pkg/component/controller/csrapprover.go
twz123: Thanks for your answer.

> Actually, it is the idea. Don't decide how to use it in the producer, let the consumer do it. The standard library has some exceptions for different reasons, but they are still exceptions.

That kinda contradicts how I think about types. While reading through all the people's comments on that topic, I clearly found voices who were saying that "preemptive" interfaces are just the wrong approach for Go, but not really giving any reason, besides that in Go, in contrast to some other languages, one can define interfaces after the fact on the consumer side. I don't buy that as is. Just because one can do it a certain way, it doesn't mean that that way is better than another one. I'd argue that interfaces neither belong to a consumer or a provider, but they literally stand in between of both of them. They're contracts, after all. So it makes perfect sense to me to think in interfaces first, to understand in how systems interact with each other, especially in more complex scenarios. Then I can start to look at actual implementations of both consumers and providers.

Go structs also have public interfaces, of course, but those are by far less easy to grasp from looking at the source code, as compared to interfaces. Go structs public methods are scattered across lengthy method implementations, where often enough public methods and private utility methods are mixed up, so there's no way to easily grasp the actual intent of the public API surface of a struct. One can address this when thinking about where to place which method in a source file, but most often, there are colliding interests: Either group the methods to better outline the public API, or group methods to better reflect the actual implementation. Usually the second one wins for good reasons.

Almost all of the component structs are just that: `Component`s. Some of them `ReconcileComponent`s. They need to implement those interfaces anyways, and, in fact, there are lot of interface checks in the code to ensure this. For me, it feels most natural to use those interface types in the factory methods. This communicates very clearly: No other interaction possible with the returned value. I don't have to check the code to find all public methods and check if there's anything outside the component interfaces that's used by k0s.

So, all in all, I prefer to return interfaces here for a single reason: Readability of intent. But that's of course just how I wrap my head around things. I am definitely on the idiomatic side of things and I now accept that, in this case, the idiomatic way of Go is not directly aligning with my train of thoughts :blush:

I'll change the return values to struct pointers again (also including the LeaderElector, for consistency :v:). It might make sense to make the structs private later on, though, since they are never used directly, just via the interfaces. Thanks again for pointing out that mantra. Definitely made me think a lot.

pkg/component/controller/leaderelector.go
twz123: :-/

* Lint: pkg/component/controller/leaderelector.go#L52
  unexported-return: exported func NewLeaderElector returns unexported type *controller.leaderElector, which can be annoying to use (revive)
* Lint: pkg/component/controller/k0scloudprovider.go#L40
  unexported-return: exported func NewK0sCloudProvider returns unexported type *controller.k0sCloudProvider, which can be annoying to use (revive)

I am inclined to either silence the linter in those cases via a nolint comment, or return interfaces instead :disappointed: WDYT @makhov?

soider: I would definitely vote for named interface. You write it once and read X times and each time reading the anonymous iface would require from me unnecessary concentration.

twz123: For the "silence the linter" case, there would be only a single occurrence of a composite interface left. Not sure if it would make sense to declare a type alias just for that.

For the "return interface" case, there would still be just two occurrences of the composite interface. That's at least one more, so that type alias would make some sense.

twz123:
> LGTM, but needs conflict resolving and some linter fixes

Concerning the linters, I asked about it here: https://github.com/k0sproject/k0s/pull/1681#discussion_r867910871

twz123:
I made some structs public to fix the linter errors.

makhov (approved):

PR #1683, opened 2022-05-03
Local IPs detection

### LINKED ISSUE #1661: Detect localhost address and automatically appends to API server cert SANs
Currently we hardcode both `localhost` and `127.0.0.1` as SANs into the cert we generate for API server. But in some cases (e.g. Hetzner cloud) the boxes are configured with localhost pointing to `127.0.1.1`

We should try to detect to which local IP localhost is actually pointing at and add that to SANs automatically.

The current workaround is to add these to `spec.api.sans` list manually.

makhov:

## Description

Local IPs detection to append them to the cert SANs. Currently, we use `127.0.0.1` which is not always the case.

Fixes #1661. 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -191,6 +191,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 		"kubernetes.default.svc.cluster",
 		fmt.Sprintf("kubernetes.svc.%s", c.ClusterSpec.Network.ClusterDomain),
 		"localhost",
+		"127.0.0.1",
 	}
 
 	localIPs, err := detectLocalIPs()
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -193,11 +193,11 @@ func (c *Certificates) Init(ctx context.Context) error {
 		"localhost",
 	}
 
-	localIP, err := detectLocalIP()
+	localIPs, err := detectLocalIPs()
 	if err != nil {
 		return fmt.Errorf("error detecting local IP: %w", err)
 	}
-	hostnames = append(hostnames, localIP)
+	hostnames = append(hostnames, localIPs...)
 	hostnames = append(hostnames, c.ClusterSpec.API.Sans()...)
 
 	internalAPIAddress, err := c.ClusterSpec.Network.InternalAPIAddresses()
@@ -237,18 +237,27 @@ func (c *Certificates) Init(ctx context.Context) error {
 	return eg.Wait()
 }
 
-func detectLocalIP() (string, error) {
+func detectLocalIPs() ([]string, error) {
+	var localIPs []string
 	addrs, err := net.LookupIP("localhost")
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+
+	if hostname, err := os.Hostname(); err == nil {
+		hostnameAddrs, err := net.LookupIP(hostname)
+		if err == nil {
+			addrs = append(addrs, hostnameAddrs...)
+		}
 	}
 
 	for _, addr := range addrs {
 		if addr.To4() != nil {
-			return addr.String(), nil
+			localIPs = append(localIPs, addr.String())
 		}
 	}
-	return "", fmt.Errorf("could not detect ipv4 IP")
+
+	return localIPs, nil
 }
 
 func kubeConfig(dest, url, caCert, clientCert, clientKey, owner string) error {
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


jnummelin:
I don't think this cover the Hetzner specific case as the address `127.0.1.1` is NOT actually configured on the interfaces:
```
root@k0s-ubuntu20:~# hostname -i
127.0.1.1
root@k0s-ubuntu20:~# ip add show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 96:00:01:3f:4d:32 brd ff:ff:ff:ff:ff:ff
    inet 167.235.251.210/32 scope global dynamic eth0
       valid_lft 68554sec preferred_lft 68554sec
    inet6 2a01:4f8:1c17:cf11::1/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 fe80::9400:1ff:fe3f:4d32/64 scope link 
       valid_lft forever preferred_lft forever
```

Maybe we should do the `hostname -i` equivalent in the code to detect this case?

cmd/controller/certificates.go
```diff
@@ -232,6 +237,36 @@ func (c *Certificates) Init(ctx context.Context) error {
...
+	}
+	for _, i := range ifaces {
+		addrs, err := i.Addrs()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, addr := range addrs {
+			switch v := addr.(type) {
+			case *net.IPNet:
```
twz123, line 254: Not sure if we want whole CIDRs?

cmd/controller/certificates.go
```diff
@@ -232,6 +237,36 @@ func (c *Certificates) Init(ctx context.Context) error {
...
+				if v.IP.To4() == nil {
+					continue
+				}
+				localIPs = append(localIPs, v.IP.To4().String())
+			case *net.IPAddr:
+				if v.IP.To4() == nil {
+					continue
+				}
+				localIPs = append(localIPs, v.IP.To4().String())
```
twz123, line 263: Isn't the linked issue just about localhost? Maybe we should filter for everything in the 127.0.0.0/8 subnet here?

twz123:
> Maybe we should do the `hostname -i` equivalent in the code to detect this case?

For the record: it [uses](https://sourceforge.net/p/net-tools/code/ci/v2.10/tree/hostname.c#l155) `gethostbyname` and [prints](https://sourceforge.net/p/net-tools/code/ci/v2.10/tree/hostname.c#l194) the `h_addr_list` from the `hostent` struct.

Closest in Golang is probably `net.LookupHost` for both `"localhost"` and `os.Hostname()`.

cmd/controller/certificates.go
makhov: We don't, we add only IPs

cmd/controller/certificates.go
makhov: Right, will fix

makhov:
`hostname -i` on my AWS instance returns `172.31.28.213`, and the same IP for `host $(hostname)`
I think `net.LookupHost("localhost")` will be a good option

cmd/controller/certificates.go
```diff
@@ -232,6 +237,20 @@ func (c *Certificates) Init(ctx context.Context) error {
 	return eg.Wait()
 }
 
+func detectLocalIP() (string, error) {
+	addrs, err := net.LookupIP("localhost")
```
twz123, line 241: I suspect that we need to do it twice, once for localhost, and once for the hostname itself? On my box, localhost resolves to 127.0.0.1, and my hostname resolves to 127.0.0.2

cmd/controller/certificates.go
```diff
@@ -189,10 +190,14 @@ func (c *Certificates) Init(ctx context.Context) error {
 		"kubernetes.default.svc",
 		"kubernetes.default.svc.cluster",
 		fmt.Sprintf("kubernetes.svc.%s", c.ClusterSpec.Network.ClusterDomain),
-		"127.0.0.1",
```
twz123, line 192: Might be worthwhile to include 127.0.0.1 in any case, even if localhost and the hostname resolve to something different.

twz123 (dismissed):

cmd/controller/certificates.go
makhov: Done

twz123 (approved):

PR #1685, opened 2022-05-03
[release-1.22] Remove quoting when parsing flags for service unit files

jnummelin:
The way we quoted the flag values broke down in older systemd (219 at least) setups.

## Description

Cherry pick #1665 (the systemd fix only)


ncopa (approved):
LGTM

PR #1727, opened 2022-05-16
Autopilot

makhov:
## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


s0j:
I realize this is draft, but there are some things here that I think that need sorting out first before taking on a PR of this size. I've mentioned a few of these to @makhov offline, and some after thinking about this.

* With a diff this large, its a tough review even with passing tests/CI.
* The new code of commonizing of the autopilot 'root' controller doesn't buy us anything when we're adding in a switch in `pkg/autopilot/autopilot.go` based off of controller/worker.
  * Also, since a 'controller+worker' is the same binary, we only need to add the 'controller' logic, as that would implicitly update the worker parts of the logic on restart.
* We need to be aware of when autopilot applies its CRDs and how that should appear in the k0s startup (or component startup?)
* The `./hack/crd-gen/crd-gen.sh` is something that can be a valuable addition to k0s before this integration.
* How will the autopilot integration tests get integrated?
  * Need to be careful here because autopilot has copied + modified the original `footloosesuite.go` in order to allow for replacing the k0s binary + other updates.
  * I don't think we can just drop in a new footloosesuite without fallout (and not needed for an initial PR)
* Where are we getting the autopilot command line arguments from?
  * k0s.yaml
  * command line arguments to k0s?

## Alternative 

What I see is a number of phases to this that can run in parallel.

* Add crd-gen (a standalone PR)
* Figure out how to marry the two footloosesuite implementations.
* Figure out how the CRD will get applied at startup, using the statically committed CRD files we provide.
* Move over autopilot `pkg/*` to k0s `pkg/autopilot/*' (unwired, but unit-tests work)
* Figure out autopilot command line argument adoption to k0s

These three once sorted out would give the groundwork for actually moving over autopilot code, and make our lives (and PRs) much easier.

From there, additional parallel items:
* Add worker code
* Add controller code

Both of these will probably look similar, but thats fine for now -- can optimize later if it makes sense.

Finally:
* Add autopilot integration tests with the footloosesuite approach figured out during the groundwork stage.

Looking for comments on this, but I really think we need to chop this up.

makhov:
Yeah, I agree, that it's a huge change, but I think we can't merge non-working parts into the main, so I'd propose to do all integration work in a separate branch and create PRs against it. 
At the first glance, we need to do three big steps: code, integration tests, and tooling. The order doesn't matter much, but we can't make tests work unless we have code integrated. 
That's why I started with this PR, then we can create PRs with much smaller changes against this branch

jnummelin:
> I think we can't merge non-working parts into the main

I wouldn't be 100% strict on that. We can have parts in main that do not work IF they are not wired into k0s itself yet. What that essentially means is that we can do some groundwork in a way that's not wired to anything yet.

PR #1732, opened 2022-05-18
Autopilot integration. pkg, internal and inttest packages

makhov:

## Description

pkg, internal and inttest packages moved to the k0s repo. Nothing more is done here.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


go.mod
```diff
@@ -68,6 +68,17 @@ require (
...
 
+require (
+	github.com/Masterminds/semver/v3 v3.1.1
+	github.com/cavaliergopher/grab/v3 v3.0.1
+	github.com/go-playground/validator/v10 v10.9.0
+	github.com/google/go-cmp v0.5.7
+	github.com/robfig/cron v1.2.0
+	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/apiextensions-apiserver v0.24.0
+	k8s.io/kubernetes v1.13.0
```
s0j, line 79: This doesn't look right (`go mod tidy` is trying to trick you)

I'd expect this to be `v1.24.0`, and doing that there should be a number of `replace` entries for each of the submodules pinning them to `v1.24.0`. This was all needed just to use the k8s ready-probing code.

From past experiences with this, I expect it to be tricky.

s0j (changes requested):

.gitignore
```diff
@@ -1,6 +1,6 @@
 /build/cache/*
 pkg/assets/zz_generated_offsets*.go
-k0s
+./k0s
```
twz123, line 3: Sure this doesn't inflict any harm on the docs build? IIRC It's downloading the k0s binary into the docs folder.

go.mod
```diff
@@ -68,6 +68,17 @@ require (
 	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
 )
 
+require (
```
twz123, line 71: Can you merge this block into the k0s block in the beginning?

twz123 (changes requested):

.gitignore
s0j: This should be `/k0s` to exclude the built `k0s` binary. There are others that can be scoped to the root of the project, but this one alone is easily the most needed change.

I just hit this problem again as autopilot has paths with `/k0s/` in them, resulting in them being incorrectly ignored.

static/autopilot.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+var (
+	//go:embed manifests/autopilot.k0sproject.io/v1beta2/*.yaml
+	crds embed.FS
+)
+
+// LoadCustomResourceDefinitions extracts all of the embedded CRDs into a map, keyed
+// by the real CRD name. Failing to load or parse a CRD will result in an error.
+func LoadCustomResourceDefinitions() (map[string]string, error) {
```
s0j, line 38: This code made sense with autopilot running standalone -- the CRD was embedded in the binary and applied on startup. Embedding this into `k0s` now, this should probably be done at `pkg/component/controller/crd.go`, or maybe even better in the new component that we add for autopilot-controllers (a different PR).

Either way, you can easily ignore this + manifests from this change, and remove the `LoadCustomResourceDefinitions()` usage.

go.mod
makhov: Done

.gitignore
twz123: Okay, I suggest to change this as follows, then:

```diff
diff --git a/.gitignore b/.gitignore
index c54ccd97a..a9fd75e57 100644
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,9 @@
 /build/cache/*
-pkg/assets/zz_generated_offsets*.go
-./k0s
-k0s.code
-k0s.exe
-k0s.exe.code
+/k0s
+/k0s.code
+/k0s.exe
+/k0s.exe.code
+/pkg/assets/zz_generated_offsets*.go
 bindata*
 .*.stamp
 .tmp
diff --git a/docs/.gitignore b/docs/.gitignore
index e71dd39e6..4f2b5aedd 100644
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,2 +1,4 @@
 /*.etag
 /.docker-image.*
+/k0s
+/k0s.exe
```

go.mod
twz123: > It's the version we use in the k0s currently.

You mean in autopilot, not in k0s? In k0s, we definitely use 1.24.0.

Since the code is currently not yet active and k8s.io/kubernetes is just used in a single file for a HTTP probe, what about commenting that piece of code out and remove the dependency, until we can re-add it properly?

s0j: Life will be much easier if you reference the `require` + `replace` directives in autopilot as this work has already been done.

Autopilot uses `k8s.io/kubernetes v1.24.0` -- `go mod tidy` is picking up the wrong version dependency because of the lack of replacement entries.

makhov: @twz123 no, we don't: https://github.com/k0sproject/k0s/blob/main/go.sum#L2291

I'm a bit losing the point of having a separate branch and keeping PRs small if we start making random improvements. In this PR I've put the code of autopilot packages, it doesn't pass our checks and doesn't work, so I'd suggest doing all the improvements in the separate small PRs. Otherwise, it will impossible to track the changes.

.gitignore
makhov: Done

static/autopilot.go
makhov: Removed

go.mod
s0j: This still needs to be `k8s.io/kubernetes v1.24.0` in `go.mod`.

The failing unit tests are the reason why
https://github.com/k0sproject/k0s/runs/6569146415?check_suite_focus=true#step:5:127

static/autopilot.go
s0j: The autopilot manifests are still in this diff, they can be removed.

makhov: They are needed exactly for `pkg/component/controller/crd.go` and are used in https://github.com/k0sproject/k0s/pull/1739

go.mod
s0j: So there are 4 PRs in motion (#1732, #1733, #1737, and #1739) with ~150 changed files that are nearly identical with each other, where three PRs are used to fix shortcomings with the first uncommitted PR? I don't understand this approach, or even how others can work with it.

I can't do anything until this stack of PRs is unwound.

s0j (approved):
Rubber-stamping this to unblock it.

PR #1742, opened 2022-05-19
Fix default build target (`build` vs. `all`)

s0j:
## Description

The previous change moved the 'all' target to be the first target, resulting
in `k0s.exe` being built unnecessarily on linux.

This change moves the `build` target to be first as it was originally.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,15 +65,15 @@ GO_ENV ?= docker run --rm \
 	k0sbuild.docker-image.k0s
 GO ?= $(GO_ENV) go
 
-.PHONY: all
-all: k0s k0s.exe
-
 .PHONY: build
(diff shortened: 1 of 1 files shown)


twz123 (approved):
Oops. I didn't think about the first goal being the default goal. Question is whether there is much value in having the all goal in the first place.

s0j:
> Oops. I didn't think about the first goal being the default goal. Question is whether there is much value in having the all goal in the first place.

Yeah, its subtle. I think there still is value for an `all` for those times you really _do_ want to build everything. I've used it on occasion, but mainly its just been `make` without targets.

PR #1791, opened 2022-06-02
Refactor Debouncer

twz123:
## Description

* Make Debouncer a generic struct over its channel items.
* Add a filter function, so that uninteresting items can be suppressed before resetting the debouncing timer.
* Combine the Start/Stop methods into a single Run method that exits when the given context is done. The Start method was blocking anyways, exiting when the Stop method was called. This way, it's also no longer required to have a factory method for Debouncer and it can be created directly as a struct literal.
* Clear potentially pending timer ticks when resetting it.
* Use the new Debouncer in StackApplier. Use a context instad of a "done" channel, so that it's compatible to the new Debouncer API.
* Update to the latest fsnotify library.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/stackapplier.go
+++ b/pkg/applier/stackapplier.go
@@ -23,8 +23,9 @@ import (
 
 	"github.com/k0sproject/k0s/pkg/debounce"
 	"github.com/k0sproject/k0s/pkg/kubernetes"
+
+	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
-	"gopkg.in/fsnotify.v1"
 )
 
 // StackApplier handles each directory as a Stack and watches for changes
@@ -34,11 +35,13 @@ type StackApplier struct {
 	fsWatcher *fsnotify.Watcher
 	applier   Applier
 	log       *logrus.Entry
-	done      chan bool
+
+	ctx    context.Context
+	cancel context.CancelFunc
 }
 
 // NewStackApplier crates new stack applier to manage a stack
-func NewStackApplier(path string, kubeClientFactory kubernetes.ClientFactoryInterface) (*StackApplier, error) {
+func NewStackApplier(ctx context.Context, path string, kubeClientFactory kubernetes.ClientFactoryInterface) (*StackApplier, error) {
 	watcher, err := fsnotify.NewWatcher()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/applier/stackapplier.go
```diff
@@ -100,3 +105,17 @@ func (s *StackApplier) DeleteStack(ctx context.Context) error {
...
 // Health-check interface
 func (s *StackApplier) Healthy() error { return nil }
+
+func (*StackApplier) triggersApply(event fsnotify.Event) bool {
+	// always let the initial apply happen
+	if event == (fsnotify.Event{}) {
+		return true
+	}
+
+	// ignore chmods (3845479a0)
```
s0j, line 115: Is `3845479a0` a commit of interest?

pkg/debounce/debounce.go
```diff
@@ -13,75 +13,90 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+			timer.Reset(d.Timeout)
+
+			// Save the current item to be sent after the timeout.
+			pendingItem = &item
+
+			// Now it is safe to restart the timer.
+			timer.Reset(d.Timeout)
+
+		case <-timer.C:
+			d.Callback(*pendingItem)
```
s0j, line 96: Is it possible to get the timer before the input? I'm looking at the dereferencing of `pendingItem` here.

pkg/applier/stackapplier.go
twz123: That was the commit that initially introduced the filtering for chmods.

pkg/debounce/debounce.go
twz123: It shouldn't be possible; I consider reaching this code here when `pendingItem` is `nil` a logic error.

When the select loop enters, it will enter with `pendingItem == nil` and an expired timer and this branch won't be reached. When there's a new item stored into `pendingItem`, the previous branch ensures that this branch will fire exactly 0 or 1 times, depending on the time between the incoming items. The previous branch stores the current item into `pendingItem` after it resets the timer, but since the timer channel is only polled in the exact same loop, its tick is guaranteed to be consumed after `pendingItem` has been updated.

s0j (approved):
LGTM

PR #1792, opened 2022-06-02
Add configurable k0s launch modes for integration tests

s0j:
## Description

In integration tests, you can now control how `k0s` will be launched,
either `standalone` (the default), or `openrc`.

This new `openrc` launch mode will start `k0s` via an openrc service.
The main motivation behind this is for the enabling of autopilot
integration tests which will need `k0s` to terminate and restart,
all handled by the service.

The `k0scloudprovider` integration test has been updated to use
`openrc` as an example.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -67,6 +67,11 @@ const (
 type FootlooseSuite struct {
 	suite.Suite
 
+	// Provide alternate launch functionalities as-needed.
+
+	LaunchMode     LaunchMode
+	launchDelegate *LaunchDelegate
+
 	/* config knobs (initialized via `initializeDefaults`) */
 
 	ControllerCount       int
@@ -110,6 +115,18 @@ func (s *FootlooseSuite) initializeDefaults() {
 	}
 	if s.KubeAPIExternalPort == 0 {
 		s.KubeAPIExternalPort = 6443
+	}
+	if s.LaunchMode == "" {
+		s.LaunchMode = LaunchModeStandalone
+	}
+
+	switch s.LaunchMode {
+	case LaunchModeStandalone:
+		s.launchDelegate = s.StandaloneLaunchDelegate()
+	case LaunchModeOpenRC:
+		s.launchDelegate = s.OpenRCLaunchDelegate()
+	default:
+		s.Require().Fail("Missing launch delegate (standalone, openrc)")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


inttest/common/footloosesuite.go
```diff
@@ -111,6 +116,18 @@ func (s *FootlooseSuite) initializeDefaults() {
...
+		s.LaunchMode = LaunchModeStandalone
+	}
+
+	switch s.LaunchMode {
+	case LaunchModeStandalone:
+		s.launchDelegate = s.StandaloneLaunchDelegate()
+	case LaunchModeOpenRC:
+		s.launchDelegate = s.OpenRCLaunchDelegate()
+	default:
+		panic("Missing launch delegate (standalone, openrc)")
```
twz123, line 129: Would it be possible to use `s.Require().Fail(...)` instead of panicking?

inttest/common/footloosesuite.go
```diff
@@ -412,19 +429,80 @@ func (s *FootlooseSuite) InitController(idx int, k0sArgs ...string) error {
...
+	if _, err := conn.ExecWithOutput(cmd); err != nil {
+		return fmt.Errorf("unable to execute '%s': %w", cmd, err)
+	}
+
+	return nil
+}
+
+// initControllerOpenRC initializes a controller in 'openrc' mode, meaning that
+// the k0s executable is launched as a service managed by openrc.
+func (s *FootlooseSuite) initControllerOpenRC(conn *SSHConnection, k0sArgs ...string) error {
```
twz123, line 461: Always asking myself when seeing those test method signatures: Since we have access to the testing.T here, wouldn't it be better to not return an error, but use `s.Require().NoError(...)` directly.  Is there a situation in which a test shouldn't abort if this method fails?

s0j: Its mainly about not having to do `s.Require().NoError(...)` through a number of layers of functions. These also aren't explicitly test functions, but supporting 'code' for setting up the test. I like to have the actual assertions done in the proper `Test*(...)` functions instead of the supporting code. Just a matter of preference.

inttest/common/footloosesuite.go
s0j: Yes, a much better idea. Done.

inttest/common/footloosesuite.go
```diff
@@ -412,19 +429,80 @@ func (s *FootlooseSuite) InitController(idx int, k0sArgs ...string) error {
...
+		return fmt.Errorf("unable to execute '%s': %w", cmd, err)
+	}
+
+	return nil
+}
+
+// installK0sServiceOpenRC will install an openrc k0s-type service (controller/worker)
+// if it does not already exist.
+func (s *FootlooseSuite) installK0sServiceOpenRC(conn *SSHConnection, k0sType string) error {
+	if _, err := os.Stat(fmt.Sprintf("/etc/init.d/k0s%s", k0sType)); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
s0j: oof, correct. I'll address that.

jnummelin (approved):

makhov (approved):

PR #1801, opened 2022-06-07
[release-1.23] Refactor k0sctl inttest

twz123:
## Description

* Use k0sctl v0.13.0, so that k0sctl supports more recent SSH key algorithms, which in turn allows for upgrading the base Alpine Docker image to a recent version.
* Hard-code the k0sctl version, instead of fetching the latest. This makes the tests more reproducible by not relying on external data that changes over time. Moreover, this allows for removing the explicit Golang dependency on k0sctl, which was only used for this test.
* Explicitly add the k0s version to the k0sctl configuration data. More recent versions of k0sctl check if the configured version and the version that the binary reports are matching.
* Fetch the full history during CI build, so that the k0s binary's version is populated correctly.
* Invoke k0sctl with its own environment. Helps to isolate test runs from regular k0sctl runs.
* Read both stdin and stderr when invoking k0sctl apply, just in case it prints something to stderr that may aid in debugging test failures.
* Remove some error return values from functions in favor of testify's require functionality.

See #1779.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


kke:
There's 0.13.0 now!

twz123:
> IMO makes sense to do this in the same PR here

So then this needs to wait on #1796.

twz123:
Okay, done!

ncopa (approved):

Issue #1814, opened 2022-06-08
More resilient manifest application


twz123:
### Is your feature request related to a problem? Please describe.

The [Manifest Deployer](https://docs.k0sproject.io/v1.23.6+k0s.2/manifests/) part of k0s could be more resilient and its performance relies on the current way in which components are initialized and run, even if it shouldn't matter.

* [x] Sometimes new stacks are missed: #1781
* [x] fsnotify events trigger a stack application even if those files would be ignored by the applier later on: #1815
* [ ] Files are written directly into the manifest folders, without any precautions that there might be files that are not fully written yet. This leads to situations where the stack applier behaves strangely (3845479a0): #1816

### Describe the solution you would like

Refactor the code to iron out edge cases.

### Describe alternatives you've considered

Leave it as is. Works 99% of the time. The 1% will be hard to debug.

PR #1815, opened 2022-06-08
Refactor StackApplier

twz123:
## Description

Part of #1814.

Changes to StackApplier:

* Combine Start/Stop methods into a simple Run method that exits when the given context is done. The start method blocked anyways, exiting when the Stop method was called. The context approach is somewhat more straight forward.
* Only consider fsnotify events for manifest files.
* Also log fsnotify watcher errors in the StackApplier.
* Synchronize access to the internal applier, as calls to it might happen concurrently.
* Remove the unused Healthy method.

Changes to the Manager:

* Always remove a stack from internal map, even if its deletion failed. The directory is gone anyways, and a readdition wouldn't work otherwise, because of the old, stopped stack still being in the map.
* Retry stack running on failure.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"time"
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/pkg/component"
@@ -40,12 +41,17 @@ type Manager struct {
 	bundlePath    string
 	cancelWatcher context.CancelFunc
 	log           *logrus.Entry
-	stacks        map[string]*StackApplier
+	stacks        map[string]stack
 
 	LeaderElector controller.LeaderElector
 }
 
 var _ component.Component = (*Manager)(nil)
+
+type stack = struct {
+	context.CancelFunc
+	*StackApplier
+}
 
 // Init initializes the Manager
 func (m *Manager) Init(ctx context.Context) error {
@@ -54,7 +60,7 @@ func (m *Manager) Init(ctx context.Context) error {
 		return fmt.Errorf("failed to create manifest bundle dir %s: %w", m.K0sVars.ManifestsDir, err)
 	}
 	m.log = logrus.WithField("component", "applier-manager")
-	m.stacks = make(map[string]*StackApplier)
+	m.stacks = make(map[string]stack)
 	m.bundlePath = m.K0sVars.ManifestsDir
 
 	m.applier = NewApplier(m.K0sVars.ManifestsDir, m.KubeClientFactory)
@@ -113,10 +119,7 @@ func (m *Manager) runWatchers(ctx context.Context) error {
 	}
 
 	for _, dir := range dirs {
-		if err := m.createStack(ctx, path.Join(m.bundlePath, dir)); err != nil {
-			log.WithError(err).Error("failed to create stack")
-			return err
-		}
+		m.createStack(ctx, path.Join(m.bundlePath, dir))
 	}
 
 	for {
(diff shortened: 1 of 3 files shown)


pkg/applier/manager.go
```diff
@@ -148,44 +149,53 @@ func (m *Manager) runWatchers(ctx context.Context) error {
...
-		return err
+		return
 	}
-	m.log.WithField("stack", name).Info("stack deleted succesfully")
+
 	delete(m.stacks, name)
+	stack.CancelFunc()
 
-	return nil
+	log := m.log.WithField("stack", name)
```
soider, line 193: nitpick: more idiomatic is not inverse compassion and make a return in negative case. That flattens the code flow and keeps every unsuccessful code branched while keeping successful flow on the top level.

e.g.
```
if err := stack.DeleteStack(ctx); err != nil {
   log.WithError(err)
   return
}
// successful branc

File: pkg/applier/manager.go [hunk 2]
```diff
@@ -148,44 +149,53 @@ func (m *Manager) runWatchers(ctx context.Context) error {
...
 
+	stackCtx, cancelStack := context.WithCancel(ctx)
+	stack := stack{cancelStack, NewStackApplier(name, m.KubeClientFactory)}
+	m.stacks[name] = stack
+
 	go func() {
-		_ = sa.Start()
-	}()
+		log := m.log.WithField("stack", name)
+		for timer := time.NewTimer(0 * time.Second); ; timer.Reset(10 * time.Second) {
```
makhov, line 164: What is the idea behind this? Maybe `time.Ticker` would be more readable and straightforward.

pkg/applier/stackapplier.go
```diff
@@ -17,95 +17,95 @@ package applier
...
-		log:       log,
-	}
 
-	sa.ctx, sa.cancel = context.WithCancel(ctx)
+	return &StackApplier{
+		log:  logrus.WithField("component", "applier-"+applier.Name),
+		path: path,
 
-	return sa, nil
+		doApply: func(ctx context.Context) error {
```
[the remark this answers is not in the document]
twz123: The intent wasn't to make it configurable, but to synchronize access to the Applier via a Mutex. Alternatively, we can put the Mutex and the Applier into the StackApplier struct, allowing the Applier to be called without locking the Mutex and making it less obvious which parts of the struct are actually guarded by that Mutex.

Using the function pointers was basically a shorthand for something like this:


```go
type StackApplier struct {
	// ...
	applier applier
}

type applier interface {
	Apply(context.Context) error
	Delete(context.Context) error
}

type syncApplier struct {
	sync.Mutex
	applier applier
}

func (a *syncApplier) Apply(ctx context.Context) error {
	a.Lock()
	defer a.Unlock()
	return a.applier.Apply(ctx)
}

func (a *syncApplier) Delete(ctx context.Context) error {
	a.Lock()
	defer a.Unlock()
	return a.applier.Delete(ctx)
}

func NewStackApplier(path string, kubeClientFactory kubernetes.ClientFactoryInterface) *StackApplier {
	applier := NewApplier(path, kubeClientFactory)
	return &StackApplier{
		// ...
		applier: &syncApplier{applier: &applier},
	}
}
```


(In theory, a unit test could leverage the doApply, doDelete stuff as well, but that was not the goal.)

pkg/applier/manager.go
twz123: This code should run the stack, as long as the context is not yet done. The stack's `Run` method will only return if either the context is done, or if an error occurs. In the latter case, the stack should be re-run after some timeout (10 seconds in this case).

I think a more readable version of the above would be this:

```go

go func() {
	log := m.log.WithField("stack", name)
	for {
		log.Info("Running stack")
		if err := stack.Run(stackCtx); err != nil {
			log.WithError(err).Error("Failed to run stack")
		}

		select {
		case <-time.After(10 * time.Second):
			continue
		case <-stackCtx.Done():
			log.Info("Stack done")
			return
		}
	}
}()
```

twz123: Done.

pkg/applier/manager.go
twz123: Done.

PR #1816, opened 2022-06-08
Write manifests atomically

twz123:
## Description

Part of #1814.

Use a temporary-write-and-rename approach to write manifests to disk, so that the appliers never see partially written files. Google's renameio library does this by creating temorary files whose names start with a dot. So let the appliers ignore anything that starts with a dot. That way, there's also no more need to ignore chmod events. They actually *can* matter, when files become readable or unreadable for the current process.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


internal/pkg/file/file.go
```diff
@@ -33,6 +36,49 @@ func Exists(fileName string) bool {
...
+	case RelativeBase:
+		return "RelativeBase"
+	}
+
+	return strconv.FormatInt(int64(d), 10)
+}
+
+// IsDottedBaseName checks if the given path is considered to be hidden on
+// Unix-like operating systems, i.e. if its base name starts with a dot.
+func IsDottedBaseName(path string) DottedBaseName {
```
jnummelin, line 71: The naming feels bit off. IMO the logic is more like converting a string path to `DottedBaseName` type so in my head something like `ToDottedBase` feels more natural.

jnummelin (approved):
Minor nitpick on some func naming, otherwise LGTM

internal/pkg/file/file.go
twz123: I see your point. Not too fond of the name (and the API) either. I previously called this `IsHiddenPath`, but that was also not feeling right. The initial implementation used a boolean return type, until I realized during implementation that a tri-state return value would be better. I refrained from using `(bool, bool)` which also felt awkward, somehow.

I don't think that the `To` prefix would be a better fit, since it's not really a conversion, too. What about `Get` though? Maybe `GetDotPrefixStyle` or something in that direction?

jnummelin: > Maybe GetDotPrefixStyle or something in that direction?

sounds much better.

twz123:
Turned into draft. Will revisit that, but not right now.

twz123:
Thanks! There are some [linter errors](https://github.com/k0sproject/k0s/actions/runs/3097871754/jobs/5027564267#step:4:8).

s0j (approved):

PR #1874, opened 2022-06-20
Refactor terraform code for check-network / conformance testing

### LINKED ISSUE #1831: Refactor conformance terraform scripts to remove unneeded module usage
Currently there's un-needed inversion between `intttest/conformance` which uses `inttest/terraform/test-cluster` as a module. IMO there's no real need for this and just makes the whole TF setup very confusing and hard to work with.

Myabe at the same time we could modify the TF outputs to be usable with k0sctl so we could use k0sctl to spin up the cluster instead of the test code spinning up stuff.

Now if the k0s cluster is spun up with k0sctl, why do we even need the Golang test suite for this, we're only essentially running sonobuoy on a ready-made cluster.

trawler:

## Description

Simplification and refactoring of sonobuoy-run checks for k0s:
 - Remove the need for a terraform module in the conformance tests terraform code.
 - Stop using smoke infrastruction (go tests) for running sonobuoy.
 - Use k0sctl for deploying the k0s cluster on the virtual hosts (and remove the bash scripts that were previously used)
 - Updated documentation for running full conformance tests on k0s.

Fixes #1831

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test updated

## Checklist:


inttest/sonobuoy/terraform/k0sctl.tf
```diff
@@ -0,0 +1,40 @@
...
+      }
+    }
+  }
+}
+
+// Save the private key to filesystem
+resource "local_file" "k0sctl_config" {
+  file_permission = "600"
+  filename        = format("%s/%s", path.module, "k0sctl.yaml")
+  content         = replace(yamlencode(local.k0sctl_tmpl), "/((?:^|\n)[\\s-]*)\"([\\w-]+)\":/", "$1$2:")
```
jnummelin, line 39: This feels overly complex, why do we need to do such replaces?

On most other cases with TF & k0sctl, we do something like this:
```
output "k0s_cluster" {
    value = yamlencode(local.k0s_tmpl)
}
```
which we can feed into k0sctl:
```
terraform output -raw k0s_cluster | k0sctl apply --config -
```
IMO this is bit cleaner than storing the (unneeded) temp k0sctl.yaml file and all the possible replaces etc. like above

inttest/sonobuoy/terraform/main.tf
```diff
@@ -2,25 +2,38 @@ provider "aws" {
   region = "eu-west-1"
 }
 
+# variable "sonobuoy_version" {
+#   type    = string
+#   default = "0.53.2"
+# }
```
twz123, line 8: Should this stay in here?

inttest/sonobuoy/terraform/variables.tf
```diff
@@ -0,0 +1,31 @@
...
+  type    = number
+  default = 1
+}
+
+variable "worker_count" {
+  type    = number
+  default = 2
+}
+
+variable "cluster_flavor" {
```
twz123, line 15: Wouldn't be "instance_type" a more idiomatic name for this?

twz123, line 15: Wouldn't be "instance_type" a more idiomatic name for this?

inttest/sonobuoy/terraform/k0sctl.tf
trawler: I feel the file-based approach is better for when needing to run the conformance test manually. it's easier to see what k0sctl is actually doing. The regex replace here is used to output a better looking yaml file, since terrafor's yamlencode outputs a yaml that has all values and keys quoted. I felt the output after the regex replacement looks more like our standard yaml configuration files (see https://github.com/hashicorp/terraform/issues/23322).

.github/workflows/release.yml
```diff
@@ -474,15 +491,16 @@ jobs:
...
       - name: Clean-Up Environment
         env:
           K0S_VER: ${{ needs.release.outputs.tag_name }}
         run: |
-          export TF_VAR_k0s_version=${K0S_VER}
+          export TF_VAR_k0s_binary_path="${{ github.workspace }}/k0s"
+          export TF_VAR_k0s_version=${KOS_VER}
 
           terraform destroy -auto-approve
         if: ${{ always() }}
```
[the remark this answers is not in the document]
trawler: not always. we've had cases where launch test environment broke half-way, and allowing the cleanup step to run anyway helps to remove any infrastructure that might be left from a broken run.

inttest/sonobuoy/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+```shell
+➜ k0sctl apply -c k0sctl.yaml
+
+⠀⣿⣿⡇⠀⠀⢀⣴⣾⣿⠟⠁⢸⣿⣿⣿⣿⣿⣿⣿⡿⠛⠁⠀⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀█████████ █████████ ███
+⠀⣿⣿⡇⣠⣶⣿⡿⠋⠀⠀⠀⢸⣿⡇⠀⠀⠀⣠⠀⠀⢀⣠⡆⢸⣿⣿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀███          ███    ███
+⠀⣿⣿⣿⣿⣟⠋⠀⠀⠀⠀⠀⢸⣿⡇⠀⢰⣾⣿⠀⠀⣿⣿⡇⢸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠀███          ███    ███
+⠀⣿⣿⡏⠻⣿⣷⣤⡀⠀⠀⠀⠸⠛⠁⠀⠸⠋⠁⠀⠀⣿⣿⡇⠈⠉⠉⠉⠉⠉⠉⠉⠉⢹⣿⣿⠀███          ███    ███
+⠀⣿⣿⡇⠀⠀⠙⢿⣿⣦⣀⠀⠀⠀⣠⣶⣶⣶⣶⣶⣶⣿⣿⡇⢰⣶⣶⣶⣶⣶⣶⣶⣶⣾⣿⣿⠀█████████    ███    ██████████
+k0sctl 0.0.0 Copyright 2021, k0sctl authors.
```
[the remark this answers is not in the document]
trawler: ?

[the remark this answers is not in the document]
twz123: I was trying to refer to `k0sctl 0.0.0` :smiley:

inttest/sonobuoy/terraform/k0sctl.tf
twz123: Given that the regex is rather a cosmetic thing, I'm not sure if it's carrying its own weight. It's already on the verge of being intimidating :smiling_face_with_tear:

inttest/sonobuoy/README.md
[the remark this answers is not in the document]
trawler: I guess that's what happens when you build k0sctl from latest 🤷‍♀️

.github/workflows/check-network.yaml
```diff
@@ -59,53 +67,92 @@ jobs:
...
-          eval `ssh-agent -s`
-
-          terraform apply -auto-approve
-
-      - name: Bindata
+      - name: Cache embedded binaries
         uses: actions/cache@v3
-        id: generated-bindata
         with:
+          key: ${{ runner.os }}-embedded-bins-linux-${{ hashFiles('embedded-bins/**/*') }}
```
twz123, line 73: Not sure if we are okay with this cross-workflow cache sharing. If yes, we need to keep in mind that stuff built by the usual branch builds may end up here, and vice versa stuff that's been built here will potentially end up in branch builds. Should be _relatively_ safe, though, since we're hashing across everything below embedded-bins and there's no crucial variable overriding going on, neither in the Makefile nor in the workflows.

.github/workflows/check-network.yaml
```diff
@@ -59,53 +67,92 @@ jobs:
...
+
+          terraform apply -auto-approve
+
+      - name: Create k0s Cluster using k0sctl
+        id: k0sctl
+        run: |
+          # download k0sctl
+          curl --silent -L "https://github.com/k0sproject/k0sctl/releases/download/v${KOSCTL_VERSION}/k0sctl-linux-x64" -o k0sctl
+
+          ./k0sctl apply -c k0sctl.yaml
```
twz123, line 140: Don't we need to set the executable flag first?

twz123 (approved):

inttest/Makefile
```diff
@@ -39,17 +40,20 @@ bin/sonobuoy: | bin
...
 
-check-network: bin/sonobuoy .footloose-alpine.stamp
-	K0S_PATH=$(realpath ../k0s) SONOBUOY_PATH=$(realpath bin/sonobuoy) \
-		go test -count=1 -v -timeout 20m github.com/k0sproject/k0s/inttest/sonobuoy/ -run ^TestFootlooseNetworkSuite$
-
-check-network-vm: bin/sonobuoy
-	K0S_PATH=$(realpath ../k0s) SONOBUOY_PATH=$(realpath bin/sonobuoy) \
-		go test -count=1 -v -timeout 30m github.com/k0sproject/k0s/inttest/sonobuoy/ -run ^TestVMNetworkSuite -kubernetes-version=v$(kubernetes_version)
+check-network: bin/sonobuoy
+	$(realpath bin/sonobuoy) run --wait=1200 --plugin=e2e --plugin-env=e2e.E2E_USE_GO_RUNNER=true \
```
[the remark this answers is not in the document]
twz123: Pretty sure `realpath` is not required here. 
```suggestion
	bin/sonobuoy run --wait=1200 --plugin=e2e --plugin-env=e2e.E2E_USE_GO_RUNNER=true \
```

twz123 (approved):

PR #1952, opened 2022-07-28
Fix airgap arm64 build

makhov:
Signed-off-by: Alexey Makhov <amakhov@mirantis.com>

## Description

Fixes #1939

For airgap bundles, we export images for the platform into the tar archive and then try to import them on k0s startup. 
The problem here is that by default during import containerd matches platforms not strictly, so for arm64 it tries to import both arm64 and armv7 images, but there are no armv7 blobs in the archive.

So, the actual fix is to change the `Only` matcher to the `OnlyStrict` one.

There is the PR with a similar fix for `ctr import` command: https://github.com/containerd/containerd/pull/6906

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #1969, opened 2022-08-01
Cache airgap image bundle in all builds

twz123:
## Description

The airgap image bundles are now also checked on ARM builds (see #1952). This means a lot of image pulls from docker.io, which is rate limited. Mitigate this by adding a caching layer for airgap image bundles for the regular amd64 and the custom armv7/arm64 builds.

Embed the TARGET_PLATFORM setting into the Makefile, add the platform name to the file names and add some auto-detection of the architecture. This way, the TARGET_PLATFORM doesn't need to be specified explicitly. Calculate a cache key for the image bundle based on the hash of the Makefile, the airgap image bundle list and the image-bundler Docker image that is used to produce the bundles.

Move the image-bundler Docker image to the hack/ folder and remove the necessity to embed the image list into the image or to mount folders into the container. Let the bundler script fail in case of an error and suppress the very noisy output of the ctr CLI tool.

Change the integration tests to only mount the airgap image bundle if this is explicitly requested by an integration test. Support multiple mount-points, too.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


twz123:
Oops, closed unintentionally :see_no_evil:

.github/workflows/go.yml
```diff
@@ -210,20 +235,28 @@ jobs:
...
           chmod +x k0s
           ./k0s sysinfo
 
+      - name: Cache airgap image bundle
+        id: cache-airgap-image-bundle
+        if: contains(matrix.smoke-suite, 'airgap')
+        uses: actions/cache@v3
+        with:
+          key: airgap-image-bundle-linux-amd64.tar-${{ needs.build.outputs.airgap-image-bundle-hash-linux }}
+          path: airgap-image-bundle-linux-amd64.tar
```
makhov, line 249: Can't we save the bundle under the `image-bundle/bundle.tar` path and keep the other code as is?

trawler: I personally think this change makes the code easier and clearer to follow by depicting exactly which image bundle we are using.

trawler (approved):

makhov (approved):

PR #2001, opened 2022-08-05
Authenticate Docker Hub pulls on self-hosted build jobs.

jnummelin:
By authing we can double the rate limits and hopefully will not hit limits too often.

## Description

Our self hosted runner have been hitting Docker Hub rate limits quite often in last week or so. By authenticating we will double the rate limits and thus hopefully will not hit limits too often.


jnummelin:
We don't get secrets into workflows triggered from forks. Thus we need some other way to do this.

PR #2014, opened 2022-08-09
Fix k0s failure due to partial write of embedded binaries.

### LINKED ISSUE #1989: K0s does not recover if embedded binaries are written only partially to a disk

### Version

v1.23.8+k0s.0

### Platform

```shell
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

### What happened?

k0s wrote empty (or partial) binary files to disk which blocks k0s from starting properly (if for example kubelet binary is broken). Restart does not help because these broken binaries do have correct timestamp. Only way to recover is manual removal of broken binaries and then k0s restart.

### Steps to reproduce

1.
2.
3.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Screenshots and logs

_No response_

### Additional context

_No response_

makhov:

## Description

Fixes #1989

Writes the original size of the embedded files to the static assets and unpacks files if the size doesn't match.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


Makefile
```diff
@@ -14,7 +14,7 @@ else
   endif
   HOST_ARCH := amd64
 endif
-undefine HOST_HARDWARE
+unexport HOST_HARDWARE
```
twz123, line 17: That's interesting. What's the reason for this?

pkg/assets/stage.go
```diff
@@ -29,11 +29,11 @@ import (
 
 // EmbeddedBinaryNeedsUpdate returns true if the provided embedded binary file should
 // be updated. This determination is based on the modification times of both the provided
-// executable, and embedded binary executable. It is expected that the embedded binary
+// executable, and embedded binary executable, and on the binary size. It is expected that the embedded binary
```
twz123, line 32: Maybe this:

> This determination is based on the modification times and file sizes of both the provided executable and the embedded executable.

twz123 (approved):
Would be cool if we'd have some tests for this.

Makefile
makhov: I had an old version of the GNU make and it didn't work. Changed it back

pkg/assets/stage.go
makhov: Done

Makefile
twz123: Ah, good to know. Technically, the else/undefine block isn't even necessary. So if that remains a problem we could also just remove it completely.

twz123 (approved):

Issue #2011, opened 2022-08-08
Cluster Deployment using Hostnames with Capitial Letters Breaks during Konnectivity Service Installation Stage


Renegade-Master:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Version

1.24.3+k0s.0

### Platform

```shell
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
Architecture:   x86_64
Hostname:       renegade-VB <-- IMPORTANT

---

Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04
Codename:       jammy
Architecture:   aarch64
Hostname:       renegade-desktop
```

### What happened?

When attempting to install a K0s Cluster using Hosts with capital letters in their Hostnames, installation will succeed, and pods will be schedulable. However, it is not possible to retrieve logs from pods, nor to exec into them, nor is it possible to port-forward any services.

### Steps to reproduce

1. Create a Multi-Node Cluster with at least one Host using capital letters in their Hostname.
2. Wait for the installation to fully apply.
3. Attempt to exec into a pod, or get the logs from a pod, or to port-forward a Service.

### Expected behavior

A Multi-Node Cluster should be capable of integrating Hosts without the requirement that their Hostname is lowercase.

### Actual behavior

When attempting any of the aforementioned three activities (retrieve logs from pods, exec into them, port-forward any services), they are all met with the same error:
```shell
/var/run/k0s/konnectivity-server/konnectivity-server.sock: connect: no such file or directory
``` 

And true enough, the `/var/run/k0s/konnectivity-server/` does exist:
```shell
$ ls -lAuhF /var/run/k0s/
total 36K
drwx--x--x 5 root                root  100 Aug  8 18:33 containerd/
-rw-r--r-- 1 root                root    5 Aug  8 18:33 containerd.pid
srw-rw---- 1 root                root    0 Aug  8 18:33 containerd.sock=
srw-rw---- 1 root                root    0 Aug  8 18:33 containerd.sock.ttrpc=
-rw-r--r-- 1 root                root    5 Aug  8 18:33 etcd.pid
[...]

PR #2048, opened 2022-08-16
[Backport release-1.23] Use "object name safe" hostnames for leases

twz123:
Backport of #2044 to `release-1.23`. See #2011.

### DIFF
--- a/inttest/Makefile.variables
+++ b/inttest/Makefile.variables
@@ -28,4 +28,5 @@ smoketests := \
 	check-upgrade \
 	check-workerrestart \
 	check-kubectl \
-	check-k0sctl
+	check-k0sctl \
+	check-capitalhostnames \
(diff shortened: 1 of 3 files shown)


twz123:
@jnummelin Oh. There's even more to it than just removing the autopilot parts.

 `k8s.io/kubernetes/pkg/util/node` is not part of the official Go API. We don't import k/k directly in 1.23 and below.

jnummelin:
maybe we just do our own helper func for 1.23 and below with the same semantics as `k8s.io/kubernetes/pkg/util/node` has?

twz123:
d0f69643c60ae3caab82ba8255e2a9b4630d693e looks good. (I cannot approve my own PRs :smile:)

s0j (approved):

PR #2059, opened 2022-08-18
Minor component interface cleanup

### LINKED ISSUE #1844: Disassemble the Component interface

The `component.Component` interface is the backbone of k0s's internal structure:

https://github.com/k0sproject/k0s/blob/71a2c707b7dd64a622160c27c0b72e1bc574a775/pkg/component/component.go#L25-L30

It serves as a lifecycle abstraction for all the different things in k0s. However, its contract is not very well defined, which makes it hard to judge what to do when implementing it.

1. What is the anticipated order in which the lifecycle methods are being called and what are the expectations about concurrency safety?
    There has been some effort in documenting the lifecycle in k0sproject/k0s#1657, but it's still not clear enough. Are components expected to be concurrency safe, or is it the responsibility of its users to ensure that access to the lifecycle methods is synchronized?
2. Many components don't need `Init`, `Stop` or `Healthy` implementations.
    Some even don't need a `Run` method, since they only act on `Reconcile`. Especially the `Healthy` method is only used in the [`Manager`](https://sourcegraph.com/github.com/k0sproject/k0s@v1.23.6+k0s.2/-/blob/pkg/component/component.go?L29:2#tab=references) when [starting components](https://github.com/k0sproject/k0s/blob/71a2c707b7dd64a622160c27c0b72e1bc574a775/pkg/component/manager.go#L90-L93).
3. Context handling and its interaction with `Init`/`Run`/`Stop` behavior.
    Currently, both the `Init` and the `Run` method take a context parameter. It is not entirely clear how those contexts should be used, given that there's also the `Stop` method. Are components expected to stop by themselves if any of those contexts is cancelled (a.k.a the ["Merge"](https://github.com/golang/go/issues/36503) problem)? If yes, why does the `Stop` method exist?

1. The key purpose of the `Component` interface is to drive the lifecycle of multiple components via the `Manager`. It is responsible to call the "main" lifecycle methods `Init`, `Run`, `Stop` and `Healthy`. It does it in a sequential manner w.r.t. each individual component. For those methods, it seems reasonable to document that they don't need to be concurrency safe. Then there's the `Reconcile` method. It's special in a way that it's not really a lifecycle method, but a special method to be called "out of band" on each registered component. Some more complex components have even more public methods, all of them may be called at different times, potentially concurrently. Let's clearly document that: as long as a component only exposes the fundamental lifecycle methods, it doesn't need to care about concurrency. But as soon as it's doing more than that, it should.
2. Instead of combining all of the methods into one interface, let's split them up into individual interfaces. There's already the specialized `ReconcilerComponent` interface for those components that require/support reconciliation. The `Manager` simply uses type assertions to figure out if a component implements it or not. How about having an `Initializable` and `Stoppable` interface for those components which support those lifecycle phases, and let the `Manager` use type assertions to figure it out. This would remove a lot of empty function implementations from the codebase, and would also obliterate the question of whether a component should actually fail if `Stop` was called before `Run` when it cannot be stopped anyway (same for `Run` before `Init`, when `Init` is a no-op). This approach is also extensible, when there are more lifecycle phases to be added in the future, there could be just a new interface for those. Not all existing components would need to implement it.
    Concerning the `Healthy` method: Given that only the `Manager` uses it when starting components, I would argue that we can remove it completely. Instead, for those components that have a reasonable `Healthy` implementation, it should be embedded into the `Run` method. `Run` can then block until the component becomes healthy. The timeout loop implemented in the `Manager`'s `waitForHealthy` method could be transformed into a context that gets cancelled if `Run` doesn't return in time.

     ```go
    type Initializable interface {
    	Init(context.Context) error
    }

    type Component interface {
    	Run(context.Context) error
    }
    
    type Stoppable interface {
    	Stop() error
    }

    type Reconcilable interface {
    	Reconcile(context.Context, *v1beta1.ClusterConfig) error
    }
    ```

3. Let's define the context passed to `Init` as a short-lived one that may get cancelled after `Init` returns. The `Run` method would better be called `Start`, since it's actually starting something, not running it (i.e. it's not blocking until stopped). The stop method is currently being used to block the `Manager` until a component is actually stopped. It combines both the cancellation request and the blocking until cancelled into one method. How about using the context passed to `Start` as the cancellation signal, and returning an error channel that may be used to block until the component has actually been stopped? This would require us to make the `Manager` implementation a bit more elaborate, but It would make implementing Components easier. I consider this a plus, since there's only one `Manager`, but many components.

    ```go
    type Initializable interface {
    	Init(context.Context) error
    }

    type Component interface {
    	Start(context.Context) (<-chan error, error)
    }

    type Reconcilable interface {
    	Reconcile(context.Context, *v1beta1.ClusterConfig) error
    }
    ```

_No response_

### Additional context

_No response_

makhov:

## Description

`Run` renamed to `Start` since it's much closer to what it actually does.
Removed embedding the `Component` from the `Reconciler` interface since it's not needed  

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -83,7 +83,7 @@ func (m *Manager) Init(ctx context.Context) error {
 }
 
 // Run runs the Manager
-func (m *Manager) Run(_ context.Context) error {
+func (m *Manager) Start(_ context.Context) error {
 	return nil
 }
 
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -24,10 +24,10 @@ import (
 
 // Component defines the lifecycle of managed components.
 //
-//	Created ――――――――――――――――――――――►(Stop)―――╮
-//	╰―(Init)―► Initialized ―――――――►(Stop)――╮│
-//	           ╰―(Run)―► Running ―►(Stop)―╮││
-//	              ╭(Healthy)╯▲            ▼▼▼
+//	Created ――――――――――――――――――――――►(Stop)―――――╮
+//	╰―(Init)―► Initialized ―――――――►(Stop)――――╮│
+//	           ╰―(Start)―► Running ―►(Stop)―╮││
+//	              ╭(Healthy)╯▲              ▼▼▼
 //	              ╰――――――――――╯         ╭► Terminated╮
 //	                                   ╰――――(Stop)――╯
 type Component interface {
@@ -38,12 +38,12 @@ type Component interface {
 	// component's lifecycle. Init must be called only once.
 	Init(context.Context) error
 
-	// Run starts the component. When Run returns, the component is executing in
+	// Start starts the component. When Start returns, the component is executing in
 	// the background. Run may be called only once after Init. If the component
-	// is a ReconcilerComponent, a call to Reconcile may be required before Run
+	// is a Reconciler, a call to Reconcile may be required before Start
 	// can be called. The given context is not intended to replace a call to
-	// Stop when cancelled. It's merely used to cancel the component's startup.
-	Run(context.Context) error
+	// Stop when canceled. It's merely used to cancel the component's startup.
+	Start(context.Context) error
 
 	// Stop stops this component, potentially cleaning up any temporary
 	// resources attached to it. Stop itself may be called in any lifecycle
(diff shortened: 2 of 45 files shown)


jnummelin:
I don't think this completely "fixes" #1844 yet. Yes, it's partially implementing what has been specced in there but not fully, thus maybe remove that keyword from the PR description so we don't accidentally close the issue yet. 😄

makhov:
Yeah, removed it, thanks. But I'm not really sure, that we need to do anything else here now. I left the comment, describing my thoughts: https://github.com/k0sproject/k0s/issues/1844#issuecomment-1219157095

twz123 (approved):

PR #2153, opened 2022-09-12
Add validation message for the TunneledNetworkingMode

soider:
Config field spec.api.tunneledNetworkingMode set to true and
node role equals to controller+worker mode leads to running
konnectivity-agent and kube-api on the same port.
Due to the fact, that konnectivity-agent runs in host network, in case
of enabled tunneledNetworkingMode, it leads to panic during the
bootstrap.
There is no way to validate spec.api with respect to node role,
so it is easier to just disallow running kube-api with default port
in case if we have tunneledNetworkingMode enabled.

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/inttest/tunneledkas/suite_test.go
+++ b/inttest/tunneledkas/suite_test.go
@@ -35,6 +35,7 @@ type Suite struct {
 const config = `
 spec:
   api:
+    port: 7443
     tunneledNetworkingMode: true
 `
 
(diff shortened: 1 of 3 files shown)


trawler (changes requested):
IMO, maybe it would be a better user experience if simply add a function that checks if `a.TunneledNetworkingMode == true` and then pick an alternative kube-api port (7443, for example), if no other port was defined.
Seems to me that unless we either automate this process or highlight this limitation in the documentation - any user who tries out TunneledNetworkingMode will fail in this check.

soider:
>Seems to me that unless we either automate this process or highlight this limitation in the documentation - any user who tries out TunneledNetworkingMode will fail in this check.

That's the point to have it failed here with validation message rather than later with cryptic bind error in logs.

About the random port: I also thought that way but I am not sure, if it is a good solution. Sounds unsafe to run kube-api on random port.

makhov (approved):

jnummelin:
> Sounds unsafe to run kube-api on random port.

I agree. Also we'd need to be able to pick the same port always in case k0s or the node restarts.

PR #2201, opened 2022-09-22
Early logging configuration

twz123:
## Description

The long-running subcommands `controller`, `worker` and `api` use stdout and info logging by default. The current setup activates those settings quite late in the Cobra pipeline, potentially logging to the wrong destination, or slurping away interesting debug logs.

Fix this by setting the defaults early for those subcommands via Cobra's `PersistentPreRun` facility. Don't call `config.GetCmdOpts()` unnecessarily in the root command's `PersistentPreRun`, but simply use the global variables directly.

This relates to #2189 in order to help making problems in the early execution stages more visible.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -63,15 +63,13 @@ func NewAPICmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "api",
 		Short: "Run the controller API",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			c := command{CLIOptions: config.GetCmdOpts()}
-
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			logrus.SetOutput(os.Stdout)
-			if !c.Debug {
-				logrus.SetLevel(logrus.InfoLevel)
-			}
-
-			return c.start()
+			logrus.SetLevel(logrus.InfoLevel)
+			return config.CallParentPersistentPreRun(cmd, args)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return (&command{CLIOptions: config.GetCmdOpts()}).start()
 		},
 	}
 	cmd.SilenceUsage = true
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -71,7 +71,7 @@ func NewControllerCmd() *cobra.Command {
 	$ k0s controller --token-file [path_to_file]
 	Note: Token can be passed either as a CLI argument or as a flag`,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			logrus.SetOutput(os.Stdout)
+			logrus.SetOutput(cmd.OutOrStdout())
 			logrus.SetLevel(logrus.InfoLevel)
 			return config.CallParentPersistentPreRun(cmd, args)
 		},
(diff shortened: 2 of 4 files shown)


cmd/api/api.go
```diff
@@ -63,15 +63,13 @@ func NewAPICmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "api",
 		Short: "Run the controller API",
-		RunE: func(cmd *cobra.Command, args []string) error {
-			c := command{CLIOptions: config.GetCmdOpts()}
-
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			logrus.SetOutput(os.Stdout)
```
kke, line 67: (this should be `logrus.SetOutput(cmd.OutOrStdout())`)

cmd/controller/controller.go
```diff
@@ -70,14 +70,14 @@ func NewControllerCmd() *cobra.Command {
 	or CLI flag:
 	$ k0s controller --token-file [path_to_file]
 	Note: Token can be passed either as a CLI argument or as a flag`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			logrus.SetOutput(os.Stdout)
```
kke, line 74:
```suggestion
			logrus.SetOutput(cmd.OutOrStdout())
```

cmd/worker/worker.go
```diff
@@ -53,14 +53,13 @@ func NewWorkerCmd() *cobra.Command {
 	or CLI flag:
 	$ k0s worker --token-file [path_to_file]
 	Note: Token can be passed either as a CLI argument or as a flag`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			logrus.SetOutput(os.Stdout)
```
kke, line 57:
```suggestion
			logrus.SetOutput(cmd.OutOrStdout())
```

kke: Hmm, shouldn't it be enough to do these in the root prerun?

cmd/worker/worker.go
```diff
@@ -53,14 +53,13 @@ func NewWorkerCmd() *cobra.Command {
 	or CLI flag:
 	$ k0s worker --token-file [path_to_file]
 	Note: Token can be passed either as a CLI argument or as a flag`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			logrus.SetOutput(cmd.OutOrStdout())
+			logrus.SetLevel(logrus.InfoLevel)
```
kke, line 58: If root's prerun does this, I don't think the subcommands need to do it on their own

kke (approved):

cmd/worker/worker.go
twz123: Bummer that I've forgotten that I've set it to auto-merge :see_no_evil: Will fix this as well.

twz123: Ah, no. I remember why it's there: Those three subcommands have info as their default level, whereas all others have warn as default. Setting the level before the call to the parent persistent prerun saves some extra logic to enable debug logging when requested.

cmd/api/api.go
twz123: #2206

PR #2205, opened 2022-09-27
Fix k0s kubectl --kubeconfig flag

kke:

## Description

It was noticed in https://github.com/k0sproject/k0sctl/issues/327 that the `k0s kubectl --kubeconfig` flag does not seem to work.

This PR addresses that.

The `fallbackToK0sKubeconfig` function was going through `args []string`, which at that point should not contain any flags as they have been parsed already. The fallback is now changed to look at the flag value directly through `cmd.Flags().Lookup("kubeconfig")`. The k0s admin kubeconfig path is set using `flag.Value.Set` instead of putting it into os env.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/kubectl/kubectl.go
+++ b/cmd/kubectl/kubectl.go
@@ -130,12 +130,11 @@ func fallbackToK0sKubeconfig(cmd *cobra.Command) error {
 		return nil
 	}
 
-	kubeconfig, envSet := os.LookupEnv("KUBECONFIG")
-	if envSet {
-		_ = os.Unsetenv("KUBECONFIG")
-	} else {
-		kubeconfig = config.GetCmdOpts().K0sVars.AdminKubeConfigPath
+	if _, ok := os.LookupEnv("KUBECONFIG"); ok {
+		return nil
 	}
+
+	kubeconfig := config.GetCmdOpts().K0sVars.AdminKubeConfigPath
 
 	// verify that k0s's kubeconfig is readable before pushing it to the env
 	if _, err := os.Stat(kubeconfig); err != nil {
--- a/cmd/kubectl/kubectl.go
+++ b/cmd/kubectl/kubectl.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
-	"strings"
 	"syscall"
 
 	"github.com/k0sproject/k0s/pkg/config"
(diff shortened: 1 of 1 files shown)


cmd/kubectl/kubectl.go
```diff
@@ -119,34 +118,32 @@ func NewK0sKubectlCmd() *cobra.Command {
...
-	for _, arg := range args {
-		if arg == "--" {
-			// no more options
-			break
-		}
-		if arg == "--kubeconfig" || strings.HasPrefix(arg, "--kubeconfig=") {
-			// kubeconfig set via args, no need to check if k0s kubeconfig is readable
-			return nil
-		}
+func fallbackToK0sKubeconfig(cmd *cobra.Command) error {
```
twz123, line 121: I think we don't need to set the flag if the environment variable is present. Kubectl will do this on its own.

An alternative proposal:

```go
func fallbackToK0sKubeconfig(cmd *cobra.Command) error {
	kubeconfigFlag := cmd.Flags().Lookup("kubeconfig")
	if kubeconfigFlag == nil {
		return fmt.Errorf("kubeconfig flag not found")
	}

	if kubeconfigFlag.Changed {
		return nil
	}

	if os.Getenv("KUBECONFIG") != "" {
		return nil
	}

	kubeconfig := config.GetCmdOpts().K0sVars.AdminKubeConfigPath

	// verify that k0s's kubeconfig is readable before setting the flag
	if _, err := os.Stat(kubeconfig); err != nil {
		return fmt.Errorf("cannot stat k0s kubeconfig, is the server running?: %w", err)
	}

	if err := kubeconfigFlag.Value.Set(kubeconfig); err != nil {
		return fmt.Errorf("failed to set kubeconfig flag: %w", err)
	}
	return nil
}
```

cmd/kubectl/kubectl.go
```diff
@@ -119,34 +118,32 @@ func NewK0sKubectlCmd() *cobra.Command {
...
+	} else {
+		kubeconfig = config.GetCmdOpts().K0sVars.AdminKubeConfigPath
 	}
-	file.Close()
 
-	if err := os.Setenv("KUBECONFIG", kubeconfig); err != nil {
-		return fmt.Errorf("failed to set k0s kubeconfig as default: %w", err)
+	// verify that k0s's kubeconfig is readable before pushing it to the env
+	if _, err := os.Stat(kubeconfig); err != nil {
+		return fmt.Errorf("cannot stat k0s kubeconfig, is the server running?: %w", err)
```
twz123, line 142: The file existence check was originally only performed when the k0s admin kubeconfig was used. Now it's also used for the environment variable. The env variable can actually hold a [list of kubeconfig files](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#the-kubeconfig-environment-variable), which gets merged. Even if there's not a single existing file in that list, it will still fallback to the default config (localhost:6443), IIIRC.

I'd only do the extra existence check if we're setting the k0s default.

twz123 (approved):

PR #2234, opened 2022-10-06
Add simple Kubernetes watch facility

twz123:
## Description

Watcher offers a convenient way of watching Kubernetes resources. An ephemeral alternative to Reflectors and Indexers, useful for one-shot tasks when no caching is required. It performs an initial list of all the resources and then starts watching them.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/kubernetes/watch/watcher.go
+++ b/pkg/kubernetes/watch/watcher.go
@@ -1,0 +1,318 @@
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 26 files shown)


pkg/kubernetes/watch/watcher.go
```diff
@@ -0,0 +1,318 @@
...
+		if ok, err := listCallback(items); err != nil {
+			return noRetry{err}
+		} else if ok {
+			return nil
+		}
+
+		initialResourceVersion = resourceVersion
+	}
+
+	watcher, err := watchtools.NewRetryWatcher(initialResourceVersion, &cache.ListWatch{
```
s0j, line 209: This was used in the original autopilot code, but had to be abandoned in favor for the polling that you're removing due to extremely hard-to-reproduce scenarios where the watcher wouldn't trigger (despite data being there).

It is the better direction, so I'm hopeful that this diff (+ time) has more success than I did (PEBKAC).

twz123: Sounds a little bit concerning. I thought it would be good to use the watcher everywhere, so it gets "battle tested", sorta. Are there some references left to the original watching code, maybe in the git history? I would love to have a look at it. Don't want to shoot ourselves into the foot with this change.

Do you have some scenarios at hand that triggered the bad behavior?

PR #2242, opened 2022-10-07
Correct error logging in ClusterConfig reconciler

twz123:
## Description

The wrong error was logged when the cluster config wasn't found. Fix that and polish the other log statements.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/clusterConfig.go
+++ b/pkg/component/controller/clusterConfig.go
@@ -102,35 +102,36 @@ func (r *ClusterConfigReconciler) Init(_ context.Context) error {
 func (r *ClusterConfigReconciler) Start(ctx context.Context) error {
 	if r.configSource.NeedToStoreInitialConfig() {
 		// We need to wait until we either succees getting the object or creating it
-		err := wait.Poll(1*time.Second, 20*time.Second, func() (done bool, err error) {
+		err := wait.Poll(1*time.Second, 20*time.Second, func() (bool, error) {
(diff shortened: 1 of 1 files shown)


pkg/component/controller/clusterConfig.go
```diff
@@ -102,35 +102,36 @@ func (r *ClusterConfigReconciler) Init(_ context.Context) error {
...
-				} else {
-					r.log.Errorf("error getting cluster-config: %v", err)
-					return false, nil
-				}
+			if _, err := r.configClient.Get(timeoutCtx, constant.ClusterConfigObjectName, getOpts); err == nil {
+				return true, nil
+			} else if !errors.IsNotFound(err) {
+				r.log.WithError(err).Error("Failed to get cluster configuration")
+				return false, nil
 			}
```
s0j, line 115: I find the structure of the original code easier to read. Specifically, I don't have to mentally jump around the double-negative `!errors.IsNotFound()`, and introducing an `else if` makes me feel that there is another condition that is missing.

What about a combination of the two -- keeping the structure, but switching to the one-liner error checks and the updated logging messages?

twz123: I think it'd be clearer if we just try to create the cluster config and check for `errors.IsAlreadyExists` instead. Saves one API call...

twz123:
Closing in favor of #2248.

PR #2255, opened 2022-10-12
Added airgap bundle support for `tool`

s0j:
## Description

For both the AWS `ha` and `havpc` commands, a collection of airgap-specific arguments have been added which allow for seeding a k0s cluster via an airgap bundle.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/hack/tool/cmd/aws/havpc/havpc.go
+++ b/hack/tool/cmd/aws/havpc/havpc.go
@@ -41,6 +41,8 @@ type options struct {
 	Workers               int
 	K0sBinary             string
 	K0sVersion            string
+	K0sAirgapBundle       string
+	K0sAirgapBundleConfig string
 	K0sUpdateBinary       string
 	K0sUpdateAirgapBundle string
 }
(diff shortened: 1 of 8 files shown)


hack/tool/cmd/aws/havpc/havpc.go
```diff
@@ -146,14 +150,22 @@ func newCommandDestroy() *cobra.Command {
...
-			cmd.SilenceUsage = true
+					return nil
+				},
+				Destroy: func(ctx context.Context) error {
+					return provider.ClusterHAVpcDestroy(ctx, opts.VpcId, opts.SubnetIdx, opts.ClusterName, opts.K0sBinary, opts.K0sUpdateBinary, opts.K0sVersion, opts.K0sAirgapBundle, opts.K0sAirgapBundleConfig, opts.K0sUpdateAirgapBundle, opts.Controllers, opts.Workers, opts.Region)
+				},
+			}
 
-			return provider.ClusterHAVpcDestroy(ctx, opts.VpcId, opts.SubnetIdx, opts.ClusterName, opts.K0sBinary, opts.K0sUpdateBinary, opts.K0sVersion, opts.K0sUpdateAirgapBundle, opts.Controllers, opts.Workers, opts.Region)
+			return provision.Deprovision(context.Background(), config)
```
twz123, line 168:
```suggestion
			return provision.Deprovision(cmd.Context(), config)
```

hack/tool/terraform/aws/commands/ha/outputs.tf
```diff
@@ -27,6 +37,7 @@ locals {
             api = {
               externalAddress = module.k0sinfra.loadbalancer_dns
             }
+            images = var.k0s_airgap_bundle != "" ? local.k0s_spec_images_airgap : null
```
twz123, line 40: I wonder if it would be better to have the possibility to have a real overlay for the k0s configuration This is not really airgap specific, as one could just use that image setting to use a self-hosted registry or any other things.

What about having a generic thing that can be used to pass in k0s configuration?

```tf
spec = merge(
  var.k0s_config_spec,
  { api = { externalAddress = module.k0sinfra.loadbalancer_dns, }, },
)


```

s0j: Right, that is the end goal, and would allow for us to makesome really obscure testable k0s cluster configurations.

As it stands now, all of the autopilot inttests as well as the external build don't need this level of customization, so I'm going to focus on this minimal functionality first.

hack/tool/cmd/aws/havpc/havpc.go
s0j: Thanks, missed this.

hack/tool/terraform/aws/commands/ha/outputs.tf
twz123: The "airgap" suffix in the config name just feels a bit off to me here.

twz123 (approved):

PR #2260, opened 2022-10-13
Calico iptables backend config parameter

makhov:

## Description

A new config parameter is added to set which iptables backend Calico should use.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
```diff
@@ -64,6 +67,7 @@ func DefaultCalico() *Calico {
 		Overlay:                 "Always",
 		IPAutodetectionMethod:   "",
 		IPv6AutodetectionMethod: "",
+		IPTablesBackend:         "auto",
```
jnummelin, line 70: Calico docs says the default is `[Default: Auto]`, note the capital `A`. dunno if it matter though but to be on safe side I'd change it to the exact same in k0s side. See https://projectcalico.docs.tigera.io/reference/felix/configuration#iptables-dataplane-configuration

makhov: Yeah, fixed. Thanks!

jnummelin:
This does makes me wonder if we'd need to generic way for a user to pass in ANY set of env variables. there's quite a few envs one can configure with calico

makhov:
Yeah, we definitely need it. The question is whether to do it in this PR or in a separate one.

My first idea was that since we set `FELIX_IPTABLESBACKEND` in the calico-node daemonset, we need to have a parameter to change the value of the var.
On the other hand, the default value is `Auto`, so we can just remove it and let the user decide to set `FELIX_IPTABLESBACKEND` or not.

jnummelin:
> The question is whether to do it in this PR or in a separate one.

Maybe better to do it in separate PR. This PR might need to get backported to release branches.

jnummelin:
Related https://github.com/projectcalico/calico/issues/5760

makhov:
Closing, since it can be done by setting env var (needs #2269 to be merged)

PR #2269, opened 2022-10-14
Passing env vars to the calico-node daemonset

makhov:

## Description

Allows setting any environment variable to the calico-node daemonset. 

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


docs/configuration.md
```diff
@@ -154,15 +154,16 @@ spec:
...
 | `ipAutodetectionMethod` | Use to force Calico to pick up the interface for pod network inter-node routing (default: `""`, meaning not set, so that Calico will instead use its defaults). For more information, refer to the [Calico documentation](https://docs.projectcalico.org/reference/node/configuration#ip-autodetection-methods).                                                                                           |
+| `envVars`               | Map of key-values (strings) for any calico-node [environment variable](https://docs.projectcalico.org/reference/node/configuration#ip-autodetection-methods).                                                                                                                                                                                                                                              |
```
twz123, line 166: Since k0s specifies quite a bunch of the variables already out of the box, it might be good to mention this here. Users may break the networking when overriding things here that k0s expects to be different. (Assuming that the envVars here override the k0s defaults.)

pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
```diff
@@ -23,6 +23,9 @@ type Calico struct {
 	// Enable wireguard-based encryption (default: false)
 	EnableWireguard bool `json:"wireguard"`
 
+	// Environment variables to configure Calico node (see https://docs.projectcalico.org/reference/node/configuration)
+	EnvVars map[string]string `json:"envVars,omitempty"`
```
twz123, line 27: Would be good to have some validation for these à la [`ValidateEnv`](https://pkg.go.dev/k8s.io/kubernetes@v1.25.3/pkg/apis/core/validation#ValidateEnv). Maybe there's a way to funnel these values through that function.

docs/configuration.md
twz123, line 166: Would make sense to add some notes about the allowed format. Something along the lines of "The values must be suitable to be used as Container Environment Variables". Might be quite obvious, but still. Would be cool if we could add some link to upstream docs... Unfortunately I failed to find some comprehensive docs about this real quick. The comments on the k8s API type seem a bit too technical and detailed: https://github.com/kubernetes/kubernetes/blob/v1.25.3/staging/src/k8s.io/api/core/v1/types.go#L1988-L2003

Ah, there's also variable name expansion involved...

twz123 (dismissed):

inttest/calico/calico_test.go
```diff
@@ -44,6 +44,17 @@ func (s *CalicoSuite) TestK0sGetsUp() {
 	err = s.WaitForNodeReady("worker1", kc)
 	s.NoError(err)
 
+	calicoDaemonset, err := kc.AppsV1().DaemonSets("kube-system").Get(context.TODO(), "calico-node", v1.GetOptions{})
+	s.NoError(err)
```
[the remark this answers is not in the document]
makhov: Done

static/manifests/calico/DaemonSet/calico-node.yaml
```diff
@@ -231,6 +231,10 @@ spec:
               value: "true"
             - name: FELIX_PROMETHEUSMETRICSENABLED
               value: "true"
+            {{ range $name, $value := .EnvVars }}
```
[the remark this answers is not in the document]
makhov: > Variables below will overwrite the ones specified above them

This one is correct. I've added a comment. Covering it in the inttests will be a bit tricky since we need to check variables inside the container. Also, it's a Kubernetes feature, so I'm not sure _we_ should test it.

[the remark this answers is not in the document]
twz123: We are responsible to keep the correct order for the behavior that k0s desires. But a comment in here would do, I guess.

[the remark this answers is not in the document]
twz123: :+1:

docs/configuration.md
makhov: Added some description on existing variables

makhov: I added a comment, that all vars must be strings. I believe that people in general understand how env vars work and how to use them.

pkg/apis/k0s.k0sproject.io/v1beta1/calico.go
makhov: TBH, I don't see how we could validate here anything. Env var's name and value are just strings, and only strings are allowed here.

twz123: Env vars cannot have arbitrary names. The function I mentioned basically calls this code snippet:

https://github.com/kubernetes/kubernetes/blob/1ffc17e329f9ba3dbe840a7f37513eb8ef074740/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L403-L417

I though about something along those lines:

```go
import (
	"k8s.io/apimachinery/pkg/util/validation"
	"k8s.io/apimachinery/pkg/util/validation/field"
)

var invalidNames []string
for name := range c.EnvVars {
	if msgs := validation.IsEnvVarName(name); len(msgs) > 0 {
		invalidNames = append(invalidNames, name)
	}
}
if invalidNames != nil {
	return field.Invalid(field.NewPath("envVars"), invalidNames, "invalid environment variable names")
}
```

docs/configuration.md
twz123: Env var names in Kubernetes are constrained to `^[-._a-zA-Z][-._a-zA-Z0-9]*$`, for example.

twz123 (dismissed):

twz123 (approved):

PR #2272, opened 2022-10-17
kine healtz check

### LINKED ISSUE #2135: Proper Kine health check missing
We've seen some cases of kine/sqlite failing to startup properly. This shows up as kube-apiserver hitting into issues like:
```
Sep 08 06:39:34 lima-lens-vm k0s[1073]: time="2022-09-08 06:39:34" level=info msg="time=\"2022-09-08T06:39:34Z\" level=error msg=\"error while range on /registry/flowschemas/ /registry/flowschemas/: disk I/O error: no such file or directory\"" component=kine
```

While we do not yet know what is the root cause for this, we've identified that restarting kine does fix the issue.

Thus if we'd have a proper health check for kine, k0s could detect this early and restart kine. Currently kine _seems_ to run fine but kube-api will hit lot of issues as it cannot read/write anything to the DB. This results into k0s restarting kube-api for a while untill it gives up. Finally systemd will restart whole k0s but that currently takes like ~5 mins.

Maybe we could do some similar query as we see kube-api failing with as a health check?

xrefs:
https://github.com/k3s-io/kine/issues/135

soider:

## Description

Health check on startup, try to write and read value through kine etcd interface

Fixes #2135  
## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/kine.go
+++ b/pkg/component/controller/kine.go
@@ -25,6 +25,8 @@ import (
 
 	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
 	"github.com/k0sproject/k0s/pkg/component"
+	"github.com/k0sproject/k0s/pkg/etcd"
+	clientv3 "go.etcd.io/etcd/client/v3"
 
 	"github.com/sirupsen/logrus"
 
@@ -37,14 +39,17 @@ import (
 
 // Kine implement the component interface to run kine
 type Kine struct {
-	Config     *v1beta1.KineConfig
-	gid        int
-	K0sVars    constant.CfgVars
-	supervisor supervisor.Supervisor
-	uid        int
+	Config       *v1beta1.KineConfig
+	gid          int
+	K0sVars      constant.CfgVars
+	supervisor   supervisor.Supervisor
+	uid          int
+	bypassClient *etcd.Client
+	ctx          context.Context
 }
 
 var _ component.Component = (*Kine)(nil)
+var _ component.Healthz = (*Kine)(nil)
 
 // Init extracts the needed binaries
 func (k *Kine) Init(_ context.Context) error {
@@ -82,14 +87,21 @@ func (k *Kine) Init(_ context.Context) error {
 			logrus.Warningf("datasource file %s does not exist", dsURL.Path)
 		}
 	}
+
+	k.bypassClient, err = etcd.NewClientWithConfig(clientv3.Config{
+		Endpoints: []string{fmt.Sprintf("unix://%s", k.K0sVars.KineSocketPath)},
+	})
+	if err != nil {
+		return fmt.Errorf("can't create bypass etcd client: %w", err)
+	}
 	return assets.Stage(k.K0sVars.BinDir, "kine", constant.BinDirMode)
 }
 
 // Run runs kine
-func (k *Kine) Start(_ context.Context) error {
+func (k *Kine) Start(ctx context.Context) error {
 	logrus.Info("Starting kine")
 	logrus.Debugf("datasource: %s", k.Config.DataSource)
-
+	k.ctx = ctx
 	k.supervisor = supervisor.Supervisor{
 		Name:    "kine",
 		BinPath: assets.BinPath("kine", k.K0sVars.BinDir),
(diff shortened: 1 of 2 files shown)


pkg/component/controller/kine.go
```diff
@@ -110,3 +122,25 @@ func (k *Kine) Start(_ context.Context) error {
 func (k *Kine) Stop() error {
 	return k.supervisor.Stop()
 }
+
+const hcKey = "/k0s-health-check"
+const hcValue = "value"
+
+func (k *Kine) Healthy() error {
+	ok, err := k.bypassClient.Write(k.ctx, hcKey, hcValue, 64)
```
twz123, line 130: What about using a value that's (reasonably) unique value for each call to `Healthy`? Say, the current timestamp.

pkg/etcd/client.go
```diff
@@ -142,3 +151,37 @@ func (c *Client) Health(ctx context.Context) error {
...
+func (c *Client) Write(ctx context.Context, key string, value string, ttl int64) (bool, error) {
+
+	leaseResp, err := c.client.Lease.Grant(ctx, ttl)
+
+	if err != nil {
+		return false, fmt.Errorf("can't get TTL lease: %w", err)
+	}
+
+	txnResp, err := c.client.KV.Txn(ctx).If(
+		notFound(key),
```
twz123, line 165: I reckon this is some "write if absent" thing? Why don't we always overwrite the value? If we want to keep this behavior, I'd suggest to rename the method to `Create`, `PutIfAbsent` or similar and add a note about this to the godoc of the method.

pkg/component/controller/kine.go
```diff
@@ -82,14 +87,21 @@ func (k *Kine) Init(_ context.Context) error {
...
 	return assets.Stage(k.K0sVars.BinDir, "kine", constant.BinDirMode)
 }
 
 // Run runs kine
-func (k *Kine) Start(_ context.Context) error {
+func (k *Kine) Start(ctx context.Context) error {
 	logrus.Info("Starting kine")
 	logrus.Debugf("datasource: %s", k.Config.DataSource)
-
+	k.ctx = ctx
```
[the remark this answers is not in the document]
soider: it does require change interface and I'd rather do it during the interface rework #2273

pkg/component/controller/kine.go
```diff
@@ -110,3 +122,25 @@ func (k *Kine) Start(_ context.Context) error {
...
+const hcKey = "/k0s-health-check"
+const hcValue = "value"
+
+func (k *Kine) Healthy() error {
+	ok, err := k.bypassClient.Write(k.ctx, hcKey, hcValue, 64)
+	if err != nil {
+		return fmt.Errorf("kine-etcd-health: %w", err)
+	}
+	if !ok {
+		logrus.Warningf("kine-etcd-health: health-check value was not written")
```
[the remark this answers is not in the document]
soider: If it is not written it means that ttl for the previous run is not expired, might be logically correct.

pkg/etcd/client.go
```diff
@@ -142,3 +151,37 @@ func (c *Client) Health(ctx context.Context) error {
 	return err
 
 }
+
+// Write tries to write value with a given key and returns indicator if write operation succeed
+func (c *Client) Write(ctx context.Context, key string, value string, ttl int64) (bool, error) {
```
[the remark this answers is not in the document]
soider: I could add it, but it just adds additional conversion layer between time.Duration and int64, as the etcd client accepts int64.

pkg/etcd/client.go
soider: Hm, but why? the idea is just to check, if communication between kine and non-kine worlds works, why to complicate it? I did the writing to the "etcd" in the easiest possible way, mostly because kine etcd api implementation is limited (see kine source code).

pkg/component/controller/kine.go
soider: I think that makes sense.

pkg/etcd/client.go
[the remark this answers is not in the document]
soider: Sorry, haven't answered the main question. The ttl is a duration of record lifetime.

pkg/etcd/client.go
soider: So I checked with the history, if I do not use "notFound" guard, the underlying lib goes with etcd PUT method which is not implemented in kine.

pkg/etcd/client.go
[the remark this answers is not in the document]
twz123: A duration would have seemed more idiomatic. Not sure if this is seconds, milliseconds, nanoseconds. The Write function would have had to do the conversion, but it would have provided more clarity and ease of use at the call site.

```go
c.Write(ctx, "foo", "bar", 1*time.Minute)
// versus
c.Write(ctx, "foo", "bar", 60 /* seconds */)
```

pkg/component/controller/kine.go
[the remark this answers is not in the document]
twz123: My previous comment stemmed from the fact that I didn't anticipate that the Write function wouldn't overwrite already existing values. That's why I was surprised. Given the current logic, I'd demote the logging to debug, probably. Gettting a `!ok` value is totally legit when the Healthy method is called more often than the value's TTL, so nothing to worry about.

pkg/component/controller/kine.go
twz123: Given the current logic, I think we should either do it for the key, or not at all, since the Write method doesn't overwrite already existing values. That would make the health test fail later on when it reads a previously written value that didn't expire yet.

pkg/etcd/client.go
twz123: > So I checked with the history, if I do not use "notFound" guard, the underlying lib goes with etcd PUT method which is not implemented in kine.

Ah, that makes perfect sense then! Might be worthwhile to add this as a comment here.

twz123 (dismissed):

makhov (approved):

PR #2277, opened 2022-10-17
Fix iptables mode detection and fallback

ncopa:
## Description

nf_tables does not seem to work well with older kernels (4.15).

Improve the selection of iptables mode to be more conservative.
- consult `iptables --version` on the host if it exists. Even if k0s will not use this it gives us a hint on the OS default mode.
- when no iptables is found on the host, use legacy for kernels older than 5.0

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/iptablesutils/iptables.go
+++ b/internal/pkg/iptablesutils/iptables.go
@@ -61,8 +61,20 @@ func DetectIPTablesMode(k0sBinPath string) (string, error) {
 		logrus.Info("kube-related iptables entries not found, go with iptables-legacy")
 		return ModeLegacy, nil
 	}
-	logrus.Info("kube-related iptables entries not found, go with iptables-nft")
-	return ModeNFT, nil
+
+	iptablesPath, err := exec.LookPath("iptables")
+	if err != nil {
+		return "", err
+	}
+	out, err := exec.Command(iptablesPath, "--version").CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	outStr := string(out)
+	if strings.Contains(outStr, "nf_tables") {
+		return ModeNFT, nil
+	}
+	return ModeLegacy, nil
 }
 
 func iptablesEntriesCount(k0sBinPath string, xtablesCmdName string, entries []string) (entriesFound int, total int, err error) {
--- a/internal/pkg/iptablesutils/iptables.go
+++ b/internal/pkg/iptablesutils/iptables.go
@@ -32,10 +32,10 @@ const (
 	ModeLegacy = "legacy"
 )
 
-// DetectIPTablesMode figure out whether iptables-legacy or iptables-nft is in use on the host.
+// DetectHostIPTablesMode figure out whether iptables-legacy or iptables-nft is in use on the host.
 // Follows the same logic as kube-proxy/kube-route.
 // See: https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/iptables-wrapper-installer.sh
-func DetectIPTablesMode(k0sBinPath string) (string, error) {
+func DetectHostIPTablesMode(k0sBinPath string) (string, error) {
 	logrus.Info("Trying to detect iptables mode")
 	logrus.Info("Checking iptables-nft for kube-related entries")
 	nftEntriesCount, nftTotalCount, err := iptablesEntriesCount(k0sBinPath, "xtables-nft-multi", []string{"KUBE-IPTABLES-HINT", "KUBE-KUBELET-CANARY"})
--- a/pkg/component/worker/kernelsetup_linux.go
+++ b/pkg/component/worker/kernelsetup_linux.go
@@ -83,3 +83,12 @@ func KernelSetup() {
 	enableSysCtl("net/bridge/bridge-nf-call-iptables")
 	enableSysCtl("net/bridge/bridge-nf-call-ip6tables")
 }
+
+// KernelMajorVersion returns the major version number of the running kernel
+func KernelMajorVersion() byte {
+	data, err := os.ReadFile("/proc/sys/kernel/osrelease")
+	if err != nil {
+		return 0
+	}
+	return data[0] - '0'
+}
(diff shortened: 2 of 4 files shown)


s0j (approved):
LGTM

internal/pkg/iptablesutils/iptables.go
```diff
@@ -61,8 +61,20 @@ func DetectIPTablesMode(k0sBinPath string) (string, error) {
 		logrus.Info("kube-related iptables entries not found, go with iptables-legacy")
 		return ModeLegacy, nil
 	}
-	logrus.Info("kube-related iptables entries not found, go with iptables-nft")
-	return ModeNFT, nil
+
+	iptablesPath, err := exec.LookPath("iptables")
+	if err != nil {
+		return "", err
```
twz123, line 67: Do we want to err out here? I'd have expected that the err is ignored and `ModeLegacy, nil` gets returned.

pkg/component/worker/kernelsetup_linux.go
```diff
@@ -83,3 +83,12 @@ func KernelSetup() {
 	enableSysCtl("net/bridge/bridge-nf-call-iptables")
 	enableSysCtl("net/bridge/bridge-nf-call-ip6tables")
 }
+
+// KernelMajorVersion returns the major version number of the running kernel
+func KernelMajorVersion() byte {
```
twz123, line 88: At some point, we should consolidate that with the code of the sysprobes.

pkg/component/worker/kubelet.go
```diff
@@ -96,8 +96,12 @@ func (k *Kubelet) Init(_ context.Context) error {
...
 			iptablesMode, err = iptablesutils.DetectIPTablesMode(k.K0sVars.BinDir)
 			if err != nil {
-				logrus.Errorf("error detecting iptables mode: %v, using iptables-nft by default", err)
-				iptablesMode = "nft"
+				if KernelMajorVersion() < 5 {
+					iptablesMode = "legacy"
+				} else {
+					iptablesMode = "nft"
+				}
+				logrus.Errorf("error detecting iptables mode: %v, using iptables-%s by default", err, iptablesMode)
```
twz123, line 104: I'd rather log this at info level.

internal/pkg/iptablesutils/iptables.go
ncopa: The idea here is that if there is no working `iptables` binary on the host, then it means we failed to detect the mode and 
`DetectIPTablesMode` return error. It means we don't know what the mode is on the host.

pkg/component/worker/kubelet.go
ncopa: Yeah. that makes sense.

pkg/component/worker/kernelsetup_linux.go
ncopa: I was not aware we had `osRelease` there. Is there any convenient way to get kernel major version (or just `osrelease` string) using sysprobes?

internal/pkg/iptablesutils/iptables.go
jnummelin: > if there is no working iptables binary on the host, then it means we failed to detect the mode

that would mean you cannot run k0s at all on a host that does not have iptables bin on the host. Which is not what we want.

pkg/component/worker/kernelsetup_linux.go
twz123: No, not really ATM. This was more like a "note to self" / "raise awareness" comment that we should probably make the sysprobes stuff a bit more usable outside of the startup probe context, as we start to rely more and more on those things in core k0s code. This is definitely something to be addressed separately.

internal/pkg/iptablesutils/iptables.go
makhov: > that would mean you cannot run k0s at all on a host that does not have iptables bin on the host. Which is not what we want.

Yes, but in case `DetectIPTablesMode` returns an error we stick to the defaults [here](https://github.com/k0sproject/k0s/pull/2277/files#diff-c4db305f5cd0d057a727228694bf6fd480aeb5dead2c8fff79cdd73f0c5c9326R99), so everything works as expected. 

But maybe we need to refactor it a bit to make the code easier to understand.

pkg/component/worker/kernelsetup_linux.go
twz123: For reference: https://github.com/k0sproject/k0s/blob/v1.25.2%2Bk0s.0/internal/pkg/sysinfo/probes/linux/kernel.go#L37-L52

Totally unusable outside of a sysprobe, currently.

internal/pkg/iptablesutils/iptables.go
```diff
@@ -61,8 +61,20 @@ func DetectIPTablesMode(k0sBinPath string) (string, error) {
...
+	}
+	out, err := exec.Command(iptablesPath, "--version").CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	outStr := string(out)
+	if strings.Contains(outStr, "nf_tables") {
+		return ModeNFT, nil
+	}
+	return ModeLegacy, nil
```
makhov, line 77: Do we want to use `ModeLegacy` here regardless of the kernel version?

ncopa: Yes. If `iptables --version` successfully executes, but does not return "nf_tables", it uses legacy mode.

I think we want use legacy mode if there are existing iptables on the host in legacy mode. Otherwise things will go wrong if user adds firewall rules after k0s has been started.

internal/pkg/iptablesutils/iptables.go
ncopa: > that would mean you cannot run k0s at all on a host that does not have iptables bin on the host. Which is not what we want.

No. It meant we fail to detect the host's iptables mode, in which case we fallback to kernel major version check and use nftables unless kernel is older than 5.x.y

pkg/component/worker/kernelsetup_linux.go
ncopa: ok. maybe we just leave it as it is for now then.

twz123: Sure!

makhov (dismissed):

makhov (approved):

PR #2310, opened 2022-10-25
Don't suppress retries for server errors in watch

twz123:
## Description

During Kubernetes startup it may happen that the API server is not yet fully initialized and answers with 503 and a message of "the request has been made before all known HTTP paths have been installed, please try again".

It seems reasonable to not forcefully bypass the retry logic for all server errors that specify a retry-after hint.

The described error occurs on CI from time to time. Inttests probably need to specify a custom error callback via `WithErrorCallback` in order to actually retry this. But this is left out for another PR.
 
Reference: https://github.com/kubernetes/kubernetes/blob/v1.25.3/staging/src/k8s.io/apiserver/pkg/util/notfoundhandler/not_found_handler.go#L48-L53

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/kubernetes/watch/watcher.go
+++ b/pkg/kubernetes/watch/watcher.go
@@ -198,6 +198,23 @@ func (w *Watcher[T]) run(ctx context.Context, listCallback func([]T) (bool, erro
 			return err
 		}
 		if ok, err := listCallback(items); err != nil {
+			// Check if the err is a "proper" API server error
+			if status, ok := err.(apierrors.APIStatus); ok || errors.As(err, &status) {
+				status := status.Status()
+				// If this is a server error and the server specified some
+				// retry-after hint, don't suppress any retries. This matches
+				// e.g. the API server's "not found handler" which has a message
+				// of "the request has been made before all known HTTP paths
+				// have been installed, please try again".
+				if status.Code >= 500 && status.Code <= 599 {
+					details := status.Details
+					if details != nil {
+						if details.RetryAfterSeconds > 0 {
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/watch/watcher.go
```diff
@@ -198,6 +198,23 @@ func (w *Watcher[T]) run(ctx context.Context, listCallback func([]T) (bool, erro
 			return err
 		}
 		if ok, err := listCallback(items); err != nil {
+			// Check if the err is a "proper" API server error
+			if status, ok := err.(apierrors.APIStatus); ok || errors.As(err, &status) {
```
s0j, line 202: Do you really need the type assertion? Couldn't this just be a test to `errors.As()`?

twz123: I was wondering the same when taking over this pattern from [`k8s.io/apimachinery/pkg/api/errors`](https://github.com/kubernetes/kubernetes/blob/v1.25.3/staging/src/k8s.io/apimachinery/pkg/api/errors/errors.go) where it's used quite frequently.

I just double-checked with the k8s git history. Seems this was [a recent performance improvement in Kubernetes](https://github.com/kubernetes/kubernetes/commit/13ebe3b454d0739ab1069cfbc37e26c1773874b1). Arguably this won't have measurable impact on k0s, so I'm happy to change that if you prefer.

s0j (approved):

Issue #2372, opened 2022-11-09
Builds and smoketests on macOS with apple silicon


juanluisvaladas:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
macOS Monterey with apple silicon. Most certainly Ventura affected as well
```

### Version

main branch

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
```text
➡️ Please replace this text with the output of `k0s sysinfo`. ⬅️
```
</details>

### What happened?

It's impossible to compile and run smoketests in macOS with apple silicon

### Steps to reproduce

1. Clone the repo in a computer with macOS, apple silicon and all the requirements installed
2. Run make
3. Run make smoketests

### Expected behavior

Everything works

### Actual behavior

Can't get to the tests.

PR #2373, opened 2022-11-09
Fix builds and ap-airgap inttests on mac m1

juanluisvaladas:
## Description

k0s is unable to build and test on apple silicon. This changes fixes partially some of the issues https://github.com/k0sproject/k0s/issues/2372

WIth this PR the airgap image bundle can be generated correctly and airgap and ap-airgap tests pass.
The scope of tests that remain broken is unknown at this point but at least ap-ha3x3 is still broken.

Also I have snuck a commit to add .DS_Store in the .gitignore, but since that's mac related and very small I think it makes sense

Fixes # (issue)
Partially https://github.com/k0sproject/k0s/issues/2372
## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,8 +215,8 @@ lint: .k0sbuild.docker-image.k0s go.sum codegen
 	CGO_ENABLED=0 $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v$(golangci-lint_version)
 	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_DIRS)
 
-airgap-images.txt: k0s
-	./k0s airgap list-images > '$@' || { \
+airgap-images.txt: k0s .k0sbuild.docker-image.k0s 
+	$(GO_ENV) ./k0s airgap list-images > '$@' || { \
 	  code=$$? && \
 	  rm -f -- '$@' && \
 	  exit $$code ; \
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ static/zz_generated_assets.go
 docs/cli
 site
 k0sctl.yaml
+.DS_Store
```
twz123, line 25: I'd suggest to put those things into `$XDG_CONFIG_HOME/git/ignore`, so we don't have to accumulate all sorts of OS specific stuff here.

Makefile
```diff
@@ -216,7 +216,7 @@ lint: .k0sbuild.docker-image.k0s go.sum codegen
 	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_DIRS)
 
 airgap-images.txt: k0s
```
twz123, line 218:
```suggestion
airgap-images.txt: k0s .k0sbuild.docker-image.k0s
```

If the Docker container is used to execute the binary, it's better to list it as a make prerequisite.

twz123 (changes requested):

.gitignore
juanluisvaladas: Makes perfect sense, I removed the commit

Makefile
juanluisvaladas: Done

twz123 (approved):

PR #2374, opened 2022-11-09
Let k0s exchange the kubelet bootstrap kubeconfig

twz123:
## Description

Instead of passing the bootstrap kubeconfig to the kubelet in order to exchange that for a regular certificate-based kubeconfig, let k0s do this. This way, k0s has full control over the bootstrapping process and can safely delete the bootstrap kubeconfig after a successful exchange.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/worker/utils.go
+++ b/pkg/component/worker/utils.go
@@ -17,31 +17,76 @@ limitations under the License.
 package worker
 
 import (
+	"context"
 	"fmt"
+	"os"
 	"path"
-
-	"k8s.io/client-go/tools/clientcmd"
+	"path/filepath"
+	"runtime"
+	"time"
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/internal/pkg/file"
+	"github.com/k0sproject/k0s/internal/pkg/flags"
+	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
 	"github.com/k0sproject/k0s/pkg/token"
+
+	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/kubernetes/pkg/kubelet/certificate/bootstrap"
+	nodeutil "k8s.io/kubernetes/pkg/util/node"
+
+	"github.com/avast/retry-go"
+	"github.com/sirupsen/logrus"
+	"go.uber.org/multierr"
 )
 
-func HandleKubeletBootstrapToken(encodedToken string, k0sVars constant.CfgVars) error {
-	kubeconfig, err := token.DecodeJoinToken(encodedToken)
-	if err != nil {
-		return fmt.Errorf("failed to decode token: %w", err)
-	}
+func BootstrapKubeletKubeconfig(ctx context.Context, k0sVars constant.CfgVars, workerOpts *config.WorkerOptions) error {
+	var bootstrapKubeconfigPath string
+	var bootstrapKubeconfig *clientcmdapi.Config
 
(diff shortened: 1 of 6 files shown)


pkg/component/worker/utils.go
```diff
@@ -17,62 +17,154 @@ limitations under the License.
...
+	}
+
+	certDir := filepath.Join(k0sVars.DataDir, "kubelet", "pki")
+	if err := dir.Init(certDir, constant.DataDirMode); err != nil {
+		return fmt.Errorf("failed to initialize kubelet certificate directory: %w", err)
+	}
+
+	// FIXME: This is not completely on par with kubelet code. The kubelet has
+	// some means of retrieving the node name via the cloud provider, if one is
+	// configured.
```
twz123, line 106: The k0s kubelet gets built without the in-tree cloud providers, so this might not apply, after all.

pkg/component/worker/utils.go
```diff
@@ -17,62 +17,154 @@ limitations under the License.
...
+	nodeName, err := nodeutil.GetHostname("")
```
twz123, line 107: Do we want to retrieve the hostname override somehow? Not sure how important that might be. Seems doable, but would require quite some code shuffling, i.e. we'd need to pull out the kubelet config part out of `kubelet.go` in order to load the worker profile early.

As far as I can tell, the node name here is only used as suffix to the user name in the certs, so that one can see which node authenticated with the API server.

jnummelin: > As far as I can tell, the node name here is only used as suffix to the user name in the certs, so that one can see which node authenticated with the API server.

Does the token -> cert exchange already create the node object in the API? If it does, then we need to worry about the hostname override. If not then I think we do not have to worry about that.

twz123: The bootstrapping won't create any node objects. But there's still something that requires that the cert's node name and the kubelet's node name match. I tried to join a worker via `k0s worker --kubelet-extra-args='--hostname-override=foobar'`. The kubelet won't be able to make progress and spams a lot of "Error getting node" err="node \"foobar\" not found" in the logs. Not sure why, but if the host names match, everything is well.

Maybe this is related to how the [Node authorization](https://kubernetes.io/docs/reference/access-authn-authz/node/) works (although the error message doesn't indicate any authz problems?). The docs there say:

> The value of `<nodeName>` **must** match precisely the name of the node as registered by the kubelet. By default, this is the host name as provided by `hostname`, or overridden via the [kubelet option](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) `--hostname-override`. However, when using the `--cloud-provider` kubelet option, the specific hostname may be determined by the cloud provider, ignoring the local `hostname` and the `--hostname-override` option. For specifics about how the kubelet determines the hostname, see the [kubelet options reference](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/).

So I guess k0s needs to make sure as best as possible that the node names match. I realized that the hostname override solely comes from k0s args, not from any dynamic source, so it's much more straight forward to use that setting also during bootstrapping. Will add that to the PR, along with an inttest that covers this case.

I'm not sure about the cloud provider stuff. In that case, kubelet will apparently _always_ use the node name from the cloud provider, even if the hostname override option is given. Assuming that kubelet will also fail if the names don't match, as in the non-cloud provider case, the only way to make k0s work with a cloud provider is to specify the right hostname override manually, so that k0s picks it up (even if kubelet will ignore it).

twz123: I double-checked the kubelet code. K0s only supports external cloud providers, which actually seems to be a [dead code path](https://github.com/kubernetes/kubernetes/blob/v1.25.4/cmd/kubelet/app/server.go#L547-L559) in kubelet anyways. So I don't see any way k0s could come to a different conclusion about the node name than kubelet.

makhov (approved):

PR #2376, opened 2022-11-09
Don't create the test network if it already exists

jnummelin:
Also don't try to remove it if we're leaving the footloose containers running after the tests as the `docker network rm` call would fail.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -1289,16 +1289,34 @@ func (s *FootlooseSuite) getIPAddress(nodeName string) string {
 // CreateNetwork creates a docker network with the provided name, destroying
 // any network that has the same name first.
 func (s *FootlooseSuite) CreateNetwork(name string) error {
-	_ = s.DestroyNetwork(name)
+	// Don't create the network if it already exists as it might be in use from previous tests
+	// ran with K0S_KEEP_AFTER_TEST variable set.
+	if s.NetworkExists(name) {
+		return nil
+	}
 
 	cmd := exec.Command("docker", "network", "create", name)
 	return cmd.Run()
 }
 
-// DestroyNetwork removes a docker network with the provided name.
-func (s *FootlooseSuite) DestroyNetwork(name string) error {
+// MaybeDestroyNetwork removes a docker network with the provided name.
+// The network might not get removed if there's still containers attached to it.
+// This is the case for example when running the tests with K0S_KEEP_AFTER_TEST=always.
+func (s *FootlooseSuite) MaybeDestroyNetwork(name string) error {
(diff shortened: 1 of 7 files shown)


juanluisvaladas (approved):
Looks good to me.

inttest/common/footloosesuite.go
```diff
@@ -1289,18 +1289,34 @@ func (s *FootlooseSuite) getIPAddress(nodeName string) string {
...
+	}
 
 	cmd := exec.Command("docker", "network", "create", name)
 	return cmd.Run()
 }
 
 // DestroyNetwork removes a docker network with the provided name.
 func (s *FootlooseSuite) DestroyNetwork(name string) error {
+	// If we're supposed to leave the footloose containers running, don't try to destroy the network
+	if keepEnvironment(s.T()) {
```
twz123, line 1305: I'd have preferred to have this check made by the callers of this method. Seems kinda unexpected that the network won't be deleted by this function under certain circumstances. But yeah, I can currently see 6 call sites, so this is a pragmatic choice.

Ideally, the network lifecycle management would be done by the `FootlooseSuite` itself, i.e. it keeps track of the created networks and destroys them conditionally along with the rest of the cluster somewhere [here](https://github.com/k0sproject/k0s/blob/7ff1e012fd1cd02dab352e80bd3110691596133e/inttest/common/footloosesuite.go#L347-L356). This function would then become obsolete, as the inttests could simply rely on the fact that `FootlooseSuite` will take care of all the networks that it has created via `CreateNetwork`, instead of having to call `DestroyNetwork` manually in `TearDownSuite`.

twz123 (approved):

inttest/common/footloosesuite.go
juanluisvaladas: I don't like that a function called `DestroyNetwork` may be expected not to detroy the network, but changing this in every test makes it prone to forget to add this in some tests, not just now but also in the future.

A somewhat acceptable option which kind of solves both problems would be renaming the function to something like `MaybeDestroyNetwork` would solve the problem. It's not as good as handling this in `TearDownSuite`, but it's less work.

jnummelin: Yeah, handling the conditional network removal on caller side would be bit painful. I renamed the func to `MaybeDestroyNetwork`

juanluisvaladas: Works for me. What do you think @twz123 ?

twz123: Sure.

twz123 (approved):

PR #2387, opened 2022-11-11
Calico test improvement: added pod-to-pod connectivity check

makhov:

## Description

Added pod-to-pod connectivity check

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

- [z] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 

### DIFF
--- a/inttest/calico/calico_test.go
+++ b/inttest/calico/calico_test.go
@@ -18,10 +18,13 @@ package calico
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/k0sproject/k0s/inttest/common"
 )
@@ -37,6 +40,8 @@ func (s *CalicoSuite) TestK0sGetsUp() {
 
 	kc, err := s.KubeClient("controller0", "")
 	s.Require().NoError(err)
+	restConfig, err := s.GetKubeConfig("controller0", "")
+	s.NoError(err)
 
 	err = s.WaitForNodeReady("worker0", kc)
 	s.NoError(err)
@@ -44,7 +49,7 @@ func (s *CalicoSuite) TestK0sGetsUp() {
 	err = s.WaitForNodeReady("worker1", kc)
 	s.NoError(err)
 
-	calicoDaemonset, err := kc.AppsV1().DaemonSets("kube-system").Get(context.TODO(), "calico-node", v1.GetOptions{})
+	calicoDaemonset, err := kc.AppsV1().DaemonSets("kube-system").Get(context.TODO(), "calico-node", metav1.GetOptions{})
 	s.Require().NoError(err)
 	var calicoCustomEnvVarsFound int
 	for _, v := range calicoDaemonset.Spec.Template.Spec.Containers[0].Env {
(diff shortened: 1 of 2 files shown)


inttest/calico/calico_test.go
```diff
@@ -58,6 +62,40 @@ func (s *CalicoSuite) TestK0sGetsUp() {
...
+			}},
+			NodeSelector: map[string]string{
+				"kubernetes.io/hostname": "worker1",
+			},
+		},
+	}, metav1.CreateOptions{})
+	s.Require().NoError(err)
+	s.NoError(common.WaitForPod(s.Context(), kc, "busybox", "default"), "busybox pod did not start")
+
+	out, err := common.PodExecCmdOutput(kc, restConfig, sourcePod.Name, sourcePod.Namespace, fmt.Sprintf("/usr/bin/wget -qO- %s:9376", targetPod.Status.PodIP))
```
jnummelin, line 96: This might be bit flaky as such. I think we should wait till konnectivity has its tunnels setup before we do pod exec. One "trick" done in some other smokes is to wait to see some pod logs as that exercises the konnecivity tunnel.

makhov: I don't think it will be flaky since first we wait for the calico daemonset to be ready and then spin up two pods, but yeah, I added `WaitForLogs`.

jnummelin (approved):

inttest/calico/calico_test.go
```diff
@@ -58,6 +63,44 @@ func (s *CalicoSuite) TestK0sGetsUp() {
...
+	targetPod, err := kc.CoreV1().Pods(createdTargetPod.Namespace).Get(s.Context(), createdTargetPod.Name, metav1.GetOptions{})
+	s.Require().NoError(err)
+
+	sourcePod, err := kc.CoreV1().Pods("default").Create(s.Context(), &corev1.Pod{
+		TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "busybox"},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{{
+				Name:    "busybox",
+				Image:   "gcr.io/google-containers/busybox",
```
juanluisvaladas, line 90: This image is not available for arm64:

➜  ~ docker run -it --rm  gcr.io/google-containers/busybox bash
<snip>
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

In the kuberouter hairpin test we have the image docker.io/curlimages/curl:7.84.0 which has sleep and curl and is available in arm64

I'm not quite sure if we will ever want to test this on more architectures than amd64, if this is an acceptable constraint I think we can leave it as is.

jnummelin: @juanluisvaladas that's a very good point. I think it makes sense to change the image to something which is multiarch. Maybe even just plain `alpine:3.16` as that also has both `sleep` and `wget` available

makhov: Changed the image to `alpine:3.16`. Thanks!

inttest/calico/calico_test.go
jnummelin: Well, the pod exec's etc. being flaky has nothing to do with if calico is ready or not. exec goes via konnectivity tunnel and if that's not fully working the execs will fail. But waiting to get logs has turned to be a good signal that the tunnels are up properly.

jnummelin (approved):

juanluisvaladas (approved):

PR #2396, opened 2022-11-11
Add inttest that reboots worker and controller

### LINKED ISSUE #1223: Add integration test that reboots workers and controllers
We should test that the worker and contrller comes back up after a reboot.

juanluisvaladas:
## Description

Before merging, I think we should discuss if it makes sense to have both this and restartworker since I believe this test also covers it and restartworker would be redundant now., probably the restartworker test should be removed in another commit 

Fixes #1223

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

- [x} I have commented my code, particularly in hard-to-understand areas


inttest/clusterreboot/clusterreboot_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	s.Require().NoError(s.WaitForSSH(s.ControllerNode(0), 1*time.Minute, 1*time.Second))
+	s.Require().NoError(s.WaitForSSH(s.WorkerNode(0), 1*time.Minute, 1*time.Second))
+}
+
+func TestClusterRebootSuite(t *testing.T) {
+	s := ClusterRebootSuite{
+		common.FootlooseSuite{
+			ControllerCount: 1,
+			WorkerCount:     1,
```
jnummelin, line 81: I think we should set the `LaunchMode: common.LaunchModeOpenRC` so that k0s is set as a system service. That way we can just reboot the nodes and wait that everything comes up without the need to re-initialize.

inttest/clusterreboot/clusterreboot_test.go
```diff
@@ -0,0 +1,85 @@
...
+	err = s.WaitForNodeReady(s.WorkerNode(0), kc)
+	s.NoError(err)
+
+	// reboot the cluster:
+	s.T().Log("Rebooting cluster")
+	s.rebootCluster()
+
+	// Restart k0s controller
+	s.T().Log("Restarting k0s")
+	s.NoError(s.InitController(0))
```
jnummelin, line 49: If we use `LaunchMode: common.LaunchModeOpenRC` then we do not have to re-init controller, openRC should just start it on reboot automatically.

inttest/clusterreboot/clusterreboot_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	kc, err = s.KubeClient(s.ControllerNode(0))
+	s.Require().NoError(err)
+	s.Require().NoError(s.WaitForKubeAPI(s.ControllerNode(0)))
+
+	// restart k0s worker and make sure it comes up
+	s.NoError(s.RunWorkers())
```
jnummelin, line 56: If we use `LaunchMode: common.LaunchModeOpenRC` then we do not have to re-init workers, openRC should just start it on reboot automatically.

inttest/clusterreboot/clusterreboot_test.go
juanluisvaladas: Fixed, thanks!

Now the question is, should we maintain the workerrestart test? I think this covers it.

jnummelin: > Now the question is, should we maintain the workerrestart test? I think this covers it.

I think that's obsolete now with this getting in. so maybe just remove it in separate commit.

jnummelin (approved):

PR #2407, opened 2022-11-15
[Backport release-1.25] Switch from k8s.gcr.io to registry.k8s.io

twz123:
## Description

Backport of #2346.

Kubernetes switched to the new registry in k8s 1.25. Do the same for k0s.

Reference: kubernetes/k8s.io#1834 and kubernetes/kubernetes#109938.

This also fixes a discrepancy between the default pause image used in containerd and the one used by k0s.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


inttest/airgap/airgap_test.go
```diff
@@ -36,7 +36,7 @@ spec:
 const etcHosts = `
 127.0.0.8 docker.io
 127.0.0.8 gcr.io
-127.0.0.8 k8s.gcr.io
+127.0.0.8 registry.k8s.io
```
jnummelin, line 39: I'd also leave `k8s.gcr.io` into the "block" list to be sure we block any accidental access to it too

docs/configuration.md
```diff
@@ -84,17 +84,17 @@ spec:
...
+      version: v0.6.1
     kubeproxy:
-      image: k8s.gcr.io/kube-proxy
+      image: registry.k8s.io/kube-proxy
       version: v1.25.4
     coredns:
-      image: k8s.gcr.io/coredns/coredns
-      version: v1.7.0
+      image: docker.io/coredns/coredns
+      version: 1.9.4
```
jnummelin, line 97: Do we want to bump up coreDNS this much in a backport?

jnummelin (commented):
I don't think we can backport this as-is to 1.24 as that uses very different image versions

inttest/airgap/airgap_test.go
twz123: I'd like to file a separate PR for this, as this is a backport, which wouldn't be on par with the original PR. Alternatively, I can add a commit here, which we then cherry-pick back to main.

docs/configuration.md
twz123: This is the docs, which have been updated in the original PR to match the versions that are actually used by k0s. The 1.25 release is currently on 1.9.4.

jnummelin: ah 🤦  all good then :)

inttest/airgap/airgap_test.go
jnummelin: good point. separate PR makes sense.

PR #2413, opened 2022-11-16
Write k0s's pause image into the default containerd config

twz123:
## Description

This is to avoid potential mismatches between k0s's assumptions and containerd's defaults. If the pause images differ, the outputs of `k0s airgap list-images` would be inaccurate, and airgapped environments might not work correctly.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/worker/containerd.go
+++ b/pkg/component/worker/containerd.go
@@ -26,17 +26,23 @@ import (
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/internal/pkg/file"
+	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
+	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
 	"github.com/k0sproject/k0s/pkg/assets"
 	"github.com/k0sproject/k0s/pkg/component"
 	"github.com/k0sproject/k0s/pkg/constant"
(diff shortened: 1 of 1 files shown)


pkg/component/worker/containerd.go
```diff
@@ -26,17 +26,23 @@ import (
...
-# For reference see https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md
+const confTmpl = `# This is a placeholder configuration for k0s managed containerd.
+# If you wish to customize the configuration, replace this file with your own configuration.
+# For reference, see https://github.com/containerd/containerd/blob/v1.6.9/docs/man/containerd-config.toml.5.md.
+
 version = 2
+
+[plugins]
+  [plugins."io.containerd.grpc.v1.cri"]
+    sandbox_image = {{ .PauseImage | quote }}
```
twz123, line 45: Maybe this is not the right approach, in the end. K0s will never replace this file, so this won't be updated during updates, which smells like trouble.

An alternative would be to add an inttest that checks the default config of containerd instead, so that k0s always references the same pause image as containerd in its default config.

jnummelin: > Maybe this is not the right approach, in the end. K0s will never replace this file, so this won't be updated during updates, which smells like trouble

Yes, that would mean that once k0s is installed with e.g. `pause:3.2` image, it'll be "stuck" with that version till end of times.

PR #2415, opened 2022-11-16
Embedded binaries test

makhov:

## Description

The test checks that we unpacked embedded binaries are operational


twz123 (approved):
I thought more about a test at runtime, after k0s staged the binaries into its bindir.

makhov:
The test waits for k0s to unpack the binaries into `/var/lib/k0s/bin` and then checks if they work. Is that what you suggested or am I missing something?

makhov:
Merging now, if we have better ideas, we can improve it in a separate PR.

PR #2417, opened 2022-11-16
Change kube hairpin configuration settings and defaults

### LINKED ISSUE #1953: Should k0s enable hairpin-mode by default for kube-router?

When using kube-router, it is not possible for a pod that is part of a service to reach out to its service IP. This setup is called hairpinning and is [supported in kube-router](https://github.com/cloudnativelabs/kube-router/blob/v1.5.0/docs/user-guide.md#hairpin-mode) via some config flag which defaults to off. The docs say

> Communication from a Pod that is behind a Service to its own ClusterIP:Port is not supported by default.

 I suspect this is quite surprising for users, as they will probably assume that hairpinning just works.

There's #1937 which is about to add support for this config option to k0s. Would it be a better UX if k0s would enable hairpinning by default?

It's not entirely clear to me what this entails. There's probably a reason why kube-router has chosen to disable it by default. The doc string on the `--hairpin-mode` CLI flag says

> Add iptables rules for every Service Endpoint to support hairpin traffic.

So this sounds costly, as managing iptables rules will slow doẃn significantly as soon as the number of rules reaches a certain size. This might start to matter on larger setups, with a lot of Services and NetworkPolicies.

Don't change the default. Folks can simply enable it if they need it, either on a service-level via annotations (i.e. `kubectl annotate service my-service "kube-router.io/service.hairpin="`), or cluster-wide via the new knobs introduced in #1937.

In my experiments, as soon as I add a second Pod to a Service, I am able to connect to the ClusterIP, although connection times vary and feel sluggish (think: seconds instead of milliseconds), so the actual impact of hairpinning being disabled is apparently even weirder.

In any case, this needs some investigation and IMO it's in order to write some docs about this, so that k0s users are aware.

### Additional context

_No response_

juanluisvaladas:
## Description

Enable hairpin for kube-router by default

We agreed that the most logical decision was to enable hairpin by default. The old configuration doesn't reflect our real needs, for this reason we decided to deprecate the old configuration and add a new configuration. This commit handles all these details.

Also I spotted a bug related to reconciling, since I noticed it in one of the files which I spotted and it's just a one liner I slipped the commit in this PR.

Fixes #1953

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -66,7 +66,7 @@ spec:
       peerRouterIPs: ""
       peerRouterASNs: ""
       autoMTU: true
-      hairpinMode: false
+      hairpin: Enabled
     kubeProxy:
       disabled: false
       mode: iptables
@@ -211,14 +211,15 @@ CALICO_IPV6POOL_CIDR: "{{ spec.network.dualStack.IPv6podCIDR }}"
 
 #### `spec.network.kuberouter`
 
-| Element          | Description                                                                                                                                        |
-| ---------------- |----------------------------------------------------------------------------------------------------------------------------------------------------|
-| `autoMTU`        | Autodetection of used MTU (default: `true`).                                                                                                       |
-| `mtu`            | Override MTU setting, if `autoMTU` must be set to `false`).                                                                                        |
-| `metricsPort`    | Kube-router metrics server port. Set to 0 to disable metrics  (default: `8080`).                                                                   |
-| `peerRouterIPs`  | Comma-separated list of [global peer addresses](https://github.com/cloudnativelabs/kube-router/blob/master/docs/bgp.md#global-external-bgp-peers). |
-| `peerRouterASNs` | Comma-separated list of [global peer ASNs](https://github.com/cloudnativelabs/kube-router/blob/master/docs/bgp.md#global-external-bgp-peers).      |
-| `hairpinMode`    | Activate hairpinMode (default: `false`) (https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#hairpin-mode)               |
+| Element          | Description                                                                                                                                                                                                                                                               |
+| ---------------- |---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `autoMTU`        | Autodetection of used MTU (default: `true`).                                                                                                                                                                                                                              |
+| `mtu`            | Override MTU setting, if `autoMTU` must be set to `false`).                                                                                                                                                                                                               |
+| `metricsPort`    | Kube-router metrics server port. Set to 0 to disable metrics  (default: `8080`).                                                                                                                                                                                          |
+| `peerRouterIPs`  | Comma-separated list of [global peer addresses](https://github.com/cloudnativelabs/kube-router/blob/master/docs/bgp.md#global-external-bgp-peers).                                                                                                                        |
+| `peerRouterASNs` | Comma-separated list of [global peer ASNs](https://github.com/cloudnativelabs/kube-router/blob/master/docs/bgp.md#global-external-bgp-peers).                                                                                                                             |
+| `hairpin`        | Hairpin mode, supported modes `Enabled`: enabled cluster wide, `Allowed`: must be allowed per service [using annotations](https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#hairpin-mode), `Disabled`: doesn't work at all (default: Enabled) |
+| `hairpinMode`    | **Deprecated** Use `hairpin` instead. If both `hairpin` and `hairpinMode` are defined, this is ignored. If only hairpinMode is configured explicitly activates hairpinMode (https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md#hairpin-mode).  |
 
 **Note**: Kube-router allows many networking aspects to be configured per node, service, and pod (for more information, refer to the [Kube-router user guide](https://github.com/cloudnativelabs/kube-router/blob/master/docs/user-guide.md)).
 
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
@@ -24,13 +24,28 @@ type KubeRouter struct {
 	MTU int `json:"mtu"`
 	// Kube-router metrics server port. Set to 0 to disable metrics  (default: 8080)
 	MetricsPort int `json:"metricsPort"`
-	// Activate Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
-	HairpinMode bool `json:"hairpinMode"`
+	// Admits three values: "Enabled" enables it globally, "Allowed" allows but services must be annotated explicitly and "Disabled"
+	// Defaults to "Enabled"
+	// +kubebuilder:default=Enabled
+	Hairpin Hairpin `json:"hairpin"`
+	// DEPRECATED: Use hairpin instead. Activates Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
+	HairpinMode bool `json:"hairpinMode,omitempty"`
 	// Comma-separated list of global peer addresses
 	PeerRouterASNs string `json:"peerRouterASNs"`
 	// Comma-separated list of global peer ASNs
 	PeerRouterIPs string `json:"peerRouterIPs"`
 }
+
+// +kubebuilder:validation:Enum=Enabled;Allowed;Disabled
+type Hairpin string
+
+const (
+	HairpinEnabled  Hairpin = "Enabled"
+	HairpinAllowed  Hairpin = "Allowed"
+	HairpinDisabled Hairpin = "Disabled"
+	// Necessary for backwards compatibility with HairpinMode
+	HairpinUndefined Hairpin = ""
+)
 
 // DefaultKubeRouter returns the default config for kube-router
 func DefaultKubeRouter() *KubeRouter {
@@ -38,5 +53,6 @@ func DefaultKubeRouter() *KubeRouter {
 		MTU:         0,
 		AutoMTU:     true,
 		MetricsPort: 8080,
+		Hairpin:     HairpinEnabled,
 	}
 }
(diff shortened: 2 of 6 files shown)


juanluisvaladas:
@hercherf  @mtb-xt since you commented on #1953 I'd like your feedback as well please.

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
...
+type Hairpin string
+
+const (
+	HairpinEnabled  Hairpin = "enabled"
+	HairpinAllowed  Hairpin = "allowed"
+	HairpinDisabled Hairpin = "disabled"
+	// Necessary for backwards compatibility with HairpinMode
+	HairpinUndefined Hairpin = ""
+)
+
```
twz123, line 46: I'd uppercase the values, and I'd add an enum hint for kubebuilder.

```suggestion
// +kubebuilder:validation:Enum=Enabled;Allowed;Disabled
type HairpinMode string

const (
	HairpinEnabled  HairpinMode = "Enabled"
	HairpinAllowed  HairpinMode = "Allowed"
	HairpinDisabled HairpinMode = "Disabled"
	// Necessary for backwards compatibility with HairpinMode
	HairpinUndefined HairpinMode = ""
)

```

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
 	MTU int `json:"mtu"`
 	// Kube-router metrics server port. Set to 0 to disable metrics  (default: 8080)
 	MetricsPort int `json:"metricsPort"`
-	// Activate Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
-	HairpinMode bool `json:"hairpinMode"`
+	// Admits three values: "enabled" enables it globaly, "allowed" allows but services must be annotated explictly and "disabled"
+	Hairpin Hairpin `json:"hairpin"`
+	// DEPRECATED: Use hairpining instead. Activates Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
+	HairpinMode bool `json:"hairpinMode,omitempty"`
```
twz123, line 30: It's a pity that the better name is already taken by the old boolean setting. I wonder if we could merge those into one. Either accepting booeans or strings, just like apimachinery's `IntOrString` does it for ints and strings.

In any case I think we should rename both the type and the field name to `HairpinMode`, but just leave the json annotations untouched.

```suggestion
	// Admits three values: "Enabled" enables it globally, "Allowed" allows but services must be annotated explicitly and "Disabled"
	// +optional
	HairpinMode HairpinMode `json:"hairpin,omitempty"`
	// DEPRECATED: Use hairpining instead. Activates Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
	HairpinEnabled bool `json:"hairpinMode,omitempty"`
```

pkg/component/controller/kuberouter.go
```diff
@@ -70,6 +71,27 @@ func (k *KubeRouter) Init(_ context.Context) error { return nil }
 // Stop no-op as nothing running
 func (k *KubeRouter) Stop() error { return nil }
 
+func getHairpinConfig(cfg *kubeRouterConfig, krc *v1beta1.KubeRouter) {
```
twz123, line 74: Maybe a better name this `applyHairpinConfig`, since this is not idempotently getting something, but updating the cfg in place.

twz123 (commented):
We should probably implement `json.Unmarshaller` for `KubeRouter`, now that we have a default which doesn't equal the zero value of its type.

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
jnummelin: > Either accepting booleans or strings, just like apimachinery's IntOrString does it for ints and strings.
I like this idea very much

mtb-xt:
There was an edge case about the CNI config - if it was already generated, then activating hairpin mode in k0s won't update the CNI config. 
If someone now creates the cluster with hairpin mode disabled, but then decides to enable it - it won't be updated as well.
I realize this is a rare scenario now it's enabled by default, but do we want to do anything about it or not?

https://github.com/k0sproject/k0s/blob/78b4567f78abbff122a2cbf4ac88e1e51c737b9c/pkg/component/controller/kuberouter.go#L232 <- because of this check, if there is a CNI config present, it won't be touched.

juanluisvaladas:
Having one field accepting multiple types breaks dynamic configuration because of https://github.com/kubernetes-sigs/controller-tools/issues/735

I'm currently waiting for controller-manager people to give me the OK to start implementing this so in the meantime I'll put this on hold. If the proposed change wouldn't be accepted  then we'll have to consider a strategy different than using the same field to accept multiple types.

juanluisvaladas:
@jnummelin @twz123 I don't think I can achieve this even by patching the CRD manually.

The CRDs with `IntOrString` work because the porperties of the schema include the field `x-kubernetes-int-or-string`. The apiserver schema validation [explictly skips some validations for IntOrString](https://github.com/kubernetes/kubernetes/blob/v1.26.0/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/schema/validation.go#L45-L62), so we cannot modify this.

I've tried to workaround this in many ways but I don't think it's actually possible. The only option that I see here would be adding x-kubernetes-preserve-unknown-fields to the KubeRouter struct and removing the HairpinMode field of the CRD but then this adds some additional problems in validation...

I see the following options:
1- Issue a new version of the clusterconfigs crd including a new version v1beta2 in which the hairpinMode is a string. This is probably the best one.
2- Add a new field with a different name and deprecate HairpinMode, which isn't great because HairpinMode is the best name.
3- Allowing unknown fields and removing the field from the CRD, which brings issues with the validation.

twz123:
I would vote for option 2 (new field with a different name). This is what this PR initially aimed at, and it's a fair compromise compared to the alternatives. Introducing a new API version for this seems like a quite heavy thing to do. Maybe it's better to accumulate some changes over time and then issue a new API version that contains more improvements than just this one. Allowing unknown fields OTOH seems to me like removing a safety net for very little gain.

I didn't know that IntOrString was such a custom thing. Otherwise I wouldn't have brought it up as an example :see_no_evil:.

juanluisvaladas:
> I would vote for option 2 (new field with a different name).

And there is a big advantage that it's already implemented, I just had to go through the reflog :) and patch some little thing.

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
...
 	// Comma-separated list of global peer addresses
 	PeerRouterASNs string `json:"peerRouterASNs"`
 	// Comma-separated list of global peer ASNs
 	PeerRouterIPs string `json:"peerRouterIPs"`
 }
 
+type Hairpin string
```
twz123, line 37: I'd add an enum hint for kubebuilder.

```suggestion
// +kubebuilder:validation:Enum=Enabled;Allowed;Disabled
type Hairpin string
```

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
 	MTU int `json:"mtu"`
 	// Kube-router metrics server port. Set to 0 to disable metrics  (default: 8080)
 	MetricsPort int `json:"metricsPort"`
-	// Activate Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
-	HairpinMode bool `json:"hairpinMode"`
+	// Admits three values: "enabled" enables it globaly, "allowed" allows but services must be annotated explictly and "disabled"
```
twz123, line 27: Some small corrections.

```suggestion
	// Admits three values: "Enabled" enables it globally, "Allowed" allows but services must be annotated explicitly and "Disabled".
```

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
...
+	Hairpin Hairpin `json:"hairpin"`
```
twz123, line 28: Worth mentioning the default value here as well.

```suggestion
	// Defaults to "Enabled".
	// +kubebuilder:default=Enabled
	Hairpin Hairpin `json:"hairpin"`
```

docs/configuration.md
```diff
@@ -66,7 +66,7 @@ spec:
       peerRouterIPs: ""
       peerRouterASNs: ""
       autoMTU: true
-      hairpinMode: false
+      hairpin: enabled
```
twz123, line 69:
```suggestion
      hairpin: Enabled
```

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
```diff
@@ -24,19 +24,32 @@ type KubeRouter struct {
...
+	// DEPRECATED: Use hairpining instead. Activates Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
```
twz123, line 29:
```suggestion
	// DEPRECATED: Use hairpin instead. Activates Hairpin Mode (allow a Pod behind a Service to communicate to its own ClusterIP:Port)
```

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
juanluisvaladas: Ugh stupid cherry-pick. This was done. I need to review this more carefully.

juanluisvaladas:
@twz123 I addressed everything and I also changed the table in docs/configuration.md. Apparently I didn´t find the latest version if the reflog and there were a couple things missing even if the code was working as expected.

I went file by file and I believe everything is good now.

pkg/apis/k0s.k0sproject.io/v1beta1/kuberouter.go
twz123: The kubebuilder default hint is still missing. But that's a purely cosmetic thing...

twz123 (approved):

juanluisvaladas:
Oh sorry I missed that one. That's added now.

twz123 (approved):

PR #2419, opened 2022-11-17
Upload sonobuoy test results

makhov:

## Description

Sonobuoy test results are uploaded as artifacts. Helps debug failures using the `sonobuoy` binary.

### DIFF
--- a/inttest/network-conformance/network_test.go
+++ b/inttest/network-conformance/network_test.go
@@ -18,6 +18,7 @@ package network
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


inttest/network-conformance/network_test.go
```diff
@@ -100,12 +102,37 @@ func (s *networkSuite) TestK0sGetsUp() {
...
 	s.Require().Equal("complete", status.Status)
 	s.Require().Len(status.Plugins, 1)
 	s.Require().Equal("passed", status.Plugins[0].ResultStatus)
 }
 
+func retrieveResults(r io.Reader, ec <-chan error) error {
+	eg := &errgroup.Group{}
+	eg.Go(func() error { return <-ec })
+	eg.Go(func() error {
+		filesCreated, err := sc.UntarAll(r, os.TempDir(), "")
```
twz123, line 123: We could also just write the tar file as-is to disk, so we have to deal with a single file only, instead of a bunch of them.

twz123 (approved):

PR #2422, opened 2022-11-17
Add unstructured and metadata utils to manifest applier

twz123:
## Description

Preparation for #2425.

Introduce some utilities to deal with the conversion of client-go types into unstructured objects, as required by the manifest applier. Add some functions to deal with common metadata, as well.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/unstructured.go
+++ b/pkg/applier/unstructured.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package applier
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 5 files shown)


pkg/applier/unstructured.go
```diff
@@ -0,0 +1,68 @@
...
+	if gvk.Group == "" || gvk.Kind == "" {
+		if s == nil {
+			s = scheme.Scheme
+		}
+
+		kinds, _, err := s.ObjectKinds(object)
+		if err != nil {
+			return nil, err
+		}
+		apiVersion, kind := kinds[0].ToAPIVersionAndKind()
```
juanluisvaladas, line 47: I'm not quite sure if I'm right, but I think s.ObjectKinds called in line 43 return the apiVersion as well. This would make calling this function redundant and if I'm not mistaken this would do the same:

```
		kinds, apiVersion, err := s.ObjectKinds(object)
		if err != nil {
			return nil, err
		}
		u.SetAPIVersion(apiVersion)
		u.SetKind(kinds[0])
	}

	return u, nil
```

Again I want to stress that I'm not 100% certain about this and I haven't tested it myself.

Other than this, the code looks good to me.

twz123: [runtime#Scheme.ObjectKinds](https://pkg.go.dev/k8s.io/apimachinery@v0.25.4/pkg/runtime#Scheme.ObjectKinds) returns `([]schema.GroupVersionKind, bool, error)`, basically a slice of possible GVKs.

juanluisvaladas: Ugh I should have checked better, sorry.

juanluisvaladas (approved):

jnummelin (approved):

PR #2425, opened 2022-11-18
Replace kubelet-config with worker-config component

twz123:
## Description

Preparation for #2429.

There will be the need to stuff more things into the worker profiles in the future. Prepare for that by refactoring the kubelet-config component into the new worker-config component.

Unify / move the worker profile loading bits found in `worker.Kubelet` and `worker.KubeletConfigClient` into the new workerconfig.Loader and use this at the beginning of the k0s worker bootstrapping. Have a new `workerconfig.Profile` struct that holds the `KubeletConfiguration` (and more stuff in the future). That struct is somewhat a counterpart to `v1beta1.ClusterSpec` for workers.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


cmd/controller/controller.go
```diff
@@ -430,8 +431,18 @@ func (c *command) start(ctx context.Context) error {
 		c.ClusterComponents.Add(ctx, metrics)
 	}
 
-	if !slices.Contains(c.DisableComponents, constant.KubeletConfigComponentName) {
-		c.ClusterComponents.Add(ctx, controller.NewKubeletConfig(c.K0sVars, adminClientFactory))
+	if !slices.Contains(c.DisableComponents, constant.WorkerConfigComponentName) {
```
soider, line 434: Let's keep ident level closer to the main code flow and split the branches:
- one if branch for instantiating and adding workerconfig reconciler
- one if branch for logging warning

Something like:

```go
if !slices.Contains(c.DisableComponents, constant.WorkerConfigComponentName) {
	reconciler, err := workerconfig.NewReconciler(c.K0sVars, c.NodeConfig.Spec, adminClientFactory, leaderElector)
	if err != nil {
		return err
	}
	c.ClusterComponents.Add(ctx, reconciler)
}
if slices.Contains(c.DisableComponents, constant.KubeletConfigComponentName) {
		logrus.Warnf("Usage of deprecated component name %q, please switch to %q",
				constant.KubeletConfigComponentName, constant.WorkerConfigComponentName,
			)
}
```

pkg/component/controller/workerconfig/cleaner.go
```diff
@@ -0,0 +1,114 @@
...
+		c.log.WithError(multierr.Combine(errs...)).Warn("Failed to remove some deprecated paths")
+	}
+}
+
+func (c *kubeletConfigCleaner) reconciled(ctx context.Context) {
+	ctx, cancel := context.WithCancel(ctx)
+	done := make(chan struct{})
+
+	// Execute this only once ...
+	if !c.doStop.CompareAndSwap(nil, func() {
```
soider, line 71: I would recommend to use https://pkg.go.dev/sync instead, since it does exactly what is required and it is a part of std library.


```go
c.Once.Do(func() {...})

**Review by soider — CHANGES_REQUESTED** [review]

File: cmd/controller/controller.go (hunk 1 above)
Comment (twz123) [reply cmd/controller/controller.go:434]: I think in this case, the deprecated label wouldn't disable the component, would it?

File: pkg/component/controller/workerconfig/cleaner.go (hunk 2 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/cleaner.go:71]: Yes, sync.Once makes it more straight forward, but it doesn't have anything built-in for async cancellation. How would the stop call cancel a concurrent reconciliation then? Do you have a suggestion?

```diff
diff --git a/pkg/component/controller/workerconfig/cleaner.go b/pkg/component/controller/workerconfig/cleaner.go
--- a/pkg/component/controller/workerconfig/cleaner.go
+++ b/pkg/component/controller/workerconfig/cleaner.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
-	"sync/atomic"
+	"sync"
 
 	"github.com/k0sproject/k0s/pkg/applier"
 	"github.com/k0sproject/k0s/pkg/kubernetes"
@@ -43,7 +43,7 @@ type kubeletConfigCleaner struct {
 	dir           string
 	clientFactory kubernetes.ClientFactoryInterface
 
-	doStop atomic.Value
+	reconcileOnce sync.Once
 }
 
 func (c *kubeletConfigCleaner) init() {
@@ -64,34 +64,18 @@ func (c *kubeletConfigCleaner) init() {
 }
 
 func (c *kubeletConfigCleaner) reconciled(ctx context.Context) {
-	ctx, cancel := context.WithCancel(ctx)
-	done := make(chan struct{})
-
-	// Execute this only once ...
-	if !c.doStop.CompareAndSwap(nil, func() {
-		cancel()
-		<-done
-	}) {
-		return
-	}
-
-	go func() {
-		defer func() {
-			close(done)
-			c.stop()
-		}()
-
+	go c.reconcileOnce.Do(func() {
 		if err := c.removeKubeletStack(ctx); err != nil {
 			c.log.WithError(err).Warn("Failed to remove deprecated kubelet stack")
 		}
-	}()
+	})
 }
 
 func (c *kubeletConfigCleaner) stop() {
-	if stop, ok := c.doStop.Swap(func() {}).(func()); ok {
-		stop()
-		return
-	}
+	c.reconcileOnce.Do(func() {
+		// This call either blocks until a concurrent reconciliation has
+		// finished, or prevents it from starting.
+	})
 }
 
 func (c *kubeletConfigCleaner) removeKubeletStack(ctx context.Context) error {
```

pkg/config/cli.go
```diff
@@ -173,6 +173,7 @@ func AvailableComponents() []string {
 		constant.HelmComponentName,
 		constant.MetricsServerComponentName,
 		constant.KubeletConfigComponentName,
```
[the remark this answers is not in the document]
twz123: I'll have a look. Needs a bit of refactoring, since this function is also used to validate the CLI args.

cmd/controller/controller.go
soider: if it is a necessity, (though personal I think it is better to fail with validation message during the arg parse step) we could modify c.DisableComponents accordingly by appending proper component name there.

pkg/component/controller/workerconfig/cleaner.go
makhov: What is happening here exactly? 
I don't fully follow, what the code does. It looks like there is no need for either `atomic.Value` and  `sync.Once`

pkg/component/worker/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+	kubeletv1beta1 "k8s.io/kubelet/config/v1beta1"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestToConfigMapData(t *testing.T) {
+	t.Parallel()
+
+	for _, test := range append(roundtripTests,
```
makhov, line 33: Let's just copy-paste all needed testcases. It's much easier to maintain and keeps the test much cleaner

pkg/component/worker/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+		roundtripTest{"nil_produces_zero", nil, map[string]string{}},
+	) {
+		t.Run(test.name, func(t *testing.T) {
+			test := test
+			t.Parallel()
```
makhov, line 38: Let's avoid using `t.Parallel` in the table tests: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

pkg/component/worker/config/loader.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Logger to use during loading of the worker configuration. Defaults to the
+	// standard logger if nil.
+	Log logrus.FieldLogger
+}
+
+// Load loads the configuration with the given profile name from Kubernetes.
+func (l *Loader) Load(ctx context.Context, profileName string) (*Profile, error) {
+	log := l.Log
+	if log == nil {
```
makhov, line 55: This and the `l.ClientFactory` check below could be in the constructor. Also, it helps to avoid constructions like this:
```
workerConfig, err := (&workerconfig.Loader{
	Kubeconfig: kubernetes.KubeconfigFromFile(kubeletKubeconfigPath),
}).Load(ctx, c.WorkerProfile)
```

pkg/component/worker/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+	if errs != nil {
+		return nil, errs
+	}
+
+	return data, nil
+}
+
+func forEachConfigMapEntry(profile *Profile, f func(fieldName string, ptr any)) {
+	for fieldName, ptr := range map[string]any{
+		"kubeletConfiguration": &profile.KubeletConfiguration,
```
makhov, line 110: What this function is supposed to do? Looks like it iterates only once, so why do we need it?

pkg/component/worker/config/config.go
```diff
@@ -0,0 +1,114 @@
...
+	p.DeepCopyInto(out)
+	return out
+}
+
+func (p *Profile) DeepCopyInto(out *Profile) {
+	*out = *p
+	p.KubeletConfiguration.DeepCopyInto(&out.KubeletConfiguration)
+}
+
+func (p *Profile) Validate(path *field.Path) (errs field.ErrorList) {
```
makhov, line 49: Looks like this function does nothing. If it's needed to implement an interface, it could be just `return nil`

pkg/component/controller/workerconfig/snapshot.go
```diff
@@ -0,0 +1,67 @@
...
+package workerconfig
+
+import (
+	"github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/v1beta1"
+
+	"github.com/sirupsen/logrus"
+)
+
+// snapshot holds a snapshot of a cluster's worker config.
+type snapshot struct {
```
makhov, line 26: Why it's needed? I fully follow, how it's used

pkg/component/controller/workerconfig/reconciler_test.go
```diff
@@ -0,0 +1,756 @@
...
+					DefaultPullPolicy: string(corev1.PullNever),
+				},
+			},
+		}))
+		assert.NotEmpty(t, applied(), "Expected some resources to be applied")
+		cleaner.awaitState(t, "reconciled")
+	})
+
+	t.Run("stops", func(t *testing.T) {
+		stopTest(t)
```
makhov, line 135: What is tested here? One run, zero asserts.

pkg/component/controller/workerconfig/reconciler_test.go
```diff
@@ -0,0 +1,756 @@
...
+			once.Do(func() {
+				assert.NoError(t, underTest.Stop())
+				assert.Equal(t, "stop", cleaner.state.Load())
+			})
+		}
+	}()
+
+	t.Run("starts", func(runT *testing.T) {
+		require.NoError(runT, underTest.Start(context.TODO()))
+		t.Cleanup(func() { stopTest(t) })
```
makhov, line 107: It would be called when the whole test will be finished, not just this subtest.

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,606 @@
...
+
+	reconcilersCtx, cancelReconcilers := context.WithCancel(context.Background())
+	var reconcilersDone sync.WaitGroup
+
+	// Will be locked during callbacks of the leader elector.
+	// TODO: Replace this with some "callback removal" procedure, once available.
+	var leaderElectorMu sync.Mutex
+
+	// Set to true once the reconciler gets stopped.
+	var stopCalled atomic.Bool
```
makhov, line 190: Well, three different async patterns in a row make this code really complicated. Is there any way to simplify it?

makhov (changes requested):
TBH, behind all these small fixes and refactorings, I'm a little lost in what the essence of this PR is, what problem it solves, and how it unblocks NLLB. 
Also, we are working on quite a complex system and need to keep the code as simple as possible, could we simplify it a bit?

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,606 @@
...
+// Reconciler maintains ConfigMaps that hold configuration to be
+// used on k0s worker nodes, depending on their selected worker profile.
+type Reconciler struct {
+	log logrus.FieldLogger
+
+	cleaner       cleaner
+	leaderElector leaderelector.Interface
+
+	mu            mtex
+	state         interface{}
```
soider, line 67: can be done instead by a custom type over string, like usually custom keys for context are implemented, for example. Easier to read + some type of safery from compiler. 

https://medium.com/@matryer/context-keys-in-go-5312346a868d

Though there is only one option for state field, why not make it more specific?

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,606 @@
...
+	"k8s.io/utils/pointer"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
+	"sigs.k8s.io/yaml"
+)
+
+type resources = []*unstructured.Unstructured
+
+type mtex = sync.Mutex
```
soider, line 56: I would suggest keep default name here. Though it saves some time during the writing code, it makes harder to understand.

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,606 @@
...
+	client kubernetes.Interface
+	apply  func(context.Context, resources) error
+}
+
+func (r *Reconciler) Init(context.Context) error {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	var created reconcilerCreated
+	switch state := r.state.(type) {
```
soider, line 131: Not sure, I follow, but what is the reason for this abstraction?

soider (changes requested):

pkg/component/controller/workerconfig/reconciler.go
twz123: That's a leftover from debugging. Will remove!

pkg/component/controller/workerconfig/reconciler_test.go
twz123: That was the intention. What about this?

```diff
diff --git a/pkg/component/controller/workerconfig/reconciler_test.go b/pkg/component/controller/workerconfig/reconciler_test.go
--- a/pkg/component/controller/workerconfig/reconciler_test.go
+++ b/pkg/component/controller/workerconfig/reconciler_test.go
@@ -94,8 +94,9 @@ func TestReconciler_Lifecycle(t *testing.T) {
 		}
 	}()
 
-	t.Run("starts", func(runT *testing.T) {
-		require.NoError(runT, underTest.Start(context.TODO()))
+	t.Run("starts", func(startT *testing.T) {
+		require.NoError(startT, underTest.Start(context.TODO()))
+		// Ensure the component gets stopped when the parent test exits
 		t.Cleanup(func() { stopTest(t) })
 	})

File: pkg/component/controller/workerconfig/reconciler_test.go (hunk 10 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler_test.go:135]: It's tested that the component stops cleanly; stopTest will assert that.

File: pkg/component/worker/config/loader.go (hunk 6 above)
Comment (twz123) [reply pkg/component/worker/config/loader.go:55]: I'll check out how it looks like if refactored into a free-standing function without a struct. Would probably be a better fit anyways.

File: pkg/component/controller/workerconfig/reconciler.go (hunk 12 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler.go:190]: I think I found a better solution. Will update the PR accordingly.

File: cmd/controller/controller.go (hunk 1 above)
Comment (twz123) [reply cmd/controller/controller.go:433]: Moving this to a validation step makes absolute sense. I think accepting/rewriting the old flag and issuing a warning is the best way for a deprecation. If k0s would stop accepting the old value, this would be a breaking change. I also discovered that some other components were missing from the list of available components. I opened #2474 for those things.

File: pkg/component/controller/workerconfig/cleaner.go (hunk 2 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/cleaner.go]: This code was intended to remove the deprecated stack from the cluster. As this is rather long-running, the intent was to run it once in a goroutine that should be "cancellable" and "awaitable" for completion.

However, the better and much easier approach is not to remove the old stack, but add some deprecation notes to it. Otherwise, old 1.25 workers wouldn't work with a 1.26 control plane. The actual removal of the stack can be done in 1.27, by simply removing (or renaming) all `*.yaml` files in the `manifests/kubelet` directory.

So all of this is gone by now.

File: pkg/config/cli.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (twz123) [reply pkg/config/cli.go]: Done: the deprecated component name is no longer listed in the CLI help text.

File: pkg/component/worker/config/loader.go (hunk 6 above)
Comment (twz123) [reply pkg/component/worker/config/loader.go]: Done by converting the `Loader` struct into the `LoadProfile` function.

File: pkg/component/controller/workerconfig/reconciler.go (hunk 12 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler.go]: Done. No more atomics. Please have another look.

File: pkg/component/controller/workerconfig/reconciler.go (hunk 14 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler.go]: Done.

File: pkg/component/controller/workerconfig/reconciler_test.go (hunk 10 above)
Comment (makhov) [reply pkg/component/controller/workerconfig/reconciler_test.go:120]: The good pattern for unit testing is AAA: Arrange-Act-Assert. In the ideal world, every test should have all three As.

File: pkg/component/controller/workerconfig/reconciler_test.go (hunk 11 above)
Comment (makhov) [reply pkg/component/controller/workerconfig/reconciler_test.go:95]: It's still unclear, why we call it here and what it's supposed to do. Again, back to AAA, what is the assert here?

Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler_test.go:95]: The arrangement has been done by the previous subtests. The action is `underTest.Start(context.TODO())`. The assertion is that the component could be started, i.e. the Start method didn't return an error. The cleanup function on the parent test is to ensure that the (now started) component will be stopped when the lifecycle test finishes.

The lifecycle test ensures that the reconciler adheres to the component lifecycle. Each subtest builds upon the previous ones. Maybe it helps to make it clear what's going on when looking at the test structure only:

```
--- PASS: TestReconciler_Lifecycle (0.00s)
    --- PASS: TestReconciler_Lifecycle/fails_to_start_without_init (0.00s)
    --- PASS: TestReconciler_Lifecycle/init (0.00s)
    --- PASS: TestReconciler_Lifecycle/another_init_fails (0.00s)
    --- PASS: TestReconciler_Lifecycle/starts (0.00s)
    --- PASS: TestReconciler_Lifecycle/another_start_fails (0.00s)
    --- PASS: TestReconciler_Lifecycle/reconciles (0.00s)
    --- PASS: TestReconciler_Lifecycle/stops (0.00s)
    --- PASS: TestReconciler_Lifecycle/stop_may_be_called_again (0.00s)
    --- PASS: TestReconciler_Lifecycle/reinit_fails (0.00s)
    --- PASS: TestReconciler_Lifecycle/restart_fails (0.00s)
```

In this case the subtests are not independent of each other, but more like consecutive steps. An alternative would be to replace the subtests with some block comments and additional message params for the assertions.

File: pkg/component/controller/workerconfig/reconciler.go (hunk 15 above)
Comment (twz123) [reply pkg/component/controller/workerconfig/reconciler.go]: `state` encapsulates the current state of the component state machine. State can be one of created, initialized, started and stopped. The Init method performs the state transition from created to initialized. The switch statement was used to assert that the Reconciler is in the right state for the transition (the "created" state).

By having an opaque state variable and one type for each possible state, one needs to use a type assertion to access the data associated with a given state. So there's no way to accidentally use something which hasn't been initialized or is destroyed already, since one needs to prove to the compiler first that the state to be accessed is actually the state we're in.

A concrete example for the Reconciler is the update channel that receives state updates. Having this in the reconcilerStarted struct instead of the Reconciler struct itself expresses that the channel is only valid as long as the Reconciler is started, not before or after. There's no way of accessing it without checking that we're in the right state beforehand, i.e. by `if state, ok := r.state.(reconcilerStarted); ok { /* safely access the update channel */ } `. At the same time, having a reference to the concrete state type can be a proof that the check has already been performed and doesn't need to be repeated, e.g. by passing on the concrete state when calling other helper functions that assume the Reconciler to be in that same state.

Returning to the code at hand: Since there's only one valid source state in which Init is allowed to be called, the switch isn't necessary. It just needs to check if the current state is "created". So I replaced this and similar checks with a simple if, e.g. like this:

```go
	if _, ok := r.state.(reconcilerCreated); !ok {
		return fmt.Errorf("cannot initialize: %T", r.state)
	}
```

pkg/component/controller/workerconfig/reconciler.go
twz123: There are actually four states: created, initialized, started and stopped. They each contain different data. Newtypes over strings will be able to express the current state's "display name", but wouldn't be able to hold the data associated with each state.

Since the states themselves don't share any commonalities, there's no natural way of letting them implement a common interface that could be used as the `state` variable's type to make it more strict. There's of course the possibility of using something along the lines of `type reconcilerState interface { isReconcilerState() }` and implement that for each of the states to mimic some sort of a "marker" interface, but not sure if that's a meaningful approach.

Btw: What's the correct name for those custom `type Foo Bar` alias types in Golang? This pattern is called "newtypes" in other programming languages. Does Go use the same name, or is there another term for those?

cmd/controller/controller.go
twz123: Done by normalizing the deprecated component name to the new component name in the config package.

pkg/component/controller/workerconfig/reconciler_test.go
makhov: > Each subtest builds upon the previous ones

> In this case the subtests are not independent of each other, but more like consecutive steps

That's exactly the problem. Tests should be independent and self-sufficient. Anyone should be able to run the single subtest without any trouble.

twz123: Done.

pkg/component/controller/workerconfig/reconciler_test.go
twz123: Done.

pkg/component/worker/config/config_test.go
twz123: Will do. That's a remnant of the past. There used to be more tests here.

pkg/component/worker/config/config.go
twz123: Removed.

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,539 @@
...
+type Reconciler struct {
+	log logrus.FieldLogger
+
+	clusterDomain string
+	clusterDNSIP  net.IP
+	clientFactory kubeutil.ClientFactoryInterface
+	leaderElector leaderelector.Interface
+
+	mu    sync.Mutex
+	state any
```
soider, line 64: I still have concerns about untyped state.
Basically what we are trying to do here is somehow implement union type in the language which has no union types. Trying to implement feature which do not exists in the language is always fishy to me and IMO leads to over engineered code. 

My suggestion here just to use type state string and use it for explicit compare like 
```
const (
stateStopped = "stateStopped"
)


if r.state == stateStopped {
    go t.requestStop() // or somehow else
}
```


That would require somehow simplify the code at line 363
```
	case reconcilerStarted:
			r.state = reconcilerStopped{state.stopped}
			go state.requestStop()
			return state.stopped, nil
		case reconcilerStopped:
			return state.stopped, nil
```

but from my perspective it would a beneficial to us, because it is hard to read now (I still not sure I fully understand it tbh).


Also, not related to this particular line the comment left for, but may I suggest if we start targetting less generic, but more specific type of a code? it is internal thing, we read it more often than write it, and I would suggest to try aiming for the simpler code.

CC: @makhov @jnummelin wdyt?

soider (changes requested):
Left a suggestion for state management and opener for a wider discussion.

pkg/component/controller/workerconfig/reconciler.go
twz123: Done. Using newtyped strings to track the current state now.

soider (dismissed):
there are still some minor discussions not covered, at least the one about snapshot structure, but generally lgtm.

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -0,0 +1,535 @@
...
+}
+
+var errStoppedConcurrently = errors.New("stopped concurrently")
+
+// reconcile enqueues the given update and awaits its reconciliation.
+func reconcile(ctx context.Context, updates chan<- updateFunc, stopped <-chan struct{}, update func(*snapshot)) error {
+	recoDone := make(chan error, 1)
+
+	select {
+	case updates <- func(s *snapshot) chan<- error { update(s); return recoDone }:
```
makhov, line 328: IIUC we are sending a func that returns a channel and wraps another func that is passed to the `reconcile` func, to the channel, and another function in a goroutine will read it from the channel and execute. By executing the function we either update a counter or store a new clusterConfig to a variable. And after this, we finally do the real reconciliation (eg creating and writing configmaps to disk). 

Am I correct?

To add to that, I think that the complexity here is quite high since it took me a significant amount of time to wrap my head around the code and even then I'm not quite sure I got it right.

I think it's worth putting some effort into streamlining the code to ensure that I (and others) can clearly grasp what's happening here in the future. I personally agree with the ideas from https://google.github.io/eng-practices/review/reviewer/looking-for.html#complexity when it comes down to judging code complexity.

But if it does block the release, we may do so in a separate PR though. WDYT, @jnummelin?

pkg/component/worker/config/config_test.go
makhov: Well, actually I'm ok with the version in the follow-up PR. Using `append` instead of duplicating a couple of cases just looks sloppy here

pkg/component/controller/workerconfig/reconciler.go
jnummelin: I do agree that this is quite complex to wrap head around.

And I also do agree that we can sort this out in subsequent PR(s), given we properly add tracking issue(s) as needed.

makhov (dismissed):

docs/configuration.md
```diff
@@ -403,21 +410,25 @@ spec:
...
+k0s allows to completely disable some of the system components. This allows
+users to build a minimal Kubernetes control plane and use what ever components
+they need to fulfill their need for the control plane. Disabling the system
+components happens through a command line flag for the controller process:
 
 ```sh
---disable-components strings                     disable components (valid items: api-config,autopilot,control-api,coredns,csr-approver,endpoint-reconciler,helm,konnectivity-server,kube-controller-manager,kube-proxy,kube-scheduler,kubelet-config,metrics-server,network-provider,node-role,system-rbac)
-
+--disable-components strings                     disable components (valid items: api-config,autopilot,control-api,coredns,csr-approver,endpoint-reconciler,helm,konnectivity-server,kube-controller-manager,kube-proxy,kube-scheduler,metrics-server,network-provider,node-role,system-rbac,worker-config)
```
juanluisvaladas, line 419: I think it would make sense to add here a deprecation warning of the kubelet-config as well .

pkg/config/cli_test.go
```diff
@@ -54,4 +54,15 @@ func TestControllerOptions_Normalize(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, expected, underTest.DisableComponents)
 	})
+
+	t.Run("replacesDeprecation", func(t *testing.T) {
+		disabled := []string{"helm", "kubelet-config", "coredns", "kubelet-config", "worker-config", "autopilot"}
```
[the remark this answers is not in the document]
twz123: Yeah, this is more "replaceDeprecationAndRemoveDupes". I think it makes sense to have this in one test case, i.e. it tests that replacing the deprecation doesn't leave the new component twice in the resulting slice. I do see the issue with the not absolutely accurate test naming, though. Any advice on a better name, or how to split it up?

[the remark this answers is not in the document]
juanluisvaladas: My concern isn't so much the name and not the test with its current implementation but detecting breaking changes in the future. I think this test case is good and valid but it would make sense also test the string slice without worker-config.

docs/configuration.md
twz123: Added a note about kubelet-config being replaced by worker-config.

pkg/config/cli_test.go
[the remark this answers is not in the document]
twz123: Added a third test case as suggested.

pkg/token/manager.go
```diff
@@ -21,14 +21,15 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/sirupsen/logrus"
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	"github.com/k0sproject/k0s/internal/pkg/random"
```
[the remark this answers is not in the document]
twz123: Done. Reverted the superfluous edits.

pkg/kubernetes/lease.go
```diff
@@ -38,7 +38,7 @@ func IsValidLease(lease coordinationv1.Lease) bool {
 func GetControlPlaneNodeCount(ctx context.Context, kubeClient kubernetes.Interface) (int, error) {
 	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
-	count := 0
+	var count int
```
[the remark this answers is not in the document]
twz123: Reverted.

makhov (approved):

PR #2428, opened 2022-11-18
Enhanced config CRD validation

twz123:
## Description

Preparation for #2429.

Add more validation to several CRD types and improve the error messages.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
+++ b/pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
@@ -19,6 +19,7 @@ package v1beta1
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"reflect"
 
@@ -107,17 +108,25 @@ type InstallSpec struct {
 	SystemUsers *SystemUser `json:"users,omitempty"`
 }
 
+var _ Validateable = (*InstallSpec)(nil)
+
+func (*InstallSpec) Validate() []error { return nil }
+
 // ControllerManagerSpec defines the fields for the ControllerManager
 type ControllerManagerSpec struct {
 	// Map of key-values (strings) for any extra arguments you want to pass down to the Kubernetes controller manager process
 	ExtraArgs map[string]string `json:"extraArgs,omitempty"`
 }
 
+var _ Validateable = (*ControllerManagerSpec)(nil)
+
 func DefaultControllerManagerSpec() *ControllerManagerSpec {
 	return &ControllerManagerSpec{
 		ExtraArgs: make(map[string]string),
 	}
 }
+
+func (c *ControllerManagerSpec) Validate() []error { return nil }
 
 // SchedulerSpec defines the fields for the Scheduler
 type SchedulerSpec struct {
@@ -130,6 +139,10 @@ func DefaultSchedulerSpec() *SchedulerSpec {
 		ExtraArgs: make(map[string]string),
 	}
 }
+
+var _ Validateable = (*SchedulerSpec)(nil)
+
+func (*SchedulerSpec) Validate() []error { return nil }
 
 // +kubebuilder:object:root=true
 // +genclient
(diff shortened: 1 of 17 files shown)


pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -277,45 +282,50 @@ func DefaultClusterSpec(defaultStorage ...*StorageSpec) *ClusterSpec {
...
+		"controllerManager": s.ControllerManager,
+		"scheduler":         s.Scheduler,
+		"storage":           s.Storage,
+		"network":           s.Network,
+		"workerProfiles":    s.WorkerProfiles,
+		"telemetry":         s.Telemetry,
+		"install":           s.Install,
+		"extensions":        s.Extensions,
+		"konnectivity":      s.Konnectivity,
+	} {
```
jnummelin, line 307: Would it be overkill to use reflection here to loop over all the fields instead of manually maintaining this list?

twz123: I think the code wouldn't look nice. There's a lot of possible errs that need to be checked, and there's potential for panics. Eventually it might be worthwhile to have a look at kubebuilder. I know for example that it can handle the defaulting stuff that's currently done done manually in k0s. Maybe there's something in there to generate validations, too.

makhov: A cheap solution to make sure all fields are in the list is to add a test that checks for this using reflection.

makhov (dismissed):

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
soider: ...so we would have both reflection and manually maintained list at the same time :trollface:

soider (dismissed):
good stuff, though I am not sure I follow how it blocks the #2429

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
makhov: Precisely my point 😄
But jokes aside, the code would be easy to read and follow, and a test will keep us from forgetting new fields.

twz123:
> good stuff, though I am not sure I follow how it blocks the #2429

This is not necessarily blocking, but it validates every piece of configuration that's used in #2429. So config errors should pop up early, instead of breaking nllb stuff later on in potentially hard-to-debug ways.

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
makhov: If we want to use reflection, this would be enough:
```
val := reflect.ValueOf(s).Elem()
for pos := 0; pos <= val.NumField() - 1; pos++ {
	tag, ok := val.Type().Field(pos).Tag.Lookup("json")
	if ok {
		for _, err := range val.Field(pos).Interface().(Validateable).Validate() {
			errs = append(errs, fmt.Errorf("%s: %w", tag, err))
		}
	}
}
```
But I'm not still sure that we need it

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -277,45 +282,50 @@ func DefaultClusterSpec(defaultStorage ...*StorageSpec) *ClusterSpec {
 	}
 }
 
-func (c *ControllerManagerSpec) Validate() []error {
-	return nil
+// Validateable interface to ensure that all config components implement Validate function
+// +k8s:deepcopy-gen=false
+type Validateable interface {
```
makhov, line 287: Btw, it's probably a typo

pkg/apis/k0s.k0sproject.io/v1beta1/clusterconfig_types.go
```diff
@@ -277,45 +282,50 @@ func DefaultClusterSpec(defaultStorage ...*StorageSpec) *ClusterSpec {
...
+	Validate() []error
 }
 
-var _ Validateable = (*SchedulerSpec)(nil)
```
makhov, line 284: Why is it removed?

twz123: I suppose this is a remnant from previous undone edits. SchedulerSpec's Validate method is a no-op, and I probably thought of removing it completely, but then decided against it for consistency. Will add it back.

twz123: Done

makhov (approved):

PR #2429, opened 2022-11-18
Node-local load balancing

twz123:
## Description

Node-local load balancing.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


cmd/controller/controller.go
```diff
@@ -430,8 +431,18 @@ func (c *command) start(ctx context.Context) error {
 		c.ClusterComponents.Add(ctx, metrics)
 	}
 
-	if !slices.Contains(c.DisableComponents, constant.KubeletConfigComponentName) {
-		c.ClusterComponents.Add(ctx, controller.NewKubeletConfig(c.K0sVars, adminClientFactory))
+	if !slices.Contains(c.DisableComponents, constant.WorkerConfigComponentName) {
```
soider, line 434: same suggestion as in https://github.com/k0sproject/k0s/pull/2425#discussion_r1033538534

pkg/apis/k0s.k0sproject.io/v1beta1/network.go
```diff
@@ -47,59 +55,74 @@ type Network struct {
...
+		errors = append(errors, field.NotSupported(field.NewPath("provider"), n.Provider, []string{"kuberouter", "calico", "custom"}))
 	}
 
 	_, _, err := net.ParseCIDR(n.PodCIDR)
 	if err != nil {
-		errors = append(errors, fmt.Errorf("invalid pod CIDR %s", n.PodCIDR))
+		errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "invalid CIDR address"))
 	}
 
+	// if n.ServiceCIDR != "" {
```
soider, line 88: leftover

pkg/apis/k0s.k0sproject.io/v1beta1/network.go
```diff
@@ -47,59 +55,74 @@ type Network struct {
...
 // Validate validates all the settings make sense and should work
 func (n *Network) Validate() []error {
+	if n == nil {
+		return nil
+	}
+
 	var errors []error
-	if n.Provider != "calico" && n.Provider != "custom" && n.Provider != "kuberouter" {
-		errors = append(errors, fmt.Errorf("unsupported network provider: %s", n.Provider))
+
```
soider, line 76: may we could utilize early return here, instead of trying to collect all errors in one run

soider (changes requested):

pkg/apis/k0s.k0sproject.io/v1beta1/network.go
twz123: Removed

cmd/controller/controller.go
twz123: Done in the other PR.

pkg/apis/k0s.k0sproject.io/v1beta1/network.go
twz123: This pattern is the standard in the Kubernetes code base, i.e. all the resource validation methods return error lists. I think this is convenient for users. When you have three errors in your resource, it's easier if all three errors are reported in one go. Otherwise you end up fixing the first, retry, get a new error, fix the second, retry, ...

inttest/nllb/nllb_test.go
```diff
@@ -0,0 +1,333 @@
...
+		s.T().Logf("Checking if HA cluster is ready")
+		s.Require().NoError(s.checkClusterReadiness(ctx, clients, s.ControllerCount))
+	}))
+
+	workerNameToRestart := s.WorkerNode(0)
+	s.Require().True(s.Run("CycleControllers", func() {
+		for i := 0; i < s.ControllerCount; i++ {
+			controllerName := s.ControllerNode(i)
+			s.Require().True(s.Run(controllerName, func() {
+				s.Require().True(s.Run("StopWorker", func() {
```
jnummelin, line 137: Why all the nesting of `s.Require.True(s.Run(...`? I do get the "flow" of the test case here but I do not get the "nesting" part.

inttest/common/launchdelegate.go
```diff
@@ -55,29 +58,37 @@ var _ launchDelegate = (*standaloneLaunchDelegate)(nil)
 // InitController initializes a controller in "standalone" mode, meaning that
 // the k0s executable is launched directly (vs. started via a service manager).
 func (s *standaloneLaunchDelegate) InitController(ctx context.Context, conn *SSHConnection, k0sArgs ...string) error {
```
twz123, line 60: When adding support for stopping and restarting, I asked myself if it's still meaningful to keep the standalone mode around? Is there any situation in which standalone mode would enable something that OpenRC doesn't? Adding the new features somehow felt a bit like reimplementing an init system on the fly.

OTOH, standalone mode could be kept, but not supporting all of the operations and return an error for those that are unsupported.

soider: Do you mean for test purposes or in general?


In general I am strictly against dropping standalone mode (if be standalone you mean running binary directly without init system). It would be strange, if we'd claim to be single binary distro and require init system at the same time

inttest/nllb/nllb_test.go
soider: I have the question but not about nested .Run. 

What is the reason to have explicit checks for the result t.Run?

twz123: > What is the reason to have explicit checks for the result t.Run?

This is to be fail-fast and stop the parent test when the first subtest fails. Otherwise, the suite would continue with the next subtest, which would be unlikely to succeed given that its predecessor(s) failed. I found this to be the least cumbersome solution in conjunction with subtests. An alternative would be to not use subtests at all and rely more on s.T().Log() to report progress. Personally, I find the subtest output a bit more appealing, and it also gives timings for each step out of the box.

soider: We use Require in those sub-tests.
According to the documentation Require should fail fast already. 

Doesn't it work as expected?

twz123: The code currently uses `s.Run(string, func())` instead of `s.T().Run(string, func(*testing.T))`. The former adjusts the suite's test pointer to the subtest as long as the subtest function is being executed, hence eliding the additional `*testing.T` from that function's signature. So all assertions via `s` called from within the subtest function will operate on the subtest, not the parent test.

That behavior can be bypassed by using `s.T().Run(string, func(*testing.T))` instead and not using the subtest's `t`. This gives weirder errors and log statements, though. Because the subtest is never used, everything operates on the parent test, making the subtest rather obsolete. Especially calling `FailNow` (that's what testify's require does) on a parent test from within a subtest always gives messages like "testing.go:1343: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test".

To clarify the current behavior, it could be beneficial to introduce a helper function on the suite, like so:

```go
func (s *Suite) runOrFailNow(name string, f func()) {
	if !s.Run(name, f) {
		s.T().FailNow()
	}
}
```

Then the current

```go
s.Require().True(s.Run("subtest", func() {
	// ...
}))
```

would become

```go
s.runOrFailNow("subtest", func() {
	// ...
})
```

Removing the subtests alltogether remains another option: adding more `s.T().Logf(...)` calls or augmenting existing log statements with additional context.

soider: what if you use the parent t in the subtest instead of its own t? for example by naming the argument as _ to not overshadow the parent one?

twz123: > what if you use the parent t in the subtest instead of its own t? for example by naming the argument as _ to not overshadow the parent one?

Using the parent test from within the subtest yields to the stuff I tried to describe here:

> This gives weirder errors and log statements, though. [...] Especially calling FailNow (that's what testify's require does) on a parent test from within a subtest always gives messages like "testing.go:1343: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test".

For example, testify errors will reference the parent test instead of the subtest as the failed test, and when using t.Log, this will always look like below, giving the impression that the subtest would do nothing, or run silently in parallel:

```text
=== RUN   parent/subtest
=== CONT  parent
    ...
```

Moreover, having references to both the parent test and the subtest around will always lead to the question of which one to use. Logging could be done via subtests, but assertions via the parent? This is going to get complicated pretty quickly.

A final option besides the helper method or the omission of subtests would be to just not be fail-fast and let subsequent subtests run and have them fail as well. This will be noisier when test failures actually occur, as each subsequent subtest will report its own failures, but should be more or less fine otherwise. (There might be some problems during test timeouts, though.)

soider: >A final option besides the helper method or the omission of subtests would be to just not be fail-fast and let subsequent subtests run and have them fail as well

I vote for this. 

Another option to consider is instead of t.Run use s.Run which seems to be merged with that PR https://github.com/stretchr/testify/pull/655/files

>Adds a .Run(..) method onto test suites, mimicking the semantics of [testing.Run](https://godoc.org/testing#T.Run). This will make writing table-driven tests in test suites cleaner and simpler, by handling the management of *testing.T instances transparently so that sub-tests can use the same suite.Assert() and suite.Require() methods, and have all output and test logic properly fall under the correct subtest.

@jnummelin @makhov what's your opinion here, folks?

soider: Oh we are using s.Run already... I wonder why it doesn't fall as expected... Well, then I vote for not having fail-fast approach now.

makhov: I don't understand, what is the problem here. `require` fails the current test and I'd say it's exactly the expected behavior. I do want to know which tests are failing and which are passing.

> This is to be fail-fast and stop the parent test when the first subtest fails

I don't see why we need to be fail-fast in unit tests. But if we need to, it may indicate the misuse of subtests

twz123: > I wonder why it doesn't fall as expected

I think it behaves as intended. It returns the outcome of the subtest as a bool, just the same way as `t.Run` does it. So we need to look at that bool ourselves.

>  vote for not having fail-fast approach now.

Will do.

inttest/common/launchdelegate.go
```diff
@@ -38,8 +38,11 @@ const (
 // [FootlooseSuite] so that alternate behavior can be performed.
 type launchDelegate interface {
 	InitController(ctx context.Context, conn *SSHConnection, k0sArgs ...string) error
+	StartController(ctx context.Context, conn *SSHConnection) error
```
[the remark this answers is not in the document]
twz123: This addition is used for the NLLB inttest, yes. I can open a separate PR with just those changes if you prefer.

inttest/nllb/nllb_test.go
twz123: Done. Flattened the test structure and removed the "fail-fastness".

makhov (approved):
The conformance test run has finished successfully:
```
…
13:28:06       e2e   global   running            Passed:359, Failed:  0
...
...
13:29:06       e2e   global   running            Passed:360, Failed:  0
13:29:26       e2e   global   running            Passed:368, Failed:  0
13:29:46       e2e   global   complete   passed   Passed:368, Failed:  0
```

mikhail-sakhnov (approved):

PR #2442, opened 2022-11-22
Allow to configure etcd extraArgs

### LINKED ISSUE #1543: I want extraArgs for etcd as well.

I don't think I need to explain the reason in depth, but sometimes I want to tune the timeout and limit of etcd.

Make sure that etcd can add arguments freely with extraArgs.

_No response_

### Additional context

_No response_

### LINKED ISSUE #2369: Make etcd configurable

We are running k0s cluster across datacenters. K0s runs the etcd cluster with default settings which work well for local network where the average latency is low, but doesn't work well for our case (see https://etcd.io/docs/v3.5/tuning/ for more details)

The etcd cluster should be configurable/customizable, either via `extraArgs` and/or by exposing some important configuration options via k0s config directly. 

In our opinion `--heartbeat-interval` and `--election-timeout` would be important enough to deserve being exposed directly as config options. 

Using `ETCD_HEARTBEAT_INTERVAL` and `ETCD_ELECTION_TIMEOUT` to configure etcd via [environment variables](https://docs.k0sproject.io/v1.25.3+k0s.0/environment-variables/) should work, but is not that frictionless as k0s promises. 

Using that approach also has the risk that one can have conflicting configuration with what k0s does. Integrating etcd configuration options directly into the config would have the benefit that some validation could be performed.  

### Additional context

_No response_

juanluisvaladas:
## Description

This allows the user to configure random flags in the etcd process.

Fixes #1543
Fixes #2369 

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


docs/configuration.md
```diff
@@ -53,6 +53,7 @@ spec:
     type: etcd
     etcd:
       peerAddress: 192.168.68.104
+      extraArgs: []
```
jnummelin, line 56:
```suggestion
      extraArgs: {}
```
After all, it's an `map[string]string`, right?

juanluisvaladas: Ah yes. every other extraArgs was also wrong, so I'm fixing the rest in a separate commit.

pkg/component/controller/etcd.go
```diff
@@ -195,6 +195,14 @@ func (e *Etcd) Start(ctx context.Context) error {
 		args["--auth-token"] = auth
 	}
 
+	for name, value := range e.Config.ExtraArgs {
+		argName := fmt.Sprintf("--%s", name)
+		if args[argName] != "" {
+			logrus.Warnf("overriding apiserver flag with user provided value: %s", argName)
```
jnummelin, line 201:
```suggestion
			logrus.Warnf("overriding etcd flag with user provided value: %s", argName)
```

jnummelin (dismissed):
Minor nit on the logging :)

pkg/component/controller/etcd.go
```diff
@@ -195,6 +195,14 @@ func (e *Etcd) Start(ctx context.Context) error {
 		args["--auth-token"] = auth
 	}
 
+	for name, value := range e.Config.ExtraArgs {
+		argName := fmt.Sprintf("--%s", name)
+		if args[argName] != "" {
```
twz123, line 200: Better use ok in case we have valueless flags?

```suggestion
		if _, ok := args[argName]; ok {
```

juanluisvaladas: That's a very valid point, since you noticed this I'll also fix the other three instances where this happens.

soider (approved):

PR #2471, opened 2022-12-01
Add basic scaffolding to deploy k0s on libvirt

juanluisvaladas:
## Description
Usually using footloose is enough to deploy test clusters, however sometimes you need a real cluster with real machines.

A real cluster can be provided with `hack/tool` on AWS but this is fairly slow, AWS is expensive and troubleshooting this may be difficult.

This PR uses kcli (a FOSS tool to manage VMs using, among other things, libvirt) to deploy libvirt virtual machines and even full clusters with **literally** one command . Right now only an all in one cluster is supported with literally one single command but eventually with little work we'll be able to deploy multi clusters, load balancers, etc. To show how fast this is:

```
$ date; kcli create plan -f aio.yml
jue 01 dic 2022 23:43:48 CET
Using stupefied-citellus as name of the plan
Deploying Vms...
Injecting private key for allinone
Hypervisor not compatible with nesting. Skipping
allinone deployed on local
<snip>
$ kcli ssh allinone
<snip>
ubuntu@allinone:~$ sudo tail -1 /var/log/k0sctl-install.log 
Thu Dec  1 22:44:53 UTC 2022
```
In just one minute and 10 seconds k0sctl has finished running. Of course there are extra steps when installing a cluster until it's actually utilizable but it's not much longer

This PR as is has some values hardcoded for my machine and the documentation isn't 100% correct.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


jnummelin:
[note]Did not dive too deep into this yet, so might be bit hazy on the details...[/note]

As `hack/tool` is predominantly using Terraform as the infra level tooling, would it not make sense to extend that with support for libvirt instead? Introducing different `hack/xyz` tools makes everything a little brittle when each tool has it's own UX and quirks.

juanluisvaladas:
> As hack/tool is predominantly using Terraform as the infra level tooling, would it not make sense to extend that with support for libvirt instead? 

The problem with terraform in a container is that it downloads the providers in every run and it's extremely slow. By the time terraform has finished downloading providers this has deployed a full cluster.

PR #2474, opened 2022-12-02
Introduce Normalize method for controller options

twz123:
## Description

Move the existing validation out of controller.go into a method on ControllerOptions, so that the functionality becomes easier to test. Moving the logic also allows for replacing AvailableCommands() with a private global variable availableCommands, since it's no longer required to be visible outside the config package.

Also: Add missing components to the list of available components:

* api-config
* autopilot
* node-role

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/config/cli_test.go
+++ b/pkg/config/cli_test.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"testing"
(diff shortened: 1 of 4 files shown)


soider (approved):

pkg/config/cli_test.go
```diff
@@ -0,0 +1,59 @@
...
+	expected := slices.Clone(availableComponents)
+	slices.Sort(expected)
+
+	assert.Equal(t, expected, availableComponents, "Available components aren't sorted")
+
+	slices.Compact(expected)
+	assert.Equal(t, expected, availableComponents, "Available components contain duplicates")
+}
+
+func TestControllerOptions_Normalize(t *testing.T) {
```
juanluisvaladas, line 39: Since the other test is run with `t.Parallel()`, perhaps this one should be also ran with `t.Parallel()` as well.
Other than that, LGTM.

twz123: Doh. Will remove the other Parallel as well to make it homogeneous again. Technically those tests could be parallelized, but since t.Parallel is sometimes tricky (depending on the test setup), it might make sense to not use it at all.

juanluisvaladas (approved):
LGTM

pkg/config/cli_test.go
juanluisvaladas: Fair enough

PR #2494, opened 2022-12-09
[Backport release-1.25] Fix propagation of default storage class setting

makhov:
Backport to release-1.25, see https://github.com/k0sproject/k0s/pull/2482.

### DIFF
--- a/pkg/constant/constant_shared.go
+++ b/pkg/constant/constant_shared.go
@@ -93,7 +93,7 @@ const (
 	KubeRouterCNIInstallerImage        = "quay.io/k0sproject/cni-node"
 	KubeRouterCNIInstallerImageVersion = "1.1.1-k0s.0"
 	OpenEBSRepository                  = "https://openebs.github.io/charts"
-	OpenEBSVersion                     = "3.0.3"
+	OpenEBSVersion                     = "3.3.0"
 
 	// Controller component names
 	APIConfigComponentName             = "api-config"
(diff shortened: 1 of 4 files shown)


twz123 (approved):

pkg/constant/constant_shared.go
```diff
@@ -93,7 +93,7 @@ const (
 	KubeRouterCNIInstallerImage        = "quay.io/k0sproject/cni-node"
 	KubeRouterCNIInstallerImageVersion = "1.1.1-k0s.0"
 	OpenEBSRepository                  = "https://openebs.github.io/charts"
-	OpenEBSVersion                     = "3.0.3"
+	OpenEBSVersion                     = "3.3.0"
```
twz123, line 96: Given that this is not a patch but a minor version bump:

1. Do we need that version bump for the default storage class?
2. Are we okay with a minor dependency update in a k0s patch release?

jnummelin: > Do we need that version bump for the default storage class?
AFAIK no

> Are we okay with a minor dependency update in a k0s patch release?
I think we should stick with `3.0.3`

cc/ @makhov

makhov: The problem here is that OpenEBS 3.0.* doesn't have `isDefaultClass` option and couldn't be set as the default class

jnummelin: Ah. That mean that this is good as-is then. 👍

jnummelin (approved):

PR #2544, opened 2022-12-21
Fix the test custom-domain flakyness

juanluisvaladas:
## Description
The test custom-domain has become flaky. I'm not quite sure when did this happen, but it was recently. We can see in many executions such as: 

https://github.com/k0sproject/k0s/actions/runs/3742332086/jobs/6359020341
https://github.com/k0sproject/k0s/actions/runs/3740780796/jobs/6349616080
https://github.com/k0sproject/k0s/actions/runs/3731323166/attempts/2

We can solve this simply by waiting for konnectivity to be ready before actually starting the test.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/inttest/customdomain/customdomain_test.go
+++ b/inttest/customdomain/customdomain_test.go
@@ -57,6 +57,7 @@ func (s *CustomDomainSuite) TestK0sGetsUpWithCustomDomain() {
 		_, err = ssh.ExecWithOutput(s.Context(), "/usr/local/bin/k0s kc run nginx --image docker.io/nginx:1-alpine")
 		s.Require().NoError(err)
 		s.NoError(common.WaitForPod(s.Context(), kc, "nginx", "default"))
+		s.NoError(common.WaitForPodLogs(s.Context(), kc, "default"))
 		output, err := ssh.ExecWithOutput(s.Context(), "/usr/local/bin/k0s kc exec nginx -- cat /etc/resolv.conf")
 		s.Require().NoError(err)
 		s.Contains(output, "search default.svc.something.local svc.something.local something.local")


soider (approved):

inttest/customdomain/customdomain_test.go
```diff
@@ -45,8 +45,9 @@ func (s *CustomDomainSuite) TestK0sGetsUpWithCustomDomain() {
 
 	s.AssertSomeKubeSystemPods(kc)
 
-	s.T().Log("waiting to see CNI pods ready")
+	s.T().Log("waiting to see CNI and konnectivity-agent pods ready")
 	s.NoError(common.WaitForKubeRouterReady(s.Context(), kc), "CNI did not start")
+	s.NoError(common.WaitForDaemonSet(s.Context(), kc, "konnectivity-agent"), "Konnectivity-agent did not start")
```
twz123, line 50: Not sure this will have the desired effect. konnetivity-agent reports readiness even it is not connected to all API servers. It might be better to either have the "pod logs" check here (that's how it's done in other inttests), or add a retry to the failing assertion on line 61.

juanluisvaladas: Oh I didn't know that, but there is only one apiserver in this test. So I think it will work as expected.

soider:
By the way, there is a way to check if agent is ready (like actually ready), but it is a little bit awkard. The konnectivity-server (the process on the controller node) has /ready endpoint which gives a list of registred agents, IIRC.

juanluisvaladas:
> By the way, there is a way to check if agent is ready (like actually ready), but it is a little bit awkard. The konnectivity-server (the process on the controller node) has /ready endpoint which gives a list of registred agents, IIRC.

I don't think we need to think about this that much because in this test it's guaranteed to be only one controller and therefore if Konnectivity is ready it's working.

soider (approved):

PR #2548, opened 2022-12-21
Make Supervisor kill pidFile before Start

### LINKED ISSUE #2537: Supervisor will start the process of a component even if the process was started
This happens on the main branch and I believe it has always been like this, I will fix this soon.

## How to reproduce

```
# cat /run/k0s/containerd.pid 
803
# ps aux | grep -e /usr/local/bin/k0s -e /var/lib/k0s/bin/containerd
root         783  0.1  7.3 797716 72336 ?        Ssl  19:47   0:00 /usr/local/bin/k0s worker --kubelet-extra-args=--node-ip=192.168.122.74
root         803  0.2  4.1 754680 40800 ?        Sl   19:47   0:00 /var/lib/k0s/bin/containerd --root=/var/lib/k0s/containerd --state=/run/k0s/containerd --address=/run/k0s/containerd.sock --log-level=info --config=/etc/k0s/containerd.toml
root         894  0.0  0.2   6208  2112 pts/0    S+   19:48   0:00 grep --color=auto -e /usr/local/bin/k0s -e /var/lib/k0s/bin/containerd
# kill -9 783
# # wait for the k0s to come up
# cat /run/k0s/containerd.pid 
2242
# ps aux | grep -e 803 -e 2242
root         803  0.5  5.1 756916 50832 ?        Sl   19:47   0:01 /var/lib/k0s/bin/containerd --root=/var/lib/k0s/containerd --state=/run/k0s/containerd --address=/run/k0s/containerd.sock --log-level=info --config=/etc/k0s/containerd.toml
root        2242  0.2  3.2 752604 31636 ?        Sl   19:52   0:00 /var/lib/k0s/bin/containerd --root=/var/lib/k0s/containerd --state=/run/k0s/containerd --address=/run/k0s/containerd.sock --log-level=info --config=/etc/k0s/containerd.toml
```
## Expected behavior

The process should not be started twice.
The supervisor should be smart enough to check if the PID exists. If it exists check if it's the same command and true kill it and start a new one.

juanluisvaladas:
## Description

Prior to this commit the pidFiles were written but ignored. This avoids starting the same process twice which could be a problem.

The PR is marked as a draft because the tests need to be refactored because there is a lot of duplicate code. But other than that I'm quite happy with the current result

Fixes #2537

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/supervisor/supervisor.go
+++ b/pkg/supervisor/supervisor.go
@@ -49,6 +49,10 @@ type Supervisor struct {
 	TimeoutRespawn time.Duration
 	// For those components having env prefix convention such as ETCD_xxx, we should keep the prefix.
 	KeepEnvPrefix bool
+	// ProcFSPath is only used for testing
+	ProcFSPath string
+	// KillFunction is only used for testing
+	KillFunction func(int, syscall.Signal) error
 	// A function to clean some leftovers before starting or restarting the supervised process
 	CleanBeforeFn func() error
 
@@ -59,6 +63,8 @@ type Supervisor struct {
 	startStopMutex sync.Mutex
 	cancel         context.CancelFunc
 }
+
+const k0sManaged = "_K0S_MANAGED=yes"
 
 // processWaitQuit waits for a process to exit or a shut down signal
 // returns true if shutdown is requested
@@ -121,6 +127,10 @@ func (s *Supervisor) Supervise() error {
 	}
 	if s.TimeoutRespawn == 0 {
 		s.TimeoutRespawn = 5 * time.Second
+	}
+
+	if err := s.maybeKillPidFile(nil, nil); err != nil {
+		return err
 	}
 
 	var ctx context.Context
@@ -254,6 +264,8 @@ func getEnv(dataDir, component string, keepEnvPrefix bool) []string {
 		}
 		i++
 	}
+	env = append([]string{k0sManaged}, env...)
+	i++
 
 	return env[:i]
 }
--- a/pkg/supervisor/supervisor_unix.go
+++ b/pkg/supervisor/supervisor_unix.go
@@ -1,0 +1,148 @@
+//go:build unix
+
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package supervisor
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+)
+
+const (
+	exitCheckInterval = 200 * time.Millisecond
+)
+
+// killPid signals SIGTERM to a PID and if it's still running after
+// s.TimeoutStop sends SIGKILL.
+func (s *Supervisor) killPid(pid int, check <-chan time.Time, deadline <-chan time.Time) error {
+	if s.KillFunction == nil {
+		s.KillFunction = syscall.Kill
+	}
+	// Kill the process pid
+	if deadline == nil {
+		deadlineTicker := time.NewTicker(s.TimeoutStop)
+		deadline = deadlineTicker.C
+		defer deadlineTicker.Stop()
+	}
+	if check == nil {
+		checkTicker := time.NewTicker(exitCheckInterval)
+		check = checkTicker.C
+		defer checkTicker.Stop()
+
+	}
+
+	// Using two tickers is not very elegant but makes testing easier...
+Loop:
+	for {
+		select {
+		case <-check:
+			shouldKill, err := s.shouldKillProcess(pid)
+			if err != nil {
+				return err
+			}
+			if !shouldKill {
+				return nil
+			}
+
+			err = s.KillFunction(pid, syscall.SIGTERM)
+			if err == syscall.ESRCH {
+				return nil
+			} else if err != nil {
+				return fmt.Errorf("failed to send SIGTERM to pid %d: %s", s.cmd.Process.Pid, err)
+			}
+		case <-deadline:
+			break Loop
+		}
+	}
+
+	shouldKill, err := s.shouldKillProcess(pid)
+	if err != nil {
+		return err
+	}
+	if !shouldKill {
+		return nil
+	}
+
+	err = s.KillFunction(pid, syscall.SIGKILL)
+	if err == syscall.ESRCH {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to send SIGKILL to pid %d: %s", s.cmd.Process.Pid, err)
+	}
+	return nil
+}
+
+// maybeKillPidFile checks kills the process in the pidFile if it's has
+// the same binary as the supervisor's and also checks that the env
+// `_KOS_MANAGED=yes`. This function does not delete the old pidFile as
+// this is done by the caller.
+// The tickers are used for testing purposes, otherwise set them to nil.
+func (s *Supervisor) maybeKillPidFile(check <-chan time.Time, deadline <-chan time.Time) error {
+	if s.ProcFSPath == "" {
+		s.ProcFSPath = "/proc"
+	}
+
+	pid, err := os.ReadFile(s.PidFile)
+	if os.IsNotExist(err) {
+		return nil
+	} else if err != nil {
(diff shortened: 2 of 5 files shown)


juanluisvaladas:
Uhg I hadn't thought about windows, how can I spin up a cluster with a windows machine?

pkg/supervisor/supervisor_nix.go
```diff
@@ -0,0 +1,98 @@
+//go:build !windows
```
soider, line 1: I think it is not necessary to have explicit build tags and the OS specific suffix in the filename. So probably having files supervisor_posix and supervisor_windows COULD work here :)

Btw, why do we have two `go:build` and `build:`? 
We can safely use only the one which works for the latest go version since we don't provide our code as a library.
Otherwise we have three different build constraint sources: filename, new style tag, old style tag and we need to keep them more or less in sync.

What do you think, if we try to have only one build tag style or even check, may be with a correct naming (_posix and _windows) it would not require tags at all?

soider (changes requested):
Generally looks good, there is a minor suggestion to improve build tags handling.

pkg/supervisor/supervisor.go
```diff
@@ -123,6 +127,14 @@ func (s *Supervisor) Supervise() error {
 		s.TimeoutRespawn = 5 * time.Second
 	}
 
+	_, err := os.Stat(s.PidFile)
+	if !os.IsNotExist(err) {
```
ncopa, line 131: This check is not necessary. We check the existence of the file when we open it.

pkg/supervisor/supervisor_nix.go
```diff
@@ -0,0 +1,98 @@
...
+// maybeKillPidFile checks kills the process in the pidFile if it's has
+// the same binary as the supervisor's. This function does not delete
+// the old pidFile as this is done by the caller.
+func (s *Supervisor) maybeKillPidFile() error {
+	if s.ProcFSPath == "" {
+		s.ProcFSPath = "/proc"
+	}
+	pid, err := os.ReadFile(s.PidFile)
+	if err != nil {
+		return fmt.Errorf("failed to read pid file %s: %v", s.PidFile, err)
```
ncopa, line 75: I think we could return nil if the error is IsNotExist.

pkg/supervisor/supervisor_nix.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	p, err := strconv.Atoi(strings.TrimSuffix(string(pid), "\n"))
+	if err != nil {
+		return fmt.Errorf("failed to parse pid file %s: %v", s.PidFile, err)
+	}
+
+	if !s.isRunning(p) {
```
ncopa, line 83: Keep in mind that the process could die directly after this check.

pkg/supervisor/supervisor_nix.go
```diff
@@ -0,0 +1,98 @@
...
+
+// killPid signals SIGTERM to a PID and if it's still running after
+// s.TimeoutStop sends SIGKILL.
+func (s *Supervisor) killPid(pid int) error {
+	if s.KillFunction == nil {
+		s.KillFunction = syscall.Kill
+	}
+	// Kill the process pid
+	if err := s.KillFunction(pid, syscall.SIGTERM); err != nil {
+		return fmt.Errorf("failed to send SIGTERM to pid %d: %s", pid, err)
```
ncopa, line 44: the process may already be dead, in which case kill will return [ESRCH](https://man7.org/linux/man-pages/man2/kill.2.html#ERRORS) error.

pkg/supervisor/supervisor_nix.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	if err := s.KillFunction(pid, syscall.SIGKILL); err != nil {
+		return fmt.Errorf("failed to send SIGKILL to pid %d: %s", s.cmd.Process.Pid, err)
+	}
+	return nil
+}
+
+func (s *Supervisor) isRunning(pid int) bool {
+	_, err := os.Stat(filepath.Join(s.ProcFSPath, strconv.Itoa(pid)))
```
ncopa, line 62: `s.KillFunction(pid, 0)` can also be used to test if process is running.

juanluisvaladas: I like the idea of having the same code handling both the test and the real cases, however that's certainly cleaner and allows us to remove Supervisor.ProcFSPath...

pkg/supervisor/supervisor_nix.go
juanluisvaladas: > Btw, why do we have two go:build and build:?
Quite frankly, I just grepped and copied it from other places :) 

> I think it is not necessary to have explicit build tags and the OS specific suffix in the filename. So probably having files supervisor_posix and supervisor_windows COULD work here :)

Turns out _windows.go is respected by default but _posix.go / _unix.go is not :/ . Although _linux.go is honored.
I added the //build:  https://go-review.googlesource.com/c/go/+/389935 because I fear calling it _linux.go will break macOS and I don't really want to add yet another dummy file  for darwin...

juanluisvaladas:
I'm still addressing the comments, this push is just to see if it fixes the windows builds.

juanluisvaladas:
I believe all the concerns are addressed:
1- There are no sleeps in the tests
2- We verify for signal ESRCH errors
3- We don't run unnecessary checks

I tested it manually and seems fine. It's just missing checking for the env variable but I want the tests to run before implementing that so I'm still having this as a draft.

juanluisvaladas:
Everything should be addressed now:
1- There are no sleeps in the tests
2- We verify for signal ESRCH errors
3- We don't run unnecessary checks
4- We verify cmdline and environ before killing.

pkg/supervisor/supervisor.go
```diff
@@ -60,6 +64,8 @@ type Supervisor struct {
 	cancel         context.CancelFunc
 }
 
+const k0sManaged = "_K0S_MANAGED=yes"
```
mikhail-sakhnov, line 67: does the underscore have some meaning here?

juanluisvaladas: Not really. The underscore is there because it's the value we decided during the discussion in the call.

pkg/supervisor/supervisor_unix.go
```diff
@@ -0,0 +1,126 @@
...
+	cmd := strings.Split(string(cmdline), "\x00")
+	if cmd[0] != s.BinPath {
+		return nil
+	}
+
+	//only kill process if it has the _KOS_MANAGED env set
+	env, err := os.ReadFile(filepath.Join(s.ProcFSPath, strconv.Itoa(p), "environ"))
+	if err != nil {
+		return fmt.Errorf("failed to read process %d environ: %v", p, err)
+	}
```
ncopa, line 118: The process may be terminated after reading `cmdline`.
```suggestion
	if os.IsNotExist(err) {
		return nil
	} else if err != nil {
		return fmt.Errorf("failed to read process %d environ: %v", p, err)
	}
```

ncopa (changes requested):

juanluisvaladas:
@mikhail-sakhnov I believe you were OK with the variable being called `_K0S_MANAGED=yes` because it's what was agreed on the call. Please let me know if that's OK with you or if you want to change it to `K0S_MANAGED=yes` instead. 

@ncopa I made a slight change in the function maybeKillPid and killPid. Now it checks for the name of the process and the environment variables right before killing the process.

juanluisvaladas:
@mikhail-sakhnov I believe you were OK with the variable being called `_K0S_MANAGED=yes` because it's what was agreed on the call. Please let me know if that's OK with you or if you want to change it to `K0S_MANAGED=yes` instead. 

@ncopa I made a slight change in the function maybeKillPid and killPid. Now it checks for the name of the process and the environment variables right before killing the process.

pkg/supervisor/supervisor_windows.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"time"
+)
+
+// maybeKillPidFile checks kills the process in the pidFile if it's has
+// the same binary as the supervisor's. This function does not delete
+// the old pidFile as this is done by the caller.
+func (s *Supervisor) maybeKillPidFile(check <-chan time.Time, deadline <-chan time.Time) error {
+	s.log.Warnf("maybeKillPidFile is not implemented on Windows")
```
twz123, line 30: I wonder why that is. There's still the possibility of sending `os.Kill` to terminate the process. Anyhow, we can address this later on.

juanluisvaladas: I think it's worth merging this as is and merging your patch for this for several reasons:
1- I don't have a windows environment to properly test this and we need to get this merged ASAP because of the release.
2- Windows is broken at the moment anyway.
3- Your solution is almost implemented right now and it's a good implementation. Duplicating the effort seems unreasonable to me.

Anyway we need a way to get windows workers for our local environments.

ncopa (approved):

PR #2585, opened 2023-01-10
Remove Docker replacements from go.mod

twz123:
## Description

No longer needed. Kubernetes doesn't have them anymore, and the manifest-tool dependency doesn't require it either in the currently used version.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -98,7 +98,6 @@ require (
 	github.com/bgentry/speakeasy v0.1.0 // indirect
 	github.com/blang/semver v3.5.1+incompatible // indirect
 	github.com/blang/semver/v4 v4.0.0 // indirect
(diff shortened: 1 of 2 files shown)


juanluisvaladas (approved):
I think it's still required by some dependencies because I see both dependencies added to go.sum but cleaning go.mod which is what humans use is always nice.

twz123:
> I think it's still required by some dependencies because I see both dependencies added to go.sum but cleaning go.mod which is what humans use is always nice.

I should have been clearer: The dependencies are still used, but the replacement directives in go.mod are no longer needed, i.e. go mod fill figure it out the correct dependency tree by itself. 🙂

PR #2586, opened 2023-01-10
Reduce telemetry heartbeat workerData size

juanluisvaladas:
## Description

Prior to this commit we send every single node, now we group the nodces which have the same cpu, memory, arch, OS and container runtime making it smaller for large clusters.

Signed-off-by: Juan Luis de Sousa-Valadas Castaño <jvaladas@mirantis.com>

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -44,8 +44,9 @@ type telemetryData struct {
 type workerData map[string]interface{}
 
 type workerSums struct {
-	cpuTotal int64
-	memTotal int64
+	cpuTotal   int64
+	memTotal   int64
+	nodesTotal int
 }
 
 func (td telemetryData) asProperties() analytics.Properties {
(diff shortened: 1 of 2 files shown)


makhov (approved):

pkg/telemetry/telemetry.go
```diff
@@ -105,29 +106,52 @@ func (c Component) getClusterID(ctx context.Context) (string, error) {
 	return fmt.Sprintf("kube-system:%s", ns.UID), nil
 }
 
+func (c Component) getWorkerGroup(os string, arch string, cpus int64, mem int64, runtime string) string {
```
kke, line 109: this could return a struct like:

```go
type workerGroup struct {
    OS string
    Arch string,
    CPUs    int64
    Mem     int64
    Runtime string
}
```

this is just basic types so it is comparable and can be used as a map key. 

After this, the `wm` below would change from:

```go
	wm := make(map[string]workerData)
```

to:

```go
	wm := make(map[workerGroup]int)
```

I think this will make it possible to just: `wm[wg]++` without any kind of nil/existence checks.

PR #2290, opened 2022-10-21
Fail inttests when the deadline is exceeded

twz123:
## Description

Some tests may not mark themselves as failed when the test context's deadline is exceeded. This can result in green builds, even if the test has been aborted. Moreover, the cluster state wasn't collected in those cases, since the check if the test actually failed was likely to happen before the integration tests recorded the failure.

Record a separate test failure whenever the context timed out, so that the test is stamped as failed in any case and the cluster state will be recorded.

This reveals #2289, so in order to merge this one, we need to either address #2289 or disable the `cnichange` inttest until that is fixed.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2612, opened 2023-01-19
Fix CNI change test 

### LINKED ISSUE #2289: Buggy integration test: `cnichange`
The `cnichange` integration test is buggy.

The test checks if the k0s controller startup fails after the CNI provider changed. It launches k0s via SSH and expects to see an error:
https://github.com/k0sproject/k0s/blob/v1.25.2+k0s.0/inttest/cnichange/cnichange_test.go#L45-L46

But: k0s actually won't fail to startup, it will simply stick to the previous provider. This is due to the fact that the checks reside in the `Reconcile` methods of the calico/kuberouter controller components nowadays, which won't trigger a termination of the controller, but rather log an error and post a Kubernetes event.

The test still succeeds because the SSH connection will be terminated as soon as the integration test's context deadline is exceeded. The SSH call will return with a context error (presumably "deadline exceeded"), which satisfies the test assertion just fine.

jnummelin:
by checking that config reconciler gives out proper warning event

Previously we used to fail the controller bootstrap but since we've moved the reconcilers to be more "async" that behaviour has changed which was not reflected in this test.

Introducing #2290 surfaces the bugs in this test case.

Fixes #2289

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/inttest/cnichange/cnichange_test.go
+++ b/inttest/cnichange/cnichange_test.go
@@ -17,11 +17,17 @@ limitations under the License.
 package cnichange
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/suite"
 
 	"github.com/k0sproject/k0s/inttest/common"
+
+	"github.com/k0sproject/k0s/pkg/kubernetes/watch"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
 )
 
 type CNIChangeSuite struct {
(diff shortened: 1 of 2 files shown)


twz123 (approved):

inttest/cnichange/cnichange_test.go
```diff
@@ -17,48 +17,79 @@ limitations under the License.
...
+
+	kc, err := s.KubeClient(s.ControllerNode(0))
+	s.Require().NoError(err)
+
+	// Wait till we see kube-router DS created
+	watch.DaemonSets(kc.AppsV1().DaemonSets("kube-system")).
+		WithObjectName("kube-router").
+		Until(s.Context(), func(ds *appsv1.DaemonSet) (bool, error) {
+			return true, nil
+		})
```
juanluisvaladas, line 51: Perhaps use `common.WaitForKubeRouterReady` instead?

jnummelin: WE cannot in this test case as there's no workers --> it will not get ready ever. :)

PR #2614, opened 2023-01-19
Add local IPv6 addresses to certificates

twz123:
## Description

The same has already been done for IPv4 addresses in #1683. Moreover, forward the context to detectLocalIPs(), so that the
resolution may be cancelled.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"os"
@@ -173,7 +174,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 		"127.0.0.1",
 	}
 
-	localIPs, err := detectLocalIPs()
+	localIPs, err := detectLocalIPs(ctx)
 	if err != nil {
 		return fmt.Errorf("error detecting local IP: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


cmd/controller/certificates.go
```diff
@@ -217,23 +218,28 @@ func (c *Certificates) Init(ctx context.Context) error {
...
+			return nil, err
 		}
 	}
 
+	var localIPs []string
 	for _, addr := range addrs {
-		if addr.To4() != nil {
-			localIPs = append(localIPs, addr.String())
+		ip := addr.IP
+		if ip.To4() != nil || ip.To16() != nil {
```
juanluisvaladas, line 241: Isn't this the only change needed? net.LookupIP calls DefaultResolver.LookupIPAddr, so I don't think the other changes are doing anything at all.

Anyway the code looks correct to me, so LGTM

juanluisvaladas (approved):

cmd/controller/certificates.go
twz123: Explanation in the PR description:

> Moreover, forward the context to detectLocalIPs(), so that the resolution may be cancelled.

So it's now using the context and DNS lookups can be cancelled in case they're hanging due to some misconfiguration.

juanluisvaladas: Ah good point.

PR #2621, opened 2023-01-20
Removes images in k0s config create by default

### LINKED ISSUE #2587: k0s config defines the versions of the images
This is a problem because plenty of users will use k0s config create and leave those fields as they are. If they perform an upgrade with this configuration the images version will not be upgraded.

I belive the right approach should be not defining .spec.images by default and adding a flag which defines the version explicictly. We could call it `--include-images` or `--set-images`

juanluisvaladas:
Fixes #2587

Signed-off-by: Juan Luis de Sousa-Valadas Castaño <jvaladas@mirantis.com>

## Description

Fixes # (issue)

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -56,40 +56,6 @@ spec:
     storage:
       create_default_storage_class: false
       type: external_storage
-  images:
-    default_pull_policy: IfNotPresent
-    calico:
-      cni:
-        image: docker.io/calico/cni
-        version: v3.24.5
-      kubecontrollers:
-        image: docker.io/calico/kube-controllers
-        version: v3.24.5
-      node:
-        image: docker.io/calico/node
-        version: v3.24.5
-    coredns:
-      image: docker.io/coredns/coredns
-      version: 1.10.0
-    konnectivity:
-      image: quay.io/k0sproject/apiserver-network-proxy-agent
-      version: 0.0.33-k0s
-    kubeproxy:
-      image: registry.k8s.io/kube-proxy
-      version: v1.26.1
-    kuberouter:
-      cni:
-        image: docker.io/cloudnativelabs/kube-router
-        version: v1.5.1
-      cniInstaller:
-        image: quay.io/k0sproject/cni-node
-        version: 1.1.1-k0s.0
-    metricsserver:
-      image: registry.k8s.io/metrics-server/metrics-server
-      version: v0.6.2
-    pushgateway:
-      image: quay.io/k0sproject/pushgateway-ttl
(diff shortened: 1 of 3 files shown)


juanluisvaladas:
Default behavior:
```
$ ./k0s config create --include-images | grep images:
  images:
```

With flag:
```
$ ./k0s config create --include-images | grep images: -A33
  images:
    calico:
      cni:
        image: docker.io/calico/cni
        version: v3.24.5
      kubecontrollers:
        image: docker.io/calico/kube-controllers
        version: v3.24.5
      node:
        image: docker.io/calico/node
        version: v3.24.5
    coredns:
      image: docker.io/coredns/coredns
      version: 1.10.0
    default_pull_policy: IfNotPresent
    konnectivity:
      image: quay.io/k0sproject/apiserver-network-proxy-agent
      version: 0.0.33-k0s
    kubeproxy:
      image: registry.k8s.io/kube-proxy
      version: v1.26.1
    kuberouter:
      cni:
        image: docker.io/cloudnativelabs/kube-router
        version: v1.5.1
      cniInstaller:
        image: quay.io/k0sproject/cni-node
        version: 1.1.1-k0s.0
    metricsserver:
      image: registry.k8s.io/metrics-server/metrics-server
      version: v0.6.2
    pushgateway:
      image: quay.io/k0sproject/pushgateway-ttl
      version: edge@sha256:7031f6bf6c957e2fdb496161fe3bea0a5bde3de800deeba7b2155187196ecbd9
```

docs/configuration.md
```diff
@@ -56,40 +56,6 @@ spec:
     storage:
       create_default_storage_class: false
       type: external_storage
-  images:
```
twz123, line 59: Are there any other places in the docs that list all the images that k0s will use? I wonder if it would make sense to have _some_ place in the docs where they're listed.

twz123 (approved):
I was under the impression that only the versions should be suppressed, not the whole images field. Despite that, LGTM.

docs/configuration.md
```diff
@@ -412,6 +378,8 @@ spec:
       version: v1.7.0
 ```
 
+If you want to get the default values or the full struct, use `k0s config create --include-images`.
```
kke, line 381:
```suggestion
If you want the list of default images and their versions to be included, use `k0s config create --include-images`.
```

or something similar, "default values or the full struct" seems wrong.

juanluisvaladas:
> I was under the impression that only the versions should be suppressed, not the whole images field. Despite that, LGTM.

The problem with that approach is that if we change the image we will have the same problem, and we don't host most of the images so it's an added risk.

PR #2604, opened 2023-01-16
feat: add possibility to customize containerd with additional files

alegrey91:
## Description

Based on the feature request opened here: https://github.com/k0sproject/k0s/issues/2550 the PR add the possibility to customize containerd without editing the original `/etc/k0s/containerd.toml` file. This is possible due to containerd's feature that allows importing additional configurations using the `import` keyword.

Fixes #2550

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
(since this is a containerd feature, I don't think tests are needed)
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2637, opened 2023-01-24
Containerd imports handled by k0s

jnummelin:
## Description

As noted in https://github.com/k0sproject/k0s/pull/2604#issuecomment-1400173518 the current way containerd handles imports for CRI configs produces less than expected results.

In this PR I propose an alternative where k0s handles the imports:
- k0s reads all "partials" from `/etc/k0s/containerd.d/*.toml`
- if a partial has CRI configs k0s will append them in memory with the default CRI config (using runc)
- if a partial does NOT have CRI config k0s will add it as normal `import` in containerd config
- once all CRI configs are appended in memory, k0s will drop the cri configs into `/run/k0s/containerd-cri.toml` file and add that as normal import in main containerd config toml

As noted in our proposed fix in upstream [containerd PR](https://github.com/containerd/containerd/pull/8056), it's not going to be included in 1.x releases. 😞  Hence the need to handle this at k0s side.

### DIFF
--- a/pkg/component/worker/containerd.go
+++ b/pkg/component/worker/containerd.go
@@ -17,6 +17,7 @@ limitations under the License.
 package worker
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"path/filepath"
@@ -26,6 +27,7 @@ import (
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/internal/pkg/file"
+	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
 	"github.com/k0sproject/k0s/pkg/assets"
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/constant"
@@ -37,8 +39,12 @@ const confTmpl = `
 # If you wish to customize the config replace this file with your custom configuration.
 # For reference see https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md
 version = 2
+imports = [
+	"{{ .ImportsGlob }}"
+]
 `
 const confPath = "/etc/k0s/containerd.toml"
+const importsPath = "/etc/k0s/containerd_imports/*.toml"
 
 // ContainerD implement the component interface to manage containerd as k0s component
 type ContainerD struct {
@@ -98,7 +104,23 @@ func (c *ContainerD) setupConfig() error {
 	if err := dir.Init(filepath.Dir(confPath), 0755); err != nil {
 		return err
 	}
-	return file.WriteContentAtomically(confPath, []byte(confTmpl), 0644)
+	if err := dir.Init(filepath.Dir(importsPath), 0755); err != nil {
+		return err
+	}
+	output := bytes.NewBuffer([]byte{})
+	tw := templatewriter.TemplateWriter{
+		Name:     "containerdconfig",
+		Template: confTmpl,
+		Data: struct {
+			ImportsGlob string
+		}{
+			ImportsGlob: importsPath,
+		},
+	}
+	if err := tw.WriteToBuffer(output); err != nil {
+		return fmt.Errorf("can't create containerd config: %v", err)
+	}
+	return file.WriteContentAtomically(confPath, output.Bytes(), 0644)
 }
 
 // Stop stops containerD
--- a/pkg/component/worker/containerd.go
+++ b/pkg/component/worker/containerd.go
@@ -27,7 +27,10 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"syscall"
+	"time"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 
@@ -38,6 +41,7 @@ import (
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/component/worker/containerd"
 	"github.com/k0sproject/k0s/pkg/constant"
+	"github.com/k0sproject/k0s/pkg/debounce"
 	"github.com/k0sproject/k0s/pkg/supervisor"
 )
 
@@ -83,7 +87,7 @@ func (c *ContainerD) Init(ctx context.Context) error {
 }
 
 // Run runs containerD
-func (c *ContainerD) Start(_ context.Context) error {
+func (c *ContainerD) Start(ctx context.Context) error {
 	logrus.Info("Starting containerD")
 
 	if err := c.setupConfig(); err != nil {
@@ -104,7 +108,13 @@ func (c *ContainerD) Start(_ context.Context) error {
 		},
 	}
 
-	return c.supervisor.Supervise()
+	if err := c.supervisor.Supervise(); err != nil {
+		return err
+	}
+
+	go c.watchDropinConfigs(ctx)
+
+	return nil
 }
 
 func (c *ContainerD) setupConfig() error {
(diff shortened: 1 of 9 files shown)


alegrey91:
Hi @jnummelin,
IMHO the idea is fine and works properly, but it would not be preferable to try to fix this in **containerd**? In this way, you could have a better separation of duties. WDYT? Am I missing something?

jnummelin:
> IMHO the idea is fine and works properly, but it would not be preferable to try to fix this in containerd?

Yes, of course. This PR is kinda like plan-B or an intermediary solution in case we cannot fix containerD side of things. Or if our proposal will not get traction in containerD. @juanluisvaladas is already working on the fix on containerD side.

pkg/component/worker/containerd.go
```diff
@@ -91,14 +100,43 @@ func (c *ContainerD) Start(_ context.Context) error {
 
 func (c *ContainerD) setupConfig() error {
 	// If the config file exists, use it as-is
-	if file.Exists(confPath) {
-		return nil
-	}
+	// FIXME We need to figure out how to overwrite this only if it's fully k0s managed
```
makhov, line 103: Do we need it? I think everything in the `/etc/k0s` directory is k0s managed

pkg/component/worker/containerd/runtime.go
```diff
@@ -0,0 +1,178 @@
...
+            IoGid = 0
+            IoUid = 0
+            NoNewKeyring = false
+            NoPivotRoot = false
+            Root = ""
+            ShimCgroup = ""
+            SystemdCgroup = false
+`
+
+const importsPath = "/etc/k0s/containerd.d/*.toml"
```
makhov, line 98: We use the same path here and in the `containerd.go`. Probably will need to refactor this and pass the path as an argument or config value.

pkg/component/worker/containerd/runtime.go
```diff
@@ -0,0 +1,178 @@
...
+			c.log.Debugf("found CRI plugin config in %s, appending to CRI config", file)
+			// Append all CRI plugin configs into single file
+			// and add it to imports
+			criConfigBuffer.WriteString(fmt.Sprintf("# appended from %s\n", file))
+			_, err := criConfigBuffer.Write(data)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			c.log.Debugf("adding %s as-is to imports", file)
```
makhov, line 150: I think it's ok for now, but in the future, we may run into the same issue with another section.

pkg/component/worker/containerd/runtime.go
```diff
@@ -0,0 +1,178 @@
...
+			// Add file to imports
+			imports = append(imports, file)
+		}
+	}
+
+	// Write the CRI config to a file and add it to imports
+	err = os.WriteFile(containerdCRIConfigPath, criConfigBuffer.Bytes(), 0644)
```
makhov, line 157: Looks a bit fishy to me. By the name of the function, I wouldn't expect a file writing here. Not sure, how to improve it tho, so we can leave it as is for now.

makhov (commented):
LGTM in general, left a few comments

jnummelin:
Upstream containerd is not willing to accept our proposed fix into 1.x series and is targeting the fix for only 2.0

Hence I'll start to work on this to make it more robust and prettify the code a bit. 😄

pkg/component/worker/containerd.go
```diff
@@ -91,14 +100,43 @@ func (c *ContainerD) Start(_ context.Context) error {
...
+			Imports []string
+		}{
+			Imports: imports,
+		},
+	}
+	if err := tw.WriteToBuffer(output); err != nil {
+		return fmt.Errorf("can't create containerd config: %v", err)
+	}
+	if err := file.WriteContentAtomically(confPath, output.Bytes(), 0644); err != nil {
+		return err
```
mikhail-sakhnov, line 135: I'd suggest use wrapping (preferable) or return as is without if.

pkg/component/worker/containerd.go
jnummelin: Yes, we need the logic to handle this. All previous versions (before this PR) will treat any existing `/etc/k0s/containerd.toml` file as user managed and we should NOT overwrite those by default.

pkg/component/worker/containerd/runtime.go
jnummelin: > we may run into the same issue with another section

Not sure I get what you mean by this? Do you mean some other section than CRI configs? We _may_ run into those cases for sure, but untill we know how those sections work, or are expected to work, I would not put any logic in place.

mikhail-sakhnov:
looks reasonably good, though there are couple of linting issues and non-addressed FIXME comment

pkg/component/worker/containerd.go
```diff
@@ -90,18 +108,109 @@ func (c *ContainerD) Start(_ context.Context) error {
 }
 
 func (c *ContainerD) setupConfig() error {
-	// If the config file exists, use it as-is
-	if file.Exists(confPath) {
+	// Check if the config file is user managed
+	// If it is, we should not touch it
+	k0sManaged, err := isK0sManagedConfig(confPath)
+	if err != nil {
+		return err
```
mikhail-sakhnov, line 115: I suggest either wrapping errors here or on the caller level (#L89)

pkg/component/worker/containerd.go
```diff
@@ -90,18 +108,109 @@ func (c *ContainerD) Start(_ context.Context) error {
...
+		if strings.Contains(scanner.Text(), magicMarker) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func isPre1_27ManagedConfig(path string) (bool, error) {
+	// Check MD5 sum of the config file
+	// If it matches the pre 1.26 config, it's k0s managed
```
mikhail-sakhnov, line 192: either comment or variable name should be changed (pre 1.26, pre1_27)

mikhail-sakhnov (changes requested):

jnummelin:
Marked as a draft till I sort out how to document this

juanluisvaladas (approved):
LGTM.

Docs are clear and I think there haven't been more recent changes.

jnummelin:
@juanluisvaladas Sorry, need a re-review as I had stale (== non pushed) commit for the dynamic re-loading of containerd in my local branch... 🤦

pkg/component/worker/containerd.go
```diff
@@ -148,6 +158,69 @@ func (c *ContainerD) setupConfig() error {
...
+	log := logrus.WithFields(logrus.Fields{"component": "containerd", "phase": "restart"})
+
+	log.Info("restart requested")
+	if err := c.setupConfig(); err != nil {
+		log.WithError(err).Warn("failed to resolve config")
+		return
+	}
+
+	p := c.supervisor.GetProcess()
+	if err := p.Signal(syscall.SIGHUP); err != nil {
```
juanluisvaladas, line 219: By sending SIGHUP we're killing the process rather than reloading it, so I would send SIGTERM but it's a nitpick. If you agree and want to change it ping me on slack after the change.

juanluisvaladas (approved):
The nitpick is really a nit so LGTM

pkg/component/worker/containerd.go
jnummelin: I guess my rationale for `SIGHUP` was that it is the "standard" for reloading. In containerd there's no realoading support per se, the only way is to restart the process. But having `SIGHUP` do that kinda makes sense, especially if containerd ever implement actual reload using `SIGHUP`. Based on the comments in https://github.com/containerd/containerd/issues/4478, it seems there's some "plans" to actually implement reloading.

TL;DR; I'd keep it as `SIGHUP` for "forward compatibility"

juanluisvaladas: OK that's fair enough.

jnummelin:
@mikhail-sakhnov Could you re-review, cannot merge as you had requested changes (which are addressed now)

mikhail-sakhnov (approved):

PR #2639, opened 2023-01-24
test(supervisor): use `t.Setenv` to set env vars in tests

Juneezee:
## Description

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	os.Setenv(key, "new value")
	defer os.Unsetenv(key)
	
	// after
	t.Setenv(key, "new value")
}
```

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/supervisor/supervisor_test.go
+++ b/pkg/supervisor/supervisor_test.go
@@ -99,20 +99,26 @@ func TestSupervisorStart(t *testing.T) {
 }
 
 func TestGetEnv(t *testing.T) {
-	// backup environment vars
+	// backup environment vars, and restore them when test finishes
 	oldEnv := os.Environ()
+	t.Cleanup(func() {
+		for _, e := range oldEnv {
+			key, val, _ := strings.Cut(e, "=")
+			assert.NoError(t, os.Setenv(key, val))
+		}
+	})
 
 	os.Clearenv()
-	os.Setenv("k3", "v3")
-	os.Setenv("PATH", "/bin")
-	os.Setenv("k2", "v2")
-	os.Setenv("FOO_k3", "foo_v3")
-	os.Setenv("k4", "v4")
-	os.Setenv("FOO_k2", "foo_v2")
-	os.Setenv("FOO_HTTPS_PROXY", "a.b.c:1080")
-	os.Setenv("HTTPS_PROXY", "1.2.3.4:8888")
-	os.Setenv("k1", "v1")
-	os.Setenv("FOO_PATH", "/usr/local/bin")
+	t.Setenv("k3", "v3")
+	t.Setenv("PATH", "/bin")
+	t.Setenv("k2", "v2")
+	t.Setenv("FOO_k3", "foo_v3")
+	t.Setenv("k4", "v4")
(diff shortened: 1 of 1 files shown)


pkg/supervisor/supervisor_test.go
```diff
@@ -99,20 +99,26 @@ func TestSupervisorStart(t *testing.T) {
 }
 
 func TestGetEnv(t *testing.T) {
-	// backup environment vars
+	// backup environment vars, and restore them when test finishes
 	oldEnv := os.Environ()
+	t.Cleanup(func() {
+		for _, e := range oldEnv {
+			kv := strings.SplitN(e, "=", 2)
+			assert.NoError(t, os.Setenv(kv[0], kv[1]))
```
twz123, line 107: I think the main reason why this test doesn't use `t.Setenv()` is because it wants to ensure a clean environment. So that any environment variables that have been set outside of the test cannot influence its outcome.

We still need to reset/restore the env ourselves, so we didn't win anything here by using the new functionality.

A way to get rid of the OS env in testing would be to pass in the env map as a function parameter to `getEnv()`.

pkg/supervisor/supervisor_test.go
```diff
@@ -99,20 +99,26 @@ func TestSupervisorStart(t *testing.T) {
 }
 
 func TestGetEnv(t *testing.T) {
-	// backup environment vars
+	// backup environment vars, and restore them when test finishes
 	oldEnv := os.Environ()
+	t.Cleanup(func() {
+		for _, e := range oldEnv {
+			kv := strings.SplitN(e, "=", 2)
```
twz123, line 106: This would be a perfect use case for `strings.Cut()`.

pkg/supervisor/supervisor_test.go
juanluisvaladas: > A way to get rid of the OS env in testing would be to pass in the env map as a function parameter to getEnv().

Agreed, that would also make possible in the future to use `t.Parallel`. We aren't using right now and using `t.Parallel` is problematic, but could be somehting that we want in the future.

ncopa (approved):

juanluisvaladas:
It's unrelated to the change, but since you're here perhaps change the actual != expected in the line 135  for `assert.Equal`

Here's 
https://github.com/k0sproject/k0s/blob/5d5fa1161efd675a8eacc7aa02fe22a8e2ed56d8/internal/pkg/stringmap/stringmap_test.go

twz123 (approved):
The test still needs to manage the OS env by itself, so there's no gain in using `t.Setenv` over `os.Setenv`. Anyhow, it doesn't hurt and the PR brings in improvements other than that.

twz123:
Thanks @Juneezee!

PR #2647, opened 2023-01-26
Add possibility to inject custom telemetry attributes via ConfigMap

jnummelin:

## Description

This allows us to add certain "pre-defined" data e.g. from Lens Desktop use case so that we can better identify these uses of k0s in telemetry data.

So essentially, if a ConfigMap `kube-system/k0s-telemetry` exists, we will include all KV pairs to telemetry data.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -147,6 +147,7 @@ func (c Component) sendTelemetry(ctx context.Context) {
 	hostData.Extra["cpuArch"] = runtime.GOARCH
 
 	addSysInfo(&hostData)
+	c.addCustomData(ctx, &hostData)
 
 	c.log.WithField("data", data).WithField("hostdata", hostData).Info("sending telemetry")
 	if err := c.analyticsClient.Enqueue(analytics.Track{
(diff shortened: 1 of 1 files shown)


pkg/telemetry/telemetry.go
```diff
@@ -159,6 +160,17 @@ func (c Component) sendTelemetry(ctx context.Context) {
...
 }
 
+func (c Component) addCustomData(ctx context.Context, analyticCtx *analytics.Context) {
+	cm, err := c.kubernetesClient.CoreV1().ConfigMaps("kube-system").Get(ctx, "k0s-telemetry", metav1.GetOptions{})
+	if err != nil {
+		c.log.WithError(err).Info("can't get k0s-telemetry configmap, not adding custom data")
+		return
+	}
+	for k, v := range cm.Data {
+		analyticCtx.Extra[k] = v
```
kke, line 170: As the "extra" is a fake field and its contents are just serialized in the json, this makes it possible to overwrite existing data in the analytics json, accidentally or, maybe in some far fetched scenario, on purpose.

Maybe prefix the keys from the custom data with something like `custom.`?

pkg/telemetry/telemetry.go
```diff
@@ -159,6 +160,17 @@ func (c Component) sendTelemetry(ctx context.Context) {
 	}
 }
 
+func (c Component) addCustomData(ctx context.Context, analyticCtx *analytics.Context) {
+	cm, err := c.kubernetesClient.CoreV1().ConfigMaps("kube-system").Get(ctx, "k0s-telemetry", metav1.GetOptions{})
+	if err != nil {
+		c.log.WithError(err).Info("can't get k0s-telemetry configmap, not adding custom data")
```
kke, line 166: Constantly seeing this error in the logs may bother users

pkg/telemetry/telemetry.go
kke: Hmm, I wonder if Extra could take the cm.Data as is, something like `analyticsCtx.Extra["custom"] = cm.Data`

jnummelin: > Maybe prefix the keys from the custom data with something like custom.?

This sounds like a good idea to me 👍

kke (approved):

PR #2660, opened 2023-01-31
Makefile improvements

### LINKED ISSUE #2654: Build: `make -j$(nproc)` fails on first build

### Platform

```shell
$ uname -srvmo; cat /etc/os-release || lsb_release -a
Linux 6.1.7-200.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jan 18 17:11:49 UTC 2023 x86_64 GNU/Linux
NAME="Fedora Linux"
VERSION="37 (Workstation Edition)"
ID=fedora
VERSION_ID=37
VERSION_CODENAME=""
PLATFORM_ID="platform:f37"
PRETTY_NAME="Fedora Linux 37 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:37"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f37/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=37
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=37
SUPPORT_END=2023-11-14
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

### Version

```sh
./k0s version     
v1.26.0+k0s.0-91-gc3d7c1b32
```

### Sysinfo

```sh
$ /k0s sysinfo
Machine ID: "6085114bbae18d873810a7c9d2e085be46818b4fc30282364ced5a6188a93dee" (from machine) (pass)
Total memory: 62.7 GiB (pass)
Disk space available for /var/lib/k0s: 851.1 GiB (pass)
  Linux kernel release: 6.1.7-200.fc37.x86_64 (pass)
  Max. file descriptors per process: current: 524288 / max: 524288 (pass)
  Executable in path: modprobe: /usr/sbin/modprobe (pass)
    cgroup controller "cpu": available (pass)
    cgroup controller "cpuset": available (pass)
    cgroup controller "memory": available (pass)
    cgroup controller "devices": available (assumed) (pass)
    cgroup controller "freezer": available (assumed) (pass)
    cgroup controller "pids": available (pass)
    cgroup controller "hugetlb": available (pass)
    CONFIG_CGROUP_FREEZER: Freezer cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_PIDS: PIDs cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_DEVICE: Device controller for cgroups: built-in (pass)
    CONFIG_CPUSETS: Cpuset support: built-in (pass)
    CONFIG_CGROUP_CPUACCT: Simple CPU accounting cgroup subsystem: built-in (pass)
    CONFIG_MEMCG: Memory Resource Controller for Control Groups: built-in (pass)
    CONFIG_CGROUP_HUGETLB: HugeTLB Resource Controller for Control Groups: built-in (pass)
      CONFIG_NETFILTER_XTABLES: Netfilter Xtables support: built-in (pass)
      CONFIG_NF_NAT: module (pass)
      CONFIG_NF_CONNTRACK_IPV4: IPv4 connetion tracking support (required for NAT): unknown (warning)
      CONFIG_NF_CONNTRACK_IPV6: IPv6 connetion tracking support (required for NAT): unknown (warning)
      CONFIG_LLC: module (pass)
      CONFIG_STP: module (pass)
  CONFIG_PROC_FS: /proc file system support: built-in (pass
```

### What happened?

`make -j$(nproc)` fails on first, and possibly second build.

```sh
$ git rev-parse HEAD        
c3d7c1b32f1af461192ad5ad9689c8f74d8fe71f
```

### Steps to reproduce

1. Checkout `main` at `c3d7c1b32f1af461192ad5ad9689c8f74d8fe71f`.
2. Run `make -j$(nproc)`.

### Expected behavior

_No response_

### Actual behavior

Too long to paste the entire log, so please see https://gist.github.com/mbana/c8e9cad7ef89b580afe48896eb88e34f.

Summary:

```sh
```sh
$ make -j$(nproc)
...
touch -- 'pkg/apis/k0s.k0sproject.io/v1beta1/.client-gen.stamp'
touch -- 'pkg/apis/autopilot.k0sproject.io/v1beta2/.client-gen.stamp'
github.com/k0sproject/k0s/internal/autopilot/testutil imports
	github.com/k0sproject/k0s/pkg/apis/autopilot.k0sproject.io/v1beta2/clientset: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
github.com/k0sproject/k0s/internal/autopilot/testutil imports
	github.com/k0sproject/k0s/pkg/apis/autopilot.k0sproject.io/v1beta2/clientset/fake: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
github.com/k0sproject/k0s/internal/testutil imports
	github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/clientset/fake: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
github.com/k0sproject/k0s/internal/testutil imports
	github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/clientset/typed/k0s.k0sproject.io/v1beta1: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
github.com/k0sproject/k0s/pkg/autopilot/controller imports
	github.com/k0sproject/k0s/pkg/apis/autopilot.k0sproject.io/v1beta2/clientset/scheme: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
github.com/k0sproject/k0s/pkg/config imports
	github.com/k0sproject/k0s/pkg/apis/k0s.k0sproject.io/clientset: git init --bare in /run/k0s-build/go/mod/cache/vcs/ae919b7ca7cc6029449e16c2a89eb1a5fedae32c585157fb07019da5434cb570: exec: "git": executable file not found in $PATH
make: *** [Makefile:107: go.sum] Error 1
```
```

### PR DESCRIPTION
**Pull request by twz123**
## Description

* Improve Docker export in embedded-binaries Makefile
  Ensure that the binaries are moved into place, instead of sequentially written. Use a temporary untar folder for that, and touch the file there before moving it.
* Remove unused "buildmode" make variable

* Don't remove auto-generated files during builds
  Place client-gen and controller-gen generated files into their own temporary folders and move them into their proper places afterwards. Don't delete any files from the source directories. This ensures that parallel builds work as expected and tools won't encounter any "missing" Go packages during builds.

  (There's still a small potential race for client-gen. The whole directory needs to be replaced, which mv refuses to do. So move the to- be-replaced directory out of the way, move the new directory in place, and delete the old directory afterwards.)
* Let generated assets depend on controller-gen output
  Amongst other things, the generated assets embed the controller-gen generated CRDs into the k0s binary. Ensure that all the CRDs have been generated _before_ generating the embedded assets, so that everything is guaranteed to be up to date, especially when using parallel make builds.
* Let CI use parallel builds

Fixes #2654.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,27 +126,32 @@ pkg/apis/autopilot/v1beta2/.controller-gen.stamp: gen_output_dir = autopilot
 
 pkg/apis/%/.controller-gen.stamp: .k0sbuild.docker-image.k0s hack/tools/boilerplate.go.txt hack/tools/Makefile.variables
 	rm -rf 'static/manifests/$(gen_output_dir)/CustomResourceDefinition'
-	rm -f -- '$(dir $@)'zz_*.go
-	CGO_ENABLED=0 $(GO) run sigs.k8s.io/controller-tools/cmd/controller-gen@v$(controller-gen_version) \
-	  crd \
-	  paths="./$(dir $@)..." \
-	  output:crd:artifacts:config=./static/manifests/$(gen_output_dir)/CustomResourceDefinition \
-	  object:headerFile=hack/tools/boilerplate.go.txt
+	mkdir -p 'static/manifests/$(gen_output_dir)'
+	gendir="$$(mktemp -d .controller-gen.XXXXXX.tmp)" \
+	  && trap "rm -rf -- $$gendir" INT EXIT \
+	  && CGO_ENABLED=0 $(GO) run sigs.k8s.io/controller-tools/cmd/controller-gen@v$(controller-gen_version) \
+	    paths="./$(dir $@)..." \
+	    object:headerFile=hack/tools/boilerplate.go.txt output:object:dir="$$gendir" \
+	    crd output:crd:dir='static/manifests/$(gen_output_dir)/CustomResourceDefinition' \
+	  && mv -f -- "$$gendir"/zz_generated.deepcopy.go '$(dir $@).'
 	touch -- '$@'
 
 clientset_input_dirs := pkg/apis/autopilot/v1beta2 pkg/apis/k0s/v1beta1 pkg/apis/helm/v1beta1
 codegen_targets += pkg/client/clientset/.client-gen.stamp
 pkg/client/clientset/.client-gen.stamp: $(shell find $(clientset_input_dirs) -type f -name \*.go -not -name \*_test.go -not -name zz_\*)
 pkg/client/clientset/.client-gen.stamp: .k0sbuild.docker-image.k0s hack/tools/boilerplate.go.txt embedded-bins/Makefile.variables
-	rm -rf -- '$(dir $@)'
-	CGO_ENABLED=0 $(GO) run k8s.io/code-generator/cmd/client-gen@v$(kubernetes_version:1.%=0.%) \
-	  --go-header-file=hack/tools/boilerplate.go.txt \
-	  --input-base='' \
-	  --input=$(subst $(space),$(comma),$(clientset_input_dirs:%=github.com/k0sproject/k0s/%)) \
-	  --output-package=github.com/k0sproject/k0s/$(patsubst %/,%,$(dir $(patsubst %/,%,$(dir $@)))) \
-	  --clientset-name=$(notdir $(patsubst %/,%,$(dir $@))) \
-	  --output-base=. \
-	  --trim-path-prefix=github.com/k0sproject/k0s
+	gendir="$$(mktemp -d .client-gen.XXXXXX.tmp)" \
+	  && trap "rm -rf -- $$gendir" INT EXIT \
+	  && CGO_ENABLED=0 $(GO) run k8s.io/code-generator/cmd/client-gen@v$(kubernetes_version:1.%=0.%) \
+	    --go-header-file=hack/tools/boilerplate.go.txt \
+	    --input-base='' \
+	    --input=$(subst $(space),$(comma),$(clientset_input_dirs:%=github.com/k0sproject/k0s/%)) \
+	    --output-package=github.com/k0sproject/k0s/pkg/client \
+	    --clientset-name=clientset \
+	    --output-base="$$gendir/out" \
+	    --trim-path-prefix=github.com/k0sproject/k0s \
+	  && { [ ! -e pkg/client/clientset ] || mv -- pkg/client/clientset "$$gendir/old"; } \
+	  && mv -f -- "$$gendir/out/github.com/k0sproject/k0s/pkg/client/clientset" pkg/client/.
 	touch -- '$@'
 
 codegen_targets += static/zz_generated_assets.go
--- a/Makefile
+++ b/Makefile
@@ -112,18 +112,19 @@ $(K0S_GO_BUILD_CACHE):
 go.sum: go.mod .k0sbuild.docker-image.k0s
 	$(GO) mod tidy && touch -c -- '$@'
 
-codegen_targets += pkg/apis/helm/v1beta1/.controller-gen.stamp
+controllergen_targets += pkg/apis/helm/v1beta1/.controller-gen.stamp
 pkg/apis/helm/v1beta1/.controller-gen.stamp: $(shell find pkg/apis/helm/v1beta1/ -maxdepth 1 -type f -name \*.go)
 pkg/apis/helm/v1beta1/.controller-gen.stamp: gen_output_dir = helm
 
-codegen_targets += pkg/apis/k0s/v1beta1/.controller-gen.stamp
+controllergen_targets += pkg/apis/k0s/v1beta1/.controller-gen.stamp
 pkg/apis/k0s/v1beta1/.controller-gen.stamp: $(shell find pkg/apis/k0s/v1beta1/ -maxdepth 1 -type f -name \*.go)
 pkg/apis/k0s/v1beta1/.controller-gen.stamp: gen_output_dir = v1beta1
 
-codegen_targets += pkg/apis/autopilot/v1beta2/.controller-gen.stamp
+controllergen_targets += pkg/apis/autopilot/v1beta2/.controller-gen.stamp
 pkg/apis/autopilot/v1beta2/.controller-gen.stamp: $(shell find pkg/apis/autopilot/v1beta2/ -maxdepth 1 -type f -name \*.go)
 pkg/apis/autopilot/v1beta2/.controller-gen.stamp: gen_output_dir = autopilot
 
+codegen_targets += $(controllergen_targets)
 pkg/apis/%/.controller-gen.stamp: .k0sbuild.docker-image.k0s hack/tools/boilerplate.go.txt hack/tools/Makefile.variables
 	rm -rf 'static/manifests/$(gen_output_dir)/CustomResourceDefinition'
 	mkdir -p 'static/manifests/$(gen_output_dir)'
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: .github/workflows/autopilot-branch-to-latest.yaml [hunk 1]
```diff
@@ -53,7 +53,7 @@ jobs:
     - name: Build Tool
       run: |
         cd hack/tool
-        make
+        make --jobs=$(nproc)
```
kke, line 56: I think all these `make`s are supposed to be `$(MAKE)` inside Makefiles.

mikhail-sakhnov:
@twz123  Could we keep buildmode=None behavior somehow? 
I use it once in a while for different experiments or when I need to work locally on mac os without remote machine.

twz123:
> Could we keep buildmode=None behavior somehow? I use it once in a while for different experiments or when I need to work locally on mac os without remote machine.

You mean `make EMBEDDED_BINS_BUILDMODE=none`? I don't think this PR interferes with it. If it does, this was certainly not intended. That mode is also used in some CI steps, so it should definitely remain intact. Please let me know if there are any issues with it.

mikhail-sakhnov:
@twz123 I was confused with "Remove unused "buildmode" make variable" but it seems to be about other thing, nvm

.github/workflows/autopilot-branch-to-latest.yaml
mikhail-sakhnov, line 56: I don't really understand how here and everywhere in the diff -j helps to parallelize jobs, majority of the targets are single external commands and all of the make calls uses only one target.

twz123: This is to ensure that k0s actually supports parallel builds, not so much to speed up things. Parallel builds are advertised in the README, so we should ensure that there are no regressions.

It's so easy to hack on the Makefiles and do something that works flawlessly in sequential builds, but will break when run in parallel.

mikhail-sakhnov: Other option is to change the readme and drop the claim, because in fact we do not support parallel build more than go build toolchain supports it. Makefile in our case basically just a job runner, not the build tool per se. Claiming that we support parallel build by passing -j is not something that exactly reflects the reality.

twz123: It actually speeds things up a bit when building the embedded binaries. We can also drop the support for parallel builds, but then I think we should explicitly say that in the README.

@ncopa: You added the note about parallel builds. Any thoughts on that?

mikhail-sakhnov: 15-20% is a significant difference. We could also keep the -j for the embedded binaries call (I guess it's coming from `make k0s`?) but drop the claim and don't put -j everywhere.

ncopa: I think that ensuring support for parallel builds will also help us make sure that partial builds always works. I think its worth the extra complexity.

ncopa (dismissed):
LGTM

twz123:
I'll revisit this once changes around #2909 have been sorted out.

Makefile
```diff
@@ -106,58 +106,68 @@ $(K0S_GO_BUILD_CACHE):
...
+	  && trap "rm -rf -- $$gendir" INT EXIT \
+	  && CGO_ENABLED=0 $(GO) run k8s.io/code-generator/cmd/client-gen@v$(kubernetes_version:1.%=0.%) \
+	    --go-header-file=hack/tools/boilerplate.go.txt \
+	    --input-base='' \
+	    --input=$(subst $(space),$(comma),$(clientset_input_dirs:%=github.com/k0sproject/k0s/%)) \
+	    --output-package=github.com/k0sproject/k0s/pkg/client \
+	    --clientset-name=clientset \
+	    --output-base="$$gendir/out" \
+	    --trim-path-prefix=github.com/k0sproject/k0s \
+	  && { [ ! -e pkg/client/clientset ] || mv -- pkg/client/clientset "$$gendir/old"; } \
```
ncopa, line 155: What is the reasoning behind this? As I read how this works:

- if pkg/client/clientset exists, move it to temp directory and delete it directly with the INT/EXIT trap.

Why not delete it immediatly and ignore errors? (to avoid race condition?)

```suggestion
	  && { mv -- pkg/client/clientset "$$gendir/old" 2>/dev/null || true; } \
```

It is a general good idea to avoid test if something exists before operation, as it is theoretically possible that the thing is created/removed between the test and the operation. Better to just do the operation and deal with (or ignore) the not-found error.

Makefile
```diff
@@ -106,58 +106,68 @@ $(K0S_GO_BUILD_CACHE):
...
+	  && mv -f -- "$$gendir/out/github.com/k0sproject/k0s/pkg/client/clientset" pkg/client/.
 	touch -- '$@'
 
 codegen_targets += static/zz_generated_assets.go
-static_asset_dirs := static/manifests static/misc
+static/zz_generated_assets.go: $(controllergen_targets) # to generate the CRDs into static/manifests/*/CustomResourceDefinition
+static/zz_generated_assets.go: $(shell find static/manifests/calico static/manifests/windows static/misc -type f)
```
ncopa, line 161: Not sure how many times or when the dependency is evaluated, but this ensure that we only executes the `find` command only once.

```suggestion
static_asset_files := $(shell find static/manifests/calico static/manifests/windows static/misc -type f)
static/zz_generated_assets.go: $(controllergen_targets) # to generate the CRDs into static/manifests/*/CustomResourceDefinition
static/zz_generated_assets.go: $(static_asset_files)
```

ncopa (dismissed):
Looks good enough to me

ncopa:
We could also evaluate to use a global `MAKEFLAGS="-j"` for the workflows.

Makefile
twz123: > Better to just do the operation and deal with (or ignore) the not-found error.

Agree.  However, with `mv`, there's no way to distinguish the not found case from any other error. Redirecting that to /dev/null even doesn't allow for spotting anything in the logs after the fact. I'd prefer to leave it as is, so we actually fail if `mv` failed. Even if the check is racy, nothing should remove that directory in the meantime anyways, hence I'd prefer that operation to fail in that case.

twz123:
> We could also evaluate to use a global `MAKEFLAGS="-j"` for the workflows.

On it. Will also move one or two commits into a separate PR.

Makefile
twz123: Isn't that done only once? I mean, it's the requisites. I can't imagine that make won't cache those? It would probably break everything, because they're used to construct the build graph, meaning that the build plan might change in-flight? We already have some other places with that pattern.

twz123:
@ncopa would you mind to have another look?

ncopa (approved):
LGTM

PR #2674, opened 2023-02-01
Use custom log writer for stdout/err in supervisor

twz123:
## Description

The `io.Writer` implementation provided by logrus refuses to write lines that are longer than 64k. An error is logged and the log pipe is closed. This will cause the stdout or stderr of the supervised process to be broken ("broken pipe"). Depending on the supervised process, the result may vary. At a minimum, k0s will no longer log anything from the affected processes. Some processes (e.g. the kubelet) will terminate themselves when trying to use stdout/stderr, resulting in a restart.

Use a custom `io.Writer` instead. It will do pretty much the same thing as logrus's one, but it will chunk long lines into multiple log statements.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/supervisor/logwriter.go
+++ b/pkg/supervisor/logwriter.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2022 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package supervisor
+
+import (
+	"bytes"
+	"unicode/utf8"
+
+	"github.com/sirupsen/logrus"
+)
+
+// logWriter implements [io.Writer] by forwarding whole lines to log. In case
+// the lines get too long, it logs them in multiple chunks.
+//
+// This is in contrast to logrus's implementation of io.Writer, which simply
+// errors out if the log line gets longer than 64k.
+type logWriter struct {
+	log     logrus.FieldLogger // receives (possibly chunked) log lines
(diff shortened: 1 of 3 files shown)


pkg/supervisor/logwriter.go
```diff
@@ -0,0 +1,87 @@
...
+	// Fill and drain buffer with available data until everything has been consumed.
+	for rest := in; len(rest) > 0; {
+
+		n := copy(w.buf[w.len:], rest) // fill buffer with new input data
+		rest = rest[n:]                // strip copied input data
+		w.len += n                     // increase buffer length accordingly
+
+		// Loop over buffer as long as there are newlines in it
+		for off := 0; ; {
+			idx := bytes.IndexByte(w.buf[off:w.len], '\n')
```
kke, line 53: Maybe: 

```suggestion
			idx := bytes.IndexRune(w.buf[off:w.len], '\n')
```

It could theoretically break a multibyte character that contains 0x0a when using IndexByte

twz123: True. Better use unicode-aware stuff!

twz123: OTOH, I think we assume UTF-8 or some single-byte ASCII superset here. IIUC, there's no valid UTF-8 byte sequence that may contain any 0x0a which is actually _not_ a line break. ASCII characters are encoded as single bytes as-is, everything else as byte sequence where the highest bit is set on all bytes, i.e. they're never in the ASCII plane.

Things would get quite complicated if we'd want to add proper charset support here. WDYT?

twz123: ... but chunking could actually break a multibyte character.... :thinking:

juanluisvaladas (approved):
LGTM.

pkg/supervisor/logwriter_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogWriter(t *testing.T) {
+	type entry struct {
+		chunk uint
+		msg   string
+	}
+
+	for _, test := range []struct {
```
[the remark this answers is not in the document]
twz123: Done!

juanluisvaladas (approved):
LGTM.

PR #2259, opened 2022-10-13
Add some notes in docs about iptables versions and modes

jnummelin:
Signed-off-by: Jussi Nummelin <jnummelin@mirantis.com>

## Description

So auxiliary documentation over iptables, firewalld and friends and how their configuration and versions _might_ affect k0s / k8s components.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2124, opened 2022-08-31
iptables-nft support

makhov:
Signed-off-by: Alexey Makhov <amakhov@mirantis.com>

## Description

iptables-nft support. kube-proxy and kube-router support nftables out of the box:

```
bash-5.1# k0s kc exec -it -n kube-system kube-proxy-5kmfn -- iptables -V
iptables v1.8.7 (nf_tables)
bash-5.1# k0s kc exec -it -n kube-system kube-router-v5jwv -- iptables -V
Defaulted container "kube-router" out of: kube-router, install-cni-bins (init), install-cniconf (init)
iptables v1.8.7 (nf_tables)
```

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

- [ ] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [ ] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2680, opened 2023-02-03
Bump iptables to v1.8.9

twz123:
## Description

https://www.netfilter.org/projects/iptables/files/changes-iptables-1.8.9.txt

Removed some unneeded packages and patches as well.

Need to check if versions >= 1.8.8 are okay to use, though. Unsure if the potential incompatible cases are covered by the integration tests.

See also:

- #2079
- #2124
- #2259
- **[Netfilter bug 1632:  iptables 1.8.8 breaks compatibility with older versions for some rules][nf-1632]**
- #2277

[nf-1632]: https://bugzilla.netfilter.org/show_bug.cgi?id=1632

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


embedded-bins/iptables/Dockerfile
```diff
@@ -1,12 +1,14 @@
 ARG BUILDIMAGE
 FROM $BUILDIMAGE AS build
 
-RUN apk add build-base git file curl \
-	linux-headers pkgconf bison flex libmnl-dev libmnl-static libnftnl-dev
+RUN apk add build-base curl pkgconf \
+	linux-headers \
+	libmnl-dev libmnl-static \
+	libnftnl-dev
```
twz123, line 7: Not sure if we should build those dependencies from scratch. We're not controlling the versions in any way here, and are just taking the stuff that's in the Alpine package repos at build time. This has been tackled a bit in #2124 (5391cddb8d80c8e1d3a8328ce56daf63db4ae287) for technical reasons, but has been [discarded](https://github.com/k0sproject/k0s/pull/2124#discussion_r960630241).

We're also building libseccomp from scratch when building runc, even if [libseccomp.a is available in the Alpine repos](https://pkgs.alpinelinux.org/contents?file=libseccomp.a&path=&name=&branch=v3.17).

twz123, line 7: I noticed this configure output

```text
checking for libnfnetlink >= 1.0... no
checking for libmnl >= 1.0... yes
checking for libnftnl >= 1.1.6... yes
checking for libnetfilter_conntrack >= 1.0.6... no
WARNING: libnetfilter_conntrack not found, connlabel match will not be built
```

```text
Iptables Configuration:
  IPv4 support:                         yes
  IPv6 support:                         yes
  Devel support:                        no
  IPQ support:                          no
  Large file support:                   yes
  BPF utils support:                    no
  nfsynproxy util support:              no
  nftables support:                     yes
  connlabel support:                    no
  profiling support:                    no
```

Are we okay with the missing connlabel support? The static libs for the needed dependencies are not in the Alpine 3.17 repos, but in edge.

twz123:
Currently experimenting with iptables 1.8.9 for the 1.28 release. /ref #3250

twz123:
The 1.28 kube-proxy OCI image now ships iptables 1.8.9, too. All the OS tests pass with this change. So I think this is good to go.

embedded-bins/iptables/Dockerfile
ncopa: I think we want the libnetfilter_conntrack support. I believe I have seen `--label` in some iptables rule, but I don't remember where (and I cannot find it).

IPQ seems to be user space filter/mangling of packets. I don't think we need that.

ncopa: I don't think we need build all deps from scratch unless there is a good reason for it.

libseccomp has been problematic in the past so I think it makes sense to build that.

ncopa (approved):

Issue #2653, opened 2023-01-27
IPv6 / dual stack integration testing


twz123:
There's currently no proper IPv6 / dual stack support in the k0s integration test suite. This is not ideal, as k0s explicitly supports dual stack. The current `dualstack` integration test only checks that k0s passes the correct flags to the components involved, but doesn't verify that the networking is actually operational in any way.

The upcoming kube-router major release adds support for dual stack, so both CNI implementations natively supported by k0s are now IPv6-enabled. Might be a good time to actually add some proper test coverage for these setups.

There's some challenges, though:

* Docker needs to be setup for IPv6.
* Footloose might need some more patches to support IPv6 as well.

Other things TBD:

* Configure k0s's footloose container accordingly
* Maybe find a way to extend `FootlooseSuite` and friends to select the mode (v4/v6/dual)

martencassel:
What kind of integration tests would be required for covering dualstack ? 

Would enabling IPv6 support int the docker daemon like so
   "ipv6": true,"fixed-cidr-v6": "2001:db8:1::/64"
be enough for such a test ?

twz123:
> What kind of integration tests would be required for covering dualstack ?

We [already have](https://github.com/k0sproject/k0s/blob/v1.26.0%2Bk0s.0/inttest/dualstack/dualstack_test.go) a "test" for dual stack, but this merely checks if k0s passes the right args to its components. We'd need to have some tests that actually probe IPv6 connectivity. The current tests don't do that because it's not expected to work in the current test setup anyway.

> Would enabling IPv6 support in https://docs.docker.com/config/daemon/ipv6/ be enough for such a test ?

That's basically the first bullet point. The other bullet points may require a solution as well.

PR #2711, opened 2023-02-08
Add dual stack network tests

martencassel:
fixes https://github.com/k0sproject/k0s/issues/2653

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -1439,3 +1439,21 @@ func (s *FootlooseSuite) AssertSomeKubeSystemPods(client *kubernetes.Clientset) 
 
 	return false
 }
+
+func (s *FootlooseSuite) IsDockerIPv6Enabled() (bool, error) {
+	cmd := exec.Command("docker", "inspect", "bridge", "--format", "\"{{ .EnableIPv6 }}\"")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("failed to run docker inspect: %w", err)
+	}
+	var docker_bridge_EnableIPv6 string
+	err = json.Unmarshal(output, &docker_bridge_EnableIPv6)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse default docker bridge EnableIPv6: %w", err)
+	}
+	bridge_EnableIPv6, err := strconv.ParseBool(docker_bridge_EnableIPv6)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse default docker bridge EnableIPv6: %w", err)
+	}
+	return bridge_EnableIPv6, nil
+}
--- a/inttest/dualstack/dualstack_test.go
+++ b/inttest/dualstack/dualstack_test.go
@@ -25,6 +25,7 @@ package dualstack
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"strings"
 
@@ -33,8 +34,16 @@ import (
 
 	"testing"
 
+	"time"
+
 	"github.com/k0sproject/k0s/inttest/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8s "k8s.io/client-go/kubernetes"
+
+	"context"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 type DualstackSuite struct {
@@ -77,7 +86,11 @@ func (s *DualstackSuite) cmdlineForExecutable(node, binary string) []string {
 }
 
 func (s *DualstackSuite) SetupSuite() {
+	isDockerIPv6Enabled, err := s.IsDockerIPv6Enabled()
+	s.NoError(err)
+	s.Require().True(isDockerIPv6Enabled, "Please enable IPv6 in docker before running this test")
 	s.FootlooseSuite.SetupSuite()
+
 	s.PutFile(s.ControllerNode(0), "/tmp/k0s.yaml", k0sConfigWithDualStack)
 	controllerArgs := []string{"--config=/tmp/k0s.yaml"}
 	if os.Getenv("K0S_ENABLE_DYNAMIC_CONFIG") == "true" {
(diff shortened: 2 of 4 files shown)


martencassel:
I cant see why this fails in pipeline.

jnummelin:
```
    footloosesuite.go:199: 
        	Error Trace:	/home/runner/work/k0s/k0s/inttest/dualstack/footloosesuite.go:199
        	            				/home/runner/work/k0s/k0s/inttest/dualstack/asm_amd64.s:1598
        	Error:      	Test deadline exceeded
        	Test:       	TestDualStack
    footloosesuite.go:202: Cleaning up
    footloosesuite.go:209: Test cleanup deadline: 2023-02-08 23:07:36.547129158 +0000 UTC m=+220.012969228
    dualstack_test.go:130: 
        	Error Trace:	/home/runner/work/k0s/k0s/inttest/dualstack/dualstack_test.go:130
        	            				/home/runner/work/k0s/k0s/inttest/dualstack/suite.go:132
        	            				/home/runner/work/k0s/k0s/inttest/dualstack/dualstack_test.go:241
        	Error:      	Received unexpected error:
        	            	context deadline exceeded
        	Test:       	TestDualStack
        	Messages:   	nginx pod did not start
```
For some reason the smoke was not able to gather all the logs 🤔 

Do you have the possibility to run it locally?

martencassel:
I have and see no such errors

[dualstack.log](https://github.com/k0sproject/k0s/files/10695940/dualstack.log)

inttest/common/footloosesuite.go
```diff
@@ -1444,3 +1445,37 @@ func (s *FootlooseSuite) AssertSomeKubeSystemPods(client *kubernetes.Clientset)
...
+	}
+	err = ioutil.WriteFile(tmpFile.Name(), json, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to write to temporary file: %w", err)
+	}
+	err = exec.Command("sudo", "mv", "-f", tmpFile.Name(), "/etc/docker/daemon.json").Run()
+	if err != nil {
+		return fmt.Errorf("failed to write daemon.json file: %w", err)
+	}
+	err = exec.Command("sudo", "systemctl", "reload", "docker").Run()
```
jnummelin, line 1476:
```suggestion
	err = exec.Command("sudo", "systemctl", "restart", "docker").Run()
```

IIRC `reload` just reloads the unit from disk but does NOT actually restart it

martencassel: sure we can try that
i folloed https://docs.docker.com/config/daemon/ipv6/

jnummelin:
I think there's one fundamental _thing_ we need to decide on. So the current version in this PR changes the docker config via code and restarts the unit. That has couple general challenges:
- tied to systemd; remember many people use e.g. osx as the dev env
- docker config file location hard coded

I'm afraid that there's too much variation in the envs (linux vs. osx vs. all the possible Linux distros) where we run the smokes to handle docker config at the test code.

What I'm instead of thinking is if we could detect whether docker has ipv6 support enabled in the test case. If it's enabled, then run the test case as expected. If it's NOT enabled, bail out with some generic help info for the user how to enable it in docker. Then in smokes we could enable it via actions.

So essentially we "externalize" the docker ipv6 config for the user and/or to CI setup.

@twz123 You've had some thoughts on this too, any comments?

inttest/common/footloosesuite.go
```diff
@@ -1444,3 +1445,37 @@ func (s *FootlooseSuite) AssertSomeKubeSystemPods(client *kubernetes.Clientset)
 
 	return false
 }
+
+type IPV6Config struct {
+	IPv6        bool   `json:"ipv6"`
+	FixedCidrV6 string `json:"fixed-cidr-v6"`
+}
+
+func (s *FootlooseSuite) ConfigureIPv6Docker(enabled bool) error {
```
twz123, line 1454: This is something that's host-specific. Better put it into the GitHub workflows? Maybe as a shell script that's invoked as part of a `run` script?

martencassel: That sounds as a good idea i can look into it.
restarting the docker daemon must also be done before the footloose suite creates all containers.

martencassel: One way to test if IPv6 is enabled could be to check if the default docker bridge network has EnableIPv6 set to true.

martencassel: Should we bail out with an error or not ?

martencassel:
I added a test for IPv6 config in docker, if not enabled then fail.

inttest/dualstack/dualstack_test.go
```diff
@@ -94,9 +111,122 @@ func (s *DualstackSuite) SetupSuite() {
...
+		}
+		s.T().Log("server response", out)
+		return strings.Contains(out, "Welcome to nginx"), nil
+	})
+	s.Require().NoError(err)
+
 	s.client = client
 }
 
+func getAlpineVersion(t *testing.T) string {
```
juanluisvaladas, line 220: You can remove all this logic if you use nginx for both pods. The nginx image has both wget and curl so you can remove this.

juanluisvaladas (commented):
The test in general looks quite good but it can be made faster by using nginx pods as clients.

You can create 2 ngnix pods called nginx-worker0 and nginx-worker1 and then have nginx-worker-0 curl/wget both nginx-worker-1 ip addresses. You are effectively testing the same things but you have to pull half the images and run half the containers.

Also this allows to reuse more code and would make it cleaner.

Other than that, the test looks good to me, now that docker isn't being reconfigured I don't see any red flags and the test is useful and tests what it has to test.

inttest/dualstack/dualstack_test.go
martencassel: I did as you suggested, see new commit.

twz123 (changes requested):
Thanks for working on this @martencassel. "Works on my machine." :+1: To make the integration tests pass on the GitHub runners, we need to enable IPv6 in Docker on those. That's basically the things you did [here][enable-ipv6]. We just need them as another build step in the `smoketest` job in `.github/workflows/go.yml`, somewhere after (or as part of) the `Prepare build environment` step. Would you mind to rewrite that as a shell script and add add it there?

Moreover, there are some lints concerning variable naming. We're not currently linting the integration tests during the lint CI job, nevertheless it would be great if you could address those:

```
inttest/dualstack/dualstack_test.go:147:3: var-naming: don't use underscores in Go names; var pod_ip should be podIP (revive)
                pod_ip := targetPod.Status.PodIPs[1].IP
                ^
inttest/dualstack/dualstack_test.go:148:3: var-naming: don't use underscores in Go names; var target_ip should be targetIP (revive)
                target_ip := net.ParseIP(pod_ip)
                ^
inttest/dualstack/dualstack_test.go:163:3: var-naming: don't use underscores in Go names; var pod_ip should be podIP (revive)
                pod_ip := targetPod.Status.PodIPs[0].IP
                ^
inttest/dualstack/dualstack_test.go:164:3: var-naming: don't use underscores in Go names; var target_ip should be targetIP (revive)
                target_ip := net.ParseIP(pod_ip)
                ^
inttest/common/footloosesuite.go:1454:6: var-naming: don't use underscores in Go names; var docker_bridge_EnableIPv6 should be dockerBridgeEnableIPv6 (revive)
        var docker_bridge_EnableIPv6 string
            ^
inttest/common/footloosesuite.go:1459:2: var-naming: don't use underscores in Go names; var bridge_EnableIPv6 should be bridgeEnableIPv6 (revive)
        bridge_EnableIPv6, err := strconv.ParseBool(docker_bridge_EnableIPv6)
        ^
```

[enable-ipv6]: https://github.com/k0sproject/k0s/commit/9ae95fdac6dbf8cc598734a256c5919d1f84dd66#diff-85e4a1b54e18b85233536b2d526e695deea45e09bd65ecf7e4ed55a7c1441b4eR1454-R1481

inttest/dualstack/dualstack_test.go
juanluisvaladas: Looks better, we need to do change some stuff on the CI before merging. Other than the linting stuff that Tom mentioned it looks good

twz123 (approved):

twz123:
Ha! Enabling IPv6 in Docker has side effects on the airgap integration test :tada:

I'll have a look into that.

twz123:
You could add a condition to the "Prepare docker for ipv6 dualstack tests" build step, like this: `if: contains(matrix.smoke-suite, 'dualstack')`.

That way, we can proceed here and can fix the airgap test in subsequent PRs.

twz123 (approved):

twz123:
Thanks again!

martencassel:
Thankyou it was pure fun :)

PR #2716, opened 2023-02-09
Refactor inttest for calico. Reduce number of image pulls.

martencassel:

### DIFF
--- a/inttest/calico/calico_test.go
+++ b/inttest/calico/calico_test.go
@@ -19,13 +19,10 @@ package calico
 import (
 	"context"
 	"fmt"
-	"os/exec"
-	"path/filepath"
 	"strings"
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


inttest/calico/calico_test.go
```diff
@@ -72,36 +69,32 @@ func (s *CalicoSuite) TestK0sGetsUp() {
 
 	createdTargetPod, err := kc.CoreV1().Pods("default").Create(s.Context(), &corev1.Pod{
 		TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
-		ObjectMeta: metav1.ObjectMeta{Name: "nginx"},
+		ObjectMeta: metav1.ObjectMeta{Name: "nginx-worker0"},
 		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{{Name: "nginx", Image: "docker.io/library/nginx:1.23.1-alpine"}},
+			Containers: []corev1.Container{{Name: "nginx-worker0", Image: "docker.io/library/nginx:1.23.1-alpine"}},
```
twz123, line 74: There's another flavor, which seems to be "smaller": `docker.io/library/nginx:1.23.3-alpine-slim` (but that might not include `wget`, I didn't check).

inttest/calico/calico_test.go
```diff
@@ -72,36 +69,32 @@ func (s *CalicoSuite) TestK0sGetsUp() {
...
 		TypeMeta:   metav1.TypeMeta{Kind: "Pod", APIVersion: "v1"},
-		ObjectMeta: metav1.ObjectMeta{Name: "alpine"},
+		ObjectMeta: metav1.ObjectMeta{Name: "nginx-worker1"},
 		Spec: corev1.PodSpec{
-			Containers: []corev1.Container{{
-				Name:    "alpine",
-				Image:   "docker.io/library/alpine:" + getAlpineVersion(s.T()),
-				Command: []string{"sleep", "infinity"},
-			}},
+			Containers: []corev1.Container{{Name: "nginx-worker1", Image: "docker.io/library/nginx:1.23.1-alpine"}},
```
twz123, line 90: Would be consequent to introduce a local constant for the image string, so it's clear that both workers should pull the same.

twz123 (commented):
I don't think this reduces the number of image pulls. The pods get scheduled on two different nodes, so the images get pulled nevertheless for both of them, even if they're the same. The nginx image is even slightly bigger than the vanilla alpine image.

PR #2700, opened 2023-02-07
Copyright 2023

twz123:
## Description

Update all copyright comments from 2022 to 2023.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #2718, opened 2023-02-09
Fix copyright

juanluisvaladas:
## Description

As discussed in #2700 we're not defining copyright correctly. This PR has 3 commits:
The first one adds the copyright in the files that didn't have one and should have one (this is done manually by me)
The second one fixes all copyrights, this was done programatically with a script added in the third commit
The third one adds the script and modifies the Makefile

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


juanluisvaladas:
I'm making this a draft because it works on my machine and I need to debug it in GH actions.

.go-header.txt
```diff
@@ -1,4 +1,4 @@
-Copyright 2022 k0s authors
+Copyright 202. k0s authors
```
twz123, line 1: Oh... it's a regex? Nice. Wasn't clear to me. Have to investigate if that might help fixing https://github.com/k0sproject/k0s/blob/v1.26.1+k0s.0/.golangci.yml#L34-L35 ...

twz123: That opens up the question if we should escape all the literal dots in the license header :see_no_evil:

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+
+get_year(){
+    # The head -1 has to be added because if a file has been added, removed and added
+    # again we'll get two lines with both dates. We only pick the newest one.
+    git log --diff-filter=A --pretty=format:%as -- ${1} | cut -d- -f1 | head -1 ||
+        (echo "FATAL!! ${1} doesn't seem to be commited in the repo" 1>&2 ; exit 1)
+}
+
+has_copyright(){
+	DATE=${1}
```
twz123, line 30: I prefer using `$...` by default and to fall back to `${...}` only if it's required. WDYT?

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+set -euo pipefail
+
+RESULT=0
+FIX=${FIX:=n}
+
+get_year(){
+    # The head -1 has to be added because if a file has been added, removed and added
+    # again we'll get two lines with both dates. We only pick the newest one.
+    git log --diff-filter=A --pretty=format:%as -- ${1} | cut -d- -f1 | head -1 ||
+        (echo "FATAL!! ${1} doesn't seem to be commited in the repo" 1>&2 ; exit 1)
```
twz123, line 26: * I'd also follow renames.
* I'd quote the file name, just in case.
* Wouldn't `head -1` be replaceable by adding `-1` to `git log`?
* The `cut` can be replaced by asking git to format the date as we want: `--date=format:%Y --pretty=format:%ad`
* No need for a subshell, so use curly braces instead of round ones.

```suggestion
    # The -1 has to be added because if a file has been added, removed and added
    # again we'll get two lines with both dates. We only pick the newest one.
    git log -1 --follow --diff-filter=A --date=format:%Y --pretty=format:%ad -- "$1" || {
        echo "FATAL!! $1 doesn't seem to be committed in the repository." 1>&2
        return 1
    }
```

I guess the above changes remove three process invocations.

Makefile
```diff
@@ -153,6 +153,11 @@ pkg/apis/%/.client-gen.stamp: .k0sbuild.docker-image.k0s hack/tools/boilerplate.
 	  --output-package=github.com/k0sproject/k0s/pkg/apis/$(gen_output_dir)/
 	touch -- '$@'
 
+codegen_targets += hack/.copyright.stamp
+hack/.copyright.stamp:
+	hack/copyright.sh
+	touch -- '$@'
```
twz123, line 159: I wonder if we should make this a PHONY target à la `lint-copyright`. It has no dependencies declared, so it will almost never be invoked again. We can invoke it explicitly in the lint CI job and maybe mention it in the contribution section in the docs.

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+
+	# GNU grep is pretty fast but there are loads of files.
+	# By using -m1 we limit to the first occurence and by using -F
+	# the regex should be
+	grep -q -m1 -F "Copyright ${DATE} k0s authors" "${FILE}"
+}
+
+# Deliberately do not search in docs as the date of the matches for the
+# Copyright notice aren't related to the date of the document.
+for i in $(find cmd hack internal inttest pkg static -type f -name '*.go' | grep -v 'zz_generated'); do
```
twz123, line 41:
```suggestion
for i in $(find cmd hack internal inttest pkg static -type f -name '*.go' -not -name 'zz_generated*' ); do
```

inttest/common/pod.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 k0s authors
```
twz123, line 2: Any idea why this has been missed by golangci-lint? Maybe the inttests aren't included?

pkg/cleanup/bridge_windows.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2021 k0s authors
```
twz123, line 2: Bummer! I bet this is excluded because ob build tags. Maybe we should run the golangci-lint also for Windows?

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+# Copyright 2023 k0s authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http:#www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
juanluisvaladas: I copied this from the file `pkg/component/worker/calico_installer_windows.go`. Funnily enough, it's the only file that has it wrong. I will replace this manually in that file as part of the first commit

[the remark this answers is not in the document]
twz123: Seems like this also escaped golangci-lint because of the windows build tag :roll_eyes:.

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+    git log --diff-filter=A --pretty=format:%as -- ${1} | cut -d- -f1 | head -1 ||
+        (echo "FATAL!! ${1} doesn't seem to be commited in the repo" 1>&2 ; exit 1)
+}
+
+has_copyright(){
+	DATE=${1}
+	FILE=${2}
+
+	# GNU grep is pretty fast but there are loads of files.
+	# By using -m1 we limit to the first occurence and by using -F
```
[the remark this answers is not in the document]
juanluisvaladas: That makes sense. Removing the -m1 and the comment.

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+# limitations under the License.
+
+set -euo pipefail
+
+RESULT=0
+FIX=${FIX:=n}
+
+get_year(){
+    # The head -1 has to be added because if a file has been added, removed and added
+    # again we'll get two lines with both dates. We only pick the newest one.
```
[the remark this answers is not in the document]
juanluisvaladas: I didn't know this flag. Just replaced it.

.go-header.txt
juanluisvaladas: Good point, I added them.

hack/copyright.sh
juanluisvaladas: I don't feel very strongly about this. I will change it.

pkg/cleanup/bridge_windows.go
juanluisvaladas: Makes sense, since you have a windows machine, would you mind running go lint in your host just to check it?

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
...
+    git log --diff-filter=A --pretty=format:%as -- ${1} | cut -d- -f1 | head -1 ||
```
twz123, line 25: Btw I don't think `git log` will error out in case the file is not found. Maybe we should better check for an empty string?

juanluisvaladas: Hmm you're right

hack/copyright.sh
juanluisvaladas: Done

Makefile
juanluisvaladas: > I wonder if we should make this a PHONY target à la lint-copyright

Actually I considered this, but I don't think we can do this because it's one of the codegen_targets. The executions of controller-gen create the auto generated files with the copyright in the header using the static file `hack/tools/boilerplate.go.txt`. This header has a year which this script must fix, so this is called in the target `codegen`, if this target is removed, then make codegen won't set the correct copyright.

Other option would be calling the script in both codegen and lint, but the script about 12-13 seconds in my machine so that seems wasteful.

I considered making it depend on `$(shell find cmd hack internal inttest pkg static 1 -type f -name \*.go)` but this would mean that every time we make a little change and run make k0s you're rerunning the whole code which seems fairly unreasonable because in my machine it takes around 12 seconds to complete.

What we should probably do is make it depend on `pkg/apis/*/.controller-gen.stamp`.

What do you think?

> We can invoke it explicitly in the lint CI job
I don't think this is necessary because being one of the `codegen_targets` it's already executed in the CI. We could add it but I don't think this would make a difference.

> maybe mention it in the contribution section in the docs.
Definitely we should. I'm not quite how to mention this in a legal way. The contributor license agreement of the CNCF seems like a massive overkill to me...

hack/copyright.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
juanluisvaladas: I used bash because I find the error handling much better and way more reliable, set -eu doesn't work equally in bash and sh -o pipefail is not really supported in dash. Also, and this is an excuse, I like having the `if [[  ]]`.

The CI already uses bash and plenty of CNCF projects, like kubernetes require bash... I'm not quite sure on this day and age the POSIX shell is really much more portable to be honest. If we REALLY care about portability I'll change it but quite frankly most people has bash nowadays and I certainly see some benefits in error handling.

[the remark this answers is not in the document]
juanluisvaladas: Changed

.go-header.txt
juanluisvaladas: @twz123 I had to remove the escaped dots because it broke the lint. I'm not sure why though.

pkg/cleanup/bridge_windows.go
juanluisvaladas: OK I got different results by runnign GOOS=windows make lint.
Turns out there are a couple more issues with lint and windows. I'll fix this issue in a different PR and make sure that we run the lint for windows, linux and maybe darwin

.go-header.txt
juanluisvaladas: Managed to get it fixed, by adding the regex to .golangci.yml now we only treat year as a regex.

juanluisvaladas:
We should probably get rid of the year of auto generated files.

This allows us to get this out of the codegen_targets, keeps us in line with the statement from the legal team and it's what kubernetes is doing:
https://github.com/kubernetes/gengo/issues/24#issuecomment-355695594

.golangci.yml
```diff
@@ -26,6 +26,9 @@ linters-settings:
     min-confidence: 0
   goheader:
     template-path: .go-header.txt
+    values:
+      regexp:
+        year: 202[0-9]
```
mikhail-sakhnov, line 31: <irony> doesn't look like a future proof </irony>

mikhail-sakhnov: Ouch, it lost the irony marker :)

mikhail-sakhnov (changes requested):
Team wdyt if we instead of having this PR which is kinda hard to follow (different go files have different changes, e.g. in some files the year is dropped, in some files the year is changed to be 2021, in some to be 2022, etc) we split the PR into 2 PRs: one that introduces all the machinery with GHA settings and Makefile and tooling around, and later, once we agreed on the format, the second PR with actual changes?

twz123:
Works for me. Juan?

PR #2720, opened 2023-02-10
Detangle autopilot waitfor code

twz123:
## Description

The code in waitfor.go was mainly used within the integration tests. This is "okay", but it actually pulled in the retry logic which was dedicated to the integration tests, which is - at least - unexpected.

Move the waitfor code into the inttest folder and inline the parts that are actually required by autopilot production code, accounting for a proper "production" retry logic, performing some refactorings on the way:

* Check the "Established" condition when waiting for CRDs. This is the proper way of ensuring that a CRD is actually usable.
* Remove some generic helpers and the condition varargs parameters, as they were no longer required after detangling everything and a bit of copy/paste is okay.
* Remove the timeout parameters from the inttest functions. There's already the inttest context in place which effectively interrupts the test execution when the test times out.
* Remove unnecessary return values after inlining the generic code.
* Add the autopilot API group automatically, so this isn't duplicated at all the call sites.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


static/misc/api-resources.yaml
```diff
@@ -226,7 +226,9 @@
   kind: FlowControl
   removed_in: v1.26.0
   version: v1beta1
-- group: k0s.k0sproject.io
-  kind: Test
-  removed_in: v1.42.0
+
+# This is used solely for integration testing.
```
mikhail-sakhnov, line 230: should not be that file autogenerated? 
if so, where does the comment come from?
if no, feel free to resolve.

twz123: > should not be that file autogenerated?

I'm not aware of any auto-generation. I think this needs to be maintained manually when upgrading Kubernetes versions.

> where does the comment come from?

I added the comment after I figured out that this entry is an artificial one just for the ap-removedapis integration test. Would be great if we could "mock" that list somehow for integration testing, but that would have been a bigger refactoring.

jnummelin (approved):

PR #2721, opened 2023-02-10
Integrate Troubleshoot support bundle collection into inttests

twz123:
## Description

See https://troubleshoot.sh/. This replaces `kubectl cluster-info --dump`. An advantage of the support bundle is that it can be used in conjunction with `sbctl` to spin up a fake API server that reflects the cluster's state at the time of the support bundle creation. This makes it much easier to inspect than having to deal with all the individual JSON files using `find`/`grep`/`jq`. See https://github.com/replicatedhq/sbctl for details on that.

Moreover: Combine all upload CI steps into a single one. No need to have them separately. Only collect the logs if the tests failed. The CI step is also just triggered on failure.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -274,19 +274,14 @@ jobs:
           }
           make -C inttest ${{ matrix.smoke-suite }}
 
-      - name: Collect test logs
+      - name: Collect k0s logs and support bundle
         if: failure()
         uses: actions/upload-artifact@v3
         with:
-          name: smoketest-${{ matrix.smoke-suite }}-logs
-          path: /tmp/*.log
-
-      - name: Collect cluster state
-        if: failure()
-        uses: actions/upload-artifact@v3
-        with:
-          name: smoketest-${{ matrix.smoke-suite }}-cluster-state
-          path: /tmp/cluster-state.tar
+          name: smoketest-${{ matrix.smoke-suite }}-files
+          path: |
+            /tmp/*.log
+            /tmp/support-bundle.tar.gz
 
       - name: Collect sonobuoy results
         if: failure() && contains(matrix.smoke-suite, 'conformance')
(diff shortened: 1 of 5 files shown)


juanluisvaladas:
The code looks good to me but I don't think it should be executed within the cleanup deadline.

If you check [the failed run](https://github.com/k0sproject/k0s/actions/runs/4162504458/jobs/7201918032) you can see it wasn't acquired:

```
Run actions/upload-artifact@v3
  with:
    name: autopilot-smoketest-check-ap-ha3x3-v1.24.4+k0s.0-files
    path: /tmp/*.log
  /tmp/support-bundle.tar.gz
  
    if-no-files-found: warn
  env:
    GO_VERSION: 1.20
    GOLANGCI_LINT_VERSION: 1.51.1
    PYTHON_VERSION: 3.10.9
Warning: No files were found with the provided path: /tmp/*.log
```

I think it would make sense to have one deadline to collect the support bundle (how much is the question) and once this is done having another cleanup deadline so that we don't leave garbage behind.

The rest looks good to me.

twz123:
The problem you mention is a general one that's inherent to the way the FootlooseSuite works that can't be solved that easily and is not directly related to the cleanup context.

Whenever there are "FailNows" in the parts of the inttests that are executed within the `SetupSuite` phase, the suite won't execute `TeardownSuite`. That means something will exit the Go runtime without waiting for the cleanup goroutine to finish.

> I think it would make sense to have one deadline to collect the support bundle (how much is the question) and once this is done having another cleanup deadline so that we don't leave garbage behind.

Whenever the test deadline is reached, the test context gets renewed with a new deadline for the cleanup tasks. We can discuss if we want to remove the deadline from the cleanup context completely. That would probably add another 20 secs of time for the completion of the cleanup tasks. Remember that the deadlines are deduced from the Go test deadline (the `-timeout` parameter to the test binary). If that's exceeded, the test will terminate with a panic no matter what we try to do. But still that wouldn't fix the problem you highlighted.

juanluisvaladas (approved):

juanluisvaladas:
Fair enough, I hadn't considered the go test timeout itself, so let's merge this anyway

twz123:
See #2729.

PR #2496, opened 2022-12-09
PoC: variables in documentation

mikhail-sakhnov:
Quick example of the mkdocs.yaml on how to use cutsomly defined versions to change version only in one place

PR #2724, opened 2023-02-13
Change versions to be a placeholders in the documentation

mikhail-sakhnov:
## Description
Replace k8s version and k0s version in the documentation site with placeholder to easify post-release updates.

Follow up for the #2496 2496


makhov (approved):

mkdocs.yml
```diff
@@ -116,15 +120,17 @@ markdown_extensions:
...
+      toc_depth: 3
   - pymdownx.superfences:
       custom_fences:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
 
 extra:
+  k8s_version: 1.26.1
+  k0s_version: 1.26.1+k0s.0
```
jnummelin, line 133: Hmm, isn't this bit more "dynamic" than this?

This is essentially a tag we're building right? Except for the main branch

docs/autopilot.md
```diff
@@ -96,10 +96,10 @@ spec:
 
   commands:
     - k0supdate:
-        version: v1.26.1+k0s.0
+        version: v{{{ extra.k0s_version }}}
         platforms:
           linux-amd64:
-            url: https://github.com/k0sproject/k0s/releases/download/v1.26.1+k0s.0/k0s-v1.26.1+k0s.0-amd64
+            url: https://github.com/k0sproject/k0s/releases/download/v{{{ extra.k0s_version }}}/k0s-v{{{ extra.k0s_version }}}-amd64
```
twz123, line 102: I think this should be hardcoded, so that the sha256 matches.

docs/docker-shim.md
```diff
@@ -161,6 +161,6 @@ You should now see the node Ready for scheduling with the docker Runtime:
 sudo k0s kubectl get nodes -o wide
 
 NAME                                        STATUS   ROLES           AGE    VERSION       INTERNAL-IP   EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION    CONTAINER-RUNTIME
-ip-10-0-49-188.eu-west-1.compute.internal   Ready    control-plane   119m   v1.26.1+k0s   10.0.49.188   <none>        Ubuntu 20.04.4 LTS   5.13.0-1022-aws   docker://20.10.16
-ip-10-0-62-250.eu-west-1.compute.internal   Ready    <none>          66m    v1.26.1+k0s   10.0.62.250   <none>        Ubuntu 20.04.4 LTS   5.13.0-1017-aws   docker://20.10.16
+ip-10-0-49-188.eu-west-1.compute.internal   Ready    control-plane   119m   v{{{ extra.k8s_version }}}+k0s   10.0.49.188   <none>        Ubuntu 20.04.4 LTS   5.13.0-1022-aws   docker://20.10.16
```
twz123, line 164: This will break the layout if we reach point releases >9.

docs/k0s-multi-node.md
```diff
@@ -24,13 +24,13 @@ The download script accepts the following environment variables:
 
 | Variable                    | Purpose                                           |
 |:----------------------------|:--------------------------------------------------|
-| `K0S_VERSION=v1.26.1+k0s.0` | Select the version of k0s to be installed         |
+| `K0S_VERSION=v{{{ extra.k0s_version }}}` | Select the version of k0s to be installed         |
```
twz123, line 27: The Markdown table should be adjusted.

docs/requirements.txt
```diff
@@ -1,30 +1,30 @@
-# Direct dependencies
-mkdocs==1.4.2
```
twz123, line 2: Those changes seem unrelated to the PR.

docs/docker-shim.md
mikhail-sakhnov: not sure I follow why.

docs/requirements.txt
mikhail-sakhnov: Hm, imo they are related. 

My reasoning behind:

- to have version in the documentation I use macro addon
- I install a new dependency (that implies I need to update the requirements.txt) to have that addon
- I created a new requirements.txt using `pip freeze` from my local python virtual environment to get a snapshot of dependency tree to avoid manually changing the requirements.txt

I decided to go the easier way and updated requirements.txt in one run, but if those changes are really unrelated I can split it into 2 or 3 PRs, like one pr with new dependency, one pr with new formatting through `pip freeze` (because it kinda makes sense to maintain files automatically instead of manually building them, saves from making mistakes), one PR with actual changes in mkdocs.toml and may be even one more for the changes in markdown files, but the question is, do we really want that level of granularity of actions?

twz123: Fair enough. I didn't realize from the diff that there was a new plugin. That was the missing link. :+1:

docs/docker-shim.md
twz123: The lines will be misaligned with their headers. Not a big deal, but it will probably look a little bit wonky in the docs and those console snippets will look like "edited after the fact".

mkdocs.yml
mikhail-sakhnov: changed it to actually get the version from makefile variables

docs/docker-shim.md
mikhail-sakhnov: Yeah, in fact, the raw markdown files look strange now, with placeholder instead of the actual version, not sure how bad is it (e.g. how many people read website and how many read markdown directly).

docs/k0s-multi-node.md
mikhail-sakhnov: I change the formatting but in fact the GH renderer doesn't care about it (I guess any markdown renderer).

twz123: Yeah, was just a cosmetic thing for the markdown file itself.

jnummelin (approved):

PR #2729, opened 2023-02-14
Ensure cleanup of inttests also during the setup phase

twz123:
## Description

When FootlooseSuites fail during the setup phase, testify won't invoke the TearDownSuite function, leading to missed cleanup tasks.

Fix that by registering a test cleanup func that actually waits until the cleanup tasks have completed. Refactor the signal handling so that it may also be used to interrupt the cleanup phase. Use the new CancelCause mechanism to ease the context cancelation error management.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -104,7 +104,9 @@ type FootlooseSuite struct {
 	ControllerNetworks           []string
 	WorkerNetworks               []string
 
-	sctx suiteCtx
+	ctx      context.Context
+	tearDown func()
+
 	/* footloose cluster setup */
 
 	clusterDir     string
@@ -113,11 +115,6 @@ type FootlooseSuite struct {
 	launchDelegate launchDelegate
 
 	dataDirOpt string // Data directory option of first controller, required to fetch the cluster state
-}
-
-type suiteCtx struct {
-	ctx  context.Context
-	stop func()
 }
 
 // initializeDefaults initializes any unset configuration knobs to their defaults.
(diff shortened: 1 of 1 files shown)


inttest/common/footloosesuite.go
```diff
@@ -168,10 +169,7 @@ func (s *FootlooseSuite) SetupSuite() {
 
 	ctx, cancel := newSuiteContext(t)
 	var cleanupTasks sync.WaitGroup
-	sctx := suiteCtx{ctx, func() {
-		cancel()
-		cleanupTasks.Wait()
-	}}
+	sctx := suiteCtx{ctx, cancel, cleanupTasks.Wait}
 
 	if !s.ctxPtr.CompareAndSwap(nil, &sctx) {
```
mikhail-sakhnov, line 174: side note, but why do we use cas here? I am not sure I understand but when do we run SetupSuite concurrently for the same instance?

mikhail-sakhnov:
How does it work with the `K0S_KEEP_AFTER_TESTS=always`?

inttest/common/footloosesuite.go
twz123: We could also use `Store` here, since this is currently the only place where the context gets replaced. It's more a defensive measure in case some other code paths get added in the future that'll also write to ctxPtr.

We need to sync the ctxPtr variable since it might be accessed concurrently from multiple goroutines via `s.Context()`.

twz123:
> How does it work with the `K0S_KEEP_AFTER_TESTS=always`?

That part is handled in `cleanupSuite`.

mikhail-sakhnov (dismissed):

twz123:
Will rebase this on top of #2730 when it's merged.

mikhail-sakhnov (approved):

PR #2730, opened 2023-02-15
Pass context explicitly to the SSH helper function in the footloose test suite

mikhail-sakhnov:
Pass context explicitly to the SSH helper function in the footloose test suite

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/inttest/common/footloosesuite.go
+++ b/inttest/common/footloosesuite.go
@@ -214,7 +214,7 @@ func (s *FootlooseSuite) SetupSuite() {
 			t.Log("Failed to replace suite context during cleanup")
 		}
 
-		s.cleanupSuite(t)
+		s.cleanupSuite(t, ctx)
 	}()
 
 	// set up signal handler so we teardown on Interrupt or SIGTERM
@@ -227,7 +227,7 @@ func (s *FootlooseSuite) SetupSuite() {
 		os.Exit(1)
 	}()
 
-	s.waitForSSH()
+	s.waitForSSH(context.Background())
 
 	if s.WithLB {
 		s.startHAProxy()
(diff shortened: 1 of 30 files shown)


inttest/common/footloosesuite.go
```diff
@@ -227,7 +210,7 @@ func (s *FootlooseSuite) SetupSuite() {
 		os.Exit(1)
 	}()
 
-	s.waitForSSH()
+	s.waitForSSH(context.Background())
```
twz123, line 213: What about this, to match the previous behavior?

```suggestion
	s.waitForSSH(ctx)
```

inttest/common/footloosesuite.go
```diff
@@ -204,17 +196,8 @@ func (s *FootlooseSuite) SetupSuite() {
 		// Replace the done context with a fresh one.
 		ctx, cancel := newSuiteContext(t)
```
twz123, line 197: We could probably simply use 

```suggestion
		ctx, cancel := context.WithCancel(context.Background())
```

here.

The main difference is that if the test really times out because of the `-timeout` parameter, we'd see a panic instead of "nice" error messages. The `newSuiteContext` is reserving some extra time for the cleanup, and will timeout quite some time before the Go test framework will terminate the test. That might be too early for the cluster state collection.

inttest/common/footloosesuite.go
```diff
@@ -105,10 +104,7 @@ type FootlooseSuite struct {
 	ControllerNetworks           []string
 	WorkerNetworks               []string
 
-	/* context and cancellation */
-
-	ctxPtr atomic.Pointer[suiteCtx]
-
+	sctx suiteCtx
```
twz123, line 107: Now that the atomic isn't required anymore, we could potentially simply inline the two variables from the suiteCtx struct here.

twz123 (approved):

twz123 (approved):

PR #2743, opened 2023-02-16
Check helm version consistency

twz123:
## Description

There's the new `helm_version` Makefile variable. Add a unit test that ensures this is in sync with the helm versions referenced in `go.mod`.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/constant/constant_shared_test.go
+++ b/pkg/constant/constant_shared_test.go
@@ -127,6 +127,22 @@ func TestRuncModuleVersions(t *testing.T) {
 	)
 }
 
+func TestHelmModuleVersions(t *testing.T) {
+	helmVersion := getVersion(t, "helm")
+
+	checkPackageModules(t,
+		func(modulePath string) bool {
+			return modulePath == "helm.sh/helm/v3"
+		},
+		func(t *testing.T, pkgPath string, module *packages.Module) bool {
+			return !assert.Equal(t, "v"+helmVersion, module.Version,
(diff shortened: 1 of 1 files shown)


pkg/constant/constant_shared_test.go
```diff
@@ -127,6 +127,22 @@ func TestRuncModuleVersions(t *testing.T) {
 	)
 }
 
+func TestHelmModuleVersions(t *testing.T) {
```
mikhail-sakhnov, line 130: There is no any reason to have this test. The helm binary we have is not related to the k0s binary, it is just for the use in one of the e2e tests, has 0 connection with the module path.

mikhail-sakhnov (changes requested):
I suggest to drop the pr.

pkg/constant/constant_shared_test.go
twz123: It's true that there's no direct connection between the binary used in the ITs and the Go module version. On the other hand, it might be confusing to have the versions diverge. Folks might look at the Makefiles and might make false assumptions about the version that's being used in k0s itself, or the helm_version variable will be picked up in the future by something that really has an impact on k0s.

OTOH, I'd be fine to remove the variable and this test and inline the version directly into the Dockerfile, as we do it e.g. for cri-dockerd.

PR #2756, opened 2023-02-20
Canonical var names for docker IPv6 detection

makhov:
## Description

Small fix of non-canonical var names


mikhail-sakhnov (approved):
I wonder how did it pass linter before, do we have it still?

twz123:
> I wonder how did it pass linter before, do we have it still?

We're currently not linting the inttests. See https://github.com/k0sproject/k0s/pull/2711#pullrequestreview-1297948580.

PR #2787, opened 2023-02-21
Fix external address reconciliation w.r.t. dynamic config

### LINKED ISSUE #2753: Dynamic configuration breaks installation on standalone worker

### Platform

```shell
Linux 5.15.85-1-pve #1 SMP PVE 5.15.85-1 (2023-02-01T00:00Z) x86_64 GNU/Linux
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### Version

v1.26.0+k0s.0

### Sysinfo

```text
Machine ID: "fc70219997392e455981de2719c1c8aa08ed65329c510ff5f6411b5362bff479" (from machine) (pass)
Total memory: 188.6 GiB (pass)
Disk space available for /var/lib/k0s: 7.2 GiB (pass)
  Linux kernel release: 5.15.85-1-pve (pass)
  Max. file descriptors per process: current: 524288 / max: 524288 (pass)
  Executable in path: modprobe: /usr/sbin/modprobe (pass)
    cgroup controller "cpu": available (pass)
    cgroup controller "cpuset": available (pass)
    cgroup controller "memory": available (pass)
    cgroup controller "devices": available (assumed) (pass)
    cgroup controller "freezer": available (assumed) (pass)
    cgroup controller "pids": available (pass)
    cgroup controller "hugetlb": available (pass)
    CONFIG_CGROUP_FREEZER: Freezer cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_PIDS: PIDs cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_DEVICE: Device controller for cgroups: built-in (pass)
    CONFIG_CPUSETS: Cpuset support: built-in (pass)
    CONFIG_CGROUP_CPUACCT: Simple CPU accounting cgroup subsystem: built-in (pass)
    CONFIG_MEMCG: Memory Resource Controller for Control Groups: built-in (pass)
    CONFIG_CGROUP_HUGETLB: HugeTLB Resource Controller for Control Groups: built-in (pass)
      CONFIG_NF_NAT: module (pass)
      CONFIG_NF_CONNTRACK_IPV4: IPv4 connetion tracking support (required for NAT): unknown (warning)
      CONFIG_NF_CONNTRACK_IPV6: IPv6 connetion tracking support (required for NAT): unknown (warning)
    CONFIG_BRIDGE: 802.1d Ethernet Bridging: built-in (pass)
      CONFIG_LLC: built-in (pass)
      CONFIG_STP: built-in (pass)
```
</details>

### What happened?

I'm noticing two distinct issues with configuring k0s. Although I encountered the second one while working around the first one.

#### Main issue with dynamic configuration

The primary issue that I am facing is that I cannot install k0s using k0sctl on dedicated workers when `dynamicConfig` is enabled in my `k0sctl.yaml` file. I'm using a four node setup as follows:
1. controller
2. controller+worker
3. controller+worker
4. worker

The controller and worker services come up fine on nodes 1-3, but node 4 never comes online. Using the debug flag, I've found what looks to me like the main issue. The worker service is attempting to read a config map called `worker-config-default-1.26`, which I understand to be the old name for what is now `kubelet-config-default-1.26`. With dynamic configuration turned off, this error either does not occur or does not prevent the worker from joining the cluster.

#### Secondary issue with k0sctl config

Since dynamic configuration was not working, I've tried making changes such as enabling OpenEBS in my `k0sctl.yaml` file then re-rerunning `k0sctl apply`, but the output looks like the new config has been applied to each controller, but the changes are not reflected in cluster state.

#### Final workaround

I was able to get the dynamic configuration working by manually adding `--enable-dynamic-config` to the already installed controller services, reload systemd, then restart all the controller units. Dynamic configuration (both via `k0sctl` on my machine and `k0s` on a controller node) then worked, but did not have my initial `k0sctl` config so I had to put that back, specifically my `externalAddress` and `sans` values to get the HA control plane to work. Now the dynamic configuration seems to be working as expected. I still cannot make changes by re-applying my `k0sctl.yaml` file.

### Steps to reproduce

1. Create a new `k0sctl` config with at least one standalone worker and `dynamicConfig` set to true. To more closely emulate my setup also use ah HA control plane with one standalone controller and two nodes with both a controller and ` worker.` Optionally add more debug flags. The HA control plane requires the addition of the `externalAddress` and `sans` config options.
2. Run `k0sctl apply` with the configuration.
3. Watch the logs on the standalone worker, it should just hang indefinitely. With the debug flag it shows a permissions error about the missing `worker-config-default-1.26` error, which I assume is due to it not existing at all. I've seen that error sometimes without the debug flag, but not consistently.
4. Let `k0sctl apply` timeout or SIGINT it.
5. Reset the nodes through `k0sctl reset` or some other means.
6. Disable `dynamicConfig` in the `k0sctl` configuration file.
7. Run `k0sctl apply` again with the new configuration.
8. Observe it install correctly.
9. Make a change to the `k0sctl` configuration file that will have easily observable side effects. I enabled OpenEBS personally via `spec.extensions.storage`.
10. Re-run `k0sctl apply`.
11. Look for the side effect of the new configuration, it should not be there. For me, I saw that `kubectl get sc` did not return the local volume provisioner.
12. SSH into each control node and add `--enable-dynamic-config` to the end of the `ExecStart` line in `/etc/systemd/system/k0scontroller.service`. The run `systemctl daemon-reload` and `systemctl restart k0scontroller`.
13. Run `k0sctl config edit`. Note that the configuration is missing any manually set values from the initial/edited `k0sctl` config.
14. Edit the dynamic configuration and observe the changes take effect. Especially important if you are using an HA control plane as controllers will start crashing without the load balancer address.

### Expected behavior

1. Installing with `dynamicConfig: true` should work.
2. Running `k0sctl apply` with an update config file should apply the changes to the cluster.

### Actual behavior

1. Installing with `dynamicConfig: true` breaks standalone workers joining the cluster.
2. Running `k0sctl apply` with an update config file does not appear to change cluster state.

### Screenshots and logs

I hope I did a decent job describing the errors I saw, most of the logs were noise from konnectivity and such. I can dig up specific messages if that would help.

### Additional context

This cluster is running inside of LXCs in Proxmox VE. The controller-only node is in an unprivileged LXC while the worker nodes are in privileged LXCs with specific kubes workarounds (creating `kmsg`, disabling app armor, cgroups, pushing the kernel config, enabling nesting). I ran a similar setup with a non-HA control plane for a few weeks with no issues once I got the initial install working.

twz123:
## Description

The external address is not part of the reconcilable cluster config, but part of the non-reconcilable node config. The external API endpoint reconciler was looking at the cluster config. This is fine as long as dynamic config isn't enabled, because the config splitting doesn't happen then. When dynamic config is enabled, the splitting takes place and the reconciler tries to reconcile an empty external address, resulting in log statements like these:

> level=error msg="cannot resolve api.externalAddress: lookup : no such host" component=endpointreconciler
> level=warning msg="external API address reconciliation failed: lookup : no such host" component=endpointreconciler

Fix this by removing the cluster config reconciliation from the reconciler and use the node config's API address instead.

Execute the customports inttest also with dynamic config enabled: This checks if the above fix works. It could have been its own test, but putting it into customports saves quite some duplication.

Fixes #2753.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


jnummelin (dismissed):

pkg/component/controller/apiendpointreconciler.go
```diff
@@ -146,17 +145,18 @@ func (a *APIEndpointReconciler) reconcileEndpoints(ctx context.Context) error {
...
 		}}
 
 		_, err := epClient.Update(ctx, ep, metav1.UpdateOptions{})
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to update Endpoints: %w", err)
 		}
 	}
 
+	a.logger.Debugf("Successfully updated Endpoints resource using %v", ipStrings)
```
twz123, line 159: This needs do go a few lines up I guess...

jnummelin (dismissed):

pkg/component/controller/apiendpointreconciler.go
```diff
@@ -37,26 +38,28 @@ import (
...
 	leaderElector     leaderelector.Interface
-	stopCh            chan struct{}
 	kubeClientFactory kubeutil.ClientFactoryInterface
+
+	stopCh chan struct{}
 }
 
 // NewEndpointReconciler creates new endpoint reconciler
-func NewEndpointReconciler(leaderElector leaderelector.Interface, kubeClientFactory kubeutil.ClientFactoryInterface) *APIEndpointReconciler {
+func NewEndpointReconciler(nodeConfig *v1beta1.ClusterConfig, leaderElector leaderelector.Interface, kubeClientFactory kubeutil.ClientFactoryInterface) *APIEndpointReconciler {
```
mikhail-sakhnov, line 55: I'd suggest renaming it is not a reconciler anymore since there is no Reconcile method or interface.

Plus I think, more or less the same logic is in the TunneledEndpointReconciler which is still not dropped, good to check if it is required to use node config.

twz123: > I'd suggest renaming it is not a reconciler anymore since there is no Reconcile method or interface.

It doesn't take dynamic configuration into account, but it still has the ten-second reconciliation loop in which it syncs any DNS changes to the endpoints object.

> Plus I think, more or less the same logic is in the TunneledEndpointReconciler which is still not dropped, good to check if it is required to use node config.

Indeed. The TunneledEndpointReconciler was looking at the dynamic config, too. Removed that part.

mikhail-sakhnov (approved):

PR #2834, opened 2023-03-02
Fix docker build-context creation in release

juanluisvaladas:
## Description
If docker tries to create a context that already exists it will fails. Now we check if it exists and if it doesn't we create the context.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -356,7 +356,7 @@ jobs:
         shell: bash
         id: buildx-context
         run: |
-          docker context create buildx-context
+          docker context inspect buildx-context -f ' ' || docker context create buildx-context
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2


makhov (approved):

mikhail-sakhnov:
How did it work before? 

Why do we even have a context? Should not we run each job in the clean state?

juanluisvaladas:
> How did it work before?
It was failing intermittently https://github.com/k0sproject/k0s/actions/runs/4313115964/jobs/7524993044

> Why do we even have a context? 
It's required https://github.com/k0sproject/k0s/pull/2598

> Should not we run each job in the clean state?
I think so, I'm not quite sure why this happens.

twz123:
> > Should not we run each job in the clean state?
> 
> I think so, I'm not quite sure why this happens. Anyway I don't think it's very relevant because we are doing everything in containers.

The release workflow runs on self-hosted runners. They're not ephemeral and are being reused for subsequent workflow runs. That's why the context remains on the build machines.

PR #2848, opened 2023-03-03
Remove unneeded steps from lint CI workflow

twz123:
## Description

The build environment is never used in the workflow, and the Golang steps are built inside Docker.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -38,14 +38,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Prepare build environment
-        run: .github/workflows/prepare-build-env.sh
-
-      - name: Set up Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
       - name: Go caches
         uses: actions/cache@v3
         with:


mikhail-sakhnov:
is it actually work without build? the codegen part is used to be done as a part of the build step and some linters fail without generated code.

twz123:
> is it actually work without build? the codegen part is used to be done as a part of the build step and some linters fail without generated code.

The codegen part is actually built (w/o embedded binaries). But that happens from within Docker, so no need to have the Go toolchain be installed on the GH runner.

.github/workflows/lint.yaml
```diff
@@ -38,14 +38,6 @@ jobs:
...
-        run: .github/workflows/prepare-build-env.sh
-
-      - name: Set up Go
-        uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
       - name: Go caches
         uses: actions/cache@v3
         with:
```
kke, line 43: I guess it doesn't use the cache then either, at least not like that. There's some kind of docker layer caching thing for github runners.

twz123: The cache should be used. It will be mounted into the container.

kke: Right

mikhail-sakhnov (approved):

PR #2850, opened 2023-03-06
TLS server configuration hardening

twz123:
## Description

Enforce TLS minimum versions and a fixed list of allowed TLS ciphers for all the TLS secured endpoints managed by k0s. This has already been the case for the API server, the kubelet and the konnectivity server. Unify the approach to also include etcd and the k0s API itself.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/etcd.go
+++ b/pkg/component/controller/etcd.go
@@ -163,15 +163,20 @@ func (e *Etcd) Start(ctx context.Context) error {
 		"--listen-peer-urls":            peerURL,
 		"--initial-advertise-peer-urls": peerURL,
 		"--name":                        name,
-		"--trusted-ca-file":             etcdCaCert,
-		"--cert-file":                   etcdServerCert,
-		"--key-file":                    etcdServerKey,
-		"--peer-trusted-ca-file":        etcdCaCert,
-		"--peer-key-file":               etcdPeerKey,
-		"--peer-cert-file":              etcdPeerCert,
-		"--log-level":                   e.LogLevel,
-		"--peer-client-cert-auth":       "true",
-		"--enable-pprof":                "false",
+		// Specifying a minimum TLS version is not yet possible in etcd,
+		// although support for it has already been merged upstream. Enable this
+		// flag once it's available in the etcd release that ships with k0s.
+		// https://github.com/etcd-io/etcd/pull/15156
+		// "--tls-min-version": "TLS1.2",
+		"--trusted-ca-file":       etcdCaCert,
+		"--cert-file":             etcdServerCert,
+		"--key-file":              etcdServerKey,
+		"--peer-trusted-ca-file":  etcdCaCert,
+		"--peer-key-file":         etcdPeerKey,
+		"--peer-cert-file":        etcdPeerCert,
+		"--log-level":             e.LogLevel,
+		"--peer-client-cert-auth": "true",
+		"--enable-pprof":          "false",
 	}
 
 	if file.Exists(filepath.Join(e.K0sVars.EtcdDataDir, "member", "snap", "db")) {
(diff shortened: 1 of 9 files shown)


pkg/component/controller/kubeletconfig.go
```diff
@@ -236,22 +237,18 @@ func getDefaultProfile(dnsAddress string, clusterDomain string) unstructuredYaml
 	// - it's easier to merge programatically defined structure
 	// - apart from map[string]interface there is no good way to define free-form mapping
 
+	cipherSuites := make([]string, len(constant.AllowedTLS12CipherSuiteIDs))
+	for i, cipherSuite := range constant.AllowedTLS12CipherSuiteIDs {
+		cipherSuites[i] = tls.CipherSuiteName(cipherSuite)
+	}
+
```
jnummelin, line 244: Maybe this could be a helper func in the `constant` package as it is used in couple different places now

pkg/component/controller/etcd.go
```diff
@@ -203,6 +204,11 @@ func (e *Etcd) Start(ctx context.Context) error {
 		args[argName] = value
 	}
 
+	// https://github.com/etcd-io/etcd/pull/15156/files#diff-538c79cd00ec18cb43b5dddd5f36b979d9d050cf478a241304493284739d31bfR810-R813
+	if args["--cipher-suites"] == "" && args["--tls-min-version"] != "TLS1.3" {
```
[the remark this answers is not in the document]
kke: That would probably also entail having some kind of `AllowedTLSCipherSuiteNames(tlsVersion)`

pkg/component/controller/kubeletconfig.go
twz123: It's used in two places, one being deprecated. That's why didn't do it. Technically, the changes in this file aren't necessary at all, since it only exists for backwards compatibility when upgrading from k0s 1.25.

pkg/component/controller/etcd.go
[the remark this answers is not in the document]
twz123: I've more or less copied the logic from etcd, which is referenced by the link above. The referenced etcd code path is not yet released, so the check for TLS 1.3 is kind of a future-proof thing, as etcd 3.5.7 doesn't support the tls-min-version flag yet.

pkg/component/controller/etcd.go
```diff
@@ -163,15 +163,16 @@ func (e *Etcd) Start(ctx context.Context) error {
...
-		"--trusted-ca-file":             etcdCaCert,
-		"--cert-file":                   etcdServerCert,
-		"--key-file":                    etcdServerKey,
-		"--peer-trusted-ca-file":        etcdCaCert,
-		"--peer-key-file":               etcdPeerKey,
-		"--peer-cert-file":              etcdPeerCert,
-		"--log-level":                   e.LogLevel,
-		"--peer-client-cert-auth":       "true",
-		"--enable-pprof":                "false",
+		// "--tls-min-version": "TLS1.2", // https://github.com/etcd-io/etcd/pull/15156
```
mikhail-sakhnov, line 166: leftover

mikhail-sakhnov (changes requested):

pkg/component/controller/etcd.go
twz123: I've put it here as a reminder for the future, when etcd actually supports `--tls-min-version`. This is not yet available as of etcd v3.5.7, but already merged to main. I can remove it if you prefer.

mikhail-sakhnov: Let's add a text comment with clarification for the reason, if you want, but now it looks like leftover and would be accidentally removed on the first edit of this file by somebody who doesn't have a context.

How I read this line now:

- there was an argument tls-min-version
- the given value was TLS1.2
- the reason to have this value is explained in the https://github.com/etcd-io/etcd/pull/15156
- somebody commented the whole thing out due to some reason

Sorry for nitpicking here, but I always afraid of out-dated or misguiding comments, because they always make reading code harder :(

twz123: Fair enough. Will make the doc string clearer.

twz123: Done.

mikhail-sakhnov (approved):

PR #2851, opened 2023-03-06
Unify clientset generation

twz123:
## Description

* Add doc.go to APIs
  So that the k8s code-generator can parse the tags correctly.

* Remove k0sproject.io from API package folder names and generate one clientset for all APIs
  This makes paths and imports more concise and resembles the canonical way of naming API folders and structuring clientsets in other projects that use k8s code generation.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


pkg/apis/autopilot/doc.go
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package autopilot contains API Schema definitions for the
+// autopilot.k0sproject.io API group.
+package autopilot
+
+const GroupName = "autopilot.k0sproject.io"
```
mikhail-sakhnov, line 19: I wonder why linter is not failing with a public constant without a comment. Is it not required for doc.go?

twz123: I don't think that there's a lint for that. There are quite a bunch of public constants without docs, e.g. in the constant package.

kke: [Revive](https://github.com/mgechev/revive) is enabled, it should nag about these. 🤔

twz123: What's the [rule name](https://revive.run/docs#available-rules)? I only see a rule about package comments.

kke: Ah, EXC0011 is in default exclusions list and should be enabled separately for that to happen.

twz123:
See #2909.

pkg/apis/autopilot/doc.go
juanluisvaladas: Actually I think there is indeed a problem with revive. If I run golangci-lint in my machine I get some extra errors.

I filed an issue upstream in golangci-lint, I also need to investigate this more locally to see if there is a difference.
https://github.com/golangci/golangci-lint/issues/3764

juanluisvaladas (dismissed):

twz123:
This constantly hits #3038.

juanluisvaladas (approved):
Approving again, this time without further review because there were no merging conflicts

PR #2886, opened 2023-03-15
Fix k0s status components

juanluisvaladas:
## Description

If a component.Healthy returns an error which is not nil the json would be marshalled and unmarshalled incorrectly. Since the error is only for humans we're storing it as a string.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/prober/prober.go
+++ b/pkg/component/prober/prober.go
@@ -19,6 +19,8 @@ package prober
 import (
 	"container/ring"
 	"context"
+	"encoding/json"
+	"errors"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mikhail-sakhnov (approved):

pkg/component/prober/prober.go
```diff
@@ -161,10 +161,15 @@ func (p *Prober) checkComponentsHealth(ctx context.Context, at time.Time) {
 			p.healthCheckState[name] = ring.New(p.probesTrackLength)
 		}
 		// TODO: add back-off logic
+		err := ""
```
kke, line 164: Having something other than an error in `err` is a bit confusing.

pkg/component/prober/prober.go
```diff
@@ -215,5 +220,5 @@ func (p *Prober) Register(name string, component any) {
 type ProbeResult struct {
 	Component string    `json:"component"`
 	At        time.Time `json:"at"`
-	Error     error     `json:"error"`
+	Error     string    `json:"error"`
```
kke, line 223: Maybe it needs omitempty if it isn't error interface.

pkg/component/prober/prober.go
```diff
@@ -161,10 +161,15 @@ func (p *Prober) checkComponentsHealth(ctx context.Context, at time.Time) {
 			p.healthCheckState[name] = ring.New(p.probesTrackLength)
 		}
 		// TODO: add back-off logic
+		err := ""
+		healthy := component.Healthy()
```
kke, line 165: This should be the the `err` :)

kke:
Could also be a custom MarshalJSON on ProbeResult:

```go
func (pr *ProbeResult) MarshalJSON() ([]byte, error) {
  type probeResult ProbeResult
  
  var errStr *string
  if pr.Error != nil {
    errStr = &pr.Error.Error()
  }
  
  return json.Marshal(&struct {
    *probeResult
    Error *string `json:"error"`
  }{
    probeResult: (*probeResult)(pr),
    Error:      errStr,
  })
}
```

Is this the only `Healthy()` implementation or am I not looking hard enough? 🤔 

```go
func (k *Konnectivity) Healthy() error {
	if k.clusterConfig == nil {
		return fmt.Errorf("cluster config not yet available")
	}

	return nil
}
```

pkg/component/prober/prober.go
juanluisvaladas: true

juanluisvaladas:
> Could also be a custom MarshalJSON on ProbeResult:
That's probably the sanest option we'll need both custom marshal and unmarshal. I'll do that in the plane.

> Is this the only `Healthy()` implementation or am I not looking hard enough? 🤔

It is at the moment. I have it implemented in my local branch for etcd as well but when there is an error it doesn't print it.

juanluisvaladas:
Disregard this change, yesterday it seemed cleaner than doing the custom marshalling and unmarshalling but it's not

kke (approved):

PR #2919, opened 2023-03-24
Fix runtime config permissions from 0755 to 0600

kke:
## Description

The runtime config in `{rundir}/k0s.yaml` was being generated with rwx-rw-rw permissions. 

## Type of change

This changes it to 0600 (`-rw------`).

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/config/file_config.go
+++ b/pkg/config/file_config.go
@@ -164,7 +164,7 @@ func (rules *ClientConfigLoadingRules) writeConfig(yamlData []byte, storageSpec 
 		return fmt.Errorf("failed to marshal config: %v", err)
 	}
 
-	err = file.WriteContentAtomically(rules.RuntimeConfigPath, data, 0755)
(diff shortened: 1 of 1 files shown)


mikhail-sakhnov (approved):

twz123:
> As a sidenote, I'm not convinced that the runtime config file should exist in the first place, I can't think of any use for it. If k0s is running, it already has the config in memory and another instance with the same config shouldn't be possible/make any sense and something like k0s status should use the socket. The file is deleted when k0s exits, so it's not there for restart either.

It has some use in case if dynamic configuration is enabled and controllers are restarted. But yeah, the location is less than ideal, especially given that it should survive restarts and the rundir is explicitly not guaranteed to be persistent across restarts. Should rather be somewhere in k0s's data dir IMO.

PR #2944, opened 2023-03-30
Include inttest in make lint

### LINKED ISSUE #2940: make lint does not include inttest

### Issue

I  believe we can all agree without much discussion that it's a good idea to lint inttest and that it could help preventing some bugs in the tests themselves.

I checked the linting and I don't think anyone would argue about most stuff but, but there are 2 lint errors I think will need some discussion:

```
inttest/common/footloosesuite.go:323:53: context-as-argument: context.Context should be the first parameter of a function (revive)
func (s *FootlooseSuite) cleanupSuite(t *testing.T, ctx context.Context) {
                                                    ^
inttest/common/footloosesuite.go:401:53: context-as-argument: context.Context should be the first parameter of a function (revive)
func (s *FootlooseSuite) dumpNodeLogs(t *testing.T, ctx context.Context, node, dir string) {
```

If we want linting there are two options here:
1- We make an exception and remove the argument `context-as-argument` for inttest
2- We change the function so that context.Context is the first argument,

The convention is [**MOST** functions that use a Context **SHOULD** accept it as their first parameter](https://github.com/golang/go/wiki/CodeReviewComments#contextsl) (emphasis is mine). 
This makes sense to me to some extent but at the same time I think testing.T should go before the Context. Besides the convention says most and should so I understand it's a bit open to interpretation.

Anyway I don't feel very strongly about either option.

Of course we can keep without linting the tests but honestly I think it would be a mistake.

juanluisvaladas:
## Description

This commit changes the Makefile so that inttest is linted and fixes all the linting errors.

Fixes #2940

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ endif
 K0S_GO_BUILD_CACHE ?= build/cache
 
 GO_SRCS := $(shell find . -type f -name '*.go' -not -path './$(K0S_GO_BUILD_CACHE)/*' -not -path './inttest/*' -not -name '*_test.go' -not -name 'zz_generated*')
-GO_DIRS := . ./cmd/... ./pkg/... ./internal/... ./static/... ./hack/...
+GO_CHECK_UNIT_DIRS := . ./cmd/... ./pkg/... ./internal/... ./static/... ./hack/...
 
 # EMBEDDED_BINS_BUILDMODE can be either:
 #   docker	builds the binaries in docker
@@ -210,12 +210,18 @@ codegen: $(codegen_targets)
 .PHONY: bindata
 bindata: static/zz_generated_assets.go pkg/assets/zz_generated_offsets_$(TARGET_OS).go
 
-.PHONY: lint
-lint: GOLANGCI_LINT_FLAGS ?=
-lint: .k0sbuild.docker-image.k0s go.sum codegen
+.PHONY: lint-copyright
+lint-copyright:
 	hack/copyright.sh
+
+.PHONY: lint-go
+lint-go: GOLANGCI_LINT_FLAGS ?=
+lint-go: .k0sbuild.docker-image.k0s go.sum codegen
 	CGO_ENABLED=0 $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v$(golangci-lint_version)
-	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_DIRS)
+	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_LINT_DIRS)
+
+.PHONY: lint
+lint: lint-copyright lint-go
 
 airgap-images.txt: k0s .k0sbuild.docker-image.k0s 
 	$(GO_ENV) ./k0s airgap list-images --all > '$@' || { \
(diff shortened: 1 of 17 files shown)


Makefile
```diff
@@ -215,7 +216,7 @@ lint: GOLANGCI_LINT_FLAGS ?=
 lint: .k0sbuild.docker-image.k0s go.sum codegen
 	hack/copyright.sh
 	CGO_ENABLED=0 $(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v$(golangci-lint_version)
-	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_DIRS)
+	$(GO_ENV) golangci-lint run --verbose $(GOLANGCI_LINT_FLAGS) $(GO_LINT_DIRS)
```
twz123, line 219: I wonder if we should move all of this into .golangci.yml. This has the benefit of being consistent with IDE integrations then.

inttest/cnichange/cnichange_test.go
```diff
@@ -44,11 +44,12 @@ func (s *CNIChangeSuite) TestK0sGetsUpButRejectsToChangeCNI() {
 	s.Require().NoError(err)
 
 	// Wait till we see kube-router DS created
-	watch.DaemonSets(kc.AppsV1().DaemonSets("kube-system")).
+	err = watch.DaemonSets(kc.AppsV1().DaemonSets("kube-system")).
 		WithObjectName("kube-router").
 		Until(s.Context(), func(ds *appsv1.DaemonSet) (bool, error) {
 			return true, nil
 		})
+	s.NoError(err)
```
twz123, line 52: I wonder if we wanna use Require() here and in the other places.

inttest/common/footloosesuite.go
```diff
@@ -201,7 +200,7 @@ func (s *FootlooseSuite) SetupSuite() {
 		// Get a fresh context for the cleanup tasks.
 		ctx, cancel := signalAwareCtx(context.Background())
 		defer cancel(nil)
-		s.cleanupSuite(t, ctx)
```
twz123, line 204: I'd prefer to keep the `t` parameter everywhere.

https://github.com/k0sproject/k0s/issues/2940#issuecomment-1490076989

inttest/cnichange/cnichange_test.go
mikhail-sakhnov: I wonder if we can somehow change the embedding to use Require by default, not the Assert

juanluisvaladas: I used the assert one in the errors which were ignored because we weren't failing on error, but actually it makes sense to fail early so I'm changing all of these.

inttest/common/footloosesuite.go
juanluisvaladas: I didn't know the context could change, I changed the order in the parameter.

Makefile
juanluisvaladas: I added it to .golangci.yml and kept the variable but without defining it. This way we use .golangci.yml but we still can override the directories.

Also I split make lint into lint-copyright and lint-go, but kept the target lint. I was tired of linting copyright every time

Issue #2879, opened 2023-03-10
Bump Kubernetes to v1.27


twz123:
Tracking issue, things to be done:

* [x] Ship quay.io/k0sproject/kube-proxy:v1.27.0
  * k0sproject/image-builder#25
  * k0sproject/image-builder#27
  * k0sproject/image-builder#29
* [x] Have a [Helm] version that's compatible with the Kubernetes 1.27 client libraries
  * Opened the draft PR helm/helm#11894. Seems to be working so far.
  * Using https://github.com/k0sproject/helm/releases/tag/v3.11.3%2Bk0s.0
* [x] Have a [controller-runtime] version that's compatible with the Kubernetes 1.27 client libraries
  * Currently using master, which has kubernetes-sigs/controller-runtime#2223 merged.
* [x] Check for any API version removals
  * Any removals should be added to `static/misc/api-resources.yaml`.
  * https://github.com/k0sproject/k0s/pull/2999
* [x] Check out how to properly build static Kubernetes binaries without breaking CGO  
  The proper way of building static Kubernetes binaries seems to be using the `KUBE_STATIC_OVERRIDES` environment variable. To my current understanding, this disables CGO altogether. This might backfire for FIPS builds.
  #2910
* [x] Adjust `nodeutil.GetHostname` imports
* [x] Migrate from `--container-runtime-endpoint` kubelet flag to `KubeletConfiguration.ContainerRuntimeEndpoint` kubelet configuration field  
* [x] Properly check for required key usages when approving node certificates  
  The "key encipherment" usage is only required when using RSA keys, which the kubelet doesn't by default. Kubernetes 1.27 will no longer request that usage for node certificates if not using RSA keys.
  * #3001
* [x] Get the panic on ARMv7 fixed
  * https://github.com/k0sproject/k0s/pull/2880#issuecomment-1463842878
  * kubernetes/kubernetes#116477
* [x] Have a closer look at kubernetes/kubernetes#108838 and if k0s needs to do something about that.
  * #2916
* [x] Check out [KEP-3638: Improve kubectl plugin resolution for non-shadowing subcommands](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/3638-kubectl-plugin-subcommands/README.md) and what it means for `k0s kc`. @kke
  * #2994 

[Helm]: https://github.com/helm/helm/
[controller-runtime]: https://github.com/kubernetes-sigs/controller-runtime/

PR #2994, opened 2023-04-13
Restructure kubectl plugin execution

twz123:
## Description

This PR will enable better integration with kubectl plugins and will prepare k0s for [KEP-3638](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/3638-kubectl-plugin-subcommands/README.md). See #2879.

K0s used the `NewDefaultKubectlCommandWithArgs` function to obtain a kubelet command to be embedded into k0s. That function merely executes its plugin logic unconditionally, i.e. it won't wait until that command is being executed by Cobra. This lead to a situation in which kubectl plugins not only shadowed kubectl commands, but also k0s commands. An inttest was added to prove that, before this change, an executable named `kubectl-airgap` would effectively replace the builtin `k0s airgap` command.

Address this by using a two-pass approach when executing the embedded kubectl command. First, during the Cobra bootstrap phase, call `NewKubectlCommand` to create a Cobra command without triggering the plugin execution logic. Then patch the command's callbacks, so that when the command is _actually_ invoked by Cobra, the `NewDefaultKubectlCommandWithArgs` function is called. That function will execute plugins and never return, if any. Otherwise it will just return a new Cobra command, which k0s will simply discard and use the previously created one.

Also don't print kubectl usage on unknown subcommands. The basic kubectl command will never accept any arguments. This will be handled more or less automatically by Cobra when used as a root command. When run as a subcommand, Cobra will instead pass all of the arguments to the command's run methods, which will trigger some unexpected output. Address this by specifying `cobra.NoArgs` for the kubectl command.

Reorder persistent flag registration for `k0s kc`, so that they are grouped together. Also add `InitLogs`/`FlushLogs`, which is done by k8s.io/component-base/cli, which k0s doesn't use.

Streamline kubectl integration test: The common test matrix wasn't checking for stderr contents so far. Address this by capturing both stdout and stderr when executing the test commands and forward those along with the execution error to the check function. Add asserts for stderr contents to all the tests in the matrix. Moreover, the integration test environment wasn't completely isolated. For example, the support-bundle plugin that's installed in the footloose image was leaking into the test setup. Address this by setting up a private PATH and use that when invoking the test commands. That way, the plugin-list and symlink-warning test cases can be integrated into the common test matrix as well.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/cmd/kubectl/kubectl.go
+++ b/cmd/kubectl/kubectl.go
@@ -89,6 +89,13 @@ func NewK0sKubectlCmd() *cobra.Command {
 	cmd := kubectl.NewDefaultKubectlCommandWithArgs(args)
 	cmd.Aliases = []string{"kc"}
 
+	// The basic kubectl command will never accept any arguments. This will be
+	// handled more or less automatically by Cobra when used as a root command.
+	// When run as a subcommand, Cobra will instead pass all of the arguments to
+	// the command's run methods, which will trigger some unexpected output.
+	// Address this by specifying NoArgs for the kubectl command.
+	cmd.Args = cobra.NoArgs
+
 	// Get handle on the original kubectl prerun so we can call it later
 	originalPreRunE := cmd.PersistentPreRunE
 	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
--- a/cmd/kubectl/kubectl.go
+++ b/cmd/kubectl/kubectl.go
@@ -22,6 +22,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/k0sproject/k0s/pkg/config"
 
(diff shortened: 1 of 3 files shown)


cmd/kubectl/kubectl.go
```diff
@@ -64,65 +67,105 @@ func (h *kubectlPluginHandler) Execute(executablePath string, cmdArgs, environme
...
+		// unexpected output. Address this by specifying NoArgs for the kubectl
+		// command.
+		if cmd == kubectlCmd {
+			if err := cobra.NoArgs(cmd, args); err != nil {
+				return err
+			}
+		}
+
+		// In vanilla kubectl, log initialization and flushing is handled by
+		// k8s.io/component-base/cli. But k0s doesn't use it, so it needs to
```
mikhail-sakhnov, line 123: Side note: I was totally sure we use component-base/cli, I wonder where does this false memory come from

mikhail-sakhnov (approved):

cmd/kubectl/kubectl.go
twz123: I digged a bit in the GitHub history when I wrote that comment and found a superseeded PR of yours. That might be the reason:

* #1408

PR #3072, opened 2023-05-05
Refactor configuration loading

kke:
## Description

Refactor configuration loading, eliminating the hard to follow `ClientConfigLoadingRules`.

The "runtime config" concept has been kept for now, as it is a smaller change than what was attempted before. 

Key changes:

- The runtime config has been converted into a "first class citizen" by making a dedicated struct for it. It now stores not only the config, but also the "k0svars". This allows commands like `token create` or `k0s status` to reuse all the parameters of the controller, such as `--data-dir` and `--status-socket` without the user having to supply them again. The runtime config also stores a pid, which is used to roughly check if the file belongs to an active process or if it's a leftover from a crash.
- The `k0svars`, aka `constant.CfgVars` has been moved to `config.CfgVars` as it is not a constant.
- The `k0sVars` now duplicates most of the `config.CLIOptions`, this is to reduce the number of arguments that need to be passed around, as the essentials are available from `k0svars`, which is passed around to almost everywhere already. This will help in the effort to eliminate the globals, which are now only used almost exclusively to build the CLIOptions.
- Usages of "k0svars" have been changed to use a pointer, since that struct holds a memoized nodeconfig. This removes the need to reload it everywhere.
- The `config.GetCmdOptions()` now takes the `*cobra.Command` as an argument and additionally returns an error. The cobra command is used to build the "k0svars". Returning an error is better than the surprise `os.Exit` in the old implementation did (which silently ended the test suite).
- Same with `c.NodeConfig`, which is now a function that returns `*v1beta1.ClusterConfig, error`.
- Fixes `k0s config validate --config`, even when using `-` for stdin. I think `k0s controller` and `install controller` may still be broken for `-`.
- Node- and dynamic config are loaded via `k0svars.NodeConfig()` and `k0svars.FetchDynamicConfig()`, which may be a bit counterintuitive and doesn't exactly follow the Single Responsibility Principle.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [X] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s/v1beta1/clusterconfig_types.go
+++ b/pkg/apis/k0s/v1beta1/clusterconfig_types.go
@@ -384,34 +384,6 @@ func (c *ClusterConfig) Validate() (errs []error) {
 	return errs
 }
 
-// GetBootstrappingConfig returns a ClusterConfig object stripped of Cluster-Wide Settings
-func (c *ClusterConfig) GetBootstrappingConfig(storageSpec *StorageSpec) *ClusterConfig {
-	var etcdConfig *EtcdConfig
-	if storageSpec.Type == EtcdStorageType {
-		etcdConfig = &EtcdConfig{
-			ExternalCluster: storageSpec.Etcd.ExternalCluster,
-			PeerAddress:     storageSpec.Etcd.PeerAddress,
-			ExtraArgs:       storageSpec.Etcd.ExtraArgs,
-		}
-		c.Spec.Storage.Etcd = etcdConfig
-	}
-	return &ClusterConfig{
-		ObjectMeta: c.ObjectMeta,
-		TypeMeta:   c.TypeMeta,
-		Spec: &ClusterSpec{
-			API:     c.Spec.API,
-			Storage: storageSpec,
-			Network: &Network{
-				ServiceCIDR:   c.Spec.Network.ServiceCIDR,
-				DualStack:     c.Spec.Network.DualStack,
-				ClusterDomain: c.Spec.Network.ClusterDomain,
-			},
-			Install: c.Spec.Install,
-		},
-		Status: c.Status,
-	}
-}
-
 // HACK: the current ClusterConfig struct holds both bootstrapping config & cluster-wide config
 // this hack strips away the node-specific bootstrapping config so that we write a "clean" config to the CR
 // This function accepts a standard ClusterConfig and returns the same config minus the node specific info:


twz123:
Just a general thought: Would it be feasible to break this down into more than one PR? The list of key changes reads quite nicely and it would be really helpful if there'd be dedicated commits for those. Depending on the amount of changes, those commits could then be bundled up into smaller PRs that are easier/safer to review.

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+	var apiConfig *v1beta1.ClusterConfig
+
+	client, err := kubeClientFactory.GetConfigClient()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := retry.Do(
+		func() (err error) {
+			ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
twz123, line 271: I think the old timeout was ten seconds. I'd at least restore that. Maybe remove the timeout completely and rely on the client-go timeout settings instead?

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+			defer cancel()
+			c, err := client.Get(ctx, constant.ClusterConfigObjectName, metav1.GetOptions{})
+			if err == nil {
+				apiConfig = c.GetClusterWideConfig()
+			}
+			return err
```
twz123, line 277: Intuitively, I think it's more common to write it like this:

```suggestion
			if err != nil {
				return err
			}

			apiConfig = c.GetClusterWideConfig()
			return nil
```

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+	client, err := kubeClientFactory.GetConfigClient()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := retry.Do(
+		func() (err error) {
+			ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
+			defer cancel()
+			c, err := client.Get(ctx, constant.ClusterConfigObjectName, metav1.GetOptions{})
```
twz123, line 273: I'd add some debug logs here. The cluster config loading has already bitten me back in the day and it would be helpful to have some logs.

`logrus.Debug("Getting cluster config from API")`

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	var apiConfig *v1beta1.ClusterConfig
+
+	client, err := kubeClientFactory.GetConfigClient()
+	if err != nil {
+		return nil, err
+	}
+
+	if err := retry.Do(
```
twz123, line 269: Helpful to have some logs on retry by adding a retry option:

```go
retry.OnRetry(func(attempt uint, err error) {
	logrus.WithError(err).Debugf("Failed to get cluster config from API in attempt #%d, retrying after backoff", attempt+1)
}),
```

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+	if nodeConfig.Spec.Storage.Type == v1beta1.KineStorageType && nodeConfig.Spec.Storage.Kine == nil {
+		nodeConfig.Spec.Storage.Kine = v1beta1.DefaultKineConfig(c.DataDir)
+	}
+
+	c.nodeConfig = nodeConfig
+
+	return nodeConfig, nil
+}
+
+func (c *CfgVars) ClusterConfig(ctx context.Context, kubeClientFactory kubeutil.ClientFactoryInterface) (*v1beta1.ClusterConfig, error) {
```
twz123, line 257: Since we're on it right now: What about renaming this function to `LoadClusterConfig`?

twz123 (commented):
Did a first pass on this. Just loads of nits so far. I've left some "notes to self" that I left for me to figure out when engaging with the code more thoroughly. Feel free to ignore them :stuck_out_tongue_closed_eyes:

cmd/backup/backup.go
```diff
@@ -42,17 +42,25 @@ func NewBackupCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "backup",
 		Short: "Back-Up k0s configuration. Must be run as root (or with sudo)",
+		PreRun: func(cmd *cobra.Command, args []string) {
+			// ensure logs don't mess up output
+			logrus.SetOutput(cmd.ErrOrStderr())
```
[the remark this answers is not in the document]
kke: We kind of don't. Logrus defaults to os.Stderr. Those additions are mostly safeguards because I was feeling cautious that the config loading stuff might output something new and that could mess up redirects.

cmd/airgap/listimages.go
```diff
@@ -33,13 +37,26 @@ func NewAirgapListImagesCmd() *cobra.Command {
...
 			}
-			uris := airgap.GetImageURIs(clusterConfig.Spec, all)
-			for _, uri := range uris {
+
+			adminClientFactory := kubernetes.NewAdminClientFactory(opts.K0sVars.AdminKubeConfigPath)
+
+			ctx, cancel := context.WithTimeout(cmd.Context(), 2*time.Minute)
+			defer cancel()
+
+			clusterConfig, err := opts.K0sVars.ClusterConfig(ctx, adminClientFactory)
```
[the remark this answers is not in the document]
kke: We could have a `RuntimeConfig` singleton

pkg/config/cfgvars.go
kke: The only things calling `ClusterConfig()` are 1) controller startup 2) airgap list-images. 

I think the controller could live without it. Components that require the "dynamic config" could probably just wait for config reconciliation before starting up. It will fail every time when there are no other controllers in the cluster. 

I don't know if even 10 seconds is enough for list-images if you run `sudo k0s start` and `k0s airgap list-images` right after it.  Maybe it should instead wait for status socket before proceeding.

pkg/config/cfgvars.go
kke: I'd call it `FetchClusterConfig` or similar in that case as I'd like it to be contextually as far as possible from `loading` which is usually done to files.

cmd/api/api.go
```diff
@@ -217,6 +225,11 @@ users:
 			sendError(fmt.Errorf("no calico-node-token secret found"), resp)
 			return
 		}
+		nodeConfig, err := c.K0sVars.NodeConfig()
```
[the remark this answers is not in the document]
kke: (Note to you: no, it's another function, `kubeConfigHandler()`. I think it should be called with `nodeConfig.Spec.API.APIAddressURL()` as argument as it's the only thing it uses from the config)

cmd/backup/backup.go
```diff
@@ -42,17 +42,25 @@ func NewBackupCmd() *cobra.Command {
...
+				return err
+			}
+			if nodeConfig.Spec.Storage.Etcd.IsExternalClusterUsed() {
 				return fmt.Errorf("command 'k0s backup' does not support external etcd cluster")
 			}
 			return c.backup(savePath, cmd.OutOrStdout())
 		},
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			c := command(config.GetCmdOpts())
-			return config.PreRunValidateConfig(c.K0sVars)
```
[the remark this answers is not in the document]
kke: To controller startup.

cmd/config/validate.go
```diff
@@ -29,11 +35,38 @@ func NewValidateCmd() *cobra.Command {
 		Long: `Example:
    k0s config validate --config path_to_config.yaml`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			c := config.GetCmdOpts()
+			var reader io.Reader
+
+			cfgFile, err := cmd.Flags().GetString("config")
+			if err != nil {
+				return fmt.Errorf("%w: --config is required", err)
```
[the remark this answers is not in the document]
kke: The error message is wrong anyway, the error only comes if there is no flag called `config` defined.

cmd/controller/controller.go
```diff
@@ -141,19 +157,35 @@ func (c *command) start(ctx context.Context) error {
...
+		if err := rtc.Cleanup(); err != nil {
+			logrus.Warnf("failed to cleanup runtime config: %v", err)
+		}
+	}()
+
+	// common factory to get the admin kube client that's needed in many components
+	adminClientFactory := kubernetes.NewAdminClientFactory(c.K0sVars.AdminKubeConfigPath)
+
+	if c.K0sVars.EnableDynamicConfig {
+		configCtx, configCancel := context.WithTimeout(ctx, 5*time.Second)
```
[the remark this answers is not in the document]
kke: The parent ctx waits forever.

cmd/etcd/list.go
```diff
@@ -31,10 +31,21 @@ func etcdListCmd() *cobra.Command {
...
-			etcdClient, err := etcd.NewClient(c.K0sVars.CertRootDir, c.K0sVars.EtcdCertDir, c.NodeConfig.Spec.Storage.Etcd)
+			opts, err := config.GetCmdOpts(cmd)
+			if err != nil {
+				return err
+			}
+			nodeConfig, err := opts.K0sVars.NodeConfig()
+			if err != nil {
+				return err
+			}
+			ctx := cmd.Context()
```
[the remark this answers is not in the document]
kke: Using the cmd.Context in all commands as the root context would allow us to add a global `--timeout` flag, the root persistent-prerun would do `ctx, cancel := context.WithTimeout(cmd.Ctx, time.Parse(&timeOut)); cmd.SetContext(ctx)` 🤔

cmd/restore/restore.go
```diff
@@ -39,21 +39,25 @@ import (
...
-			if len(args) != 1 {
-				return fmt.Errorf("path to backup archive expected")
+			opts, err := config.GetCmdOpts(cmd)
+			if err != nil {
+				return err
 			}
+
+			c := command{opts, restoredConfigPath}
+
 			return c.restore(args[0], cmd.OutOrStdout())
```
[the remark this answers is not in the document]
kke: There seems to be:

```go
	if c.restoredConfigPath == "" {
		c.restoredConfigPath = defaultConfigFileOutputPath(path)
	}
```

cmd/status/status.go
```diff
@@ -41,19 +37,23 @@ func NewStatusCmd() *cobra.Command {
...
-			statusInfo, err := status.GetStatusInfo(config.StatusSocket)
+			statusInfo, err := status.GetStatusInfo(opts.K0sVars.StatusSocketPath)
 			if err != nil {
 				return err
 			}
 			if statusInfo != nil {
 				printStatus(cmd.OutOrStdout(), statusInfo, output)
 			} else {
-				fmt.Fprintln(cmd.OutOrStdout(), "K0s is not running")
+				return config.ErrK0sNotRunning
```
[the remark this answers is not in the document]
kke: I think there are some other bits that go as far as to do something like `if statusInfo.Pid == 0 { return fmt.Errorf("k0s not running") }`

cmd/worker/worker.go
```diff
@@ -201,10 +203,10 @@ func (c *Command) Start(ctx context.Context) error {
 				Workloads:     true,
 				SingleNode:    false,
 				K0sVars:       c.K0sVars,
-				ClusterConfig: clusterConfig,
+				ClusterConfig: nil,
```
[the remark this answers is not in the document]
kke: I'm not sure if it's necessary to have the config in `k0s status -o json` on a worker. I think it used to work on main, here it is just nil. I set the nil explicitly just to be clear about it :)

pkg/config/cfgvars.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+	}
+
+	if dataDir == "" {
+		switch runtime.GOOS {
+		case "windows":
+			dataDir = constant.WinDataDirDefault
+		default:
+			dataDir = constant.DataDirDefault
+		}
```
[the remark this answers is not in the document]
twz123: #3078

internal/testutil/runtime_config.go
```diff
@@ -57,15 +57,12 @@ func NewConfigGetter(t *testing.T, yamlData string, isNodeConfig bool, k0sVars c
...
+	require.NoError(c.t, err, "failed to create fake runtime config")
+	defer require.NoError(c.t, os.Remove(cfgPath))
 
-	cfg, err := loadingRules.Load()
-	require.NoError(c.t, err, "failed to load fake config from file")
-	return cfg
+	return rtc.NodeConfig
 }
 
 func (c *ConfigGetter) FakeAPIConfig() *v1beta1.ClusterConfig {
```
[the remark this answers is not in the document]
twz123: #3079

kke:
> Just a general thought: Would it be feasible to break this down into more than one PR? The list of key changes reads quite nicely and it would be really helpful if there'd be dedicated commits for those. Depending on the amount of changes, those commits could then be bundled up into smaller PRs that are easier/safer to review.

Without some intelligent tool to splice it up, it would seem like a big effort to identify and extract closely related changes into separate commits when the files include multiple change sets.

pkg/config/cfgvars.go
kke: Renamed to `FetchDynamicConfig`

pkg/config/cfgvars.go
kke: Does list-images need to work with dynamic config?

kke: The config business in controller node/clustercomponents is complicated.

### node components:

- all of the nodeComponents that need clusterConfig receive it during construct, ie. &foo.Foo{Config: config} or foo.NewFoo(config)
- none of the nodeComponents have Reconcile, they will never receive a config in any other way after that

### cluster components:

- some of the clusterComponents receive a nodeconfig during constructor, some don't.
- many of them have a Reconcile function, so they will receive a new config when:
  * ClusterConfigReconcilier first starts
  * When ClusterConfigReconcilier detects a dynamic config change
  
 The startup process is somewhat exotic:
  
  1. All of the components are `Add()`ed to cluster component manager
  2. Finally component manager's `Start()` is called
  3. Start iterates over all of the `Add()`ed components and calls `Start()` for each and then waits for `Ready()` if the component has defined that function.
  4. When the iteration encounters the `ClusterConfigReconcilier` component somewhere middle of the way, it calls the `Start()` on it like for any other component, which makes ClusterConfigReconciler call component manager's `Reconcile(ctx, cfg)`
  5. Component manager's `Reconcile` iterates over all of the `Add`ed components (we're still in the middle of the iterator in step 3) and calls `Reconcile(ctx, cfg)` on them if they have defined this function.
  6. At this point, all of the components that have a Reconcile function will receive a new config (local or dynamic) 
  7. The iteration in step 3 continues to `Start()` the rest of the components
  
  So,

- the components before step 4 are started with either the local config or no config
- the components after step 4 receive a new config via `Reconcile()` (if they implement it), before they are started

  This needs to be cleaned up somehow.

kke: Some additional issues:

#### The component manager:

- It has an `Init()` function that is called once for both of the managers in the controller.
- It also has the `Add()` function. The function tries to call `Reconcile` on the components in case the manager has a config available. It never has unless there's something calling the manager's `Reconcile`,  which only happens once it has been started.
- Any component added after the `Init()` call will not be initialized when `Add()`ed.
- The `Start()` function iterates over all of the components to call `Start` on them and then returns. This happens once.
- Nothing starts or initializes any components that are added after that.
- Nothing restarts components that fail
- If a component fails, nothing kills the manager
- the "reconcile" term is used rather sloppily for multiple things and it's slightly difficult to figure out what it means for each of the components. I think the dynamic config reconciliation related functions should be named uniquely.

#### Regarding the components:

- There are multiple components that have just `return nil` in their `Init()` and `Start()` and the actual functionality is in `Reconcile()`.
- Some components have functionality in their `Start` but they are one-shot, the component "starts" but nothing is actually left running.
- There are several components that are added to the `clusterComponents` manager that do not have a `Reconcile` function, so they could just as well be in the `nodeComponents` manager that doesn't reconcile.
- The `pkg/component/controller` package defines 48 structs and 5 interfaces. The "node components" and "cluster components" are all mixed up in there in a single directory without any separation.
- Many of them have `// Run ....` doc comment on top of the `Start()` function, a remnant from some function rename I suppose

#### The prober:

- The only components that implement the prober healthcheck interface are: Konnectivity and `mockComponent`.
- The only components that implement the prober event emitter interface are: Konnectivity, OCIBundleReconcilier (which is one-shot, `Start` just unpacks stuff and returns) and `mockComponentWithEvents`.

#### The cluster config reconciler:

- Nothing restarts the watcher if the server or a watchtimeout closes it
- It does not differentiate between add, create, update, delete
- It does not check for nils. If it receives a *v1beta1.ClusterConfig typed nil, it will call `Validate()` on that, which will return `nil` and then it proceeds to reconcile using the nil.
- Nothing tells you if the component has actually stopped, not even the experimental `k0s status components` subcommand. It doesn't implement event emitter or health check interfaces. the only way you notice is that config reconciliation does not happen.

kke: The components are started and reconcilied sequentially, not sure if it's necessary.

cmd/ctr/ctr.go
```diff
@@ -32,16 +32,20 @@ const pathEnv = "PATH"
...
-		RunE: func(_ *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			opts, err := config.GetCmdOpts(cmd)
+			if err != nil {
+				return err
+			}
+			setDefaultValues(opts.K0sVars.RunDir, containerdCtr.Flags)
 			args := extractCtrCommand(os.Args)
-			newPath := fmt.Sprintf("%s%s%s", config.GetCmdOpts().K0sVars.BinDir,
+			newPath := fmt.Sprintf("%s%s%s", opts.K0sVars.BinDir,
```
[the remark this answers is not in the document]
kke: There's nothing in stdlib, https://github.com/golang/go/issues/25205

cmd/controller/controller.go
```diff
@@ -141,19 +157,35 @@ func (c *command) start(ctx context.Context) error {
...
+		defer configCancel()
 
-	// from now on, we only refer to the runtime config
-	c.CfgFile = loadingRules.RuntimeConfigPath
+		clusterConfig, err := c.K0sVars.FetchDynamicConfig(configCtx, adminClientFactory)
+		if err != nil {
+			logrus.WithError(err).Warn("Failed to load cluster config")
+		} else {
+			logrus.Info("Using cluster config from API")
+			nodeConfig = clusterConfig
```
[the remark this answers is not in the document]
kke: True

cmd/ctr/ctr.go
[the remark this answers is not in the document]
kke: Of course could do `strings.Join([]string{opts.K0sVars.BinDir, os.Getenv("PATH")}, string(os.PathListSeparator))` or something

cmd/controller/controller.go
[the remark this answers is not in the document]
kke: Removed the whole config fetching from controller startup

cmd/worker/worker.go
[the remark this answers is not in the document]
kke: The problem with having it there is that then the status component would need to be reconcilied, but there's no reconciliation on a worker or in the nodecomponents-manager on a controller where the status component is hosted, so this would be (has been) showing just the first dynamic config unless a bigger change is made. 

The function in main has been doing this:

```go
		clusterConfig, err := config.LoadClusterConfig(c.K0sVars)
		if err != nil {
			return fmt.Errorf("failed to load cluster config: %w", err)
		}
```
==>
```go
func LoadClusterConfig(k0sVars constant.CfgVars) (*v1beta1.ClusterConfig, error) {
	loadingRules := ClientConfigLoadingRules{K0sVars: k0sVars}
	return loadingRules.Load()
}
```
==>
```go
func (rules *ClientConfigLoadingRules) Load() (*v1beta1.ClusterConfig, error) {
	if rules.Nodeconfig {
		return rules.fetchNodeConfig()
	}
	if !rules.IsAPIConfig() {
		return rules.readRuntimeConfig()
	}
	if rules.IsAPIConfig() {
		nodeConfig, err := rules.BootstrapConfig()
		if err != nil {
			return nil, err
		}
		apiConfig, err := rules.ClusterConfig()
		if err != nil {
			return nil, err
		}
		// get node config from the config-file and cluster-wide settings from the API and return a combined result
		return rules.mergeNodeAndClusterconfig(nodeConfig, apiConfig)
	}
	return nil, nil
}
```


So, it seems the config has been nil on a worker all along.

For controller it works like it used to, it shows the config k0s was started with, never the dynamic config.

IMO changing this behavior can be done in a separate PR as it works here as it has worked all along.

pkg/config/cfgvars.go
kke: > The only things calling `ClusterConfig()` are 1) controller startup 2) airgap list-images.

Now only number 2.

Does "airgap list-images" need to work with dynamic config, @jnummelin ?

cmd/ctr/ctr.go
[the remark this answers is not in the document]
kke: Addressed in #3184

cmd/controller/controller.go
```diff
@@ -311,12 +329,7 @@ func (c *command) start(ctx context.Context) error {
 	if c.EnableDynamicConfig {
 		configSource, err = clusterconfig.NewAPIConfigSource(adminClientFactory)
 	} else {
-		var clusterConfig *v1beta1.ClusterConfig
-		clusterConfig, err = config.LoadClusterConfig(c.K0sVars)
-		if err != nil {
-			return fmt.Errorf("failed to load cluster config: %w", err)
-		}
-		configSource, err = clusterconfig.NewStaticSource(clusterConfig)
+		configSource, err = clusterconfig.NewStaticSource(nodeConfig)
```
jnummelin, line 332: hmm, I think we should use the full cluster config here

kke: NodeConfig is always a full cluster config when dynamic config is not enabled. It is only stripped down to BootstrapConfig when merged to a dynamic one.

Reconcile should only deal with ClusterWideConfig, any component that uses anything from the NodeConfig should  receive it as a static config in initialization, if it needs anything from the cluster wide config, it should be received from reconcile.

pkg/config/cfgvars.go
kke: Now airgap list-images will error out if dynamic is enabled. Now nothing uses FetchDynamicConfig() so it has been removed. Now configs are never merged.

cmd/controller/controller.go
kke: Now `GetBootstrappingConfig()` is never called so it too can be removed.

kke: GetClusterWideConfig() is only used upon creating the initial clusterconfig api object.

jnummelin (approved):

cmd/controller/controller.go
jnummelin: 👍

Issue #3069, opened 2023-05-04
Cgroup v2 freezer availability detection in RHEL 8 and derivatives


NeonSludge:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 4.18.0-425.19.2.el8_7.x86_64 #1 SMP Tue Apr 4 05:30:47 EDT 2023 x86_64 GNU/Linux
NAME="AlmaLinux"
VERSION="8.7 (Stone Smilodon)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="AlmaLinux 8.7 (Stone Smilodon)"
ANSI_COLOR="0;34"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:almalinux:almalinux:8::baseos"
HOME_URL="https://almalinux.org/"
DOCUMENTATION_URL="https://wiki.almalinux.org/"
BUG_REPORT_URL="https://bugs.almalinux.org/"

ALMALINUX_MANTISBT_PROJECT="AlmaLinux-8"
ALMALINUX_MANTISBT_PROJECT_VERSION="8.7"
REDHAT_SUPPORT_PRODUCT="AlmaLinux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.7"
```

### Version

`v1.27.1+k0s.0`

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "c4a752dd6f4b8e4016bdec0ed3ec76d816d864de901ce912f82fa493a5d38794" (from machine) (pass)
Total memory: 3.8 GiB (pass)
Disk space available for /var/lib/k0s: 34.5 GiB (pass)
Operating system: Linux (pass)
  Linux kernel release: 4.18.0-425.19.2.el8_7.x86_64 (pass)
  Max. file descriptors per process: current: 262144 / max: 262144 (pass)
  Executable in path: modprobe: /usr/sbin/modprobe (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (pass)
[...]

PR #1577, opened 2022-03-10
Add pre-flight checks and probes module

twz123:
## Description

Users can choose to ignore the results of the pre-flight checks by adding the `--ignore-pre-flight-checks` command line argument.

Re-implement the sysinfo subcommand without the system-validators library. Some reasons as to why this makes sense:

* The system-validators library has some weird error reporting, where callers cannot easily distinguish between validation failures and hard errors.
* The lib wasn't easily extensible with k0s-specific checks.
* It still relies on external binaries (uname) (although this is certainly fixable upstream).

Add a new internal package sysinfo/probes whose design loosely borrows from Go's testing framework and Testify's assert/require nomenclature. Those probes are being used to implement a "K0sSysSpec", which is the foundation for both the sysinfo subcommand and the pre-flight checks.

Supersedes #1526.
Fixes #1289.

The output of `k0s sysinfo` looks like this on my box now:

```text
Machine ID: "d02e3609003cf4b2d96bf16e7f1729c02fa3be9f8e0eab85d9ec38b0aa11528f" (from machine) (pass)
Total memory: 14.9 GiB (pass)
Disk space available for /var/lib/k0s: 279.7 GiB (pass)
Operating system: Linux (pass)
  Linux kernel release: 5.16.8 (pass)
  CONFIG_CGROUPS: Control Group support: built-in (pass)
    CONFIG_CGROUP_FREEZER: Freezer cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_PIDS: PIDs cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_DEVICE: Device controller for cgroups: built-in (pass)
    CONFIG_CPUSETS: Cpuset support: built-in (pass)
    CONFIG_CGROUP_CPUACCT: Simple CPU accounting cgroup subsystem: built-in (pass)
    CONFIG_MEMCG: Memory Resource Controller for Control Groups: built-in (pass)
    CONFIG_CGROUP_HUGETLB: HugeTLB Resource Controller for Control Groups: built-in (pass)
    CONFIG_CGROUP_SCHED: Group CPU scheduler: built-in (pass)
      CONFIG_FAIR_GROUP_SCHED: Group scheduling for SCHED_OTHER: built-in (pass)
        CONFIG_CFS_BANDWIDTH: CPU bandwidth provisioning for FAIR_GROUP_SCHED: built-in (pass)
    CONFIG_BLK_CGROUP: Block IO controller: built-in (pass)
[...]

PR #3077, opened 2023-05-05
Revisit devices and freezer controller detection for cgroup v2

twz123:
## Description

The cgroup v2 freezer and device controllers are not listed in the `cgroup.controllers` file, Therefore, k0s sysinfo determines their presence based on the Linux kernel version. This is problematic for old kernels that have many backported features, such as RHEL and consorts.

Introduce a hopefully more reliable way of detecting them.

* Use github.com/containerd/cgroups/v3 for cgroupv2 sysinfo. This is the same library that is used by containerd when it comes to cgroupv2 handling. k0s already links against it, and it exposes enough knobs in order to be the backend for the cgroupv2 sysinfo backend, so this seems a good choice to reduce the amount of low-level plumbing here.

* The freezer controller still has its `cgroup.freeze` file, albeit this is deliberately not available in the root cgroup. Therefore, k0s sysinfo now tries to detect the controller file in its own cgroup. In the unlikely case that k0s is running in the root cgroup, it will try to create an empty, temporary cgroup for the freezer file detection. If it doesn't have sufficient permissions to do so, it issues an "unknown" warning.

* The device controller solely available via the kernel's BPF interface. Therefore, k0s sysinfo now tries to attach a dummy device filter to an empty, temporary cgroup. In case k0s is unable to create the cgroup or attach the device filter because of missing permissions, the sysinfo probe will issue an "unknown" warning.

Fixes:
* #3069

See:
* #1577

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


internal/pkg/sysinfo/probes/linux/cgroup_controllers.go
```diff
@@ -70,12 +73,16 @@ func (c *cgroupControllerProbe) Probe(reporter probes.Reporter) error {
 }
 
 type cgroupControllerAvailable struct {
-	available bool
-	msg       string
+	available    bool
+	msg, warning string
```
kke, line 77: 😲  Didn't know that's possible. I'd prefer two separate lines.

jnummelin (approved):
LGTM

PR #3174, opened 2023-05-31
Windows containerd compilation

mikhail-sakhnov:
## Description

Add containerd.exe and containerd related shims for the windows build to allow using k0s workers without external cri instance.
Partially fixes broken windows support.
## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


pkg/component/worker/containerd.go
```diff
@@ -59,88 +60,136 @@ imports = [
...
+		}
 	}
 
 	go c.watchDropinConfigs(ctx)
 
 	return nil
 }
 
+func (c *ContainerD) windowsStart(_ context.Context) error {
+	return winExecute("Start-Service containerd")
```
twz123, line 162: Maybe wrap the error here instead of at the call site at line 134? (Same for the `windowsStop` method.)

```suggestion
	if err := winExecute("Start-Service containerd"); err != nil {
		fmt.Errorf("failed to start Windows Service %q: %w", "containerd", err)
	}
	return nil
```

pkg/component/worker/containerd.go
```diff
@@ -59,88 +60,136 @@ imports = [
...
-
-	if err := dir.Init(filepath.Dir(confPath), 0755); err != nil {
-		return err
+	if err := dir.Init(filepath.Dir(c.confPath), 0755); err != nil {
+		return fmt.Errorf("can't create containerd config dir: %v", err)
 	}
-	if err := dir.Init(filepath.Dir(importsPath), 0755); err != nil {
-		return err
+	if err := dir.Init(filepath.Dir(c.importsPath), 0755); err != nil {
+		return fmt.Errorf("can't create containerd config imports dir: %v", err)
```
twz123, line 186:
```suggestion
		return fmt.Errorf("can't create containerd config imports dir: %w", err)
```

pkg/component/worker/containerd.go
```diff
@@ -59,88 +60,136 @@ imports = [
...
 	if !k0sManaged {
-		logrus.Infof("containerd config file %s is not k0s managed, skipping config generation", confPath)
+		logrus.Infof("containerd config file %s is not k0s managed, skipping config generation", c.confPath)
 		return nil
 	}
-
-	if err := dir.Init(filepath.Dir(confPath), 0755); err != nil {
-		return err
+	if err := dir.Init(filepath.Dir(c.confPath), 0755); err != nil {
+		return fmt.Errorf("can't create containerd config dir: %v", err)
```
twz123, line 183:
```suggestion
		return fmt.Errorf("can't create containerd config dir: %w", err)
```

pkg/component/worker/containerd.go
```diff
@@ -59,88 +60,136 @@ imports = [
...
+	if err := dir.Init(filepath.Dir(c.importsPath), 0755); err != nil {
+		return fmt.Errorf("can't create containerd config imports dir: %v", err)
 	}
-	containerDConfigurer := containerd.NewConfigurer()
+	containerDConfigurer := containerd.NewConfigurer(filepath.Join(c.importsPath, "*.toml"))
 
 	imports, err := containerDConfigurer.HandleImports()
 	if err != nil {
-		return err
+		return fmt.Errorf("can't handle imports: %v", err)
```
twz123, line 192:
```suggestion
		return fmt.Errorf("can't handle imports: %w", err)
```

pkg/component/worker/utils_linux.go
```diff
@@ -0,0 +1,21 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package worker
+
+func winExecute(args ...string) error {
```
twz123, line 19: Wouldn't it be sufficient to not define this function at all? Defining it like this makes it a runtime error rather than a compile time error.

pkg/component/worker/utils_windows.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// PowerShell struct
+type PowerShell struct {
+	powerShell string
+}
+
+// New create new session
+func NewPowershell() *PowerShell {
+	ps, _ := exec.LookPath("powershell.exe")
```
twz123, line 35: Would be nice to add error handling here. Also we should take a note on PowerShell in the external runtime dependencies section of the docs. Windows still simply states "TBD" :-)

pkg/component/worker/utils_windows.go
```diff
@@ -0,0 +1,73 @@
...
+	}, retry.Delay(time.Second*5))
+	if err != nil {
+		return "", err
+	}
+	return vip, nil
+}
+
+func winExecute(args ...string) error {
+	ps := NewPowershell()
+	return ps.execute(args...)
```
twz123, line 72: Looks like we could simply inline all the `PowerShell` struct's methods here and completely remove it?

pkg/component/worker/utils_linux.go
twz123: Ah, now I understand. It's guarded via `if runtime.GOOS` on the call sites instead of using conditional compilation. I'd suggest to either use runtime guards or conditional compilation, but not both. So either capture the call sites via conditional compilation or remove this file and add an `if runtime.GOOS == "windows" { panic(...) }` at the top of the real `winExecute` definition.

mikhail-sakhnov: Why not to use both, if it solves an issue? Plus it makes code easier

embedded-bins/containerd/Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV GOPATH=/go
 RUN apk upgrade -U -a && apk add \
 	build-base git \
 	btrfs-progs-dev btrfs-progs-static \
-	protoc
+	protoc bash
```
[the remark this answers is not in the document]
mikhail-sakhnov: yeah, for some build pipeline of the containerd

twz123 (approved):

twz123 (approved):

makhov (approved):

PR #3192, opened 2023-06-06
Improve github workflow embedded binaries caching

kke:
## Description

Looks to me like the workflow embedded-bins caching can be slightly improved.

### cache key

The cache key was:

```yaml
          key: ${{ runner.os }}-embedded-bins-${{ matrix.target }}-${{ hashFiles('embedded-bins/**/*') }}
```

To me looks like that would include `embedded-bins/staging/` which is what we actually want to restore instead of rebuild. I guess it still works because at the start of a build there's nothing there, but to be explicit, I added a step that generates a hash key from `embedded-bins/**/*` excluding `embedded-bins/staging/`.

### restore key

I don't think an exact match is necessary for the cache to be beneficial, it should be ok to restore a partly outdated cache and then only rebuild bins that have changed.

### list of paths

The paths to restore included `embedded-bins/Makefile.variables` - I don't think it should be restored, especially not when restoring a partially outdated cache as described above. When restoring an exact match, it should already be identical.

### DIFF
--- a/.github/workflows/check-network.yaml
+++ b/.github/workflows/check-network.yaml
@@ -68,16 +68,25 @@ jobs:
       - name: Terraform Init
         run: terraform init
 
+      - name: Generate embedded binaries cache key
+        id: embedded-bins-hash
+        run: |
+          hash=$(find embedded-bins/ -type f -not -path "embedded-bins/staging/*" -print0 | xargs -0 cat | md5sum | awk '{print $1}')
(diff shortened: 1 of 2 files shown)


kke:
Hmm, git checkout should set the timestamp to whenever it was last changed?

twz123:
> Hmm, git checkout should set the timestamp to whenever it was last changed?

That's not the case.

kke:
I guess there should then be two separate cache steps 🤔

PR #3193, opened 2023-06-06
Remove special-case for setup-go on ARMv7

twz123:
## Description

The setup-go action should now be capable of installing Go on AArch32.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #3362, opened 2023-08-09
Use go 1.21.0

kke:
## Description

[Announcement](https://groups.google.com/g/golang-announce/c/Mk0Jar6hfhI)

[Release Notes](https://go.dev/doc/go1.21)

[Blog](https://go.dev/blog/go1.21)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -279,10 +279,27 @@ jobs:
       # Need to install Go manually: https://github.com/actions/setup-go/issues/106
       - name: Set up Go for smoke tests (armv6l)
         run: |
-          echo "Setup go stable version $GO_VERSION"
           rm -rf -- "$HOME/.local/go"
           mkdir -p -- "$HOME/.local/go"
-          curl --silent -L "https://go.dev/dl/go${GO_VERSION%%.0}.linux-armv6l.tar.gz" | tar -C "$HOME/.local" -xz
+
+          # Remove trailing .0 for go versions before 1.21.0
+          major=$(echo "${GO_VERSION}" | cut -d. -f1)
+          minor=$(echo "${GO_VERSION}" | cut -d. -f2)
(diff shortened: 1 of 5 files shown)


makhov (approved):

ncopa (approved):

twz123:
I wonder if we need to fix the release workflow for ARMv7, since Go changed its "dot zero" versioning scheme. Something similar has been done in #3193 for the PR workflows, but the release workflow has been overlooked in that PR:

https://github.com/k0sproject/k0s/blob/f3db9b6eb4daf6f5c22a7d6b2a37e0ec85050abc/.github/workflows/release.yml#L285

kke:
It would be nice if there was a download link for "latest 1.21.x". 

With 1.20 the current implementation will download 1.20.0 if you give "1.20" or "1.20.0". (1.20.tar.gz contains v1.20, not the latest v1.20.7).

kke:
> Go 1.21 now defines that if a goroutine is panicking and recover was called directly by a deferred function, the return value of recover is guaranteed not to be nil. To ensure this, calling panic with a nil interface value (or an untyped nil) causes a run-time panic of type [*runtime.PanicNilError](https://tip.golang.org/pkg/runtime/#PanicNilError).

>To support programs written for older versions of Go, nil panics can be re-enabled by setting GODEBUG=panicnil=1. This setting is enabled automatically when compiling a program whose main package is in a module that declares go 1.20 or earlier.

I think we can just keep 1.19 in go.mod for k8s <= 1.26 and 1.20 for current and build all of them using go 1.21.

ncopa:
upstream kubernetes 1.28.0 is built with go1.20.7:
```
$ go version kubernetes/client/bin/kubectl
kubernetes/client/bin/kubectl: go1.20.7
```

kke:
I think https://github.com/kubernetes/release/issues/3076 is the issue to track

kke:
Already at 1.21.6

Issue #3351, opened 2023-08-07
Failed to start k0s on arch fresh installation


hnrbs:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.4.7-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 27 Jul 2023 22:02:18 +0000 x86_64 GNU/Linux
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
PRIVACY_POLICY_URL="https://terms.archlinux.org/docs/privacy-policy/"
LOGO=archlinux-logo
```

### Version

v1.27.4+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "9053313309a029437f8acf5805b6351c0af0e835b945656df826f90d015fde8c" (from machine) (pass)
Total memory: 38.4 GiB (pass)
Disk space available for /var/lib/k0s: 735.0 GiB (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.4.7-arch1-1 (pass)
  Max. file descriptors per process: current: 524288 / max: 524288 (pass)
  Executable in path: modprobe: /usr/bin/modprobe (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (pass)
    cgroup controller "memory": available (pass)
    cgroup controller "devices": available (assumed) (pass)
    cgroup controller "freezer": available (assumed) (pass)
    cgroup controller "pids": available (pass)
    cgroup controller "hugetlb": available (pass)
    cgroup controller "blkio": available (via io in version 2) (pass)
  CONFIG_CGROUPS: Control Group support: built-in (pass)
[...]

PR #3366, opened 2023-08-10
Avoid erroring out when localhost cannot be resolved

twz123:
## Description

During certificate generation, k0s makes an effort to compile a comprehensive list of SANs for the TLS certificates. Consequently, it tries to ensure that connections from localhost will function properly: It attempts to perform a lookup for localhost to include its IP address in the list of SANs. Although this lookup is typically expected to succeed, as it's assumed to be local to the machine anyways, it may fail in specific setups, leading to cluster formation issues.

Instead of treating a failed IP lookup for localhost as a critical error, record the error in the logs and proceed with the execution. This approach is likely to be effective if localhost resolves to its standard address, 127.0.0.1.

Additionally, introduce some log statements to capture errors that would otherwise go unnoticed, such as those arising from attempts to resolve the machine's hostname.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -56,8 +56,10 @@ func (c *Certificates) Init(ctx context.Context) error {
 		return err
 	}
 
+	log := logrus.WithField("component", "certificates")
+
 	// We need CA cert loaded to generate client configs
-	logrus.Debugf("CA key and cert exists, loading")
+	log.Debug("CA key and cert exists, loading")
 	cert, err := os.ReadFile(caCertPath)
 	if err != nil {
 		return fmt.Errorf("failed to read ca cert: %w", err)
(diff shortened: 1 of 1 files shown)


kke (approved):

twz123:
As said in https://github.com/k0sproject/k0s/issues/3351#issuecomment-1673538675: I'm not sure if this patch helps much. Maybe it will make things worse, because I'm pretty sure that k0s will just not work when localhost cannot be resolved, and subsequent errors might be more obscure and less obvious. Should we merge it nonetheless?

Maybe we should have some sysinfo probe with a name lookup for localhost, at least.

ncopa:
I'm thinking that if `localhost` does not resolve, something is seriously broken and its better to fail early.

twz123:
Agree. Other parts of the code really rely on localhost being resolvable.

PR #3483, opened 2023-09-13
Capture target variable in loop body

twz123:
## Description

So that goroutines operate on the right target.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/autopilot/controller/readyprober.go
+++ b/pkg/autopilot/controller/readyprober.go
@@ -81,7 +81,7 @@ func (p readyProber) Probe() error {
 	g := errgroup.Group{}
 
 	for _, target := range p.targets {
-		// nolint:govet
+		target := target
 		g.Go(func() error {
 			return func(target apv1beta2.PlanCommandTargetStatus) error {
 				return p.probeOne(target)


pkg/autopilot/controller/readyprober.go
```diff
@@ -81,7 +81,7 @@ func (p readyProber) Probe() error {
 	g := errgroup.Group{}
 
 	for _, target := range p.targets {
-		// nolint:govet
+		target := target
 		g.Go(func() error {
 			return func(target apv1beta2.PlanCommandTargetStatus) error {
 				return p.probeOne(target)
```
ncopa, line 87:
```suggestion
		currentTarget := target
		g.Go(func() error {
			return func(target apv1beta2.PlanCommandTargetStatus) error {
				return p.probeOne(currentTarget)
```

I'm fine either way but `currentTarget` may be slightly more readable.

ncopa: also, the `errorCh` in same function appears to be completely unused.

twz123: IMO It's better to shadow the outer variable completely, as it's virtually always a bug to use it when goroutines are involved in the loop body. Go is actually considering to change the language behavior of this, because it's so error-prone: See https://github.com/golang/go/wiki/LoopvarExperiment

ncopa (approved):

pkg/autopilot/controller/readyprober.go
twz123: > also, the `errorCh` in same function appears to be completely unused.

Good catch. Will file another PR for that.

twz123: #3491

PR #3585, opened 2023-10-13
Add a PodDisruptionBudget for CoreDNS

twz123:
## Description

As CoreDNS is a rather critical component in Kubernetes clusters, k0s should ensure it remains available, even during node maintenance or other disruptions. A PodDisruptionBudget helps to maintain a minimum level of availability.

### Some additional tweaks and their considerations

* No pod anti-affinity for single replica deployments, so that CoreDNS can be rolled on a single node without downtime, as the single replica can remain operational until the new replica can take over.
* No PodDisruptionBudget for single replica deployments, so that CoreDNS may be drained from the node running the single replica. This might leave CoreDNS eligible for eviction due to node pressure, but such clusters aren't HA in the first place and it seems more desirable to not block a drain.
* Set maxUnavailable=1 only for deployments with two or three replicas. Use the Kubernetes defaults (maxUnavailable=25%, rounded down to get absolute values) for all other cases. For single replica deployments, maxUnavailable=1 meant that the deployment's available condition would be true even with zero ready replicas. For deployments with 4 to 7 replicas, this has been the same as the default, and for deployments with 8 or more replicas, this has been artificially constraining the rolling update speed.

### Basic integration test

* Use strings as identifier when detecting multiple CoreDNS pods on nodes. Makes for more readable log/error messages.
* Ignore pods without pod anti-affinity, as they are remnants when scaling up single node clusters.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/inttest/basic/basic_test.go
+++ b/inttest/basic/basic_test.go
@@ -19,7 +19,6 @@ package basic
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"testing"
@@ -33,7 +32,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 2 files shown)


jnummelin (dismissed):

twz123:
This more or less revealed CoreDNS failures in inttests: #3602

twz123:
Also revealed #3609.

juanluisvaladas (dismissed):

juanluisvaladas (approved):

PR #3640, opened 2023-10-25
Rename pkg/context to k0scontext to avoid named imports

kke:
## Description

Since stdlib's `"context"` is imported almost everywhere, having an internal package called `context` forces almost always importing it with a label like `k0sctx`.

Here it is renamed to `k0scontext` to avoid having to do so.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -50,7 +50,7 @@ import (
 	"github.com/k0sproject/k0s/pkg/component/worker"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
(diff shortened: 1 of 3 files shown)


twz123 (approved):
Personally, I think that named imports are okay. Just look at the Kubernetes API structure, where there's always the necessity of adding some name to imports, because all the packages are named "v1", "v1beta1" and so on. It would be awkward to try to disambiguate that for each package, basically repeating information that's already obvious from the package path.

Maybe there's a better name for that k0s package, or just move that stuff to pkg/config?

PR #2221, opened 2022-10-04
API resource version checker

makhov:
Signed-off-by: Alexey Makhov <amakhov@mirantis.com>

## Description

Some APIs used in the cluster might be removed in the new Kubernetes version. The added check ensures that it's safe to update the cluster and stops the plan with the warning if any removed resource found.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #3780, opened 2023-11-29
Remove the duplicate restClientGetter from the stack applier

twz123:
## Description

That struct has already been factored out as part of #2221. Makes sense to use it in all places. Also rename rest_client_getter.go to just rest.go. The new file name is more concise and sufficiently describes its content. Moreover it doesn't use underscores, which is a bit of an uncanny valley in Goland, considering they are also implicitly used for build tags and such.

See:
* #2221

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


pkg/applier/applier.go
```diff
@@ -178,28 +175,3 @@ func (a *Applier) parseFiles(files []string) ([]*unstructured.Unstructured, erro
 
 	return resources, nil
 }
-
-type restClientGetter struct {
```
makhov, line 182: I might be mistaken, but if I recall correctly, it was done on purpose to avoid some caching issue. But it could be outdated info

makhov (approved):

PR #3792, opened 2023-12-07
Patch away chart finalizers

twz123:
## Description

The previous approach was using a standard update and was prone to be failing due to optimistic locking. However, we _want_ the finalizer to disappear after the Helm release has been uninstalled, so instead use a patch that explicitly targets the finalizer, ignoring any intermediate resource updates.

Also check that chart resources get deleted in the addons inttest. The previous checks only verifed that the Helm release had been uninstalled, but didn't take into account the chart resource itself. Due to the aforementioned finalizer removal problem, the deleted chart was never purged from the API server.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/extensions_controller.go
+++ b/pkg/component/controller/extensions_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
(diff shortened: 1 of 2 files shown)


jnummelin (dismissed):

jnummelin (approved):

pkg/component/controller/extensions_controller.go
```diff
@@ -223,13 +226,36 @@ func (cr *ChartReconciler) Reconcile(ctx context.Context, req reconcile.Request)
...
+		Value string `json:"value,omitempty"`
+	}{
+		{"test", path, finalizerName},
+		{"remove", path, ""},
+	})
+	if err != nil {
+		return err
+	}
+
+	return c.Patch(ctx, chart, client.RawPatch(types.JSONPatchType, []byte(patch)))
```
kke, line 256: Isn't patch `[]byte` already?

twz123: Wondering that there's no lint for such things.

jnummelin (approved):

PR #1880, opened 2022-06-22
Add config option to set bind address for all components.

jnummelin:
This enables to bind all components, that listen for traffic, only to certain interface on the host.

Draft as we're waiting for reactions on https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/363
In the meantime konnectivity is build from my custom branch which has the bind-address flags.

Signed-off-by: Jussi Nummelin <jnummelin@mirantis.com>

## Description

Fixes #1150 , the ability to use VIP+LB on controller nodes.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

- [ ] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [ ] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

PR #3824, opened 2023-12-18
Add spec.api.onlyBindToAddress configuration

### LINKED ISSUE #957: Changing "bind-address" of kube-api

pschichtel:
I picked up #1038 and rebased it onto main. Smoke tests pass, but not entirely clear what was left here.

From #1038:

> Should a new `bindAddress` option be created and change documentation for the existing `address` option, or should the existing `address` option's behavior be changed to match the existing documentation?
> 
> The later I can see causing some issues with existing setups, `bindAddress` is probably the safest route. Currently the default for `bindAddress` is "0.0.0.0", which is the default kube-api-server uses if the `--bind-address` flag isn't used.
> 
> **Issue** Fixes #957
> 
> **What this PR Includes** Adds the ability to define a `bindAddress` for the kube-api-server.

### DIFF
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -20,16 +20,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
-	"os"
-	"path/filepath"
-
 	"github.com/k0sproject/k0s/internal/pkg/file"
 	"github.com/k0sproject/k0s/internal/pkg/users"
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 	"github.com/k0sproject/k0s/pkg/certificate"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
+	"net"
+	"os"
+	"path/filepath"
+	"strconv"
 
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
@@ -65,7 +65,8 @@ func (c *Certificates) Init(ctx context.Context) error {
 	}
 	c.CACert = string(cert)
 	// Changing the URL here also requires changes in the "k0s kubeconfig admin" subcommand.
-	kubeConfigAPIUrl := fmt.Sprintf("https://%s:%d", c.ClusterSpec.API.APIServerAddress(), c.ClusterSpec.API.Port)
+	apiAddress := net.JoinHostPort(c.ClusterSpec.API.Address, strconv.Itoa(c.ClusterSpec.API.Port))
+	kubeConfigAPIUrl := fmt.Sprintf("https://%s", apiAddress)
 
 	apiServerUID, err := users.LookupUID(constant.ApiserverUser)
 	if err != nil {
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -65,7 +65,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 	}
 	c.CACert = string(cert)
 	// Changing the URL here also requires changes in the "k0s kubeconfig admin" subcommand.
-	kubeConfigAPIUrl := fmt.Sprintf("https://localhost:%d", c.ClusterSpec.API.Port)
+	kubeConfigAPIUrl := fmt.Sprintf("https://%s:%d", c.ClusterSpec.API.APIServerAddress(), c.ClusterSpec.API.Port)
 
 	apiServerUID, err := users.LookupUID(constant.ApiserverUser)
 	if err != nil {
@@ -73,7 +73,6 @@ func (c *Certificates) Init(ctx context.Context) error {
 		apiServerUID = users.RootUID
 		logrus.WithError(err).Warn("Files with key material for kube-apiserver user will be owned by root")
 	}
-
 	eg.Go(func() error {
 		// Front proxy CA
 		if err := c.CertManager.EnsureCA("front-proxy-ca", "kubernetes-front-proxy-ca"); err != nil {
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -54,6 +54,7 @@ spec:
     address: 192.168.68.104
     k0sApiPort: 9443
     port: 6443
+    bindAddress: 192.0.2.1
     sans:
     - 192.168.68.104
   controllerManager: {}
@@ -113,9 +114,10 @@ spec:
 ### `spec.api`
 
 | Element                  | Description                                                                                                                                                                                                                 |
-| ------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|--------------------------| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `externalAddress`        | The loadbalancer address (for k0s controllers running behind a loadbalancer). Configures all cluster components to connect to this address and also configures this address for use  when joining new nodes to the cluster. |
-| `address`                | Local address on which to bind an API. Also serves as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.                                  |
+| `address`                | Local address on which to bind an API. Also serves as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.                                 |
+| `bindAddress`            | The IP address for the Kubernetes API server to to listen on. The associated interface(s) must be reachable by the rest of the cluster. Will be added as an additional subject alternative name to the API server's TLS certificate. If blank or an unspecified address (`0.0.0.0` or `::`), all interfaces and IP address families will be used. This is effectively the value for the API server's `--bind-address` CLI flag.    |
 | `sans`                   | List of additional addresses to push to API servers serving the certificate.                                                                                                                                                |
 | `extraArgs`              | Map of key-values (strings) for any extra arguments to pass down to Kubernetes api-server process.                                                                                                                          |
 | `port`¹                  | Custom port for kube-api server to listen on (default: 6443)                                                                                                                                                                |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -54,7 +54,6 @@ spec:
     address: 192.168.68.104
     k0sApiPort: 9443
     port: 6443
-    bindAddress: 192.0.2.1
     sans:
     - 192.168.68.104
   controllerManager: {}
@@ -113,15 +112,15 @@ spec:
 
 ### `spec.api`
 
-| Element                  | Description                                                                                                                                                                                                                 |
-|--------------------------| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `externalAddress`        | The loadbalancer address (for k0s controllers running behind a loadbalancer). Configures all cluster components to connect to this address and also configures this address for use  when joining new nodes to the cluster. |
-| `address`                | Local address on which to bind an API. Also serves as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.                                 |
-| `bindAddress`            | The IP address for the Kubernetes API server to to listen on. The associated interface(s) must be reachable by the rest of the cluster. Will be added as an additional subject alternative name to the API server's TLS certificate. If blank or an unspecified address (`0.0.0.0` or `::`), all interfaces and IP address families will be used. This is effectively the value for the API server's `--bind-address` CLI flag.    |
-| `sans`                   | List of additional addresses to push to API servers serving the certificate.                                                                                                                                                |
-| `extraArgs`              | Map of key-values (strings) for any extra arguments to pass down to Kubernetes api-server process.                                                                                                                          |
-| `port`¹                  | Custom port for kube-api server to listen on (default: 6443)                                                                                                                                                                |
-| `k0sApiPort`¹            | Custom port for k0s-api server to listen on (default: 9443)                                                                                                                                                                 |
+| Element             | Description                                                                                                                                                                                                                                                                |
+|---------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `address`           | IP Address used by cluster components to talk to the API server. Also serves as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.                                                       |
+| `onlyBindToAddress` | The API server binds too all interfaces by default. With this option set to `true`, the API server will only listen on the IP address configured by the `address` option (first non-local address by default). This can be necessary with multi-homed control plane nodes. |
+| `externalAddress`   | The loadbalancer address (for k0s controllers running behind a loadbalancer). Configures all cluster components to connect to this address and also configures this address for use when joining new nodes to the cluster.                                                 |
+| `sans`              | List of additional addresses to push to API servers serving the certificate.                                                                                                                                                                                               |
+| `extraArgs`         | Map of key-values (strings) for any extra arguments to pass down to Kubernetes api-server process.                                                                                                                                                                         |
+| `port`¹             | Custom port for kube-api server to listen on (default: 6443)                                                                                                                                                                                                               |
+| `k0sApiPort`¹       | Custom port for k0s-api server to listen on (default: 9443)                                                                                                                                                                                                                |
 
 ¹ If `port` and `k0sApiPort` are used with the `externalAddress` element, the loadbalancer serving at `externalAddress` must listen on the same ports.
 
--- a/pkg/apis/k0s/v1beta1/api.go
+++ b/pkg/apis/k0s/v1beta1/api.go
@@ -37,6 +37,10 @@ type APISpec struct {
 	// Address on which to connect to the API server.
 	Address string `json:"address,omitempty"`
 
+	// The IP address for the Kubernetes API server to listen on.
+	// +optional
+	BindAddress string `json:"bindAddress,omitempty"`
+
 	// The loadbalancer address (for k0s controllers running behind a loadbalancer)
 	ExternalAddress string `json:"externalAddress,omitempty"`
 
@@ -102,10 +106,21 @@ func (a *APISpec) getExternalURIForPort(port int) string {
 	return fmt.Sprintf("https://%s:%d", addr, port)
 }
 
-// Sans return the given SANS plus all local adresses and externalAddress if given
+// APIServerAddress returns the address the API is listening on
+func (a *APISpec) APIServerAddress() string {
+	if a.BindAddress == "" {
+		return "localhost"
+	}
+	return a.BindAddress
+}
+
+// Sans return the given SANS plus all local addresses and externalAddress if given
 func (a *APISpec) Sans() []string {
 	sans, _ := iface.AllAddresses()
 	sans = append(sans, a.Address)
+	if a.BindAddress != "" {
+		sans = append(sans, a.BindAddress)
+	}
 	sans = append(sans, a.SANs...)
 	if a.ExternalAddress != "" {
 		sans = append(sans, a.ExternalAddress)
@@ -150,6 +165,9 @@ func (a *APISpec) Validate() []error {
 		validateIPAddressOrDNSName(sansPath.Index(idx), san)
 	}
 
+	if a.BindAddress != "" && !govalidator.IsIP(a.BindAddress) {
+		errors = append(errors, field.Invalid(field.NewPath("bindAddress"), a.BindAddress, "invalid IP address"))
+	}
 	return errors
 }
 
--- a/pkg/apis/k0s/v1beta1/api.go
+++ b/pkg/apis/k0s/v1beta1/api.go
@@ -37,9 +37,9 @@ type APISpec struct {
 	// Address on which to connect to the API server.
 	Address string `json:"address,omitempty"`
 
-	// The IP address for the Kubernetes API server to listen on.
+	// Whether to only bind to the IP given by the address option.
 	// +optional
-	BindAddress string `json:"bindAddress,omitempty"`
+	OnlyBindToAddress bool `json:"onlyBindToAddress,omitempty"`
 
 	// The loadbalancer address (for k0s controllers running behind a loadbalancer)
 	ExternalAddress string `json:"externalAddress,omitempty"`
@@ -97,9 +97,6 @@ func (a *APISpec) K0sControlPlaneAPIAddress() string {
 
 func (a *APISpec) getExternalURIForPort(port int) string {
 	addr := a.Address
-	if a.BindAddress != "" {
-		addr = a.BindAddress
-	}
 	if a.ExternalAddress != "" {
 		addr = a.ExternalAddress
 	}
@@ -109,27 +106,20 @@ func (a *APISpec) getExternalURIForPort(port int) string {
 	return fmt.Sprintf("https://%s:%d", addr, port)
 }
 
-// APIServerAddress returns the address the API is listening on
-func (a *APISpec) APIServerAddress() string {
-	if a.BindAddress == "" {
-		return "localhost"
-	}
-	return a.BindAddress
-}
-
 // Sans return the given SANS plus all local addresses and externalAddress if given
 func (a *APISpec) Sans() []string {
 	sans, _ := iface.AllAddresses()
 	sans = append(sans, a.Address)
-	if a.BindAddress != "" {
-		sans = append(sans, a.BindAddress)
-	}
 	sans = append(sans, a.SANs...)
 	if a.ExternalAddress != "" {
 		sans = append(sans, a.ExternalAddress)
 	}
 
 	return stringslice.Unique(sans)
+}
+
+func isAnyAddress(address string) bool {
+	return address == "0.0.0.0" || address == "::"
 }
 
 // Validate validates APISpec struct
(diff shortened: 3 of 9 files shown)


pkg/apis/k0s/v1beta1/api.go
```diff
@@ -100,10 +104,20 @@ func (a *APISpec) getExternalURIForPort(port int) string {
...
 
+// APIServerAddress returns the address the API is listening on
+func (a *APISpec) APIServerAddress() string {
+	return a.getAPIServerAddress(a.BindAddress)
+}
+
+func (a *APISpec) getAPIServerAddress(address string) string {
+	return address
+}
+
```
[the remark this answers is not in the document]
twz123: OTOH, if this would fall back to `localhost` if BindAddess is empty, it would make some sense again, and it would also mean that there'd be no behavioral change at all when BindAddress is not specified.

pkg/apis/k0s/v1beta1/api.go
```diff
@@ -35,6 +35,9 @@ type APISpec struct {
 	// Local address on which to bind an API
 	Address string `json:"address"`
 
+	// Local address on which to bind the kube-apiserver. Defaults to 0.0.0.0.
+	BindAddress string `json:"bindAddress,omitempty"`
```
twz123, line 39: We might want to add some kubebuilder validation here to reject nonsensical values already at the API server level. Sure, a regex that's correctly checking for IPv4/IPv6 compliance will probably be one megabyte in size or so, but something short that rejects obviously wrong stuff might be helpful nevertheless?

twz123 (commented):
Thanks for picking this up.

I'd double-check if there are some bad interactions with node-local load balancing? I imagine the LB config needs to pick up the correct bind addresses as well?

We should think about some integration test, too, I suppose.

pkg/apis/k0s/v1beta1/api.go
pschichtel: what are you thinking of? something like `^([0-9.]+|[0-9a-f:]+)$` would at least ensure that the correct alphabet is being used while keeping it simple.

docs/configuration.md
```diff
@@ -55,6 +55,7 @@ spec:
     externalAddress: my-lb-address.example.com
     k0sApiPort: 9443
     port: 6443
+    bindAddress: 0.0.0.0
```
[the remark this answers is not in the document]
pschichtel: I guess there is no equivalent to example.org in ip addresses, maybe some private IP address? 192.168.0.1 mabye ?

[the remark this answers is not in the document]
pschichtel: nvm, there actually is: https://datatracker.ietf.org/doc/html/rfc5737

I'm going for TEST-NET-1's 192.0.2.1

twz123:
Okay, some things to proceed with:

* We need an integration test. Need to come up with a scenario for that, i.e. what is to be tested.
* We need to double-check if this plays nicely with nllb and other things. Do we need to update the docs on e.g. external load balancing? How would this affect k0sctl, if at all?

pkg/apis/k0s/v1beta1/api.go
twz123: Yeah, something like this. Although I have a gut feeling that the current regex will break IPv6 link-local addresses. Maybe my idea wasn't the best here :smile:. A pity that there's no "ip" format in the OpenAPI validator, just "ipv4" _or_ "ipv6" (at least as far as I reckon). A workaround would be to have `bindAddress: {type: IPv4, ipv4: 127.0.0.1}` and `bindAddress: {type: IPv6, ipv6: "::1"}`, just to have two distinct fields to put the format tags on. That'd be a _very_ artificial distinction, though.

Overall, we might be better off without server-side validation, unfortunately, just to not introduce unnecessary bugs.

pschichtel: > Although I have a gut feeling that the current regex will break IPv6 link-local addresses.

that's why I added the `%` in a later revision, but I'm still not familiar enough with the related RFCs to be certain it covers everything.

I also lean towards not risking it.

pschichtel:
> We need an integration test. Need to come up with a scenario for that, i.e. what is to be tested.

I'm not very familiar with the code base yet, can you point me to examples for integration tests?

> We need to double-check if this plays nicely with nllb and other things.

I'd have to look into how nllb works exactly. Depending on how the upstream connection is made its configuration probably needs adjustments.

> Do we need to update the docs on e.g. external load balancing?

I guess a disclaimer like "if you change the bind address make sure to align your external load balancer" or so. I feel like people changing bind addresses are probably aware of the impact. But that's just my opinion.

> How would this affect k0sctl, if at all?

k0sctl would need to interact with the API server on the correct IP, wouldn't it?

twz123:
Sorry for getting back to you so late @pschichtel. I'll bluntly blame the holiday season for that. :sweat_smile:

> > We need an integration test. Need to come up with a scenario for that, i.e. what is to be tested.
> 
> I'm not very familiar with the code base yet, can you point me to examples for integration tests?
> 
> > We need to double-check if this plays nicely with nllb and other things.
> 
> I'd have to look into how nllb works exactly. Depending on how the upstream connection is made its configuration probably needs adjustments.

We could try to start with the [existing nllb inttest](https://github.com/k0sproject/k0s/tree/e121de6a33b864bc3db74c8a23a9d488f9538177/inttest/nllb), copy that over into a separate directory and start hacking on the new one. You can probably delete all the cluster-restart stuff in there, and just rely on the cluster bootstrapping part. You need to add the new directory name to [`inttest/Makefile.variables`](https://github.com/k0sproject/k0s/tree/e121de6a33b864bc3db74c8a23a9d488f9538177/inttest/Makefile.variables). Then you should be able to do something like `make check-yolo` (or whatever the name is you choose.)

Might make sense to simply try to overwrite the bind addresses of every component that you can get hold of and see what happens to the cluster.

> > Do we need to update the docs on e.g. external load balancing?
> 
> I guess a disclaimer like "if you change the bind address make sure to align your external load balancer" or so. I feel like people changing bind addresses are probably aware of the impact. But that's just my opinion.

Sounds good enough for the start.

> 
> > How would this affect k0sctl, if at all?
> 
> k0sctl would need to interact with the API server on the correct IP, wouldn't it?

Yes. Would it need to look at the k0s config to figure this out? Or would it "just work"? Maybe @kke can tell.

pschichtel:
> Sorry for getting back to you so late @pschichtel. I'll bluntly blame the holiday season for that. 😅
 
No worries, I don't have any rush with this one.

> We could try to start with the [existing nllb inttest](https://github.com/k0sproject/k0s/tree/e121de6a33b864bc3db74c8a23a9d488f9538177/inttest/nllb), copy that over into a separate directory and start hacking on the new one. You can probably delete all the cluster-restart stuff in there, and just rely on the cluster bootstrapping part. You need to add the new directory name to [inttest/Makefile.variables](https://github.com/k0sproject/k0s/tree/e121de6a33b864bc3db74c8a23a9d488f9538177/inttest/Makefile.variables). Then you should be able to do something like make check-yolo (or whatever the name is you choose.)
> 
> Might make sense to simply try to overwrite the bind addresses of every component that you can get hold of and see what happens to the cluster.

Makes sense, I'll have a go at that.

twz123:
(The linter error will go away after a rebase. See #3991 for details.)

pschichtel:
yeah I plan to work on this again towards the end of this week

andycandy-de:
Any idea when this feature will be available? I also have multiple IPs and only want to bind the k0scontroller to a specific IP.

pschichtel:
@andycandy-de sadly this had to be pushed back a little in my prio list, but I still intent to eventuelly finish this up. I think some of the recent k0s/k8s releases also added additional components, which probably needs/deserves some investigation. I'm currently unable to work on it until mid of April, so no sooner than that.

pkg/apis/k0s/v1beta1/api.go
pschichtel: Is this something that could be deferred to after this PR ?

twz123: Yeah, let's just skip validation right now, and just :crossed_fingers:

pkg/apis/k0s/v1beta1/api.go
```diff
@@ -57,11 +60,12 @@ func DefaultAPISpec() *APISpec {
...
 	publicAddress, _ := iface.FirstPublicAddress()
 	return &APISpec{
-		Port:       defaultKasPort,
-		K0sAPIPort: 9443,
-		SANs:       addresses,
-		Address:    publicAddress,
-		ExtraArgs:  make(map[string]string),
+		Port:        defaultKasPort,
+		K0sAPIPort:  9443,
+		BindAddress: "0.0.0.0",
```
twz123, line 65: Any reason for defaulting this to "0.0.0.0"?

pkg/apis/k0s/v1beta1/api_test.go
```diff
@@ -35,7 +35,8 @@ func (s *APISuite) TestValidation() {
 
 	s.T().Run("accepts_ipv6_as_address", func(t *testing.T) {
 		a := APISpec{
-			Address: "2001:0db8:85a3:0000:0000:8a2e:0370:7334",
+			Address:     "2001:0db8:85a3:0000:0000:8a2e:0370:7334",
+			BindAddress: "2001:0db8:85a3:0000:0000:8a2e:0370:7334",
```
twz123, line 39: Do we need to add the BindAddress to theexisting test cases?

pkg/apis/k0s/v1beta1/api_test.go
```diff
@@ -68,6 +71,18 @@ func (s *APISuite) TestValidation() {
 			s.ErrorContains(errors[0], `sans[0]: Invalid value: "something.that.is.not.valid//(())": invalid IP address / DNS name`)
 		}
 	})
+
+	s.T().Run("invalid_api_bind_address", func(t *testing.T) {
+		a := APISpec{
+			Address:     "1.2.3.4",
+			BindAddress: "somehting.that.is.not.valid//(())",
```
twz123, line 78: :stuck_out_tongue:

```suggestion
			BindAddress: "something.that.is.not.valid//(())",
```

pkg/apis/k0s/v1beta1/api.go
pschichtel: Compared to just leaving it out?

pkg/apis/k0s/v1beta1/api_test.go
pschichtel: As in: should we leave them untouched?

pkg/apis/k0s/v1beta1/api.go
twz123: Yes.

pkg/apis/k0s/v1beta1/api.go
```diff
@@ -35,6 +35,9 @@ type APISpec struct {
 	// Local address on which to bind an API
 	Address string `json:"address"`
```
twz123, line 36: Now I feel really stupid for only realizing it now, but this parameter actually sounds like what we want to do here.

We use this parameter as the advertise address, not as the bind address. Would the following assumption hold?

"The advertise address will work when connecting to the API server locally, as well as from other peer controllers."

If yes, then the implementation could be much easier: Just fix all the places where k0s uses a hard coded localhost or 127.0.0.1 to connect to the API server by using the `Address`. This is pretty much all the places where `BindAddress` is used now. Then set `spec.api` to sth. like `{"address": "x.x.x.x", "extraArgs": {"bind-address": "x.x.x.x"}}`. And we're done. (At least that's what I'd guess :see_no_evil:)

twz123:
>But how would I go about that in the current containerized setup, given that I have limited control over what networks are being used.

Might be sufficient if we bind to something else than 0.0.0.0 or a loopback address. So what you could try to do is to bind to the Docker container's IP. In the integration test, you could get that via

```go
controller0, err := s.MachineForName(s.ControllerNode(0))
s.Require().NoError(err)
ip := controller0.Status().IP
```

pkg/apis/k0s/v1beta1/api.go
pschichtel: I noticed address myself last time I worked on this, totally forgot to add a comment about that :D I'll look into it.

pschichtel: @twz123 the latest commit replaces BindAddress with Address as suggested.

Both versions (with dedicated BindAddress and with Address used for bind address) seem to work.

@gakio12 in the original PR #1038 actually asked the question:

> Should a new bindAddress option be created and change documentation for the existing address option, or should the existing address option's behavior be changed to match the existing documentation?

To which @jnummelin answered, that he prefers the dedicated option.

How would we justify (and document) the existence of both Address BindAddress? I see one possible reason to separate them: `0.0.0.0` as a bind address ([which also came up in the previous PR.](https://github.com/k0sproject/k0s/pull/1038#pullrequestreview-817786306)). If address is something I want to use to connect to the API server, then it should never be `0.0.0.0`. If address is something the server binds to, then I'd expect to be able to set `0.0.0.0`.

pschichtel: The broken tests seem to be broken by the switch over to Address also being the bind address and I think that makes sense. I just attempted to change it so Address takes to role of BindAddress and merge its previous role with ExternalAddress, but that falls apart quickly.

My current understanding is:

* Address: Used internally by the cluster
* ExternalAddress: Used by clients external to the cluster
* BindAddress: The address the api server binds to

Clearly from this, if BindAddress is set, than Address must be compatible with its value, because we can't bind the API server to e.g. 10.0.0.1 and then advertise it's address as 10.0.0.2. But we can bind it to 0.0.0.0 and then advertise it as any address.

New proposal for the options:

* Address: IP used internally by the cluster and used for the `--bind-address` option in API server
* ExternalAddress: Address used by clients external to the cluster (does this one actually need to be an IP address?)
* OnlyBindToAddress: Boolean option (default false for compatibility) to allow binding the API server to only the IP address given by Address instead of 0.0.0.0/::

Address and ExternalAddress would also be validated to not be 0.0.0.0 or ::.

pschichtel:
I'll try to have a look at the openebs test this evening

pschichtel:
The test actually passed on my machine

pschichtel:
@twz123 @jnummelin given this is now the oldest open PR and I haven't heard back in a while: Is there still interest in this? I've been keeping it in sync with the master so far. I myself don't have a need for this anymore, but I'm still willing to push it over the finish line.

andycandy-de:
@pschichtel I am still waiting for this feature to get merged... Thank you for your efforts... ❤️

@twz123 @jnummelin any chance to get the feature merged soon?

jr200:
I'm also interested in this PR being merged. Thanks all.

pschichtel:
@jr200  @andycandy-de has any one if you guys given this PR a spin in a lab environment? It's fairly simple to use custom builds with k0s

jr200:
> @jr200 @andycandy-de has any one if you guys given this PR a spin in a lab environment? It's fairly simple to use custom builds with k0s. It would also be pretty easy to backport it to older stable releases.

Hi @pschichtel,

I had a go just now and I hit some dead-ends. I'm using `k0sctl` on macos (intel). I checked out your branch, installed docker-desktop, did `make`, which I then realised built the linux version of the binaries. I tried a manual `go` build from your folder which failed without producing any binaries on the error `pkg/component/worker/containerd/component.go:121:9: undefined: winExecute`. 

Some CIs for open-source projects are setup to automatically build binaries on all platforms which can be tested by downloading them from the CI directly. Can you generate a macos binary binary somehow - i'd be happy to test...

By the way, I think this PR only partially solves my issue. I'm using k0s over zerotier VPN, so I wanted all k0s-related ports to bind on the IP from a specific network-interface (i.e., the VPN's NIC). This patch doesn't seem to set `--advertise-address` on the `spec.api.extraArgs` - should it?

pschichtel:
> I'm using k0sctl on macos (intel). I checked out your branch, installed docker-desktop, did make, which I then realised built the linux version of the binaries. I tried a manual go build from your folder which failed without producing any binaries on the error pkg/component/worker/containerd/component.go:121:9: undefined: winExecute.

I don't have the capacity to help you with macos build issues, sorry.

> This patch doesn't seem to set --advertise-address on the spec.api.extraArgs - should it?

Isn't this where it's set: https://github.com/k0sproject/k0s/pull/3824/files#diff-e2c56849a04d7e414d5b968597251ab14c9f4229772cca9babf39600f63d3991R102 ? That would be configured by the "address" option k0s already has.

pschichtel:
it's been months of rebasing now.

docs/configuration.md
```diff
@@ -112,14 +114,15 @@ spec:
...
+| Element             | Description                                                                                                                                                                                                                  |
+|---------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `address`           | Local address on which to bind an API. Also serves as one of the addresses pushed on the k0s create service certificate on the API. Defaults to first non-local address found on the node.                                   |
+| `onlyBindToAddress` | Whether to bind only to the IP address given by the `address` option, instead of binding to all addresses.                                                                                                                   |
```
twz123, line 120: Would be great to add some details as to why this knob exists, and what may be the consequences if set to true. Basically some excerpt of https://github.com/k0sproject/k0s/pull/3824#discussion_r1622862280.

pkg/component/controller/apiserver.go
```diff
@@ -230,7 +234,7 @@ func (a *APIServer) Ready() error {
 		TLSClientConfig: tlsConfig,
 	}
 	client := &http.Client{Transport: tr}
-	resp, err := client.Get(fmt.Sprintf("https://localhost:%d/readyz?verbose", a.ClusterConfig.Spec.API.Port))
+	resp, err := client.Get(fmt.Sprintf("https://%s:%d/readyz?verbose", a.ClusterConfig.Spec.API.Address, a.ClusterConfig.Spec.API.Port))
```
twz123, line 237: Ideally, this would use `net.JoinHostPort`. I think it'll break for IPv6 addresses otherwise.

cmd/controller/certificates.go
```diff
@@ -64,7 +64,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 	}
 	c.CACert = string(cert)
 	// Changing the URL here also requires changes in the "k0s kubeconfig admin" subcommand.
-	kubeConfigAPIUrl := fmt.Sprintf("https://localhost:%d", c.ClusterSpec.API.Port)
+	kubeConfigAPIUrl := fmt.Sprintf("https://%s:%d", c.ClusterSpec.API.Address, c.ClusterSpec.API.Port)
```
twz123, line 67: Ideally, this would use `net.JoinHostPort`. I think it'll break for IPv6 addresses otherwise.

docs/configuration.md
```diff
@@ -52,8 +52,10 @@ metadata:
 spec:
   api:
     address: 192.168.68.104
+    onlyBindToAddress: false
     k0sApiPort: 9443
     port: 6443
+    bindAddress: 192.0.2.1
```
twz123, line 58: This is a remnant, right?

docs/configuration.md
```diff
@@ -52,8 +52,10 @@ metadata:
 spec:
   api:
     address: 192.168.68.104
+    onlyBindToAddress: false
```
twz123, line 55: As it defaults to `false` and it's `omitempty`, the `k0s kubeconfig create` command won't include that field.

```suggestion
```

docs/configuration.md
twz123, line 120: I wonder if we could just keep `bindAddress` instead of `onlyBindToAddress` and let the user decide which address to bind to instead of forcing it to be either the advertised address or all interfaces. This may be a niche requirement, though. I cannot really judge. Maybe some folks with more network expertise have an opinion? OTOH, the `address` field will be defaulted by k0s's address discovery if omitted. There wouldn't be an easy way to let users use the auto-detected address as the bind address, then.

Anyway, I'm fine with the `onlyBindToAddress` logic as well, but wanted to at least ask about it.

twz123, line 120: An alternative name could be `bindToAllInterfaces` which defaults to `true`. This would have to be implemented as a bool pointer though. `onlyBindToAddress` is a bit harder for me to understand.

twz123 (commented):
Thanks for your dedication and perseverance, and sorry for the very late feedback. Looks good. Just some small fixes to use `net.JoinHostPort`. I'd like to get some quick feedback from other maintainers on the name of the config knob, 👍 otherwise.

docs/configuration.md
jnummelin: IMO `bindAddress` makes more sense and is probably more common in general. At least I remember seeing such knobs on other things quite often. 😄

twz123: Agreed. However, what about

> There wouldn't be an easy way to let users use the auto-detected address as the bind address, then.

jnummelin: hmm, what's wrong in the current default behaviour of autodetecting the publish address and binding to all interfaces?

docs/configuration.md
pschichtel: right

docs/configuration.md
pschichtel: I read in other PRs that you generally prefer defaulting to false, so I'd assume other booleans are also like that. Not sure this is worth breaking the consistency, especially since this is more of a niche option.

pschichtel: > IMO bindAddress makes more sense and is probably more common in general. At least I remember seeing such knobs on other things quite often.

bind-address is a common term yes, but rarely any software asks you to correctly configure several addresses for different purposes, that may or may not interact with each other. That's why I intended to reduce the bind-address parameter to what makes sense. I think the current documentation of `address` is misleading, given nothing actually binds to that address, it's just the internal address published to other components.

> hmm, what's wrong in the current default behaviour of autodetecting the publish address and binding to all interfaces?

The original issue #957 describe where this PR comes from: Multi-homed kubernetes hosts. In my case specially, my k0s hosts was also my edge router, so one of the interfaces the API server bound to was my internet uplink. With proper firewall rules in place no big deal, but it carries some risk. My setup since changed (for the better).

> There wouldn't be an easy way to let users use the auto-detected address as the bind address, then.

if the user configured `bindToAddressOnly: true` without actually configuring the `address`, then it would just use the auto-detected address, right? (probably worth documenting too)

As I said in https://github.com/k0sproject/k0s/pull/3824#discussion_r1622862280, `address` and `bindAddress` are strongly related. Configuring them incorrectly will break the cluster. You can't bind to an address that's not internally announced (unless NAT between components is supported, which would be ridiculous), so you have 2 options: bind to all interfaces and announce one of the addresses or bind to the address you announce. I can't see any other valid configuration (again: assuming NAT is not a thing here). 

Also... I'd assume (I don't have numbers on it) fairly few people actually have a need for this. Few kubernetes hosts are multi-homed and of those most are fine with binding all interfaces.

cmd/controller/certificates.go
pschichtel: makes sense, done

docs/configuration.md
pschichtel: I improved it I think, please have a look.

pkg/apis/k0s/v1beta1/api.go
```diff
@@ -37,6 +37,10 @@ type APISpec struct {
 	// Address on which to connect to the API server.
 	Address string `json:"address,omitempty"`
 
+	// Whether to only bind to the IP given by the address option.
+	// +optional
+	OnlyToBindAddress bool `json:"onlyBindToAddress,omitempty"`
```
jnummelin, line 42:
```suggestion
	OnlyBindToAddress bool `json:"onlyBindToAddress,omitempty"`
```

IMO worth to keep the variable name and json key in sync

pschichtel: ops, yeah I'll fix that in the evening.

pschichtel:
Is that failing job just flakyness? It seems to exceeded a deadline unrelated to my change, right?

twz123:
Yep, the inttests are quite a bit flaky, unfortunately.

docs/configuration.md
kke: `bindExclusivelyToAddress`, `bindExclusive`, `bind(To)Address(Only)`, `addressBind`?

Somehow `onlyBindToAddress` sounds clunky to me.

Of course `address` (advertise) + `bindAddress: <address>` (bind) would be the best, I think there's an issue for that (#1880).

pschichtel: > Somehow onlyBindToAddress sounds clunky to me.

I agree, I'm not super happy with the name, but at least it's descriptive.

> Of course address (advertise) + bindAddress: <address> (bind) would be the best

I actually don't agree here. I think `bindAddress` is not an option that should be offered, because the additional flexibility brings no benefit, only the risk that people misconfigure it.

twz123 (approved):
Thank you!

mback2k:
`k0sctl` will need to be tweaked to support this as it currently tries to validate the API server coming up this way:
```
executing `curl -kso /dev/null --connect-timeout 20 -w \"%{http_code}\" \"https://localhost:6443/version\"`"
```
and then fails this way:
```
INFO ==> Running phase: Upgrade controllers 
INFO [ssh] ...: starting upgrade   
INFO [ssh] ...: waiting for the k0s service to start 
INFO * Running clean-up for phase: Acquire exclusive host lock 
INFO * Running clean-up for phase: Upgrade controllers 
INFO ==> Apply failed 
```

twz123:
@mback2k Mind creating a [k0sctl issue](https://github.com/k0sproject/k0sctl/issues/new) for that?

PR #3831, opened 2023-12-18
Use an informer to watch and approve CSRs

cPu1:
## Description

Uses an Informer to watch and approve CSRs, instead of polling the API server and calling the `List` API every few seconds. Additionally, the existing `TestBasicCRSApprover` test was inaccurate in that it wasn't actually testing the condition that the certificate has been approved. The call to `approveCSR` was failing the validation in `ValidateKubeletServingCSR`, but since the error is only being logged and not returned, the test was still passing. The actual certificate status check [here](https://github.com/k0sproject/k0s/blob/7d9532a9c51039ba8a2b1ddd6351ee440d9a31bd/pkg/component/controller/csrapprover_test.go#L84) was never reached because `status.Conditions` was empty.

This changelist also fixes the existing test by generating a certificate request that complies with the validation in `ValidateKubeletServingCSR`, and by mocking the fake `Clientset` to authorize all operations (by responding to creation of `SubjectAccessReviews` with `status.Allowed = true`). A new test is also added to cover the case where the CSRs already exist before `CSRApprover.Start()` is called.

This may slightly increase the memory footprint as the CSRs will need to be held in the informer's cache, but it will be limited to CSRs with `spec.signerName` set to `kubernetes.io/kubelet-serving`. This is a minor trade-off compared to polling which is slower to respond to newly-issued CSRs, makes more requests to the API server and fetches a full list of kubelet-serving CSRs in each request.

A TODO is also added for the future to share the informer factory with other components.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

The checks listed [here](https://github.com/k0sproject/k0s/blob/main/docs/contributors/testing.md#run-local-verifications) are passing. 

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -279,7 +279,9 @@ func (c *command) start(ctx context.Context) error {
 	if !slices.Contains(c.DisableComponents, constant.CsrApproverComponentName) {
 		nodeComponents.Add(ctx, controller.NewCSRApprover(nodeConfig,
 			leaderElector,
-			adminClientFactory))
+			adminClientFactory,
+			30*time.Minute),
+		)
 	}
 
 	if c.EnableK0sCloudProvider {
--- a/pkg/component/controller/csrapprover.go
+++ b/pkg/component/controller/csrapprover.go
@@ -20,23 +20,26 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
-	"sync/atomic"
 	"time"
 
+	"github.com/avast/retry-go"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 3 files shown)


twz123:
I think an informer is a big gun with all its caching. For simple use cases like this, k0s has `github.com/k0sproject/k0s/pkg/kubernetes/watch`. Might be simpler and won't cache anything.

jnummelin:
> I think an informer is a big gun with all its caching

I agree, does feel like a bit too big of a 🔫  for this use case

twz123:
Maybe I'll find some time to put together an alternative to this using `github.com/k0sproject/k0s/pkg/kubernetes/watch` in the next weeks.

PR #3898, opened 2024-01-08
Replace image validity check with shell script

twz123:
## Description

The script is more or less straight forward and saves us some go dependencies and a custom go build execution.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


twz123:
On the other hand, is this functionality still useful? K0s ships its own images anyways, so it's under the project's control to provide appropriate multi-arch images anyways.

jnummelin:
> We could simply remove it?

I wonder would we catch any missing archs in smokes? For most of the images probably yes, but we're not really running with all permutation configs for arm/arm64 smokes that would exercise all the images.

twz123:
My point here is that the default images are all built by the k0s project. How likely is it that we're forgetting to ship an architecture? I guess this was a bigger concern back in the day when k0s used upstream images verbatim.

twz123:
Plus, wouldn't the airgap image tarball generation fail when an image is unavailable for a certain architecture?

jnummelin:
> How likely is it that we're forgetting to ship an architecture? 

Quite low I'd say

> wouldn't the airgap image tarball generation fail when an image is unavailable for a certain architecture?

I think you're right, it would AFAIK fail.

So maybe we should just remove this stuff

twz123:
Superseeded by #3913.

PR #3935, opened 2024-01-16
Add k0sctl version lookup to vars.sh

twz123:
## Description

The vars.sh script is used to obtain variables from Makefile variables in scripts. Even if the k0sctl version is not maintained in a Makefile, but rather in a go.mod file, we can use the same interface for scripts. This makes the origin transparent and we can special-case anything in the vars.sh script, instead of introducing other ad-hoc mechanisms.

Moreover, include the k0sctl version in the build-env by default. It's fast and safe enough to be included by default, even if not required.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/vars.sh
+++ b/vars.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
+#shellcheck disable=SC3043
 
 set -eu
 
@@ -10,6 +11,18 @@ print_usage() {
   echo
   echo "  $0 go_version"
   echo "  $0 FROM=docs python_version"
+}
+
+version_from_go_mod() {
+  local pkg version rest
+  while read -r pkg version rest; do
+    if [ "$pkg" = "$1" ]; then
+      printf %s "$version"
+      return 0
+    fi
+  done
+
+  return 1
 }
 
 fail() {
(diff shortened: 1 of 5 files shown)


juanluisvaladas (approved):
I think using `go list -m`is cleaner than this, I leave you a suggestion.
Nonetheless this code is correct so if you have some reason to prefer that's fine too.

vars.sh
```diff
@@ -12,6 +13,18 @@ print_usage() {
...
+  while read -r pkg version rest; do
+    if [ "$pkg" = "$1" ]; then
+      printf %s "$version"
+      return 0
+    fi
+  done
+
+  return 1
+}
+
```
juanluisvaladas, line 27:
```suggestion
```

vars.sh
```diff
@@ -41,6 +54,11 @@ done
 [ -n "$var" ] || fail Makefile variable not given
 [ -n "$from" ] || from=embedded-bins
 
+if [ "$var" = k0sctl_version ]; then
+  version_from_go_mod github.com/k0sproject/k0sctl <"$from"/go.mod
```
juanluisvaladas, line 58:
```suggestion
  (cd hack/tool && go list -m -f '{{.Version}}' github.com/k0sproject/k0sctl)
```

juanluisvaladas: Then it's a good solution.

PR #3954, opened 2024-01-18
Dual-stack with Kube router 2

ncopa:
## Description

Update kube-router to 2.0.1 and enable ipv6

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/inttest/dualstack/dualstack_test.go
+++ b/inttest/dualstack/dualstack_test.go
@@ -113,6 +113,15 @@ func (s *DualstackSuite) SetupSuite() {
 	err = s.WaitForNodeReady(s.WorkerNode(1), client)
 	s.Require().NoError(err)
 
+	for i := 0; i < s.WorkerCount; i++ {
+		ssh, err := s.SSH(s.Context(), s.WorkerNode(i))
+		s.Require().NoError(err)
+		defer ssh.Disconnect()
+		output, err := ssh.ExecWithOutput(s.Context(), "cat /proc/sys/net/ipv6/conf/all/disable_ipv6")
+		s.Require().NoError(err)
+		s.T().Logf("worker%d: /proc/sys/net/ipv6/conf/all/disable_ipv6=%s", i, output)
+	}
+
 	kc, err := s.KubeClient("controller0", "")
 	s.Require().NoError(err)
 	restConfig, err := s.GetKubeConfig("controller0", "")
--- a/inttest/dualstack/dualstack_test.go
+++ b/inttest/dualstack/dualstack_test.go
@@ -91,7 +91,13 @@ func (s *DualstackSuite) SetupSuite() {
 	s.Require().True(isDockerIPv6Enabled, "Please enable IPv6 in docker before running this test")
 	s.BootlooseSuite.SetupSuite()
 
-	s.PutFile(s.ControllerNode(0), "/tmp/k0s.yaml", k0sConfigWithDualStack)
+	k0sConfig := k0sConfigWithCalicoDualStack
+
+	if os.Getenv("K0S_NETWORK") == "kube-router" {
+		s.T().Log("Using kube-router network")
+		k0sConfig = k0sConfigWithKuberouterDualStack
+	}
+	s.PutFile(s.ControllerNode(0), "/tmp/k0s.yaml", k0sConfig)
 	controllerArgs := []string{"--config=/tmp/k0s.yaml"}
 	if os.Getenv("K0S_ENABLE_DYNAMIC_CONFIG") == "true" {
 		s.T().Log("Enabling dynamic config for controller")
@@ -191,7 +197,7 @@ func TestDualStack(t *testing.T) {
 
 }
 
-const k0sConfigWithDualStack = `
+const k0sConfigWithCalicoDualStack = `
 spec:
   network:
     kubeProxy:
@@ -199,6 +205,17 @@ spec:
     provider: calico
     calico:
       mode: "bird"
+    dualStack:
+      enabled: true
+      IPv6podCIDR: "fd00::/108"
+      IPv6serviceCIDR: "fd01::/108"
+    podCIDR: 10.244.0.0/16
+    serviceCIDR: 10.96.0.0/12`
+
+const k0sConfigWithKuberouterDualStack = `
+spec:
+  network:
+    provider: kuberouter
     dualStack:
       enabled: true
       IPv6podCIDR: "fd00::/108"
--- a/pkg/component/worker/kubelet.go
+++ b/pkg/component/worker/kubelet.go
@@ -170,8 +170,9 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		args["--node-labels"] = strings.Join(k.Labels, ",")
 	}
 
+	extras := flags.Split(k.ExtraArgs)
 	if runtime.GOOS == "windows" {
-		node, err := node.GetNodename("")
+		node, err := node.GetNodename(extras["--hostname-override"])
 		if err != nil {
 			return fmt.Errorf("can't get hostname: %w", err)
 		}
@@ -200,7 +201,6 @@ func (k *Kubelet) Start(ctx context.Context) error {
 
 	// Handle the extra args as last so they can be used to override some k0s "hardcodings"
 	if k.ExtraArgs != "" {
-		extras := flags.Split(k.ExtraArgs)
 		args.Merge(extras)
 	}
 	logrus.Debugf("starting kubelet with args: %v", args)
--- a/pkg/component/worker/kubelet.go
+++ b/pkg/component/worker/kubelet.go
@@ -64,6 +64,7 @@ type Kubelet struct {
 	Taints              []string
 	ExtraArgs           string
 	IPTablesMode        string
+	DualStackEnabled    bool
 }
 
 var _ manager.Component = (*Kubelet)(nil)
@@ -197,11 +198,13 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		return fmt.Errorf("failed to get hostname: %w", err)
 	}
 
-	ipv4, ipv6, err := getIPs(ctx, nodename)
-	if err != nil {
-		logrus.WithError(err).Infof("failed to get ip address of node %s", nodename)
-	} else if err == nil && ipv4 != nil && ipv6 != nil {
-		args["--node-ip"] = ipv4.String() + "," + ipv6.String()
+	if k.DualStackEnabled && extras["--node-ip"] == "" {
+		ipv4, ipv6, err := getIPs(ctx, nodename)
+		if err != nil {
+			logrus.WithError(err).Infof("failed to get ip address of node %s", nodename)
+		} else if err == nil && ipv4 != nil && ipv6 != nil {
+			args["--node-ip"] = ipv4.String() + "," + ipv6.String()
+		}
 	}
 
 	if runtime.GOOS == "windows" {
--- a/pkg/component/worker/kubelet.go
+++ b/pkg/component/worker/kubelet.go
@@ -132,13 +132,12 @@ func (k *Kubelet) Init(_ context.Context) error {
 	return nil
 }
 
-func getIPs(ctx context.Context, hostname string) (net.IP, net.IP, error) {
+func lookupHostname(ctx context.Context, hostname string) (ipv4 net.IP, ipv6 net.IP, _ error) {
 	ipaddrs, err := net.DefaultResolver.LookupIPAddr(ctx, hostname)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	var ipv4, ipv6 net.IP
 	for _, addr := range ipaddrs {
 		if ipv4 == nil && addr.IP.To4() != nil {
 			ipv4 = addr.IP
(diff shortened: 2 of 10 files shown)


pkg/component/worker/kubelet.go
```diff
@@ -170,15 +191,21 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		args["--node-labels"] = strings.Join(k.Labels, ",")
 	}
 
+	nodename, err := node.GetNodename("")
+	if err != nil {
+		return fmt.Errorf("can't get hostname: %v", err)
+	}
```
twz123, line 197: Any reason this has moved out of the Windows specific branch? It's only used there. Moreover, this regressed to use `%v` again. Maybe some remnant of a conflict resolution?

pkg/component/worker/kubelet.go
```diff
@@ -131,6 +131,27 @@ func (k *Kubelet) Init(_ context.Context) error {
...
 }
 
+func getIPs(hostname string) (net.IP, net.IP, error) {
+	ips, err := net.LookupIP(hostname)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var ipv4, ipv6 net.IP
+	for _, ip := range ips {
```
twz123, line 141: Maybe pass on the context here?

```suggestion
func getIPs(ctx context.Context, hostname string) (net.IP, net.IP, error) {
	ips, err := net.DefaultResolver.LookupIPAddr(ctx, hostname)
	if err != nil {
		return nil, nil, err
	}

	var ipv4, ipv6 net.IP
	for _, ip := range ips {
		ip := ip.IP
```

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +191,21 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		args["--node-labels"] = strings.Join(k.Labels, ",")
 	}
 
+	nodename, err := node.GetNodename("")
+	if err != nil {
+		return fmt.Errorf("can't get hostname: %v", err)
+	}
+
+	ipv4, ipv6, err := getIPs(nodename)
+	if err == nil && ipv4 != nil && ipv6 != nil {
```
twz123, line 200: Can we log the error?

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +191,21 @@ func (k *Kubelet) Start(ctx context.Context) error {
...
+		args["--node-ip"] = ipv4.String() + "," + ipv6.String()
```
twz123, line 201: Do we want to set this unconditionally, or maybe only when k0s is running in dual stack mode?

twz123, line 201: I reckon kube-router gets in trouble without this?

twz123 (commented):
NB: This is the error that's reported by containerd in the failed inttest:
> failed to setup network for sandbox: plugin type="bridge" name="kubernetes" failed (add): failed to set bridge addr: could not add IP address to "kube-bridge": permission denied

pkg/component/worker/kubelet.go
ncopa: I use the the nodename directly after: `getIPs(nodename)`. Since I need the nodename to get the IPs, i avoid calling `node.GetNodename` twice by moving out the windows specific branch.

pkg/component/worker/kubelet.go
ncopa: For ipv6 kube-router needs that the ipv6 address is known from the node object in kubernetes. kubelet is stupid and will not set that unless you explicitly tell it to do so. It will only autodetect the ipv6 address if you disable ipv4, so if you want kubernetes to be aware of both you need to explicitly set them.

Longterm I think this should be fixed in kubelet.
ref: https://github.com/kubernetes/kubernetes/blob/2f2c1fb8116af491ed2daabc4d4a3350bf8f6fa4/pkg/kubelet/nodestatus/setters.go#L63

pkg/component/worker/kubelet.go
ncopa: Seems like a good idea. https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/net/lookup.go;l=205

pkg/component/worker/kubelet.go
ncopa: After second thought, we probably want hide it behind k0s dualstack mode

inttest/dualstack/dualstack_test.go
```diff
@@ -107,6 +113,15 @@ func (s *DualstackSuite) SetupSuite() {
 	err = s.WaitForNodeReady(s.WorkerNode(1), client)
 	s.Require().NoError(err)
 
+	for i := 0; i < 2; i++ {
```
twz123, line 116:
```suggestion
	for i := 0; i < s.WorkerCount; i++ {
```

pkg/component/worker/kubelet.go
```diff
@@ -131,6 +132,27 @@ func (k *Kubelet) Init(_ context.Context) error {
...
+func getIPs(ctx context.Context, hostname string) (net.IP, net.IP, error) {
+	ipaddrs, err := net.DefaultResolver.LookupIPAddr(ctx, hostname)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var ipv4, ipv6 net.IP
+	for _, addr := range ipaddrs {
+		if ipv4 == nil && addr.IP.To4() != nil {
+			ipv4 = addr.IP
```
twz123, line 144:
```suggestion
			ipv4 = addr.IP.To4()
```

pkg/component/worker/kubelet.go
```diff
@@ -131,6 +132,27 @@ func (k *Kubelet) Init(_ context.Context) error {
...
+		} else if ipv6 == nil && addr.IP.To16() != nil && addr.IP.To4() == nil {
+			ipv6 = addr.IP
```
twz123, line 146:
```suggestion
			ipv6 = addr.IP.To16()
```

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +192,25 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		args["--node-labels"] = strings.Join(k.Labels, ",")
 	}
 
-	if runtime.GOOS == "windows" {
-		node, err := node.GetNodename("")
+	nodename, err := node.GetNodename("")
+	if err != nil {
+		return fmt.Errorf("failed to get hostname: %w", err)
```
twz123, line 197:
```suggestion
		return fmt.Errorf("failed to get nodename: %w", err)
```

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +192,25 @@ func (k *Kubelet) Start(ctx context.Context) error {
...
+	}
+
+	if k.DualStackEnabled {
+		ipv4, ipv6, err := getIPs(ctx, nodename)
```
twz123, line 201: Can we have a comment with the link to the upstream IP detection for reference? What happens when `--hostname-override` is passed to kubelet?

pkg/component/worker/kubelet.go
```diff
@@ -131,6 +132,27 @@ func (k *Kubelet) Init(_ context.Context) error {
 	return nil
 }
 
+func getIPs(ctx context.Context, hostname string) (net.IP, net.IP, error) {
+	ipaddrs, err := net.DefaultResolver.LookupIPAddr(ctx, hostname)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var ipv4, ipv6 net.IP
```
twz123, line 141:
```suggestion
```

pkg/component/worker/kubelet.go
```diff
@@ -131,6 +132,27 @@ func (k *Kubelet) Init(_ context.Context) error {
 	return nil
 }
 
+func getIPs(ctx context.Context, hostname string) (net.IP, net.IP, error) {
```
twz123, line 135:
```suggestion
func lookupHostname(ctx context.Context, hostname string) (ipv4 net.IP, ipv6 net.IP, _ error) {
```

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +192,25 @@ func (k *Kubelet) Start(ctx context.Context) error {
...
+	nodename, err := node.GetNodename("")
+	if err != nil {
+		return fmt.Errorf("failed to get hostname: %w", err)
+	}
+
+	if k.DualStackEnabled {
+		ipv4, ipv6, err := getIPs(ctx, nodename)
 		if err != nil {
-			return fmt.Errorf("can't get hostname: %w", err)
+			logrus.WithError(err).Infof("failed to get ip address of node %s", nodename)
```
twz123, line 203:
```suggestion
			logrus.WithError(err).Errorf("failed to lookup %q", nodename)
```

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +192,25 @@ func (k *Kubelet) Start(ctx context.Context) error {
 		args["--node-labels"] = strings.Join(k.Labels, ",")
 	}
 
-	if runtime.GOOS == "windows" {
-		node, err := node.GetNodename("")
+	nodename, err := node.GetNodename("")
```
twz123, line 195: We need to honor the `--hostname-override` flag here. Maybe move up the extra args splitting here and pass the override flag value instead of the empty string?

pkg/component/worker/kubelet.go
```diff
@@ -170,15 +192,25 @@ func (k *Kubelet) Start(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get hostname: %w", err)
+	}
+
+	if k.DualStackEnabled {
```
twz123, line 200: We should only do this if the user didn't specify `--node-ip` via extra args.

twz123 (commented):
How does this work with dynamic configuration? We'd need to restart kubelet, I guess, since we're actively changing flags based on dual stack?

twz123:
This needs attention from kubectl side, as well. I'm a bit on the fence with the IP auto-detection, because this makes it so much harder for k0sctl to do the right thing, but maybe I'm overstating the problem. WDYT @kke?

.github/workflows/go.yml
```diff
@@ -236,6 +236,8 @@ jobs:
 
       - name: k0s sysinfo
         run: |
+          ip route
+          ip -6 route
```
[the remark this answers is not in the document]
ncopa: I think it can removed. was on lyl useful for debugging the ipv6 stuff in github runners.

pkg/component/worker/kubelet.go
ncopa: What would this solve?

pkg/component/worker/kubelet.go
ncopa: Trying to wrap my head around what value this would bring?

In what situation would a valid IPv6 address not already be in a 16 byte representation?

pkg/component/worker/kubelet.go
ncopa: If `--hostname-override` is an ip address, it will be used and no lookups will be made.
 https://github.com/kubernetes/kubernetes/blob/0cc57258c3f8545c8250f0e7a1307fd01b0d283d/pkg/kubelet/nodestatus/setters.go#L200

pkg/component/worker/kubelet.go
ncopa: Kubelet uses `can't get ip address of node %s. error: %v`.
https://github.com/kubernetes/kubernetes/blob/0cc57258c3f8545c8250f0e7a1307fd01b0d283d/pkg/kubelet/nodestatus/setters.go#L231

And I think calico will work just fine in this situation, which is why I made it info only.

ncopa:
> This needs attention from kubectl side, as well. I'm a bit on the fence with the IP auto-detection, because this makes it so much harder for k0sctl to do the right thing, but maybe I'm overstating the problem. WDYT @kke?

Would you prefer that we required to specify the ipv4 and ipv6 node address in k0s config? How would that work for dynamic addresses? There is no guarantee that the node gets same ipv4 address after a reboot.

pkg/component/worker/kubelet.go
twz123: It would ensure that ipv4 is actually in its four byte form, so that `len(ipv4) == IPv4len`. Otherwise, it could be in its 16 byte form, i.e. `len(ipv4) == IPv6len`. This is a nit, but it feels a bit saner to have a v4 IP returned as 4 bytes.

pkg/component/worker/kubelet.go
twz123: Yeah, this converts a v4 IP to a v6 IP, but as an earlier branch already prevents v4 IPs to reach this point, it's a no-op, indeed.

Thinking if lines 143 and following could better be written as

```go
		if v4 := addr.IP.To4(); v4 != nil {
			if ipv4 == nil {
				ipv4 = v4
			}
		} else if ipv6 == nil {
			if v6 := addr.IP.To16(); v6 != nil {
				ipv6 = v6
			}
		}
```

twz123:
> Would you prefer that we required to specify the ipv4 and ipv6 node address in k0s config?

I think there's two separate things which bother me:

First: There's this schism between single-stack and dual-stack. There is IP auto-detection in kubelet for v4 and v6, but not for both at the same time (which would be reasonable in dual-stack mode, IMO). So k0s has to do it, and is responsible for getting it right,  even it doesn't really care about the detected IPs itself. Although I would rather see someone else (i.e. upstream) maintain this :upside_down_face:, there is still the option to set the IPs via `--node-ip` as kubelet extra args if users want to set the IPs (or need to set them since the auto-detection doesn't do the right thing for them). I think that's a fair compromise.

Second, and this is something that needs clarification and probably a fix on the k0sctl or on both k0sctl and k0s side: k0sctl sets `--node-ip` based on `privateInterface`. This also needs to get dual-stack support. So people will _need_ to specify exactly zero or two IPs in `privateInterface` if they are using a dual-stack cluster. I don't think k0sctl can accept more than one IP in this setting, so this is a blocker for use cases like #4694. One way to fix this on k0s's end might be to look at the user-provided `--node-ip` value and, if it doesn't have IP addresses for v4 _and_ v6, add the missing ones via auto-detection.

> How would that work for dynamic addresses? There is no guarantee that the node gets same ipv4 address after a reboot.

Is that even a thing, workers with dynamic addresses? Maybe. Would require a k0s restart on IP address changes, though. Not sure about the running containers. They probably need to be restarted, as well.

ncopa:
> I would rather see someone else (i.e. upstream) maintain this 🙃

I agree. This is something that should be fixed in kubelet. But for now we will have to work around it in k0s.

> k0sctl sets `--node-ip` based on `privateInterface`. This also needs to get dual-stack support. So people will _need_ to specify exactly zero or two IPs in `privateInterface` if they are using a dual-stack cluster.

Yes, and this is due to upstream kubelet behavior.

> I don't think k0sctl can accept more than one IP in this setting, so this is a blocker for use cases like #4694. One way to fix this on k0s's end might be to look at the user-provided `--node-ip` value and, if it doesn't have IP addresses for v4 _and_ v6, add the missing ones via auto-detection.

I'd rather fix `kubelet`

> Is that even a thing, workers with dynamic addresses? Maybe. Would require a k0s restart on IP address changes, though. Not sure about the running containers. They probably need to be restarted, as well.

I would expect most workers use DHCP for ipv4. Not sure if we care about address changes on dhcp lease expiry runtime. But I think that address change on reboot is something to be expected. (which probably means we shouldn't set the address from k0sctl?)

pkg/component/worker/kubelet.go
ncopa: I think we only care what the address family is. How the buffers are allocated under the hood is completely irrelevant.

twz123 (approved):

Issue #781, opened 2021-03-17
Improve documentation for Zsh autocompletion


trawler:
**What happened?**
The existing documentation on how to enable zsh auto-completion is incorrect/missing information.

**Expected behavior**
In this page: https://docs.k0sproject.io/v0.11.0/install/#zsh

The following should be removed:
```
# To load completions for each session, execute once:
$ k0s completion zsh > "${fpath[1]}/_k0s"
```
and replaced by
```
$ echo "source <(k0s completion zsh)" >> ~/.zshrc
```

kke:
That would run k0s on every shell invocation, not a best practice.

The usual way is something like:

```shell
k0s completion zsh > /usr/local/share/zsh/site-functions/_k0s
```

kke:
The current example of doing:

```shell
k0s completion zsh > "${fpath[1]}/_k0s"
```

may end up writing the file somewhere weird if you're using some zsh plugin framework, for example on my system it would go to `/Users/kimmo/Library/Caches/antibody/https-COLON--SLASH--SLASH-github.com-SLASH-robbyrussell-SLASH-oh-my-zsh/functions`.

PR #3961, opened 2024-01-19
Overhaul shell completion docs

twz123:
## Description

Fixes:
* #781

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/shell-completion.md
+++ b/docs/shell-completion.md
@@ -1,45 +1,108 @@
 # Enabling Shell Completion
 
-Generate the k0s completion script using the `k0s completion <shell_name>` command, for Bash, Zsh, fish, or PowerShell.
+## Introduction
 
-Sourcing the completion script in your shell enables k0s autocompletion.
+Shell completion enhances the user experience by providing auto-completion for
+commands in the terminal. K0s supports shell completion for the following
+shells:
 
-## Bash
+- `bash`, [GNU Bash](https://www.gnu.org/software/bash/)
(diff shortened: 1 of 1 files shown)


juanluisvaladas (approved):

docs/shell-completion.md
```diff
@@ -1,45 +1,108 @@
...
+One-shot usage: `source <(k0s completion bash)`.
+
+This is a recipe to load completions for each new shell. Adjust to your personal
+needs:
+
+```bash
+mkdir ~/.bash_completion.d
+k0s completion bash >~/.bash_completion.d/k0s
+
+cat <<'EOF' >~/.bashrc
```
kke, line 31: how about something like this?

```sh
grep -q .bash_completion.d .bashrc || cat <<'EOF'
```

twz123: I tried to find some references about the idiomatic way of including user-managed bash shell completions, but I didn't really find anything. The `bash_completion.d` stuff seems to be used, but there doesn't seem to be "the way of doing it". That's why I tried to give an example with the caveat "adjust to your personal needs". Checking if the folder has already been set up in .bashrc or some other place is not a piece of cake. That's why I opted for not doing it.

If you think checking would be better, then I'd probably change this to some descriptive block where to put the snippet, similar to the examples for zsh.

 WDYT?

PR #4029, opened 2024-02-06
Wrap errors correctly

twz123:
## Description

Replace the %v format specifier in calls to fmt.Errorf() with %w whenever an error argument is used. Use errors.Join() in some cases where multiple errors were collected by their error messages. Replace %v with %s in error format strings whenever it's used for string-like arguments. Remove unnecessary fmt.Errorf() calls when logging errors.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -157,7 +157,7 @@ func (c *command) start(ctx context.Context) error {
 	}
 	// let's make sure run-dir exists
 	if err := dir.Init(c.K0sVars.RunDir, constant.RunDirMode); err != nil {
-		return fmt.Errorf("failed to initialize dir: %v", err)
+		return fmt.Errorf("failed to initialize dir: %w", err)
 	}
 
 	rtc, err := config.NewRuntimeConfig(c.K0sVars)
@@ -180,7 +180,7 @@ func (c *command) start(ctx context.Context) error {
 	if c.TokenArg != "" && c.needToJoin() {
 		joinClient, err = joinController(ctx, c.TokenArg, c.K0sVars.CertRootDir)
 		if err != nil {
-			return fmt.Errorf("failed to join controller: %v", err)
+			return fmt.Errorf("failed to join controller: %w", err)
 		}
 	}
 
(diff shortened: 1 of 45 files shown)


cmd/backup/backup.go
```diff
@@ -72,11 +72,11 @@ func (c *command) backup(savePath string, out io.Writer) error {
 	}
 
 	if savePath != "-" && !dir.IsDirectory(savePath) {
-		return fmt.Errorf("the save-path directory (%v) does not exist", savePath)
+		return fmt.Errorf("the save-path directory (%s) does not exist", savePath)
```
kke, line 75: `%s` should be fine for strings or stringers.

kke: of course the `String()` could panic if it's a `fmt.Stringer` so `%v` is safer.

For paths and things like that `%q` might be better to get a quoted string.

cmd/controller/controller.go
```diff
@@ -157,7 +157,7 @@ func (c *command) start(ctx context.Context) error {
 	}
 	// let's make sure run-dir exists
 	if err := dir.Init(c.K0sVars.RunDir, constant.RunDirMode); err != nil {
-		return fmt.Errorf("failed to initialize dir: %v", err)
+		return fmt.Errorf("failed to initialize dir: %w", err)
```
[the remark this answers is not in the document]
twz123: I'd argue, in your example, the error message should be `failed to start controller: failed to initialize dir: path can not be empty`. Seems that errors are wrapped in places where it's unnecessary or superfluous. (Stating what dir can't be created would be helpful, though.)

As for the pattern "failed to foo: failed to bar: baz", I personally find it quite readable. Something like "while doing foo: while barring baz: boom" would be fine by me, too. Do you have something specific in mind?

cmd/backup/backup.go
twz123: > of course the String() could panic if it's a fmt.Stringer so %v is safer.

That's true. If String() is expected to panic, I agree that %v is the right choice. As for unexpected panics: Those are handled within the printf implementation anyways and won't reach the caller.

> For paths and things like that %q might be better to get a quoted string.

Also agree! Didn't want to change the error/log output too much in this PR, though. This might get into another dedicated PR.

kke (approved):

PR #4052, opened 2024-02-13
Refactor for UID/GID handling

juanluisvaladas:
## Description
The PR has two commits with two directly related small refactors:

1- Log a warning when using uid 0 because the gid cannot be found for a specific user.
2- Remove references to the gid in all components. In many components we were not defining a gid in the code but we had the field in the struct and was being used. Effectively it was using gid 0. The chown calls now use gid -1, which means don't change the permissions to match the behavior of other components such as konnectivity.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/file/file.go
+++ b/internal/pkg/file/file.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 
 	"github.com/k0sproject/k0s/internal/pkg/users"
+	"github.com/sirupsen/logrus"
 
 	"go.uber.org/multierr"
 )
@@ -42,8 +43,12 @@ func Exists(fileName string) bool {
 // Chown changes file/dir mode
 func Chown(file, owner string, permissions os.FileMode) error {
 	// Chown the file properly for the owner
-	uid, _ := users.GetUID(owner)
-	err := os.Chown(file, uid, -1)
+	uid, err := users.GetUID(owner)
+	if err != nil {
+		uid = 0
+		logrus.WithError(err).Warnf("failed to get uid for %s, using uid 0", owner)
+	}
+	err = os.Chown(file, uid, -1)
 	if err != nil && os.Geteuid() == 0 {
 		return err
 	}
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -24,7 +24,8 @@ import (
 	"strings"
 )
 
-// GetUID returns uid of given username and logs a warning if its missing
+// GetUID returns uid of given username. If the user cannot be found, it
+// returns an error and UID 0.
 func GetUID(name string) (int, error) {
 	entry, err := user.Lookup(name)
 	if err == nil {
--- a/pkg/certificate/manager.go
+++ b/pkg/certificate/manager.go
@@ -105,7 +105,11 @@ func (m *Manager) EnsureCertificate(certReq Request, ownerName string) (Certific
 	keyFile := filepath.Join(m.K0sVars.CertRootDir, fmt.Sprintf("%s.key", certReq.Name))
 	certFile := filepath.Join(m.K0sVars.CertRootDir, fmt.Sprintf("%s.crt", certReq.Name))
 
-	uid, _ := users.GetUID(ownerName)
+	uid, err := users.GetUID(ownerName)
+	if err != nil {
+		uid = 0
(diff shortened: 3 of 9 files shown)


internal/pkg/file/file.go
```diff
@@ -42,8 +43,11 @@ func Exists(fileName string) bool {
 // Chown changes file/dir mode
 func Chown(file, owner string, permissions os.FileMode) error {
 	// Chown the file properly for the owner
-	uid, _ := users.GetUID(owner)
-	err := os.Chown(file, uid, -1)
+	uid, err := users.GetUID(owner)
+	if err != nil {
+		logrus.Warnf("failed to get uid for %s, using 0: %v", owner, err)
```
twz123, line 48: Can you use logrus.WithError(err).Warnf(...)?

twz123, line 48: Good catch. I wonder if it would be better to error out here? The desire to log a warning in a core utility function somehow indicates to me that it's probably a real error condition. You write in your commit message that it's correct to fallback to UID 0. It's not immediately obvious to me why that is. An alternative approach might be to return some well known error that callers can check against and implement their own fallback handling, if they desire to do so.

I'd ask for at least two things:

1. Add an explanatory comment as to why this is desired, maybe directly in the functions godoc string?
2. Explicitly hardcode the UID 0 in the subsequent call to chown, or explicitly set the uid variable to 0 here, instead of implicitly relying on the behavior of GetUID to return 0 on error.

internal/pkg/users/users.go
```diff
@@ -24,7 +24,7 @@ import (
 	"strings"
 )
 
-// GetUID returns uid of given username and logs a warning if its missing
+// GetUID returns uid of given username and returns a warning if its missing
```
twz123, line 27: Does this refer to the second error return value? Maybe better to write "returns an error" instead?

pkg/certificate/manager.go
```diff
@@ -105,7 +105,10 @@ func (m *Manager) EnsureCertificate(certReq Request, ownerName string) (Certific
 	keyFile := filepath.Join(m.K0sVars.CertRootDir, fmt.Sprintf("%s.key", certReq.Name))
 	certFile := filepath.Join(m.K0sVars.CertRootDir, fmt.Sprintf("%s.crt", certReq.Name))
 
-	uid, _ := users.GetUID(ownerName)
+	uid, err := users.GetUID(ownerName)
+	if err != nil {
+		logrus.Warnf("failed to get uid for %s, using uid 0: %v", ownerName, err)
```
twz123, line 110: Can you use logrus.WithError(err).Warnf(...)?

twz123, line 110: Can you explicitly set uid to 0 here instead of relying on the behavior of GetUID here?

twz123, line 110: There are some os.Chown calls down the line in this function which ignore errors, as well. Do we want to log those, too?

internal/pkg/users/users.go
juanluisvaladas: You're right, it's not a warning, it's an error.

pkg/certificate/manager.go
juanluisvaladas: That's a good point. I've set these to Warning, I don't feel very strongly though, so if you want me to make these an error that's fine.

I'm not quite sure if it makes more sense to have these as warnings or errors. Anyway there it is.

internal/pkg/file/file.go
juanluisvaladas: Yep, that's better.

I found over a thousand instances of us logging errors without using WithError (mind the regexes may not be 100% correct), but since logrus is in maintenance mode we probablty don't want to look into this as logrus is in maintenance mode...

juanluisvaladas:
Related to  #4062

juanluisvaladas:
@twz123 I believe all your concerns should be addressed now. Can you please check the commit "Log when a call to users.GetUID fails" ? The other commit has no modifications.

There are quite a few calls to `logrus.Warning(fmt.Errorf(` or similar around that area but these are lines I haven't touched. I'm thinking of editing those too but the problem is there are A LOT of occurrences and eventually we end up modifying unrelated stuff.

PR #4060, opened 2024-02-15
Fix linter errors in macOS

juanluisvaladas:
## Description

The linter on macOS complains about these two errors:

util_test.go uses syscall.EBADFD which does not exist in darwin. This causes some linter issues in my editor. By giving the build tag linux we shut up the linter without making any actual change.

utils_linux.go was renamed to utils_unix.go so that macOS has the function winExecute and the linter doesn't complain about its calls.


juanluisvaladas:
Nevermind, this breaks windows builds

Issue #4047, opened 2024-02-09
Updating the in-cluster autopilot config while a helm chart is being created results in the chart being created twice


laverya:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.2.0-1019-azure #19~22.04.1-Ubuntu SMP Wed Jan 10 22:57:03 UTC 2024 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.29.1+k0s.1

### Sysinfo
<details>
  <summary>k0s sysinfo</summary>
  Machine ID: "ae13f7395464d07eda6ba68f7e94b56d64b68c4b3527b66ccb7ab4868f9c13b1" (from machine) (pass)
  Total memory: 15.6 GiB (pass)
  Disk space available for /var/lib/k0s: 24.3 GiB (pass)
  Name resolution: localhost: [::1 127.0.0.1] (pass)
  Operating system: Linux (pass)
    Linux kernel release: 6.2.0-1019-azure (pass)
    Max. file descriptors per process: current: 1048576 / max: 1048576 (pass)
    AppArmor: active (pass)
    Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
    Executable in PATH: mount: /usr/bin/mount (pass)
    Executable in PATH: umount: /usr/bin/umount (pass)
    /proc file system: mounted (0x9fa0) (pass)
    Control Groups: version 2 (pass)
      cgroup controller "cpu": available (is a listed root controller) (pass)
      cgroup controller "cpuacct": available (via cpu in version 2) (pass)
[...]

PR #4064, opened 2024-02-15
bug: retry saving helm chart object status

ricardomaraschini:
## Description

if we fail to save the status, specially after installing the chart, we will try to install the chart again during the next reconcile (its Status.Status.ReleaseName will be empty). this pr attempts to mitigate this in two ways:

- before updating fetch the newest version of the chart.
- retry if it is still on conflict.

this may be overkill as it most likely won't fail after we fetch an updated version of the chart but given the importance of this operation we better be safe than sorry.

when this bug this solves happens this is what is reported in the chart status:

```
can't install loadedChart ```chart-name```: cannot re-use a name that is
still in use
```

Mitigates #4047 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/extensions_controller.go
+++ b/pkg/component/controller/extensions_controller.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


juanluisvaladas (approved):

juanluisvaladas:
Thanks for the contribution @ricardomaraschini and sorry for taking this long to review this.

PR #3633, opened 2023-10-23
Fix helm race condition

juanluisvaladas:
## Description

The PR includes three commits.

The first commit removes the field concurrencyLevel in helm controller. This field isn't honored at all and we want to remove concurrent reconciliation anyway so entirely remove the field.

The second commit makes extensions controller thread-safe. Helm is unable to reconcile multiple charts at the same time because /var/lib/k0s/helmhome/cache/ is shared between charts and helm cache was not designed to be safe to be used concurrently. To fix it we document some functions of the Commands struct are not thread safe (they couldn't have a local lock unless it was shared across all instances) and modify the controller not concurrent.

The third commit implements timeouts, we were already passing a context but this wasn't used at all. This is to prevent a reconciler being stuck forever 

Requirement to fix #3282 and #3433

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #4106, opened 2024-02-21
left-pad helm chart filenames with 0s to fix sorting

laverya:
## Description

Fixes # (issue)

Currently, helm chart apply order is determined by sorting the helm chart manifest filenames. The order is included as the first element in the filename, resulting in names like `1_helm_extension_chartname.yaml`.

Manifests are fetched [here](https://github.com/k0sproject/k0s/blob/v1.29.1%2Bk0s.1/pkg/applier/applier.go#L37-L39), which uses [filepath.Glob](https://github.com/golang/go/blob/go1.22.0/src/path/filepath/match.go#L347), which calls sort on the results.

Unfortunately, this breaks at 10 - `10_helm_extension_anothername.yaml` is alphabetically before `1_helm_extension_chartname.yaml`.

This PR adds a fmt directive to left-pad these integers to 9 digits, which allows sorting to work properly - see https://go.dev/play/p/THZg3TD6eHI.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s/v1beta1/extensions.go
+++ b/pkg/apis/k0s/v1beta1/extensions.go
@@ -102,7 +102,7 @@ type Chart struct {
 
 // ManifestFileName returns filename to use for the crd manifest
 func (c Chart) ManifestFileName() string {
-	return fmt.Sprintf("%d_helm_extension_%s.yaml", c.Order, c.Name)
+	return fmt.Sprintf("%09d_helm_extension_%s.yaml", c.Order, c.Name)
 }
 
 // Validate performs validation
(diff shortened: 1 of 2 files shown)


laverya:
I'm trying to work out how to add an auto test - is there a way to specify a config to be applied in CI? If so, I can just have two charts order `1` and `10` where `1` provides resources used by `10` 😄

twz123:
Thanks for looking into this. Can you provide an example of what you'd like to fix, so that we can cover this in an integration test? K0s does offer the order field on Helm extensions, but this may not work quite as expected. The file names control the order in which chart resources are applied to the cluster, but the actual installation happens when those resources are reconciled. Not sure if the order is guaranteed in any way on that end.

Previous related PRs:
* #2974
* #3633

laverya:
Let me get an example spec in a moment.

laverya:
(we had initially discovered this due to a demo helm chart with incorrect configuration that was applied before the ones we actually required, and never completed, blocking installation of the other charts until the timeout, and so I need to create a better example)

PR #4151, opened 2024-03-07
[Docs] Fix k0s in docker documentation

### LINKED ISSUE #3712: docker vs docker compose differes in docs
in https://docs.k0sproject.io/v1.28.3+k0s.0/k0s-in-docker/

docker instructions are missing tmpfs for `/run` and `/var/run` in docker-compose  https://docs.k0sproject.io/v1.28.3+k0s.0/k0s-in-docker/#use-docker-compose-alternative

are they needed or not? seems to work fine without?

juanluisvaladas:
This commit changes a few things

1- By default make k0s controller also a worker, IMO it's more logical to do it the way the documentation is written.

2- Remove the mount to /sys/fs/cgroup as it's not really required. I assume this was never required and comes from the 4.3.0  release notes. It's probably still required for systemd but we don't run systemd.

3- Remove unnecessary tmpfs in /run/ and /var/run in docker-compose. They are not wrong but not required either and this is consistent with what we do in the regular docker run.

4- Always make cgroup: host. This works with both cgroups v1 and v2 but not having this breaks cgroups v2.

## Description

Fixes #3712

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


twz123 (approved):
> 2- Remove the mount to /sys/fs/cgroup as it's not really required. I assume this was never required and comes from the 4.3.0 release notes. It's probably still required for systemd but we don't run systemd.

A writable `/sys/fs/cgroup` _is_ a requirement, for sure! It seems that --cgroupns=host also means that `/sys/fs/cgroup` is mounted verbatim from the host, so the additional mount is redundant.

> 4- Always make cgroup: host. This works with both cgroups v1 and v2 but not having this breaks cgroups v2.

Note that the cgroupns=host thing is a hack to get around the lack of a proper cgroupns setup in k0s's Docker image entrypoint, and breaks container isolation quite a bit, as the entire host's cgroup tree shows up unprotected in the Docker container. I agree that most systems run cgroupv2 these days, and thus most people will need to enable it, but it's really a bit wonky.

Issue #4181, opened 2024-03-17
Keepalived support


juanluisvaladas:
We're going to merge keepalived support soon. This is a pretty large change and the review will be quite difficult. For easier review and merge I'm going to split keepalived in 3 different pull requests, 2 of them with several commits:

~~- [ ] - Add a keepalived binary to k0s #4182~~ This will be added to the 2nd PR
- [x] Add keepalived, and virtual ip failover support (API, implementation, tests and docs) #4190
- [ ] Add load balancing support  (API, implementation, tests and docs) #4344 
- [ ] Add unicast support for VRRP.

PR #4190, opened 2024-03-18
Implement VIPs using Keepalived vrrp_instances 

juanluisvaladas:
## Description

Currently this is a draft PR. I'm doing a final refactor to the controller and have to review the docs, but in the meantime I want the linter CI and everything to pass.

This is part of #4181.
This PR is split into 5 commits for easier review:
1- Add keepalived to the embedded bins build
2- Create the API objects for ExternalLoadBalancer
3- Implement the controller
4- Implement an integration test that tests the functionality
5- Add docs.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


pkg/apis/k0s/v1beta1/elb.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// Defaults are keepalived's defaults.
+const defaultVirtualRouterID = 51
+const defaultAdvertInterval = 1
+const defaultInterface = "eth0"
+
+// ExternalLoadBalancingSpec defines the configuration options related to k0s's
+// keepalived feature.
+type ExternalLoadBalancingSpec struct {
```
juanluisvaladas, line 32: I'm not very happy with the ExternalLB/ExternalLoadBalancing names. It hints that it's running outside. Also it's contraction to ELB can be confused for AWS ELB which is something that I, again, dislike.

In the first implementation I called it keepalived, which works really well, however I really like that NLLB API allows of not calling it keepalived in case the software used for this changes.

I've considered  ControlPlaneLoadBalancing or CPLB, I'd like to hear opinions on this.

pkg/apis/k0s/v1beta1/elb.go
```diff
@@ -0,0 +1,116 @@
...
+	VRRPInstances VRRPInstances `json:"vrrpInstances,omitempty"`
```
juanluisvaladas, line 33: I really like this API for configuring VRRP instances, if you check how VRRP is done outside keepalived, this API would  without any change to other VRRP software [such as cisco IOS](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/addr_serv/configuration/guide/ic40crs1book_chapter10.html).  However, if we need to switch later to something that does not use VRRP?

I don't see a really good solution for this. I considered using `leaderElection` instead of `VRRPInstances`, however we **REALLY** need to be able to define the VRID and Authpass.

I would love a better name if someone can think of something.

cmd/controller/controller.go
```diff
@@ -224,6 +224,14 @@ func (c *command) start(ctx context.Context) error {
 		nodeComponents.Add(ctx, controllerLeaseCounter)
 	}
 
+	enableExternalLB := !c.SingleNode && !slices.Contains(c.DisableComponents, constant.ExternalLBComponentName)
```
juanluisvaladas, line 227: This is added before apiserver as the VIP may be a requirement for the apiserver depending on the configuration (for instance in the inttest). Hence I don't think it's a good idea to allow dynamic configuration changes for this.

There is no technical reason why we cannot, however I think we shouldn't.

pkg/apis/k0s/v1beta1/elb.go
juanluisvaladas: Renamed it to ControlPlaneLoadBalancing and I like it much more but I'm still not 100% convinced

docs/configuration.md
```diff
@@ -300,6 +300,29 @@ node-local load balancing.
...
+
+##### `spec.network.controlPlaneLoadBalancing.VRRPInstances`
+
+Configuration options required for using VRRP to configure VIPs in control plane load balancing.
+
+| Element           | Description                                                                                        |
+| ----------------- | -------------------------------------------------------------------------------------------------- |
+| `name`            | The name of the VRRP instance. If omitted it generates a predictive name shared accross all nodes. |
+| `virtualIPs`      | A list of the CIDRs handled by the VRRP instance.                                                  |
+| `interface`       | The interface used by each VRRPInstance. Default: `eth0`.                                          |
```
jnummelin, line 321: Couldn't we somehow try to autodetect the interface?

docs/configuration.md
```diff
@@ -300,6 +300,29 @@ node-local load balancing.
 | `apiServerBindPort`          | Port number on which to bind the Envoy load balancer for the Kubernetes API server to on a worker's loopback interface.  Default: `7443`. |
 | `konnectivityServerBindPort` | Port number on which to bind the Envoy load balancer for the konnectivity server to on a worker's loopback interface. Default: `7132`.    |
 
+##### `spec.network.controlPlaneLoadBalancing`
+
+Configuration options related to k0s's [control plane load balancing] feature
+
+| Element         | Description                                                                               |
+| --------------- | ----------------------------------------------------------------------------------------- |
+| `vrrpInstances` | Configuration options related to the VRRPInstances. This is used to configure virtual IPs |
```
jnummelin, line 309: Based on this explanation and the `VRRPInstances` below, I'm not really sure how I'd use this. E.g. is `vrrpInstances` an array as the name hints?

Maybe some simple example would help users.

docs/cplb.md
```diff
@@ -0,0 +1,310 @@
...
+
+In order to use control plane load balancing, the cluster needs to comply with the
+following:
+
+* K0s isn't running as a [single node](k0s-single-node.md), i.e. it isn't
+  started using the `--single` flag.
+* The cluster should have multiple controller nodes. Technically CPLBalso works
+  with a single controller node, but is only useful in conjunction with a highly
+  available control plane.
+* Unique virtualRotuerID and authPass for each cluster in the same broadcast domain.
```
jnummelin, line 27:
```suggestion
* Unique virtualRouterID and authPass for each cluster in the same broadcast domain.
```

docs/networking.md
```diff
@@ -54,6 +54,7 @@ One goal of k0s is to allow for the deployment of an isolated control plane, whi
 | TCP      | 10250 | kubelet        | controller, worker => host `*` | Authenticated kubelet API for the controller node `kube-apiserver` (and `heapster`/`metrics-server` addons) using TLS client certs
 | TCP      | 9443  | k0s-api        | controller <-> controller      | k0s controller join API, TLS with token auth
 | TCP      | 8132  | konnectivity   | worker <-> controller          | Konnectivity is used as "reverse" tunnel between kube-apiserver and worker kubelets
+| TCP      | 112   | keepalived     | controller <-> controller      | Only required for control plane load balancing vrrpInstances for ip address 224.0.0.18.
```
jnummelin, line 57: I don't get the reference to a specific IP address?

inttest/cplb/cplb_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Verify that only one controller has the VIP in eth0
+	count := 0
+	for idx := 0; idx < s.BootlooseSuite.ControllerCount; idx++ {
+		if s.hasVIP(ctx, s.ControllerNode(idx), ipAddress) {
+			count++
+		}
+	}
+	s.Require().Equal(1, count, "Expected only one controller to have the VIP")
+
```
jnummelin, line 98: We don't seem to actually test LB functionality and a failover case. Not really sure how to do that though

pkg/component/controller/cplb_unix.go
```diff
@@ -0,0 +1,328 @@
...
+func (k *Keepalived) Init(_ context.Context) error {
+	if k.Config == nil {
+		return nil
+	}
+	k.log = logrus.WithField("component", "CPLB")
+
+	var err error
+	k.uid, err = users.GetUID(constant.KeepalivedUser)
+	if err != nil {
+		k.log.Errorf("running keepalived as root: %v", err)
```
jnummelin, line 62: I don't think this is an error per se, maybe a `Warn` level?

pkg/component/controller/cplb_unix.go
```diff
@@ -0,0 +1,328 @@
...
+			k.log.Warnf("failed to get link by name %s. Attempting to create it anyway: %v", linkName, err)
+		}
+		return k.createDummyInterface(linkName)
+	}
+
+	if _, isDummy := link.(*netlink.Dummy); isDummy {
+		return nil
+	}
+
+	if err = netlink.LinkDel(link); err != nil {
```
jnummelin, line 175: Is delete done here too in case of user changes the address and we need to re-create the dummy interface?

pkg/component/controller/cplb_unix.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+const dummyLinkName = "dummyvip0"
+const keepalivedConfigTemplate = `
+{{ range .VRRPInstances }}
+vrrp_instance {{ .Name }} {
+    state BACKUP
+    # Make sure the interface is aligned with your server's network interface
+    interface {{ .Interface }}
+    # The virtual router ID must be unique to each VRRP instance that you define
```
jnummelin, line 310: Hmm, not sure what `VRRP instance` means in this context, a cluster? If it is cluster, maybe use that term here?

pkg/component/controller/cplb_unix.go
```diff
@@ -0,0 +1,328 @@
...
+    virtual_router_id {{ .VirtualRouterID }}
+    # All servers have the same priority so that when a new one comes up we don't
+    # do a failover
+    priority 200
+#   advertisement interval, 1 second
```
jnummelin, line 315:
```suggestion
#   advertisement interval, 1 second by default
```

pkg/component/controller/cplb_windows.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Keepalived doesn't work on windows, so we cannot implement it at all.
+// Just create the interface so that the CI doesn't complain.
+type Keepalived struct {
+	K0sVars *config.CfgVars
+	Config  *k0sAPI.ControlPlaneLoadBalancingSpec
+}
+
+func (k *Keepalived) Init(_ context.Context) error {
+	return nil
```
jnummelin, line 34: Shouldn't we in this case return errors in all cases?

jnummelin (changes requested):
Minor nits and some questions

docs/configuration.md
juanluisvaladas: That makes sense, I'll use the interface used for the default gateway

docs/networking.md
juanluisvaladas: It's a multicast IP address. VRRP defines it in [RFC 3768](https://datatracker.ietf.org/doc/html/rfc3768#section-5.2.2):

[5.2.2](https://datatracker.ietf.org/doc/html/rfc3768#section-5.2.2).  Destination Address

   The IP multicast address as assigned by the IANA for VRRP is:

      224.0.0.18

   This is a link local scope multicast address.  Routers MUST NOT
   forward a datagram with this destination address regardless of its
   TTL.

I will add a note in the documentation.

docs/configuration.md
juanluisvaladas: I added some more text to make it clearer, I hope this is better.

inttest/cplb/cplb_test.go
juanluisvaladas: There is no LB in this PR, just the virtual IP.

That will come in the second PR once this is merged. Adding LB is almost the same size so it makes things a bit saner to merge.

This tests that the virtual IP is working in two ways:
1- The node joins the virtual IP, so it has to be working
2- Checks that every controller has the dummy interface
3- Checks that only one controller has the VIP in the real interface.

pkg/component/controller/cplb_unix.go
juanluisvaladas: It should. Fixed.

pkg/component/controller/cplb_unix.go
juanluisvaladas: Fixed

pkg/component/controller/cplb_windows.go
juanluisvaladas: Yes, fixed.

pkg/component/controller/cplb_unix.go
juanluisvaladas: I think this is a great point but I'd say the right place to explain this is in the docs. I added a new section called `VRRP Instances` in `docs/cplb.md` which should address this. Let me know what you think about it.

pkg/component/controller/cplb_unix.go
juanluisvaladas: This would happen if the interface exists but it's not of type dummy, I don't think it can actually happen but since it's there I'd say it doesn't do any harm. I added a comment explaining why it's there.

inttest/cplb/cplb_test.go
jnummelin: Oh, right. Makes sense.

PR #4213, opened 2024-03-28
Manage Etcd peers using custom resources

jnummelin:
## Description

One of the challenges for automation (Terraform, cluster API, k0sctl) when removing nodes is that when removing a controller node one has to also remove the Etcd peer entry to it in Etcd cluster.

In this PR there's few things going on:
- Each controller creates their own `EtcdMember` object with memberID and peer address fields
- Each controller watches the changes on `EtcdMember` objects
- When the leader controller sees `EtcdMember` object being marked for leaving the etcd cluster it will remove the corresponding peer from Etcd cluster.
- The peer reconciliation status is added to the object

One thing I'm not yet sure is if the EtcdREconciler component should try to detect the conditions for peer removal or just rely on Etcd failing the removal. Like in case where you have already lost majority and are trying to remove a healthy member?

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:


static/manifests/etcd/CustomResourceDefinition/etcd.k0sproject.io_etcdmembers.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - jsonPath: .peerAddress
+      name: PeerAddress
+      type: string
+    - jsonPath: .memberID
+      name: MemberID
+      type: string
+    name: v1beta1
+    schema:
+      openAPIV3Schema:
+        description: EtcdMember dexcribs the nodes etcd membership status
```
kke, line 27: dexcribs

go.mod
twz123, line 1: Those are all bumps of indirect dependencies. Are they required?

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+				return 0, err
+			}).
+			IncludingDeletions().
+			Until(ctx, func(member *etcdv1beta1.EtcdMember) (bool, error) {
+				e.reconcileMember(ctx, member)
+				// Never stop the watch
+				return false, nil
+			})
+		if err != nil {
+			log.WithError(err).Warn("watch terminated")
```
twz123, line 118: I reckon this will log an error, even on coordinated shutdown.

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+			}
+
+			return false, nil
+		})
+
+}
+
+func (e *EtcdMemberReconciler) createMemberObject() error {
+	log := logrus.WithFields(logrus.Fields{"component": "etcdMemberReconciler", "phase": "createMemberObject"})
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
twz123, line 161: Any objections to let `createMemberObject` accept a parent context?

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+	defer cancel()
+	// find the member ID for this node
+	etcdClient, err := etcd.NewClient(e.k0sVars.CertRootDir, e.k0sVars.EtcdCertDir, e.etcdConfig)
+	if err != nil {
+		return err
+	}
+
+	peerURL := fmt.Sprintf("https://%s:2380", e.etcdConfig.PeerAddress)
```
twz123, line 169: We should probably use `net.JoinHostPort` here.

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+
+	memberID, err := etcdClient.GetPeerIDByAddress(ctx, peerURL)
+	if err != nil {
+		return err
+	}
+
+	// Convert the memberID to hex string
+	memberIDStr := fmt.Sprintf("%x", memberID)
+
+	name, err := os.Hostname()
```
twz123, line 179: Is there a way to get the name from etcd itself?

go.mod
jnummelin: Not really AFAIK. Side-product of doing non-trivial rebase for go.mod and then running `go mod tidy`. Should I get rid of those?

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+	var err error
+
+	log.Debugf("reconciling EtcdMember: %+v", member)
+
+	if !member.Leave {
+		log.Debug("member not marked for leave, no action needed")
+		return
+	}
+
+	if member.Status.JoinStatus == etcdv1beta1.JoinStatusLeft {
```
twz123, line 258: Question is if this should be checking against etcd, just in case the info in the API server is bogus or stale.

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,316 @@
...
+		log.Debug("member already left, no action needed")
+		return
+	}
+
+	// TODO Do we need to verify the conditions for removing peer? Or rely on Etcd client to error out if a peer cannot be removed?
+	// TODO I need to split up the logic a bit so the error handling is easier to understand/handle
+	defer func() {
```
twz123, line 265: I wonder if it would be clearer to have two functions. One calling the other, and then creating events depending on the outcome. Might be easier to follow mentally than by using a defer func.

go.mod
twz123: I'd prefer it, yes.

pkg/component/controller/etcd_member_reconciler.go
jnummelin: You mean something like this:
```
if err := doSomething(); err != nil {
    updateState(...)
    return
}
if err := doSomethingElse(); err != nil {
    updateState(...)
    return
}
```

I was trying to avoid that as there's so many error cases. It seemed "easier" to store the error state and handle it in defer.

pkg/component/controller/etcd_member_reconciler.go
jnummelin: I know, hence warn level only. Dunno if it makes sense to try to differentiate between a gracefull shutdown and abnormal termination

pkg/component/controller/etcd_member_reconciler.go
jnummelin: I'll add it

pschichtel:
two questions:
1. Will "left over" etcd members (e.g. from past-controller changes, happened to me once after changing node IPs) be cleaned up?
2. How would controller removal look like overall? especially when the node died without cleaning up after itself.

jnummelin:
> Will "left over" etcd members (e.g. from past-controller changes, happened to me once after changing node IPs) be cleaned up?

Not sure I get what you mean by "left over" member here? Do you mean you had a stale member in etcd cluster, i.e. a node with the old IP / hostname? If yes, then those can be also cleaned up by marking the node to leave the etcd cluster.

> How would controller removal look like overall? especially when the node died without cleaning up after itself.

Using `EtcdMember`object you'd remove a node by running something like:
```
kubectl patch etcdmember controller2 -p '{"leave":true}' --type merge
```

That triggers the current leader controller to remove the given member from etcd. Of course now if the node died without cleaning up youäd still have leftover data etc on the node.

pschichtel:
> Not sure I get what you mean by "left over" member here? Do you mean you had a stale member in etcd cluster, i.e. a node with the old IP / hostname?

yes exactly. so the "stale" members would also just show up as EtcdMember objects and I can then just delete them, is that right?

Regarding the controller removal: I was more interested in whether there would be any automation like e.g. a finalizer on Node objects, that makes sure that corresponding etcd member is deleted as well. So would "delete the etcdmember object" be an additional step in the process of removing a controller node or would it happen automatically under normal circumstances.

jnummelin:
> automation like e.g. a finalizer on Node objects

We cannot use `Node` object for this purpose, for plain controllers there's no Node object in k0s.

> So would "delete the etcdmember object" be an additional step in the process of removing a controller node or would it happen automatically under normal circumstances.

Once the member has left etcd cluster and the object status reflects that, you can safely remove the object. k0s will NOT do that automatically as we wanted to have some visibility on the actions already taken. My initial idea was actually to rely on object deletion but that does not keep any visible history what has happened.

Also in case you remove say `controller5` member and later re-join that same node (same hostname) k0s will be able to re-join it in etcd and the object status will be updated.

twz123 (approved):

cmd/controller/controller.go
```diff
@@ -325,6 +325,19 @@ func (c *command) start(ctx context.Context) error {
 		CertManager: worker.NewCertificateManager(ctx, c.K0sVars.KubeletAuthConfigPath),
 	})
 
+	if nodeConfig.Spec.Storage.Type == v1beta1.EtcdStorageType && nodeConfig.Spec.Storage.Etcd.ExternalCluster == nil {
```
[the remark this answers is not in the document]
jnummelin: It needs to be the other way around, i.e. to use `!nodeConfig.Spec.Storage.Etcd.IsExternalClusterUsed()`, but I get the pint 😄

docs/remove_controller.md
```diff
@@ -41,6 +41,51 @@ k0s reset
...
+### Declarative Etcd member management
+
+Starting from version 1.30, k0s supports also declarative way to remove a Etcd member. Since in k0s Etcd cluster is setup so that the Etcd API is **NOT** exposed outside of
+the nodes it makes it difficult for external automation like Cluster API, Terraform etc. to handle controller node replacements.
+
+Each controller manages their own `EtcdMember` object.
+
+```shell
+k0s kubectl get etcdmember
+NAME          PEERADDRESS   MEMBERID           JOINED   RECONCILESTATUS
```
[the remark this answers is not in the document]
jnummelin: They can, I'll change those

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+
+const JoinStatusJoined JoinStatus = "Joined"
+const JoinStatusLeft JoinStatus = "Left"
+
+type Status struct {
+	ReconcileStatus string `json:"reconcileStatus,omitempty"`
+	Message         string `json:"message,omitempty"`
+	// JoinStatus      JoinStatus `json:"joinStatus,omitempty"`
+
+	Conditions []JoinCondition `json:"conditions,omitempty"`
```
twz123, line 85: Picked that up by incident just today:

```suggestion
	// +listType=map
	// +listMapKey=type
	Conditions []JoinCondition `json:"conditions,omitempty"`
```

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+const (
+	ConditionTrue    ConditionStatus = "True"
+	ConditionFalse   ConditionStatus = "False"
+	ConditionUnknown ConditionStatus = "Unknown"
+)
+
+type JoinCondition struct {
+	Type ConditionType `json:"type"`
+	// +kubebuilder
+	Status ConditionStatus `json:"status"`
```
twz123, line 108:
```suggestion
	// +kubebuilder:validation:Required
	Type ConditionType `json:"type"`
	// +kubebuilder:validation:Required
	Status ConditionStatus `json:"status"`
```

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+	Conditions []JoinCondition `json:"conditions,omitempty"`
+}
+
+type ConditionType string
+
+const (
+	ConditionTypeJoined ConditionType = "Joined"
+)
+
+type ConditionStatus string
```
twz123, line 94:
```suggestion
// +kubebuilder:validation:Enum=True;False;Unknown
type ConditionStatus string
```

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+
+type Status struct {
+	ReconcileStatus string `json:"reconcileStatus,omitempty"`
+	Message         string `json:"message,omitempty"`
+	// JoinStatus      JoinStatus `json:"joinStatus,omitempty"`
+
+	Conditions []JoinCondition `json:"conditions,omitempty"`
+}
+
+type ConditionType string
```
twz123, line 88:
```suggestion
// +kubebuilder:validation:Enum=Joined
type ConditionType string
```

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,173 @@
...
+// +genclient:nonNamespaced
+type EtcdMember struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// PeerAddress is the address of the etcd peer
+	PeerAddress string `json:"peerAddress"`
+	// MemberID is the unique identifier of the etcd member.
+	// The hex form ID is stored as string
+	// +kubebuilder:validation:Pattern="^[a-fA-F0-9]+$"
```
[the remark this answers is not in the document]
jnummelin: Internally etcd uses `uint64`: https://github.com/etcd-io/etcd/blob/main/api/etcdserverpb/rpc.pb.go#L3085

The CLI output uses some generated (I assume as I didn't find the code in the repo) type casting: https://github.com/etcd-io/etcd/blob/main/etcdctl/ctlv3/command/printer_fields.go#L44

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+			log.WithError(err).Warn("watch terminated")
+		}
+	}()
+
+	return nil
+}
+
+func (e *EtcdMemberReconciler) Stop() error {
+	return nil
+}
```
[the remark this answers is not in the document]
jnummelin: I agree. But IMO not in scope of this PR. I'll spin up a separate issue on that

juanluisvaladas:
I converted this to draft accidentally and made it ready to review again, disregard my updates

twz123:
When looking at a EtcdMember custom resource, it looks like the following:

```yaml
apiVersion: etcd.k0sproject.io/v1beta1
kind: EtcdMember
memberID: 21e943ed8772de19
metadata:
  creationTimestamp: "2024-05-03T11:32:11Z"
  generation: 1
  name: etcdmember-controller-2
  resourceVersion: "362"
  uid: 641f2594-881a-4e8f-939e-94535d6b5904
peerAddress: 172.27.200.93
status:
  conditions:
  - status: "True"
    type: Joined
```

Wouldn't it be better to not have something else than apiVersion/kind/metadata/spec/status as top-level fields?

So what about this as an alternative:

```diff
--- etcdmember-controller-2
+++ etcdmember-controller-2
@@ -1,14 +1,16 @@
 apiVersion: etcd.k0sproject.io/v1beta1
 kind: EtcdMember
-memberID: 21e943ed8772de19
 metadata:
   creationTimestamp: "2024-05-03T11:32:11Z"
   generation: 1
   name: etcdmember-controller-2
   resourceVersion: "362"
   uid: 641f2594-881a-4e8f-939e-94535d6b5904
-peerAddress: 172.27.200.93
+spec:
+  leave: false
 status:
   conditions:
   - status: "True"
     type: Joined
+  memberID: 21e943ed8772de19
+  peerAddress: 172.27.200.93
```

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+	if !ok {
+		log.Debug("member marked for leave but not in actual member list, updating state to reflect that")
+		member.Status.SetCondition(etcdv1beta1.ConditionTypeJoined, etcdv1beta1.ConditionFalse, member.Status.Message, time.Now())
+		member, err = e.etcdMemberClient.UpdateStatus(ctx, member, metav1.UpdateOptions{})
+		if err != nil {
+			log.WithError(err).Error("failed to update EtcdMember status")
+		}
+	}
+
+	if member.Status.GetCondition(etcdv1beta1.ConditionTypeJoined).Status == etcdv1beta1.ConditionFalse && !ok {
```
twz123, line 318: Observed a panic during testing:

```text
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2aaa037]

goroutine 432 [running]:
github.com/k0sproject/k0s/pkg/component/controller.(*EtcdMemberReconciler).reconcileMember(0xc000d89940, {0x40de860, 0xc000e077d0}, 0xc001c08f00)
        github.com/k0sproject/k0s/pkg/component/controller/etcd_member_reconciler.go:318 +0xe37
github.com/k0sproject/k0s/pkg/component/controller.(*EtcdMemberReconciler).Start.func1.2(0x0?)
        github.com/k0sproject/k0s/pkg/component/controller/etcd_member_reconciler.go:110 +0x25
github.com/k0sproject/k0s/pkg/kubernetes/watch.(*Watcher[...]).processWatchEvent(0xc0024976ac?, 0xc001d1d6f0?, 0x0)
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:340 +0x177
github.com/k0sproject/k0s/pkg/kubernetes/watch.(*Watcher[...]).watch(0x4116d20, {0x40de898, 0xc0006b5b30}, {0xc0001ddba8, 0x5}, 0xc001d1dcd0)
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:319 +0x3f1
github.com/k0sproject/k0s/pkg/kubernetes/watch.(*Watcher[...]).run(0x4116d20, {0x40de898, 0xc0006b5b30}, 0xc001d1dcd0)
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:247 +0x76
github.com/k0sproject/k0s/pkg/component/controller.(*EtcdMemberReconciler).Start.func1.(*Watcher[...]).Until.3()
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:198 +0x79
github.com/k0sproject/k0s/pkg/kubernetes/watch.retry({0x40de860, 0xc000e077d0}, 0xc0016601a0, 0xc001d1df80)
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:390 +0x54
github.com/k0sproject/k0s/pkg/kubernetes/watch.(*Watcher[...]).Until(...)
        github.com/k0sproject/k0s/pkg/kubernetes/watch/watcher.go:195
github.com/k0sproject/k0s/pkg/component/controller.(*EtcdMemberReconciler).Start.func1()
        github.com/k0sproject/k0s/pkg/component/controller/etcd_member_reconciler.go:109 +0x4a5
created by github.com/k0sproject/k0s/pkg/component/controller.(*EtcdMemberReconciler).Start in goroutine 1
        github.com/k0sproject/k0s/pkg/component/controller/etcd_member_reconciler.go:80 +0x171
```

This occurred after deleting a removed EtcdMember resource via kubectl.

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+		member.Status.ReconcileStatus = "Failed"
+		member.Status.Message = err.Error()
+		_, err = e.etcdMemberClient.UpdateStatus(ctx, member, metav1.UpdateOptions{})
+		if err != nil {
+			log.WithError(err).Error("failed to update EtcdMember status")
+		}
+		return
+	}
+
+	// Peer removed succesfully, update status
```
twz123, line 343:
```suggestion
	// Peer removed successfully, update status
```

pkg/apis/etcd/v1beta1/types.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return nil
+}
+
+func (s *Status) SetCondition(t ConditionType, status ConditionStatus, msg string, time time.Time) {
+	var joinCondition JoinCondition
+	for i, j := range s.Conditions {
+		if j.Type == t {
+			jc := &s.Conditions[i]
+			// We found the matchin type, update it
```
twz123, line 135:
```suggestion
			// We found the matching type, update it
```

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+			log.WithError(err).Error("failed to update member state")
+		}
+
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+
+	// Verify that the member is actualy still present in etcd
```
twz123, line 295:
```suggestion
	// Verify that the member is actually still present in etcd
```

pkg/component/controller/etcd_member_reconciler.go
```diff
@@ -0,0 +1,352 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"time"
+
+	"github.com/k0sproject/k0s/inttest/common"
```
twz123, line 27: This is importing code from the integration tests.

pkg/component/controller/etcd_member_reconciler.go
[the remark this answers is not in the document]
twz123: Sure!

pkg/apis/etcd/v1beta1/types.go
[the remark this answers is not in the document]
twz123: I found `types.ID.String()`: https://github.com/etcd-io/etcd/blob/client/pkg/v3.5.13/client/pkg/types/id.go#L24-L26. It's using `strconv.FormatUint(uint64(i), 16)`. We can adjust this in a follow-up PR maybe.

docs/remove_controller.md
```diff
@@ -41,6 +41,61 @@ k0s reset
...
+$ kubectl wait etcdmember controller2 --for condition=Joined=False
+etcdmember.etcd.k0sproject.io/controller2 condition met
+```
+
+You'll see the node left etcd cluster:
+
+```console
+$ k0s kc get etcdmember
+NAME          PEER ADDRESS   MEMBER ID           JOINED   RECONCILE STATUS
+controller0   172.17.0.2     b8e14bda2255bc24    True
```
twz123, line 83: I wonder if k0s should also set the reconcile status to Success when joined is true.

twz123 (approved):

PR #4230, opened 2024-04-03
Replace the Machine ID usage with an "Invocation ID"

### LINKED ISSUE #3983: Replace machine ID with k0s-managed ID
K0s uses the [machine ID] for three purposes:

1. Leader election among controller nodes
2. Konnectivity server identification
3. Telemetry

I suggest replacing this with an ephemeral random value generated at each k0s startup, or alternatively, if it's deemed better to have this value persistent, to store and load this value in k0s's data-dir.

Telemetry aside, the value used for leader election and server identification doesn't need to be persistent at all, it just needs to be unique per k0s process. Using an ephemeral value would allow to remove a Go module dependency along with 100+ lines of code, tests and docs, I believe. It would eliminate the "disk image cloning" problem. Even if this ID needs to be persistent, it's much easier to just use some file in the data-dir for that.

[machine ID]: https://docs.k0sproject.io/v1.29.1+k0s.0/external-runtime-deps/#a-unique-machine-id-for-multi-node-setups

twz123:
## Description

Instead of relying on an externally maintained "machine identity" that may or may not exist on hosts, use a so-called Invocation ID. The Invocation ID is a randomly generated 32-byte hexadecimal string that is created during process startup. The name is inspired by systemd's INVOCATION_ID.

K0s doesn't need a unique host identifier that is stable across process invocations. It only needs to have unique identifiers per process. This removes a dependency on the k0s host machine (some machine-id file or registry entry),  a Go module dependency, Machine ID related documentation and system probes. It also makes the use of a unique ID obvious. The machine ID has been hard-coded into the guts of the leader election and telemetry components. The invocation ID is passed around instead.

Stub out the use of the anonymous ID in telemetry, as the new ID is no longer stable and would be useless for this type of use.

Resolves #3983.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


ncopa:
I think at some point kubelet would fail if the machine-id was missing. Is this no longer the case?

twz123:
> I think at some point kubelet would fail if the machine-id was missing. Is this no longer the case?

At least the inttests pass, and I removed all machine ID related setup from the bootloose test image. So The Alpine containers shouldn't have a machine ID anymore. I checked the Kubernetes sources, and there are very few traces about machine IDs, as described in https://github.com/k0sproject/k0s/issues/3983#issuecomment-2023013715.

twz123:
Checked the Kubernetes sources again.

On Linux, `MachineID` should be empty: [`MachineID:getInfoFromFiles(filepath.Join(rootFs, *machineIDFilePath)),`](https://github.com/kubernetes/kubernetes/blob/77aa9c21c295614817d5d49ea81f20783480691b/vendor/github.com/google/cadvisor/machine/info.go#L142) assigns it, [`machineIDFilePath`](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.1/vendor/github.com/google/cadvisor/machine/info.go#L40) lists the files to check, and [`klog.Warningf("Couldn't collect info from any of the files in %q", filePaths)`](https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.1/vendor/github.com/google/cadvisor/machine/info.go#L53) is logged in case none of the files could be read.

On Windows, there's even no attempt to read if from the registry, it is simply set to the hostname: https://github.com/kubernetes/kubernetes/blob/v1.30.0-rc.1/pkg/kubelet/winstats/perfcounter_nodestats.go#L162

pkg/telemetry/telemetry.go
```diff
@@ -151,7 +150,7 @@ func (c Component) sendTelemetry(ctx context.Context) {
 
 	c.log.WithField("data", data).WithField("hostdata", hostData).Info("sending telemetry")
 	if err := c.analyticsClient.Enqueue(analytics.Track{
-		AnonymousId: machineID(),
+		AnonymousId: "(removed)",
```
makhov, line 153: AnonymousID is the Segment entity: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#anonymous-ids

As far as I can see it's not used in our data pipeline now, but are we sure, that we want to remove it completely?

kke: I would think the unique cluster id is more important for telemetry than the machine id 🤔

makhov: Yeah, of course, I was wondering if it could break the pipeline, but we checked with Chen and it should be ok. 
Chen suggested testing it a bit, but in general we are fine

makhov (approved):

PR #4262, opened 2024-04-09
Rework users.GetUID

twz123:
## Description

**Rename to LookupUID**
The new name more closely matches the standard library function name. Also, "lookup" sounds less lightweight than "get", which gives a better sense of the actual runtime cost.

**Explicit usage and fallback handling**
Prefer to call utility functions directly with a UID, rather than having them accept a username and call `LookupUID` internally. This allows for a more obvious fallback handling in case certain users don't exist. Add a special `RootUID` constant that can be used to explicitly fallback to the root user. Also add log statements to every fallback assignment. Use -1 as UID return value in case of errors. The zero value for a UID happens to be root's UID. Assuming root may or may not be a good default, depending on the use case. Setting file ownership to root is a restrictive and safe default, running programs as root is not. By returning `UnknownUID` as a constant to -1, callers must explicitly decide on the fallback instead of accidentally assuming root.

**Use id program from PATH**
The `id` program is not always installed to `/usr/bin`. Rely on k0s's `PATH` to find the right executable.

**More context for errors**
There are quite some failure modes. Sort them out by providing hopefully helpful error messages. This might help when debugging things. Introduce `users.ErrNotFound`, which makes the `errors.Is()` checks nicer. On the contrary, the error message won't contain the user name anymore. Wrap the error accordingly on the caller side instead.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -38,9 +38,11 @@ const (
 	RootUID = 0 // User ID of the root user
 )
 
+var ErrNotExist = errors.New("user does not exist")
+
 // Lookup looks up a user's UID by username. If the user cannot be found, the
-// returned error is of type [user.UnknownUserError]. If an error is returned,
-// the returned UID will be [UnknownUID].
+// returned error is [ErrNotExist]. If an error is returned, the returned UID
+// will be [UnknownUID].
 func LookupUID(name string) (int, error) {
 	var uid string
 
@@ -48,6 +50,8 @@ func LookupUID(name string) (int, error) {
 		if !errors.Is(err, user.UnknownUserError(name)) {
 			return UnknownUID, err
 		}
+
+		err = ErrNotExist
 
 		// fallback to call external `id` in case NSS is used
 		out, idErr := exec.Command("id", "-u", name).Output()
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 )
 
+const RootUID = 0 // User ID of the root user
+
 // GetUID returns uid of given username and logs a warning if its missing
 func GetUID(name string) (int, error) {
 	entry, err := user.Lookup(name)
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -27,8 +27,8 @@ import (
 
 const RootUID = 0 // User ID of the root user
 
-// GetUID returns uid of given username and logs a warning if its missing
-func GetUID(name string) (int, error) {
+// LookupUID returns uid of given username and logs a warning if its missing
+func LookupUID(name string) (int, error) {
 	var uid string
 
 	if entry, err := user.Lookup(name); err != nil {
--- a/internal/pkg/users/users.go
+++ b/internal/pkg/users/users.go
@@ -25,15 +25,28 @@ import (
 	"strconv"
 )
 
-const RootUID = 0 // User ID of the root user
+const (
+	// An unknown (i.e. invalid) user ID. It's distinct from a UID's zero value,
+	// which happens to be [RootUID]. Assuming root may or may not be a good
+	// default, depending on the use case. Setting file ownership to root is a
+	// restrictive and safe default, running programs as root is not. Therefore,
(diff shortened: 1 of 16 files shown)


internal/pkg/users/users.go
```diff
@@ -17,25 +17,61 @@ limitations under the License.
...
+)
+
+// Lookup looks up a user's UID by username. If the user cannot be found, the
+// returned error is of type [user.UnknownUserError]. If an error is returned,
+// the returned UID will be [UnknownUID].
+func LookupUID(name string) (int, error) {
+	var uid string
+
+	if entry, err := user.Lookup(name); err != nil {
+		if !errors.Is(err, user.UnknownUserError(name)) {
```
kke, line 48: `errors.Is(err, user.UnknownUserError(name))` looks weird but seems to work.

Maybe returning a locally defined `UnknownUserError = errors.New("unknown user")` would make the function more pleasant to use.

https://go.dev/play/p/EsG_Q7Uagbw

twz123: I was also pondering about this. The original code re-used this standard library error in this way. Changing the returned error would trickle down to callers, which are relying on this as well.

What about having a function for this instead, e.g.

```go
func IsNotExist(err error) bool {
	var UnknownUserError user.UnknownUserError
	return errors.As(err, &UnknownUserError)
}
```

Callers could then use `if users.IsNotExist(err) { ... }` ... Which would mimic the design of the standard library's os.Is... functions.

kke: Sounds reasonable. The `os.Is...` function docs now do say `new code should use errors.Is(err, fs.ErrNotExist)` but they're not going anywhere.

twz123: Oh. Right. I'll have a look at your initial proposal first, then.

twz123: PTAL

internal/pkg/users/users.go
```diff
@@ -17,25 +17,65 @@ limitations under the License.
...
+			return UnknownUID, err
+		}
+
+		err = ErrNotExist
+
 		// fallback to call external `id` in case NSS is used
-		out, err := exec.Command("/usr/bin/id", "-u", name).CombinedOutput()
-		if err == nil {
-			return strconv.Atoi(strings.TrimSuffix(string(out), "\n"))
+		out, idErr := exec.Command("id", "-u", name).Output()
```
kke, line 57: Go's `user.Lookup` reads `/etc/passwd`.  It seems like coreutils `id` does the same.

Maybe this doesn't need a fallback?

kke: Or maybe it does in the case of nss 🤔

kke: https://pkg.go.dev/github.com/tmrts/flamingo/pkg/sys/nss uses `getent` but `id` is probably fine as well.

internal/pkg/users/users.go
```diff
@@ -17,25 +17,65 @@ limitations under the License.
...
+		if idErr != nil {
+			var exitErr *exec.ExitError
+			if errors.As(idErr, &exitErr) {
+				firstLine, _, _ := bytes.Cut(exitErr.Stderr, []byte{'\n'})
+				return UnknownUID, fmt.Errorf("%w (%w: %s)", err, idErr, string(firstLine))
+			}
+			return UnknownUID, fmt.Errorf("%w (%w)", err, idErr)
 		}
+
+		uid = string(bytes.TrimSuffix(out, []byte{'\n'}))
```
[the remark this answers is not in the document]
twz123: To be precise, this removes a trailing newline, and doesn't return only the first line, but all of them, if any. `id` is expected to only write a single line to stdout, though.

The stderr truncation is different. In case of an error, we never know how many lines will be written. Having only the first line in the error message seemed like a good idea to me, to not potentially pollute the error message but still be helpful. The standard error message of `*exec.ExitError` is not overly helpful on its own, it's usually just "exit status 1" and that's it.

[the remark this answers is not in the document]
twz123: PTAL

jnummelin (approved):
LGTM

PR #4277, opened 2024-04-11
Ignore token-file argument when value not needed

kke:
## Description

Both "k0s controller" and "k0s worker" take either `[TOKEN]` as an argument or a path to a file containing the join token via `--token-file`.

Before this PR, when `--token-file` was given, the file was read upon spin-up, even when the node has joined already and won't utilize the token for anything. As the token is not needed after join, you would expect that you can or even should remove the file, but that makes k0s refuse to start unless the `--token-file` argument is also removed from the systemd unit or other startup scripts.

This PR makes the controller and worker read the contents of the file only right before the value is getting used. This allows `--token-file` argument to be ignored when it is not needed and the file not existing will not generate an error.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [X] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -92,20 +92,13 @@ func NewControllerCmd() *cobra.Command {
 			if len(args) > 0 {
 				c.TokenArg = args[0]
 			}
-			if len(c.TokenArg) > 0 && len(c.TokenFile) > 0 {
+			if c.TokenArg != "" && c.TokenFile != "" {
 				return fmt.Errorf("you can only pass one token argument either as a CLI argument 'k0s controller [join-token]' or as a flag 'k0s controller --token-file [path]'")
 			}
 			if err := c.ControllerOptions.Normalize(); err != nil {
 				return err
 			}
 
-			if len(c.TokenFile) > 0 {
-				bytes, err := os.ReadFile(c.TokenFile)
-				if err != nil {
-					return err
-				}
-				c.TokenArg = string(bytes)
-			}
 			c.Logging = stringmap.Merge(c.CmdLogLevels, c.DefaultLogLevels)
 
 			if err := (&sysinfo.K0sSysinfoSpec{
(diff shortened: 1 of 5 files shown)


kke:
Todo: some kind of test.

cmd/controller/controller.go
```diff
@@ -177,8 +170,22 @@ func (c *command) start(ctx context.Context) error {
 
 	var joinClient *token.JoinClient
 
-	if c.TokenArg != "" && c.needToJoin() {
-		joinClient, err = joinController(ctx, c.TokenArg, c.K0sVars.CertRootDir)
+	if c.needToJoin() {
+		if c.TokenArg == "" && c.TokenFile == "" {
+			return fmt.Errorf("token argument or --token-file is required to join")
+		}
```
twz123, line 176: If we want to keep the same behavior as before it, it needs to be

```suggestion
	if (c.TokenArg != "" || c.TokenFile != "") && c.needToJoin() {
```

kke: But is that a wished behavior? :)

twz123:
This might need #4288.

inttest/upgrade/upgrade_test.go
```diff
@@ -151,6 +151,11 @@ func (s *UpgradeSuite) TestK0sGetsUp() {
 			if err != nil {
 				return err
 			}
+			// delete token file because it shouldn't be needed anymore to see that worker can join without it
```
twz123, line 154:
```suggestion
			// Delete the token file, as it shouldn't be needed after the worker has joined.
```

inttest/hacontrolplane/hacontrolplane_test.go
```diff
@@ -108,7 +109,9 @@ func (s *HAControlplaneSuite) TestDeregistration() {
 	defer sshC1.Disconnect()
 	_, err = sshC1.ExecWithOutput(s.Context(), "kill $(pidof k0s) && while pidof k0s; do sleep 0.1s; done")
 	s.Require().NoError(err)
-	s.NoError(s.InitController(1, token))
+	_, err = sshC1.ExecWithOutput(s.Context(), "rm -f /etc/k0s.token")
```
twz123, line 112:
```suggestion
	// Delete the token file, as it shouldn't be needed after the controller has joined.
	_, err = sshC1.ExecWithOutput(s.Context(), "rm -f /etc/k0s.token")
```

twz123 (approved):

twz123 (approved):

PR #4289, opened 2024-04-11
Apply some idiomatic patterns to CPLB component

twz123:
## Description

This is how other components do it.

- Use the standard RunDir/DataDir for Supervisor
- Use k0s's RunDir for keepalived config file
  The file is ephemeral and doesn't need to live on a persistent path. This also saves us from ensuring the existence of the base directory in the cplb component, as the RunDir's existence is ensured during controller startup.
- Use file.WriteAtomically to write keepalived config file

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


juanluisvaladas:
LGTM, needs to pass tests again after #4290 is merged

twz123:
Rebased.

pkg/component/controller/cplb_unix.go
```diff
@@ -274,31 +268,25 @@ func (*Keepalived) getLinkAddresses(link netlink.Link) ([]netlink.Addr, []string
...
-		return fmt.Errorf("failed to open keepalived config file: %w", err)
-	}
-	defer f.Close()
-
-	template, err := template.New("keepalived").Parse(keepalivedConfigTemplate)
-	if err != nil {
-		return fmt.Errorf("failed to parse keepalived template: %w", err)
-	}
-
+	template := template.Must(template.New("keepalived").Parse(keepalivedConfigTemplate))
```
kke, line 271:
```suggestion
	template, err := template.New("keepalived").Parse(keepalivedConfigTemplate)
	if err != nil {
		return fmt.Errorf("parse keepalived template: %w", err)
	}
```

pkg/component/controller/cplb_unix.go
```diff
@@ -274,31 +268,25 @@ func (*Keepalived) getLinkAddresses(link netlink.Link) ([]netlink.Addr, []string
...
+	if err := file.WriteAtomically(k.configFilePath, 0400, func(file io.Writer) error {
+		if err := file.(*os.File).Chown(k.uid, -1); err != nil {
+			return err
+		}
+
+		w := bufio.NewWriter(file)
+		if err := template.Execute(w, kc); err != nil {
+			return err
+		}
+		return w.Flush()
```
kke, line 285: What's the benefit of running it through the bufio.Writer?

pkg/component/controller/cplb_unix.go
```diff
@@ -274,31 +268,25 @@ func (*Keepalived) getLinkAddresses(link netlink.Link) ([]netlink.Addr, []string
...
-	}
 
-	// TODO: Do we really need to this every single time?
-	if err = os.Chown(k.K0sVars.KeepalivedConfigFile, k.uid, -1); err != nil {
-		return fmt.Errorf("failed to chown keepalived config file: %w", err)
-	}
-	if err = os.Chmod(k.K0sVars.KeepalivedConfigFile, fs.FileMode(0400)); err != nil {
-		return fmt.Errorf("failed to chmod keepalived config file: %w", err)
+	if err := file.WriteAtomically(k.configFilePath, 0400, func(file io.Writer) error {
+		if err := file.(*os.File).Chown(k.uid, -1); err != nil {
```
kke, line 277: maybe `file.WriteAtomically`s callback should be `func (file *os.File)` if it's always a file.

pkg/component/controller/cplb_unix.go
twz123: I think those "Must" style functions are specifically meant for const inputs like here. Why bother with error handling if the input is guaranteed to be valid? Using `template.Must(...)` here is no different from using e.g. `regex.MustParse("[a-z0-9]")`.

pkg/component/controller/cplb_unix.go
twz123: Buffered writes. I.e. instead of having lots of write calls that write just a couple of bytes directly to disk, buffer them until a reasonable amount of bytes has been accumulated. I think the default is to buffer 4KiB before writing to the wrapped writer (the file on disk in this case). Reduces the number of syscalls and usually speeds up things quite a bit. An alternative would be to buffer everything in memory and write it all at once using a single call to file.Write.

pkg/component/controller/cplb_unix.go
twz123: This is actually me cheating here because I know it's a file 😈

When I created the WriteAtomically function, I explicitly wanted to hide the fact that this is a file, so that nobody would do unintended things like calling file.Name(), file.Remove() and so on. The function's API has the permissions, but it's lacking ownership information. I missed that, since `os.WriteFile` is similar, probably because you cannot convey ownership information through the `open` syscall and have to do it separately? Have to think about it a bit...

twz123: That said, I'm not even 100% sure right now that os.Rename will keep ownership info. I assume, but would I bet? For Linux and Windows?

pkg/component/controller/cplb_unix.go
kke: I think it's to ease defining globals:

```go
var parsedTemplate = template.Must(template.New("keepalived").Parse(keepalivedConfigTemplate))

func main() {
   parsedTemplate.Execute(...)
}
```

At least, that seems to be the intention of `regexp.MustCompile`:

> MustCompile is like [Compile] but panics if the expression cannot be parsed.
> It simplifies safe initialization of global variables holding compiled regular
> expressions.

So, as the template is constant, it could be parsed at startup:

```diff
-const keepalivedConfigTemplate = `
+var keepalivedConfigTemplate = template.Must(template.New("keepalived").Parse(`
```

pkg/component/controller/cplb_unix.go
twz123: Reminder to self: os.Chwon/file.Chown is not a thing on Windows anyways.

juanluisvaladas (approved):

pkg/component/controller/cplb_unix.go
twz123: Follow-up in #4331.

PR #4294, opened 2024-04-12
Remove binary before building it

twz123:
## Description

The Go compiler doesn't touch the binary if it doesn't detect any code changes. (Presumably by checking the BuildID of the old and the new binary?) However, the Makefile relied on a regeneration of the binary in any case, so that it can reappend the bindata file agan. Mitigate this by deleting the binary before the build. That way, the Go compiler doesn't have a choice.

Fixes:
* #3905

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,6 +189,7 @@ k0s.exe: TARGET_OS = windows
 k0s.exe: BUILD_GO_CGO_ENABLED = 0
 
 k0s.exe k0s: $(GO_SRCS) $(codegen_targets) go.sum
+	rm -f -- '$@'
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -189,6 +189,7 @@ k0s.exe: TARGET_OS = windows
 k0s.exe: BUILD_GO_CGO_ENABLED = 0
 
 k0s.exe k0s: $(GO_SRCS) $(codegen_targets) go.sum
+	-[ ! -e '$@' ] || rm -- '$@'
```
ncopa, line 192: Why not do a simple: `rm -f -- '$@'`?

twz123: Because that will give an error when the file is not there. Such things usually catch my eyes in build outputs and then I need to spend some brain cycles to figure out that this is actually fine. Avoiding error messages is hence an investment in the future :upside_down_face:

twz123: Oh. Actually it will not give an error if the file doesn't exist. Will change that ASAP :see_no_evil:

ncopa (approved):

PR #4299, opened 2024-04-16
Drop multierr module and rely on the standard lib

twz123:
## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/internal/pkg/file/file.go
+++ b/internal/pkg/file/file.go
@@ -25,8 +25,6 @@ import (
 	"path/filepath"
 
 	"github.com/k0sproject/k0s/internal/pkg/users"
-
-	"go.uber.org/multierr"
 )
 
 // Exists checks if a file exists and is not a directory before we
@@ -60,7 +58,7 @@ func Copy(src, dst string) (err error) {
 	if err != nil {
 		return err
 	}
-	defer multierr.AppendInvoke(&err, multierr.Close(in))
+	defer func() { err = errors.Join(err, in.Close()) }()
 
 	sourceFileStat, err := in.Stat()
 	if err != nil {
@@ -97,18 +95,31 @@ func WriteAtomically(fileName string, perm os.FileMode, write func(file io.Write
 	tmpFileName := fd.Name()
 	close := true
 	defer func() {
-		remove := err != nil
-		if close {
-			err = multierr.Append(err, fd.Close())
+		var errs []error
+		if err != nil {
+			errs = append(errs, err)
 		}
-		if remove {
-			removeErr := os.Remove(tmpFileName)
+
+		if close {
+			if err := fd.Close(); err != nil {
+				errs = append(errs, err)
+			}
+		}
+
+		if err != nil {
+			err := os.Remove(tmpFileName)
 			// Don't propagate any fs.ErrNotExist errors. There is no point in
 			// doing this, since the desired state is already reached: The
 			// temporary file is no longer present on the file system.
-			if removeErr != nil && !errors.Is(err, fs.ErrNotExist) {
-				err = multierr.Append(err, removeErr)
+			if err != nil && !errors.Is(err, fs.ErrNotExist) {
+				errs = append(errs, err)
 			}
+		}
+
+		if len(errs) == 1 {
+			err = errs[0]
+		} else {
+			err = errors.Join(errs...)
 		}
 	}()
 
--- a/internal/pkg/iptablesutils/iptables.go
+++ b/internal/pkg/iptablesutils/iptables.go
@@ -18,13 +18,13 @@ package iptablesutils
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"os/exec"
 	"path/filepath"
 	"strings"
 
 	"github.com/sirupsen/logrus"
-	"go.uber.org/multierr"
 )
 
 const (
@@ -68,12 +68,12 @@ func DetectHostIPTablesMode(k0sBinPath string) (string, error) {
 
 	iptablesPath, err := exec.LookPath("iptables")
 	if err != nil {
-		return "", multierr.Combine(err, nftErr, legacyErr)
+		return "", errors.Join(err, nftErr, legacyErr)
 	}
 
 	out, err := exec.Command(iptablesPath, "--version").CombinedOutput()
 	if err != nil {
-		return "", multierr.Combine(err, nftErr, legacyErr)
+		return "", errors.Join(err, nftErr, legacyErr)
 	}
 
 	outStr := strings.TrimSpace(string(out))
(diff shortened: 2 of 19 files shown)


internal/pkg/file/file.go
```diff
@@ -60,7 +58,14 @@ func Copy(src, dst string) (err error) {
 	if err != nil {
 		return err
 	}
-	defer multierr.AppendInvoke(&err, multierr.Close(in))
+	defer func() {
+		closeErr := in.Close()
+		if err == nil {
+			err = closeErr
+		} else if closeErr != nil {
+			err = fmt.Errorf("%w; %w", err, closeErr)
```
kke, line 66: There's a `;` there.

I suppose `defer func() { err = errors.Join(err, in.Close()) }` would do the same except if the`%w: %w` formatting is important.

internal/pkg/iptablesutils/iptables.go
```diff
@@ -119,10 +119,7 @@ func findMatchingEntries(k0sBinPath, mode string, entries ...string) (entriesFou
 
 	v4Err, v6Err := findMatches("iptables-save"), findMatches("ip6tables-save")
 	if v4Err != nil && v6Err != nil {
-		return false, 0, multierr.Combine(
-			fmt.Errorf("iptables-save: %w", v4Err),
-			fmt.Errorf("ip6tables-save: %w", v6Err),
-		)
+		return false, 0, fmt.Errorf("iptables-save: %w; ip6tables-save: %w", v4Err, v6Err)
```
kke, line 122: There's a `;` again, are these intentional? :)

kke: I suppose they are. Looks a bit confusing to me.

pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -469,7 +468,11 @@ func extractAPIServerAddresses(endpoints *corev1.Endpoints) ([]k0snet.HostPort,
 	if len(apiServers) < 1 {
 		// Never update the API servers with an empty list. This cannot
 		// be right in any case, and would never recover.
-		return nil, multierr.Append(errors.New("no API server addresses discovered"), warnings)
+		const msg = "no API server addresses discovered"
+		if warnings := errors.Join(warnings...); warnings != nil {
+			return nil, fmt.Errorf("%s: %w", msg, warnings)
+		}
+		return nil, errors.New(msg)
```
kke, line 475: Hmm, `warnings := errors.Join(warnings...); warnings != nil` looks confusing. How does it even compile? `warnings` is `[]error` but `errors.Join()` returns `error`, doesn't this make warnings change type mid-sentence? 🤔 


Could the whole block just be `return nil, errors.Join(errors.New("no API server addresses discovered"), warnings...)`?

or

```go
err := errors.New("no API server addresses discovered")
if warnings := errors.Join(warnings...); warnings != nil {
  return nil, fmt.Errorf("%w: %w", err, warnings)
}
return nil, err
```

pkg/component/worker/nllb/envoy.go
```diff
@@ -242,11 +241,11 @@ func writeEnvoyConfigFiles(params *envoyParams, filesParams *envoyFilesParams) e
 			return bufferedWriter.Flush()
 		})
 		if err != nil {
-			errs = multierr.Append(errs, fmt.Errorf("failed to write %s: %w", fileName, err))
+			errs = append(errs, fmt.Errorf("failed to write %s: %w", fileName, err))
```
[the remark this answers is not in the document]
kke: Perhaps for another PR since that would change behavior

internal/pkg/file/file.go
twz123: Not exactly the same. This is touching some of the things I preferred in the multierr implementation.

1. multierr used `"; "` to join error messages, while `errors.Join(...)` uses `"\n"`, which has puzzled me since its inception. So yes, the semicolon is intentional, as this is two equitable errors, not one that's been caused by another one, where a colon would be more appropriate.
2. `errors.Join(...)` wraps the error in any case, even if there's only one, so `errors.Join(nil, err) != err`, whereas multierr did not wrap a single error: `multierr.Combine(nil, err) == err`. Same for `multierr.Append(...)`. I wonder why upstream decided to do it that way. Maybe because of 3.
3. I wonder why there isn't a public error type for "a slice of errors", which is basically what `errors.Join(...)` represents. To me, there's a fundamental difference between "multiple errors occurred" (which is "a slice of errors" / `errors.Join(...)`) and "an error occurred that has multiple causes" (which is `fmt.Errorf(...)` with multiple `%w` placeholders, or some other custom error type that implements `Unwrap() []error`). The way the Go standard library does it makes these cases indistinguishable.

Those things bother me in the way that things bother people even when they know they really shouldn't care that much 🙈

internal/pkg/iptablesutils/iptables.go
twz123: What would be a more appropriate separator? A newline, as in `errors.Join(...)`?

pkg/component/worker/nllb/envoy.go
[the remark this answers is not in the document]
twz123: > does it make sense to write the second file if writing the first already fails instead of just returning the error when it happens?

That's a good question. I think yes, it makes sense, since the templated values in those files don't depend on each other. The first defines the local bind addresses, the second defines the remote endpoints. So in any case the resulting config should remain valid, albeit partially out of date.

pkg/component/controller/workerconfig/reconciler.go
twz123: Done.

internal/pkg/file/file.go
twz123: Done.

kke: Go docs say:

> When adding additional context to an error, either with fmt.Errorf or by implementing a custom type, you need to decide whether the new error should wrap the original. There is no single answer to this question; it depends on the context in which the new error is created. Wrap an error to expose it to callers. Do not wrap an error when doing so would expose implementation details.
> 
> As one example, imagine a Parse function which reads a complex data structure from an io.Reader. If an error occurs, we wish to report the line and column number at which it occurred. If the error occurs while reading from the io.Reader, we will want to wrap that error to allow inspection of the underlying problem. Since the caller provided the io.Reader to the function, it makes sense to expose the error produced by it.
> 
> In contrast, a function which makes several calls to a database probably should not return an error which unwraps to the result of one of those calls. If the database used by the function is an implementation detail, then exposing these errors is a violation of abstraction. For example, if the LookupUser function of your package pkg uses Go’s database/sql package, then it may encounter a sql.ErrNoRows error. If you return that error with fmt.Errorf("accessing DB: %v", err) then a caller cannot look inside to find the sql.ErrNoRows. But if the function instead returns fmt.Errorf("accessing DB: %w", err), then a caller could reasonably write
> 
> ```go
> err := pkg.LookupUser(...)
> if errors.Is(err, sql.ErrNoRows) …
> ```
> At that point, the function must always return sql.ErrNoRows if you don’t want to break your clients, even if you switch to a different database package. In other words, wrapping an error makes that error part of your API. If you don’t want to commit to supporting that error as part of your API in the future, you shouldn’t wrap the error.
> 
> It’s important to remember that whether you wrap or not, the error text will be the same. A person trying to understand the error will have the same information either way; the choice to wrap is about whether to give programs additional information so they can make more informed decisions, or to withhold that information to preserve an abstraction layer.

So, in that light, `%v` is sometimes better than `%w` and `errors.Join` is not a replacement for adding context to error messages 🤔

kke (approved):

Issue #4316, opened 2024-04-18
autopilot update does not import new images if k0s version remains the same


cbodonnell:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.5.0-1016-gcp #16~22.04.1-Ubuntu SMP Sat Mar  9 00:58:37 UTC 2024 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.29.3+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "eb37358b11ef42b9066501fe5d801ae439fb533d58d35002b876896ecddff908" (from machine) (pass)
Total memory: 3.8 GiB (pass)
Disk space available for /var/lib/k0s: 179.1 GiB (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.5.0-1016-gcp (pass)
  Max. file descriptors per process: current: 1048576 / max: 1048576 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
[...]

PR #4321, opened 2024-04-22
feat: implement watcher for oci bundles

ricardomaraschini:
## Description

Implements an OCI bundle watcher. This allows k0s to load new OCI bundles without requiring a restart of the process. The watcher acts upon `Create()` or `Write()` operations happening in the OCI bundles directory and events are debounced with a timeout of 10 seconds.

Fixes [# 4316](https://github.com/k0sproject/k0s/issues/4316)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

At this stage I have updated `k0s` locally and created/deleted/updated bundles manually in the OCI bundles directory. I have been checking the results by inspecting the logs (as follow) and with `k0s ctr i ls`:

```
root@ec:/usr/local/bin# journalctl -u k0scontroller.service -f | grep -i component=OCIBundleReconciler
Apr 22 10:49:44 ec k0s[2014911]: time="2024-04-22 10:49:44" level=info msg="started to watch events on /var/lib/k0s/images" component=OCIBundleReconciler
Apr 22 10:50:06 ec k0s[2014911]: time="2024-04-22 10:50:06" level=info msg="OCI bundle directory changed, reconciling" component=OCIBundleReconciler
Apr 22 10:50:08 ec k0s[2014911]: time="2024-04-22 10:50:08" level=info msg="Imported image docker.io/library/x:latest" component=OCIBundleReconciler
Apr 22 10:50:28 ec k0s[2014911]: time="2024-04-22 10:50:28" level=info msg="OCI bundle directory changed, reconciling" component=OCIBundleReconciler
Apr 22 10:50:29 ec k0s[2014911]: time="2024-04-22 10:50:29" level=info msg="Imported image docker.io/library/x:latest" component=OCIBundleReconciler
Apr 22 10:51:03 ec k0s[2014911]: time="2024-04-22 10:51:03" level=info msg="OCI bundle directory changed, reconciling" component=OCIBundleReconciler
Apr 22 10:51:04 ec k0s[2014911]: time="2024-04-22 10:51:04" level=info msg="Imported image docker.io/library/y:latest" component=OCIBundleReconciler
Apr 22 10:51:40 ec k0s[2014911]: time="2024-04-22 10:51:40" level=info msg="OCI bundle directory changed, reconciling" component=OCIBundleReconciler
Apr 22 10:51:59 ec k0s[2014911]: time="2024-04-22 10:51:59" level=info msg="Imported image docker.io/library/rhel-9-kubernetes-images-1.29.3:latest" component=OCIBundleReconciler
Apr 22 10:51:59 ec k0s[2014911]: time="2024-04-22 10:51:59" level=info msg="Imported image docker.io/kurl/rhel-7-k8s:1.29.3" component=OCIBundleReconciler
Apr 22 10:51:59 ec k0s[2014911]: time="2024-04-22 10:51:59" level=info msg="Imported image docker.io/library/dockerout-containerd-1.6.28-ubuntu-22.04:latest" component=OCIBundleReconciler
```

## Checklist:

### DIFF
--- a/pkg/component/worker/ocibundle.go
+++ b/pkg/component/worker/ocibundle.go
@@ -58,7 +58,7 @@ func NewOCIBundleReconciler(vars *config.CfgVars) *OCIBundleReconciler {
 		log:             logrus.WithField("component", "OCIBundleReconciler"),
 		EventEmitter:    prober.NewEventEmitter(),
 		alreadyImported: map[string]time.Time{},
-		end:             make(chan struct{}, 1),
+		end:             make(chan struct{}),
 	}
 }
 
@@ -178,11 +178,13 @@ func (a *OCIBundleReconciler) installWatcher(ctx context.Context) error {
 	}()
 
 	go func() {
+		defer close(a.end)
 		a.log.Infof("Started to watch events on %s", a.k0sVars.OCIBundleDir)
 		_ = debouncer.Run(ctx)
-		watcher.Close()
+		if err := watcher.Close(); err != nil {
+			a.log.Errorf("Failed to close watcher: %s", err)
+		}
 		a.log.Info("OCI bundle watch bouncer ended")
-		a.end <- struct{}{}
 	}()
 
 	return nil
--- a/pkg/component/worker/ocibundle.go
+++ b/pkg/component/worker/ocibundle.go
@@ -89,7 +89,7 @@ func (a *OCIBundleReconciler) loadOne(ctx context.Context, fpath string) error {
 	return nil
 }
 
-// loadAll loads all OCI bundle files into containerd. Read all files from the oci bundle
+// loadAll loads all OCI bundle files into containerd. Read all files from the OCI bundle
 // directory and loads them one by one. Errors are logged but not returned, upon failure
 // in one file this function logs the error and moves to the next file.
 func (a *OCIBundleReconciler) loadAll(ctx context.Context) {
--- a/pkg/component/worker/ocibundle.go
+++ b/pkg/component/worker/ocibundle.go
@@ -26,18 +26,22 @@ import (
 	"github.com/avast/retry-go"
 	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/platforms"
+	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
+
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/component/prober"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
-	"github.com/sirupsen/logrus"
+	"github.com/k0sproject/k0s/pkg/debounce"
 )
 
 // OCIBundleReconciler tries to import OCI bundle into the running containerd instance
 type OCIBundleReconciler struct {
 	k0sVars *config.CfgVars
 	log     *logrus.Entry
+	loaded  map[string]time.Time
 	*prober.EventEmitter
 }
 
@@ -49,6 +53,7 @@ func NewOCIBundleReconciler(vars *config.CfgVars) *OCIBundleReconciler {
 		k0sVars:      vars,
 		log:          logrus.WithField("component", "OCIBundleReconciler"),
 		EventEmitter: prober.NewEventEmitter(),
+		loaded:       make(map[string]time.Time),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jnummelin:
The code might get bit simpler / cleaner if you'd separate loading a single bundle file into it's own func. Also that way when we see FS event, we do not need to go through the entire dir. Although there's probably only few files there anyways so it should not be that big of an optimization. 😄

ricardomaraschini:
> The code might get bit simpler / cleaner if you'd separate loading a single bundle file into it's own func. Also that way when we see FS event, we do not need to go through the entire dir. Although there's probably only few files there anyways so it should not be that big of an optimization. 😄

I thought about that but haven't gone that route because I noticed we are connecting / disconnecting containerd's client every time. That allied with the fact we are calling `client.ListImages(ctx)` as some kind of "connection validation" made me believe there were more things that meet my eye. 

That all being said I can certainly work to change what we have here. Do you think connecting / disconnecting from containerd for each file would be a problem ? Are you aware of any sort of problem with long live containerd connections ?

ricardomaraschini:
@jnummelin I added a function to import a single OCI bundle and now we are calling it from the watcher. Let me know what you think.

jnummelin:
> Do you think connecting / disconnecting from containerd for each file would be a problem ?

No, I don't think it'll be any issue. I mean there's not like gazillions of those files and it's "just" a local socket connection.

> Are you aware of any sort of problem with long live containerd connections ?

I haven't seen anything but I don't think we've ever implemented ones either. So honestly, dunno 🤷 

IMO it's better this way now where we can load single files based on events.

@twz123 WDYT?

twz123:
> > Do you think connecting / disconnecting from containerd for each file would be a problem ?
> 
> No, I don't think it'll be any issue. I mean there's not like gazillions of those files and it's "just" a local socket connection.
> 
> > Are you aware of any sort of problem with long live containerd connections ?
> 
> I haven't seen anything but I don't think we've ever implemented ones either. So honestly, dunno 🤷
> 
> IMO it's better this way now where we can load single files based on events.
> 
> @twz123 WDYT?

I'm not too concerned about the connect/probe/import/disconnect procedure for each individual bundle file. As you already said: there'll be just a few anyways and they won't change frequently. On the other hand, I wouldn't opt for a long-living connection. That sounds like something that breaks after being idle for x weeks, and then a new file gets added.

But: instead of doing a directory listing in the loadAll method itself, we could add a new parameter with a slice of file names to import. We could inline the loadOne method again, and reuse a single connection for the whole slice, as it was before before. Then we can still just use loadAll with a single-element slice from the watcher.

pkg/component/worker/ocibundle.go
```diff
@@ -56,46 +60,126 @@ func (a *OCIBundleReconciler) Init(_ context.Context) error {
...
+			switch item.Op {
+			case fsnotify.Create, fsnotify.Write:
+				return true
+			default:
+				return false
+			}
+		},
+		Callback: func(ev fsnotify.Event) {
+			a.log.Infof("Loading OCI bundle %s", ev.Name)
+			if err := a.loadOne(ctx, ev.Name); err != nil {
```
twz123, line 149: Unfortunately, it won't work that way with the debouncer. The debouncer will discard all but the last event, so if you're adding multiple files in one go, you'll only get the event for the file that got added last.

You can only meaningfully use it with the watcher if you inspect the whole directory. You'd need to cache which files have been imported already (e.g. based on last modified date and file size or sth. like that), or implement debouncing differently.

twz123, line 149: Just a note, since we're not expecting a lot of churn in the bundle directory, but still good to know:

If we're importing the bundles in the same goroutine that drains the file system events, we'll be stalling the queue. Depending on the kernel buffer size, we might miss events, in which case we should get an error via the Error channel then. If we want to avoid this, we'd need to move the import out of this goroutine, which adds quite a bit of complexity, though.

twz123 (commented):
Thanks for working on this!

pkg/component/worker/ocibundle.go
```diff
@@ -56,46 +60,126 @@ func (a *OCIBundleReconciler) Init(_ context.Context) error {
...
+	}()
+
+	a.log.Infof("Started to watch events on %s", a.k0sVars.OCIBundleDir)
+	if err := debouncer.Run(ctx); err != nil {
+		if errors.Is(err, context.Canceled) {
+			a.log.Info("OCI bundle watch bouncer ended")
+			return
+		}
+		a.log.WithError(err).Warn("OCI bundle watch bouncer exited with error")
+	}
```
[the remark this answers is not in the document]
ricardomaraschini: I don't understand why but I can't commit your suggestion so I had to copy & pasta. Thanks for that, it is already adjusted.

pkg/component/worker/ocibundle.go
```diff
@@ -56,46 +60,126 @@ func (a *OCIBundleReconciler) Init(_ context.Context) error {
...
+}
+
+// Starts initiate the OCI bundle loader. It does an initial load of the directory and
+// once it is done, it starts a watcher on its own goroutine.
+func (a *OCIBundleReconciler) Start(ctx context.Context) error {
+	a.loadAll(ctx)
+	go a.watch(ctx)
```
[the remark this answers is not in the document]
ricardomaraschini: I understand. My initial approach here was to run through the whole directory every time an event popped out (taking the necessary measures to avoid importing the same file twice). I have since changed to this new version as per @jnummelin suggestion.

Long story short, I will raise a new attempt at reviving the old approach + the changes you proposed here and ping you back.

pkg/component/worker/ocibundle.go
ricardomaraschini: Got it. I think we can use a buffered channel + load all files in the directory to mitigate this. I will make this part of my next commit.

pkg/component/worker/ocibundle.go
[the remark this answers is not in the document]
ricardomaraschini: Ah, I see, it got "outdated" most likely after I merge some of your other suggestions.

pkg/component/worker/ocibundle.go
ricardomaraschini: Yeah, another case where the "load everything in the directory" could help out.

ricardomaraschini:
@twz123 @jnummelin  I think this is ready for another round of review. Let me know how it looks like now. Thanks.

ajp-io:
@twz123 @jnummelin Are you able to review this PR again sometime soon? We have a core feature that has a significant known issue until we get this fixed. Thanks!

pkg/component/worker/ocibundle.go
```diff
@@ -46,60 +54,153 @@ var _ manager.Component = (*OCIBundleReconciler)(nil)
...
+				continue
+			}
+			return
+		}
+	}()
+
+	go func() {
+		a.log.Infof("Started to watch events on %s", a.k0sVars.OCIBundleDir)
+		_ = debouncer.Run(ctx)
+		watcher.Close()
```
twz123, line 183: This is returning an error that could be logged. I wonder why the linter didn't catch this.

pkg/component/worker/ocibundle.go
twz123: Alright!

pkg/component/worker/ocibundle.go
```diff
@@ -46,60 +54,153 @@ var _ manager.Component = (*OCIBundleReconciler)(nil)
...
+		a.log.Info("OCI bundle watch bouncer ended")
+		a.end <- struct{}{}
```
twz123, line 185: This would be even more robust if we'd have this as `defer close(a.end)` at the top of the func. That way, `a.end` wouldn't need to be buffered as well.

twz123 (approved):
Looks good! Left a few small nits to be improved, if there's time and inclination.

pkg/component/worker/ocibundle.go
ricardomaraschini: Implemented. Please take a look.

pkg/component/worker/ocibundle.go
ricardomaraschini: Implemented. Please take a look.

twz123 (approved):

ricardomaraschini:
@jnummelin @twz123 Thanks for your help so far, I really appreciate it. Are there any plans to back port this to v1.29 / v1.28 ? If needed I can help with that.

jnummelin:
We don't usually backport new features. One could argue though that is this one. 😆 @twz123 WDYT, could/should we backport this?

jnummelin:
@ricardomaraschini Thanks for working on this 👍

ricardomaraschini:
@jnummelin @twz123 Is there any chance of back porting this to v1.29 and v1.28 release branches ? This would be so valuable for our use case here. I tried to create a back port myself but, IIUC, I need to add the `backport/release-1.2{8,9}` label here and I don't have enough permissions to do so. Welp.

twz123:
Actually, I'm a pretty conservative guy when it comes to backports.  I wouldn't want to port it back so much because it's a pretty substantial change, adding new behavior, not fixing a broken one. Shipping your on k0s builds with the backport applied wouldn't be an option for you in this case? (Just asking :upside_down_face:)

twz123:
Another thing that comes to my mind: Eventually, we need to have a way to re-enable GC on containerd side for images that have previously been imported via k0s, but are no longer present. Otherwise, the images will pile up on the image fs, potentially leading to disk pressure in the long run.

PR #4332, opened 2024-04-23
Fix the typing of Helm chart timeout duration

### LINKED ISSUE #4172: Chart.spec.extensions.helm.charts.timeout only accepts numerical value
k0s version: v1.29.2+k0s.0
k0sctl version: version: v0.17.4 commit: 372a589

Given the following config:
```
# generated-by-k0sctl 2024-03-11T14:06:29-07:00
kind: ClusterConfig
spec:
  api:
    address: 10.81.34.223
    sans:
    - 10.81.34.223
    - 10.81.34.225
    - 10.81.34.224
    - 127.0.0.1
  extensions:
    helm:
      charts:
      - chartname: appscode/gateway-api
        name: gateway-api
        namespace: kube-system
        order: 2
        version: v1.0.0
      - chartname: metallb/metallb
        name: metallb
        namespace: metallb
        order: 1
      - chartname: oci://[redacted private registry].com/helm-charts/nginx-ingress-controller
        name: nginx-ingress-controller
        namespace: nginx-ingress
        order: 3
        timeout: 100s
```
Config validation reports the following error:
```
Error: failed to read config: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go struct field Chart.spec.extensions.helm.charts.timeout of type time.Duration
```
Changing that setting to `10000000` results in a valid config.

According to the [documentation](https://github.com/k0sproject/k0s/blob/b646e9fe45c33f9d370d727c4e7caee263240c88/docs/helm-charts.md?plain=1#L58) this field should accept a user friendly setting.

jnummelin:
## Description

The intention and docs has always been to allow e.g. 10m as the timeout value. However, due to incorrect typing used it only accepted numeric values in nanosecs. :D

This fixes the typing to use k8s `k8s.io/apimachinery/pkg/apis/meta/v1/Duration` type so it'll be correctly parsed.

Fixes #4172

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


kke:
This will of course give an error if someone already uses it without a unit.

Something like this could make it backwards compatible:

```go
type metav1Duration struct {
	metav1.Duration
}

func (m *metav1Duration) UnmarshalJSON(data []byte) error {
	var v any
	if err := json.Unmarshal(data, &v); err != nil {
		return err
	}

	switch value := v.(type) {
	case float64:
		nsData := []byte(strconv.Itoa(int64(value)) + "ns")
		return m.Duration.UnmarshalJSON(nsData)
	default:
		return m.Duration.UnmarshalJSON(data)
	}
}
```

But that may confuse the generator?

twz123 (approved):
I'm usually very reluctant to break APIs like this. But this was such an obvious oversight to begin with that I doubt anyone got it right, and may very well benefit from getting validation errors.

hedge-sparrow:
hate to dredge this one up months later but since this was a breaking change to the clusterconfig api, it might have been nicer to release this with a bump to the CR apiVersion so that software that interacts with the clusterconfig object could traverse versions without having a hard incompatibility.

PR #4344, opened 2024-04-29
Implement keepalived load balancer 

juanluisvaladas:
## Description

For easier review I split the PR in 5 commits:
1- Add API types and autogenerated code
2- Controller changes to make keepalived work
3- Updated inttest
4- Encapsulate keepalived config in a new struct
5- Documentations

I also found a typo so I sneaked in a tiny 1 line commit

This is part of #4181

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


cmd/controller/controller.go
```diff
@@ -232,11 +232,16 @@ func (c *command) start(ctx context.Context) error {
 		if c.SingleNode {
 			return errors.New("control plane load balancing cannot be used in a single-node cluster")
 		}
+		if cplb.Type != v1beta1.CPLBTypeKeepalived {
```
jnummelin, line 235: IMO we should do this in config validation already.

docs/cplb.md
```diff
@@ -3,17 +3,23 @@
 For clusters that don't have an [externally managed load balancer](high-availability.md#load-balancer) for the k0s
 control plane, there is another option to get a highly available control plane called control plane load balancing (CPLB).
 
-CPLB allows automatic assigned of predefined IP addresses using VRRP across masters.
+CPLB has two features that often will be combined, but normally will be used together: VRRP Instances, which allows
```
jnummelin, line 6: > CPLB has two features that often will be combined, but normally will be used together

This doesn't sound right 😄

docs/cplb.md
```diff
@@ -46,19 +52,23 @@ following:
   These do not provide any sort of security against ill-intentioned attacks, they are
   safety features to prevent accidental conflicts between VRRP instances in the same
   network segment.
+* If `VirtualServers` are used, the cluster configuration doesn't specify a non-empty
+  [`spec.api.externalAddress`][specapi]. `VRRPInstances` are compatible.
```
jnummelin, line 56: hmm, I don't really get what this means. Could you rephrase this a bit

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -34,9 +34,34 @@ type ControlPlaneLoadBalancingSpec struct {
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
 
+	// type indicates the type of the node-local load balancer to deploy on
+	// worker nodes. Currently, the only supported type is "Keepalived".
+	// +kubebuilder:default=Keepalived
+	// +optional
+	Type CPLBType `json:"type,omitempty"`
```
jnummelin, line 41: Do we really need the type at all since we have only one?

pkg/component/controller/cplb_reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+				switch event.Type {
+				case watch.Added, watch.Modified:
+					endpoint, ok := event.Object.(*corev1.Endpoints)
+					if ok {
+						r.log.Info("Kubernetes endpoint was updated")
+						r.maybeUpdateIPs(endpoint)
+					}
+				case watch.Deleted:
+					r.log.Error("Kubernetes endpoint was deleted, this is not supposed to happen. Ignoring the event")
```
jnummelin, line 112: Wouldn't this happen if one of the controllers get's e.g. re-created with different IP?

pkg/component/controller/cplb_reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (r *CPLBReconciler) Stop() {
+	r.log.Info("Stopping CPLB reconciler")
+	r.watchCancelFunc()
+}
+
+func (r *CPLBReconciler) watchAPIServers(ctx context.Context) {
+	// Before starting check that the API is actually responding
+	for {
```
jnummelin, line 83: Maybe we could utilize our internal `watch` helper here? IMO would be much simpler and that already handles retries etc.

pkg/component/controller/cplb_reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+				r.log.Info("Stopped watching kubernetes endpoints")
+				return
+			}
+		}
+	}
+}
+
+// maybeUpdateIPs updates the list of IP addresses if the new list has
+// different addresses
+func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
```
jnummelin, line 129: Do we really need to compare the lists and store the "state" internally? What if we'd do just a signal whenever there's changes seen from watch and then always reload all the addreses from the endpoint object itself? I mean they are not changing very often in general and thus keeping a cache kinda feels bit wonky.

pkg/component/controller/cplb_unix.go
```diff
@@ -71,20 +78,45 @@ func (k *Keepalived) Init(_ context.Context) error {
...
 		return nil
 	}
 
-	if err := k.configureDummy(); err != nil {
-		return fmt.Errorf("failed to configure dummy interface: %w", err)
+	if len(k.Config.VRRPInstances) > 0 {
+		if err := k.configureDummy(); err != nil {
+			return fmt.Errorf("failed to configure dummy interface: %w", err)
+		}
+		if err := k.Config.ValidateVRRPInstances(nil); err != nil {
```
jnummelin, line 90: Again, IMO we need to hook validation to general config validation "phase"

pkg/component/controller/cplb_unix.go
```diff
@@ -71,20 +78,45 @@ func (k *Keepalived) Init(_ context.Context) error {
...
+			return fmt.Errorf("failed to validate VRRP instances: %w", err)
+		}
+	}
+
+	if len(k.Config.VirtualServers) > 0 {
+		if k.APISpec.ExternalAddress != "" {
```
jnummelin, line 96: Again, IMO we need to hook validation to general config validation "phase"

pkg/component/controller/cplb_unix.go
```diff
@@ -71,20 +78,45 @@ func (k *Keepalived) Init(_ context.Context) error {
...
+			return errors.New("externalAddress is not supported with virtual servers")
+		}
+		if err := k.Config.ValidateVirtualServers(); err != nil {
```
jnummelin, line 99: Again, IMO we need to hook validation to general config validation "phase"

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -81,50 +106,148 @@ type VirtualIPs []string
 
 // ValidateVRRPInstances validates existing configuration and sets the default
```
jnummelin, line 107: Validating changing the state sounds bit bogus. Could we hook defaulting into the config loading "phase" like we do for other parts of the config?

juanluisvaladas:
Converting to draft while I address the concerns.

pkg/apis/k0s/v1beta1/cplb.go
twz123: This makes the API future-proof. Maybe there's alternatives to keepalived. Having the type here makes adding new stuff easier. Folks don't have to specify it in the YAMLs since Keepalived is the default. We do similar things for NLLB.

pkg/component/controller/cplb_unix.go
```diff
@@ -43,13 +45,18 @@ import (
...
 // Keepalived is the controller for the keepalived process in the control plane load balancing
 type Keepalived struct {
-	K0sVars         *config.CfgVars
-	Config          *k0sAPI.ControlPlaneLoadBalancingSpec
-	DetailedLogging bool
-	uid             int
-	supervisor      *supervisor.Supervisor
-	log             *logrus.Entry
-	configFilePath  string
+	K0sVars          *config.CfgVars
```
twz123, line 48: I'd sort all public fields before the private ones.

pkg/supervisor/supervisor.go
```diff
@@ -302,5 +302,8 @@ func getEnv(dataDir, component string, keepEnvPrefix bool) []string {
 func (s *Supervisor) GetProcess() *os.Process {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
+	if s.cmd == nil {
```
twz123, line 305: Hmm, is it okay to call GetProcess when the supervisor isn't started?

pkg/component/controller/cplb_reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+	case r.updateCh <- true:
+	default:
+	}
+}
+
+// GetIPs gets a thread-safe copy of the current list of IP addresses
+func (r *CPLBReconciler) GetIPs() []string {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.addresses
```
twz123, line 174: Might be a good idea to clone the slice here, so that there's no surprises when callers start to modify it.

```suggestion
	return slices.Clone(r.addresses)
```

pkg/apis/k0s/v1beta1/cplb.go
jnummelin: 👍

pkg/component/controller/cplb_reconciler.go
jnummelin: eh, no, it wouldn't happen. I guess the only way this to happen is someone actually manually deletes `endpoints/kubernetes` object.

docs/cplb.md
```diff
@@ -142,13 +158,18 @@ spec:
     config:
       spec:
         api:
-          externalAddress: 192.168.122.200
+          sans:
```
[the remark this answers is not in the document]
jnummelin: I think we should

cmd/controller/controller.go
juanluisvaladas: Agreed, this is done in a separate commit because rewriting the git history was quite complex...

docs/cplb.md
juanluisvaladas: It doesn't 😅 . Fixed

docs/cplb.md
juanluisvaladas: I rephrased it, please review it again, I think it's better now but I'm not quite convinced to be honest

pkg/apis/k0s/v1beta1/cplb.go
juanluisvaladas: Since we're all in the same page I'm marking this as resolved

pkg/component/controller/cplb_reconciler.go
juanluisvaladas: Changed this as a separate commit. Can be squashed.

pkg/component/controller/cplb_reconciler.go
juanluisvaladas: I had already implemented this in a slightly different way but it's pretty much the same. If you feel strongly about this instead of what I've done let me know and I'll replace it.

pkg/component/controller/cplb_unix.go
juanluisvaladas: Agreed, it's added as a separate commit because rewriting the git history wasn't easy...

pkg/component/controller/cplb_unix.go
juanluisvaladas: Same

pkg/component/controller/cplb_unix.go
juanluisvaladas: I'm resolving all of these and leave just one

pkg/component/controller/cplb_reconciler.go
twz123: No, all good!

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -81,50 +106,148 @@ type VirtualIPs []string
...
+	// LBAlgo is the load balancing algorithm. If not specified, defaults to rr.
+	// Valid values are rr, wrr, lc, wlc, lblc, dh, sh, sed, nq. For further
+	// details refer to keepalived documentation.
+	LBAlgo KeepalivedLBAlgo `json:"lbAlgo,omitempty"`
+	// LBKind is the load balancing kind. If not specified, defaults to DR.
+	// Valid values are NAT DR TUN. For further details refer to keepalived documentation.
+	LBKind KeepalivedLBKind `json:"lbKind,omitempty"`
+	// PersistenceTimeout specify a timeout value for persistent connections in
+	// seconds. If not specified, defaults to 360 (6 minutes).
+	PersistenceTimeout int `json:"persistenceTimeout,omitempty"`
```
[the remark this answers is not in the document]
juanluisvaladas: I have just seen this after doing this change...  I added a check that persistenceTimeoutSeconds is not negative.
I discarded previously `metav1.Duration` precisely it allows fractions of seconds so I think this is better.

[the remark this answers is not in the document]
twz123: Alright. Would you mind to rename it then?

```suggestion
	// kubebuilder:validation:Minimum=0
	//+kubebuilder:default=360
	// +optional
	PersistenceTimeoutSeconds int `json:"persistenceTimeoutSeconds,omitempty"`
```

pkg/supervisor/supervisor.go
juanluisvaladas: It is now :) prior to checking it the code would panic more times than it didn't.

Also it's thread-safe because s.cmd is only written in: https://github.com/k0sproject/k0s/blob/fa80f8b1fe3f76ae8d67b570a3bff53f3ac25f55/pkg/supervisor/supervisor.go#L168-L178

docs/cplb.md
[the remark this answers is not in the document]
juanluisvaladas: VERY good point, I'll do this.

pkg/supervisor/supervisor.go
twz123: My question was more of the sorts: What is Supervisor's contract? Or what should Supervisor's contract be? Is it a legal thing to call GetProcess before Supervise has been called? There's no go docs about this. I see that we could very well signal the wrong PID in certain circumstances, i.e. the cmd won't be nilled out after the process terminated. Anyhow, that's something that can be tackled another time.

> prior to checking it the code would panic more times than it didn't.

Not sure what you mean.

pkg/component/controller/cplb_unix.go
```diff
@@ -290,16 +312,47 @@ func (k *Keepalived) generateKeepalivedTemplate() error {
...
 
+func (k *Keepalived) watchReconcilerUpdates() {
+	// Wait for the supervisor to start keepalived before
+	// watching for endpoint changes
+	process := k.supervisor.GetProcess()
+	for process == nil {
+		k.log.Info("Waiting for keepalived to start")
+		time.Sleep(5 * time.Second)
+		process = k.supervisor.GetProcess()
+	}
```
twz123, line 323: It might work without this loop, if the nil check would be moved into the update loop.

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -81,50 +106,148 @@ type VirtualIPs []string
...
+// +kubebuilder:validation:Enum=NAT;DR;TUN
+type KeepalivedLBKind string
+
+const (
+	NATLBKind KeepalivedLBKind = "NAT"
+	DRLBKind  KeepalivedLBKind = "DR"
+	TUNLBKind KeepalivedLBKind = "TUN"
+)
+
+type RealServer struct {
```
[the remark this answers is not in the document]
juanluisvaladas: These are used in ControlPlaneLoadBalancingSpec.validateVirtualServers()

pkg/component/controller/cplb_reconciler.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func NewCPLBReconciler(kubeconfigPath string, updateCh chan<- bool) *CPLBReconciler {
+	return &CPLBReconciler{
+		log:            logrus.WithField("component", "cplb-reconciler"),
+		kubeconfigPath: kubeconfigPath,
+		updateCh:       updateCh,
+	}
+}
+func (r *CPLBReconciler) getClientSet() error {
```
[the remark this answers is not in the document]
juanluisvaladas: As you suggested later, I inlined it and passed the clientset as an argument to watchAPIServers. So this function is entirely removed now

pkg/component/controller/cplb_unix.go
```diff
@@ -108,6 +140,10 @@ func (k *Keepalived) Start(_ context.Context) error {
 		DataDir: k.K0sVars.DataDir,
 		UID:     k.uid,
 	}
+
+	if len(k.Config.VirtualServers) > 0 {
+		go k.watchReconcilerUpdates()
```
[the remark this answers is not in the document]
juanluisvaladas: Nope, thanks, I fixed it by closing the updateCh channel in watchAPIServers once the context is cancelled.

pkg/component/controller/cplb_unix.go
```diff
@@ -290,16 +321,47 @@ func (k *Keepalived) generateKeepalivedTemplate() error {
...
+			k.log.Errorf("failed to generate keepalived template: %v", err)
+			continue
+		}
+
+		process := k.supervisor.GetProcess()
+		if err := process.Signal(syscall.SIGHUP); err != nil {
+			k.log.Errorf("failed to send SIGHUP to keepalived: %v", err)
+		}
+	}
+	k.log.Info("stopped watching cplb-reconciler updates")
```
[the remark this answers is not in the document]
juanluisvaladas: Indeed, should be fixed now.

pkg/supervisor/supervisor.go
juanluisvaladas: > Not sure what you mean.

Before adding this condition, `s.GetProcess()` could be called before `s.cmd != nil` and this resulted in a panic, of course it was racy, but it happened extremely more than 50% of the times.

pkg/apis/k0s/v1beta1/cplb.go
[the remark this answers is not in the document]
twz123: Are they? Still not able to spot them :see_no_evil:

pkg/component/controller/cplb_unix.go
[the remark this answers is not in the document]
twz123: I cannot spot it. Can you verify?

pkg/component/controller/cplb_unix.go
twz123: > If you feel strongly about it, I will change it, but I prefer it this way.

Currently, the loop cannot be cancelled externally. That's why I figured that we could just inline the nil check below. Would be less code, too. I'm fine with keeping it, as long as it can be cancelled.

> it will just return a process with an invalid PID

Right. I'd argue that this is a current shortcoming of Supervisor, though.

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -34,21 +34,41 @@ type ControlPlaneLoadBalancingSpec struct {
 	// +optional
 	Enabled bool `json:"enabled,omitempty"`
 
+	// type indicates the type of the node-local load balancer to deploy on
+	// worker nodes. Currently, the only supported type is "Keepalived".
```
twz123, line 38:
```suggestion
	// type indicates the type of the control plane load balancer to deploy on
	// controller nodes. Currently, the only supported type is "Keepalived".
```

pkg/component/controller/cplb_unix.go
```diff
@@ -389,7 +394,7 @@ vrrp_instance k0s-vip-{{$i}} {
 {{ if gt (len $RealServers) 0 }}
 {{ range .VirtualServers }}
 virtual_server {{ .IPAddress }} {{ $APIServerPort }} {
-    delay_loop {{ .DelayLoop }}
+    delay_loop {{ delay_loop_str .DelayLoop }}
```
twz123, line 397: Wouldn't this "just work"?

```suggestion
    delay_loop {{ .DelayLoop.Seconds }}
```

juanluisvaladas: It does 🤦

juanluisvaladas:
Everything should be addressed now

pkg/component/controller/cplb_unix.go
juanluisvaladas: @twz123 I accidentally editted your comment with my reply 🤦 

The reply is: 

I added a limit of 6 times, that's 30 seconds and should be way more than enough time to start keepalived. It gets cancelled eventually, it's not the prettiest but should fix the problem.

Unfortunatelly I deleted your comment saying that it had to be possible to cancel it from the outside.

pkg/component/controller/cplb_unix.go
```diff
@@ -290,39 +317,100 @@ func (k *Keepalived) generateKeepalivedTemplate() error {
...
 }
 {{ end }}
-`
+
+{{ $APIServerPort := .APIServerPort }}
+{{ $RealServers := .RealServers }}
+{{ if gt (len $RealServers) 0 }}
+{{ range .VirtualServers }}
+virtual_server {{ .IPAddress }} {{ $APIServerPort }} {
+    delay_loop {{ .DelayLoop.Seconds }}
```
twz123, line 396:
```suggestion
    {{- if gt .DelayLoop.Seconds 0.0 }}
    delay_loop {{ .DelayLoop.Seconds }}
    {{- end }}
```

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -79,52 +102,192 @@ type VRRPInstance struct {
...
+// VirtualServer defines the configuration options for a virtual server.
+type VirtualServer struct {
+	// IPAddress is the virtual IP address used by the virtual server.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	IPAddress string `json:"ipAddress"`
+	// DelayLoop is the delay timer for check polling. If not specified, defaults to 0.
+	// DelayLoop is defined in microseconds. Further precision will be truncated without
+	// warnings.
+	// kubebuilder:validation:Minimum=0
```
twz123, line 166:
```suggestion
	// DelayLoop is the delay timer for check polling. DelayLoop accepts
	// microsecond precision. Further precision will be truncated without
	// warnings.
	// +optional
```

docs/configuration.md
```diff
@@ -304,26 +304,47 @@ node-local load balancing.
...
+
+Configuration options required for using VRRP to configure VIPs in control plane load balancing.
+
+| Element           | Description                                                                                                                      |
+| ----------------- | -------------------------------------------------------------------------------------------------------------------------------- |
+| `ipAddress`          | The load balancer's listen address.                                                                                           |
+| `delayLoop`          | Delay timer for health check polling in seconds. Default: `0`.                                                                |
+| `lbAlgo`             | Algorithm used by keepalived. Supported algorithms: `rr`, `wrr`, `lc`, `wlc`, `lblc`, `dh`, `sh`, `sed`, `nq`. Default: `rr`. |
+| `lbKind`             | Kind of ipvs load balancer. Supported values: `NAT`, `DR`, `TUN`  Default: `DR`.                                              |
+| `persistenceTimeout` | Timeout value for persistent connections in seconds. Default: `360` (6 minutes).                                              |
```
twz123, line 346:
```suggestion
| Element              | Description                                                                                                                                   |
| -------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| `ipAddress`          | The load balancer's listen address.                                                                                                           |
| `delayLoop`          | Delay timer for check polling. DelayLoop accepts microsecond precision. Further precision will be truncated without warnings. Example: `10s`. |
| `lbAlgo`             | Algorithm used by keepalived. Supported algorithms: `rr`, `wrr`, `lc`, `wlc`, `lblc`, `dh`, `sh`, `sed`, `nq`. Default: `rr`.                 |
| `lbKind`             | Kind of ipvs load balancer. Supported values: `NAT`, `DR`, `TUN`  Default: `DR`.                                                              |
| `persistenceTimeout` | Timeout value for persistent connections in seconds. Default: `360` (6 minutes).                                                              |
```

twz123 (commented):
One last thing™. I think we can just omit the delay_loop if it's zero.

pkg/component/controller/cplb_unix.go
juanluisvaladas: We can't in keepalived it defaults to 60. I set it to 0 because I think it doesn't make sense to delay it at all in CPLB because it's added after kubernetes.default.svc is reconciled and hence has passed all the relevant health local health checks.

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -79,52 +102,185 @@ type VRRPInstance struct {
...
+		if k.VirtualServers[i].DelayLoop.Duration < time.Duration(0) {
+			errs = append(errs, errors.New("DelayLoop must be a positive"))
+		}
+	}
+	return errs
+}
+
+// Validate validates the ControlPlaneLoadBalancingSpec
+func (c *ControlPlaneLoadBalancingSpec) Validate(externalAddress string) []error {
+	if c == nil || !c.Enabled {
```
twz123, line 265: We should validate the config in any case, even if it's disabled.

```suggestion
	if c == nil {
```

pkg/component/controller/cplb_unix.go
twz123: > We can't in keepalived it defaults to 60.

I see. Apparently I found yet another delay_loop in the keepalived codebase :sweat_smile:

> I set it to 0 because I think it doesn't make sense to delay it at all in CPLB because it's added after kubernetes.default.svc is reconciled and hence has passed all the relevant health local health checks.

I don't fully understand the implications, but a delay_loop of 0 is not a thing in keepalived (it will use the default of 60, then):

```console
# /var/lib/k0s/bin/keepalived --dont-fork --use-file /run/k0s/keepalived.conf --no-syslog --log-console --log-detail --dump-conf -t
(/run/k0s/keepalived.conf: Line 33) number '0' outside range [0.000001, 18446744073709.551614]
(/run/k0s/keepalived.conf: Line 33) virtual server delay loop '0' invalid - ignoring
```

But letz address this in a subsequent PR :smile:

twz123 (approved):

PR #4416, opened 2024-05-14
Use a versioned worker config stack

twz123:
## Description

The worker config ConfigMaps have the Kubernetes version in their names. This is to ensure a seamless upgrade path when upgrading minor k0s versions. However, the RBAC resources and the stack name itself didn't have the version in their names. This means that as soon as a controller with the new minor k0s version takes over, it removes all previous worker ConfigMaps from the cluster, effectively defeating the purpose of the versioned names.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/workerconfig/reconciler.go
+++ b/pkg/component/controller/workerconfig/reconciler.go
@@ -143,7 +143,7 @@ func (r *Reconciler) Init(context.Context) error {
 			return err
 		}
 		return (&applier.Stack{
-			Name:      "k0s-" + constant.WorkerConfigComponentName,
+			Name:      fmt.Sprintf("k0s-%s-%s", constant.WorkerConfigComponentName, constant.KubernetesMajorMinorVersion),
 			Client:    dynamicClient,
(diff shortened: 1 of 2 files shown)


pkg/component/controller/workerconfig/reconciler.go
```diff
@@ -143,7 +143,7 @@ func (r *Reconciler) Init(context.Context) error {
 			return err
 		}
 		return (&applier.Stack{
-			Name:      "k0s-" + constant.WorkerConfigComponentName,
+			Name:      fmt.Sprintf("k0s-%s-%s", constant.WorkerConfigComponentName, constant.KubernetesMajorMinorVersion),
```
kke, line 146: This string can be a constant, same with the `system:bootstrappers:...` below. With `fmt.Sprintf` it's regenerated/allocated at runtime. This is in `Init` so maybe it doesn't matter much though.

jnummelin (approved):

Issue #4411, opened 2024-05-13
kube-router failed to start when installation with default settings


bizhao:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 5.15.0-106-generic #116-Ubuntu SMP Wed Apr 17 09:17:56 UTC 2024 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.30.0+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Total memory: 15.6 GiB (pass)
Disk space available for /var/lib/k0s: 74.7 GiB (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 5.15.0-106-generic (pass)
  Max. file descriptors per process: current: 1048576 / max: 1048576 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
    cgroup controller "memory": available (is a listed root controller) (pass)
    cgroup controller "devices": available (device filters attachable) (pass)
[...]

PR #4420, opened 2024-05-15
Add readinessProbe/minReadySeconds to kube-router

twz123:
## Description

This allows for better feedback of kube-router health via the DaemonSet resource. Without those, it's possible to observe a "healthy" DaemonSet, even if it's not. This affects e.g. rolling updates, and, most notably k0s's own integration tests.

See:
* #4411
* #4421

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/kuberouter.go
+++ b/pkg/component/controller/kuberouter.go
@@ -243,6 +243,7 @@ spec:
     spec:
       priorityClassName: system-node-critical
       serviceAccountName: kube-router
+      minReadySeconds: 5
       initContainers:
         - name: install-cni-bins
           image: {{ .CNIInstallerImage }}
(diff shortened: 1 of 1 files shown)


pkg/component/controller/kuberouter.go
```diff
@@ -321,12 +322,25 @@ spec:
...
+        ports:
+        - name: healthz
+          containerPort: 20244
         livenessProbe:
           httpGet:
             path: /healthz
-            port: 20244
-          initialDelaySeconds: 10
+            port: healthz
+          initialDelaySeconds: 300
```
jnummelin, line 332: Seems bit excessive? What's the reasoning for such a long delay?

twz123: In my experience, liveness probes are only helpful in very few cases. Forcefully restarting a container over and over again is usually not helping much and will just increase churn/load on a cluster that is probably already busy with other things that lead to healthz answering with non-2xx responses. That's why I prefer high timeouts here. Henning [wrote a blog post about this](https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html) back in the day.

twz123: The basic difference here is that an app usually reports failures that it might be able to recover from by itself via the readiness endpoint. Restarting the app won't help and might even worse the situation in such a case. Unrecoverable errors should make an app terminate itself. This leaves the liveness probe to detect situations in which the app itself is broken due to things like deadlocks, tight endless loops, blocked on system calls that usually don't block for too long.

jnummelin: 👍

jnummelin (approved):

PR #4434, opened 2024-05-16
Make k0s reset fail if it can't reach containerd

juanluisvaladas:
## Description

Prior to this commit, if the containerd unix socket wasn't listening grpc.Dial would try to connect forever.

This makes the dial retry about 3 times and we retry the whole ListPods operation 3 times and after that we assume this step has failed.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/cleanup/containers.go
+++ b/pkg/cleanup/containers.go
@@ -17,6 +17,7 @@ limitations under the License.
 package cleanup
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -134,20 +135,20 @@ func (c *containers) stopContainerd() {
 
 func (c *containers) stopAllContainers() error {
 	var errs []error
-	logrus.Debugf("trying to list all pods")
 
 	var pods []string
+	ctx := context.TODO()
 	err := retry.Do(func() error {
+		logrus.Debugf("trying to list all pods")
 		var err error
-		pods, err = c.Config.containerRuntime.ListContainers()
+		pods, err = c.Config.containerRuntime.ListContainers(ctx)
 		if err != nil {
 			return err
 		}
 		return nil
-	})
+	}, retry.Context(ctx), retry.LastErrorOnly(true))
 	if err != nil {
-		logrus.Debugf("failed at listing pods %v", err)
-		return err
+		return fmt.Errorf("failed at listing pods %w", err)
 	}
 	if len(pods) > 0 {
 		if err := removeMount("kubelet/pods"); err != nil {
@@ -160,7 +161,7 @@ func (c *containers) stopAllContainers() error {
 
 	for _, pod := range pods {
 		logrus.Debugf("stopping container: %v", pod)
-		err := c.Config.containerRuntime.StopContainer(pod)
+		err := c.Config.containerRuntime.StopContainer(ctx, pod)
 		if err != nil {
 			if strings.Contains(err.Error(), "443: connect: connection refused") {
 				// on a single node instance, we will see "connection refused" error. this is to be expected
@@ -170,13 +171,13 @@ func (c *containers) stopAllContainers() error {
 				errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
 			}
 		}
-		err = c.Config.containerRuntime.RemoveContainer(pod)
+		err = c.Config.containerRuntime.RemoveContainer(ctx, pod)
 		if err != nil {
 			errs = append(errs, fmt.Errorf("failed to remove pod %s: %w", pod, err))
 		}
 	}
 
-	pods, err = c.Config.containerRuntime.ListContainers()
+	pods, err = c.Config.containerRuntime.ListContainers(ctx)
 	if err == nil && len(pods) == 0 {
 		logrus.Info("successfully removed k0s containers!")
 	}
--- a/pkg/container/runtime/cri.go
+++ b/pkg/container/runtime/cri.go
@@ -32,7 +32,7 @@ type CRIRuntime struct {
 	criSocketPath string
 }
 
-func (cri *CRIRuntime) ListContainers() ([]string, error) {
+func (cri *CRIRuntime) ListContainers(ctx context.Context) ([]string, error) {
 	client, conn, err := getRuntimeClient(cri.criSocketPath)
 	defer closeConnection(conn)
 	if err != nil {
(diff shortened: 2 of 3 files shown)


pkg/cleanup/containers.go
```diff
@@ -134,19 +134,19 @@ func (c *containers) stopContainerd() {
...
 		if err != nil {
 			return err
 		}
 		return nil
-	})
+	}, retry.Attempts(3))
 	if err != nil {
-		logrus.Debugf("failed at listing pods %v", err)
+		logrus.Errorf("failed at listing pods %v", err)
 		return err
```
twz123, line 150: What about  removing that log completely, since the err gets returned to the caller anyways?

```suggestion
		return fmt.Errorf("failed at listing pods %w", err)
```

pkg/container/runtime/cri.go
```diff
@@ -105,10 +106,12 @@ func getRuntimeClient(addr string) (pb.RuntimeServiceClient, *grpc.ClientConn, e
 }
 
 func getRuntimeClientConnection(addr string) (*grpc.ClientConn, error) {
-	conn, err := grpc.Dial(addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(5*time.Second))
+	defer cancel()
+	conn, err := grpc.DialContext(ctx, addr, grpc.WithBlock(), grpc.WithTransportCredentials(insecure.NewCredentials()))
```
twz123, line 111: My IDE is telling me `grpc.DialContext` is deprecated:

> Deprecated: use NewClient instead. Will be supported throughout 1.x.

pkg/container/runtime/cri.go
```diff
@@ -105,10 +106,12 @@ func getRuntimeClient(addr string) (pb.RuntimeServiceClient, *grpc.ClientConn, e
 }
 
 func getRuntimeClientConnection(addr string) (*grpc.ClientConn, error) {
-	conn, err := grpc.Dial(addr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(5*time.Second))
```
twz123, line 109: Did you consider to add proper context propagation here, i.e. adding a context to, say, `StopContainer` and friends to use it here?

pkg/cleanup/containers.go
```diff
@@ -134,19 +134,19 @@ func (c *containers) stopContainerd() {
...
 	err := retry.Do(func() error {
+		logrus.Debugf("trying to list all pods")
 		var err error
 		pods, err = c.Config.containerRuntime.ListContainers()
 		if err != nil {
 			return err
 		}
 		return nil
-	})
+	}, retry.Attempts(3))
```
twz123, line 147: The default of 10 times was too much, I guess? If we had a context here, we could support Ctrl+C from the console.

twz123 (dismissed):
> Prior to this commit, if the containerd unix socket wasn't listening grpc.Dial would try to connect forever.

What lead to this situation? Did containerd crash? I mean, the reset step is starting its own containerd instance just in order to be able to gracefully stop containers (and doesn't restart it in case it crashes. Maybe it should use the containerd component to do it "the usual way"? Maybe it should short-circuit the list-and-stop loop when it detects that containerd terminated?). Having containerd not listening on its socket file makes me think that we might have another problem here. Did you consider to use non-blocking dials? Not sure if a retry even makes sense when there's nothing listening. The only reason for retrying this is that containerd is still starting up didn't start to listen yet. On the other hand, a blocking dial would be preferable then.

Anyhow, /approve

juanluisvaladas:
After checking this further, `grpc.NewClient` ignores `grpc.WithBlock()`, so even if we pass a `grpc.WithContextDialer` it won't really have the effect we want.

I think the most sensible way of dealing with this is starting the connection in the background with `grpc.NewClient` and if it fails to connect, we'll get the error listing the pod sandbox. As there aren't practically any steps in between instantiating the client and using i, and it's instantiated on every request to the CRI I don't think this changes the logic at all, even if that bit is in the background.

Finally, I reverted the attempts to the default value of 10 the code is much faster and 10 doesn't take that long to fail anyway.

The context propagation for control+c is indeed useful, so I added it.

pkg/cleanup/containers.go
```diff
@@ -134,19 +135,20 @@ func (c *containers) stopContainerd() {
 
 func (c *containers) stopAllContainers() error {
 	var errs []error
-	logrus.Debugf("trying to list all pods")
 
 	var pods []string
+	ctx := context.Background()
```
twz123, line 140: Unfortunately, this is not sufficient to support Ctrl+C. A clean solution probably needs to use proper context propagation using a context that's setup somewhere at the beginning of the reset subcommand, which is hooked up to signal handlers that cancel it, e.g. by using `os/signal.NotifyContext(...)`. That's separate from this PR's intent, though, and can be followed up later on. We can probably leave the contexts alone here and add them at a later stage. If we'd keep them here, I'd suggest to indicate that this is not yet complete:

```suggestion
	ctx := context.TODO()
```

pkg/cleanup/containers.go
```diff
@@ -134,19 +135,20 @@ func (c *containers) stopContainerd() {
...
+		logrus.Debugf("trying to list all pods")
 		var err error
-		pods, err = c.Config.containerRuntime.ListContainers()
+		pods, err = c.Config.containerRuntime.ListContainers(ctx)
 		if err != nil {
 			return err
 		}
 		return nil
-	})
+	}, retry.Context(ctx))
```
twz123, line 149: Another useful option in this context might be LastErrorOnly, otherwise each error is collected and reported individually.

```suggestion
	}, retry.Context(ctx), retry.LastErrorOnly(true))
```

juanluisvaladas:
I made two changes:
1- Applied both suggestions. 
2- Removed gprc.WithBlock from grpc.NewClient. I apparently added it accidentally in the previous commit. This option is ignored in [grpc.NewClient](https://pkg.go.dev/google.golang.org/grpc#NewClient) so effectively this change isn't doing anything, but it shouldn't be there.

twz123 (approved):
Alright. Note to self: The culprit was that [`grpc.WithBlock()`](https://pkg.go.dev/google.golang.org/grpc@v1.63.2#WithBlock) made the call to `grpc.Dial` not return and retry forever internally, effectively making our own retry loop useless.

This can happen if the internal containerd fails to start or crashes, since there's no monitoring or supervision at all for the spawned process. Startup failures can happen, as k0s reset may very well be executed on nodes that are already broken in one way or another.

The same goes for an external runtime if one is configured but not behaving well.

@juanluisvaladas Would you mind rebasing this?

PR #4452, opened 2024-05-20
Allow CPLB to work with externalAddress

juanluisvaladas:
## Description

Allow this behavior if the endpoint-reconciler is disabled.

K0smotron needs to use external address for cluster API. The only limitation why we couldn't combine externalAddress and CPLB is the endpoint-reconciler, since it can be disabled, allow to run CPLB with externalAddress if the user manually disables it.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -228,24 +228,26 @@ func (c *command) start(ctx context.Context) error {
 		nodeComponents.Add(ctx, controllerLeaseCounter)
 	}
 
+	disableEndpointReconciler := !slices.Contains(c.DisableComponents, constant.APIEndpointReconcilerComponentName) &&
+		nodeConfig.Spec.API.ExternalAddress != ""
+
 	if cplb := nodeConfig.Spec.Network.ControlPlaneLoadBalancing; cplb != nil && cplb.Enabled {
 		if c.SingleNode {
 			return errors.New("control plane load balancing cannot be used in a single-node cluster")
 		}
 
 		nodeComponents.Add(ctx, &controller.Keepalived{
-			K0sVars:         c.K0sVars,
-			Config:          cplb.Keepalived,
-			DetailedLogging: c.Debug,
-			LogConfig:       c.Debug,
-			KubeConfigPath:  c.K0sVars.AdminKubeConfigPath,
-			APIPort:         nodeConfig.Spec.API.Port,
+			K0sVars:               c.K0sVars,
+			Config:                cplb.Keepalived,
+			DetailedLogging:       c.Debug,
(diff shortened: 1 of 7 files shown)


twz123:
> K0smotron needs to use external address for cluster API.

Can you elaborate? K0smotron can simply use the CPLB addresses without k0s using them. Is that backport really required?

juanluisvaladas:
> Can you elaborate? K0smotron can simply use the CPLB addresses without k0s using them.

Yes, but k0smotron would have to be modified as it's hardcoded to set the externalAddress. That's an option too.

> Is that backport really required?

From a purely technical point of view no, but this was requested by sales and it's important for one potential customer. I don't see many downsides to backporting so I'd say we should backport it.

twz123:
I think i slowly get what this is about :sweat_smile: Can you maybe write up an example use case to better explain what this PR actually unblocks?

pkg/component/controller/cplb_unix.go
```diff
@@ -92,6 +93,9 @@ func (k *Keepalived) Start(_ context.Context) error {
 	}
 
 	if len(k.Config.VirtualServers) > 0 {
+		if k.HasEndpointReconciler {
```
twz123, line 96: What about moving that check out the component into the controller command? Otherwise the `HasEndpointReconciler` flag is just some "I will fail to start" flag.

inttest/cplb/cplb_test.go
```diff
@@ -33,6 +33,8 @@ type keepalivedSuite struct {
 
 const haControllerConfig = `
 spec:
+  api:
```
twz123, line 36: Can we test both setups here? One with externalAddress and one with NLLB?

cmd/controller/controller.go
```diff
@@ -228,24 +228,26 @@ func (c *command) start(ctx context.Context) error {
 		nodeComponents.Add(ctx, controllerLeaseCounter)
 	}
 
+	disableEndpointReconciler := !slices.Contains(c.DisableComponents, constant.APIEndpointReconcilerComponentName) &&
```
twz123, line 231: This condition is also used at line 436/438 for the endpoint reconciler component itself. Would be cool to reuse this boolean there as well. And it could deserve a less ambiguous name: Basically it swaps endpoint reconcilers. What about naming this `preferExternalAddressEndpointReconciler`, `preferK0sEndpointReconciler` or sth. like that?

juanluisvaladas:
Closing this for now. Will reopen after merging userspace reverse proxy.

PR #4524, opened 2024-05-30
arch: add adr on autopilot basic auth + oras support

ricardomaraschini:
## Description

this pr adds an architecture decision record on adding oras and basic auth support on autopilot.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -36,9 +36,9 @@ type PlanResourceURL struct {
         // Sha256 provides an optional SHA256 hash of the URL's content for verification.
         Sha256 string `json:"sha256,omitempty"`
 
-        // ArtifactPullSecrets holds a reference to a secret where Docker or Basic Auth
-        // credentials are stored. We use these credentials when pulling the artifacts from
-        // the URL.
+        // ArtifactPullSecrets holds a reference to a secret where the credentials are
+        // stored. We use these credentials when pulling the artifacts from the provided
+        // URL using any of the supported protocols (http, https, and oci).
         ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
 
         // Insecure indicates whether certificates in the remote URL (if using TLS) can
@@ -59,7 +59,7 @@ type ArtifactPullSecret struct {
 }
 ```
 
-The secret pointed to by the provided `ArtifactPullSecret` property is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` (see below) or of type `Opaque` if protocols `http://` or `https://` are used.
+The secret pointed by the provided `ArtifactPullSecret` will be used for pulling artifacts using either HTTP[S] or OCI protocols and is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` or of type `Opaque` if protocols `http://` or `https://` are used (see below for details on the Secret layout).
 
 Example configuration for OCI:
 
@@ -79,6 +79,16 @@ sha256: e95603f167cce6e3cffef5594ef06785b3c1c00d3e27d8e4fc33824fe6c38a99
 artifactPullSecret:
   namespace: kube-system
   name: artifacts-basic-auth
+```
+
+Example configuration for HTTP:
+
+```yaml
+url: http://my.file.server/binaries/k0s-v1.30.1+k0s.0
+sha256: e95603f167cce6e3cffef5594ef06785b3c1c00d3e27d8e4fc33824fe6c38a99
+artifactPullSecret:
+  namespace: kube-system
+  name: artifacts-token-based-auth
 ```
 
 ### Secrets Layout
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -39,7 +39,7 @@ type PlanResourceURL struct {
         // ArtifactPullSecrets holds a reference to a secret where the credentials are
         // stored. We use these credentials when pulling the artifacts from the provided
         // URL using any of the supported protocols (http, https, and oci).
-        ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
+        ArtifactPullSecret *corev1.SecretReference `json:"artifactPullSecret,omitempty"`
 
         // InsecureSkipTLSVerify indicates whether certificates in the remote URL (if using
         // TLS) can be ignored.
@@ -47,19 +47,7 @@ type PlanResourceURL struct {
 }
 ```
 
-The `ArtifactPullSecret` property will be added and its struct will be defined as follow:
-
-```go
-type ArtifactPullSecret struct {
-      // Namespace of the secret.
-      Namespace string `json:"namespace"`
-
-      // Name of the secret.
-      Name string `json:"name"`
-}
-```
-
-The secret pointed by the provided `ArtifactPullSecret` will be used for pulling artifacts using either HTTP[S] or OCI protocols and is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` or of type `Opaque` if protocols `http://` or `https://` are used (see below for details on the Secret layout).
+`ArtifactPullSecret` property is of type `SecretReference` as defined by `k8s.io/api/core/v1` package. The secret pointed by the provided `ArtifactPullSecret` will be used for pulling artifacts using either HTTP[S] or OCI protocols and is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` or of type `Opaque` if protocols `http://` or `https://` are used (see below for details on the Secret layout).
 
 Example configuration for OCI:
 
@@ -130,7 +118,6 @@ Proposed
 ## Consequences
 
 - Users will have an additional protocol to be aware of.
-- Introduction of a new type (`ArtifactPullSecret`) when ideally existing types could be reused.
 - If the Secret referenced by `ArtifactPullSecret` does not exist, the download will fail.
 - Users need to be notified about different failure types (e.g., unreadable secret, invalid secret).
 - Additional configuration is required to handle authentication, ensuring secure access to resources.
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -1,4 +1,4 @@
-# ADR 1: Add Support for ORAS Protocol and Basic Authentication in Autopilot for Artifact Retrieval
+# ADR 1: Add Support for OCI Registry and Basic Authentication
 
 ## Context
 
@@ -8,7 +8,7 @@ Enhancing Autopilot to pull artifacts directly from registries will streamline w
 
 ## Decision
 
-Implement support in Autopilot for pulling artifacts, such as k0s binaries and image bundles, directly from a registry using the [ORAS](https://oras.land/docs/) protocol. Additionally, add support for basic authentication to ensure secure access to artifacts over HTTP.
+Implement support in Autopilot for pulling artifacts, such as k0s binaries and image bundles, directly from a registry using the [ORAS](https://oras.land/docs/) client. Additionally, add support for basic authentication to ensure secure access to artifacts over HTTP.
 
 ## Solution
 
@@ -59,7 +59,7 @@ type ArtifactPullSecret struct {
 
 The secret pointed to by the provided `ArtifactPullSecret` property is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` (see below) or of type `kubernetes.io/basic-auth` if protocols `http://` or `https://` are used .
 
-Example configuration for ORAS:
+Example configuration for OCI:
 
 ```yaml
 url: oci://my.registry/binaries/k0s:v1.30.1+k0s.0
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -108,7 +108,7 @@ data:
 
 - The `InsecureSkipTLSVerify` property is equivalent of defining `InsecureSkipTLSVerify` on a Go HTTP client.
 - The `InsecureSkipTLSVerify` property will be valid for both `oci://` and `https://` protocols.
-- If no protocol is defined, HTTP is used.
+- If no protocol is defined, HTTPS is used.
 - If no `ArtifactPullSecret` is defined, access will be anonymous (no authentication).
 
 ## Status
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -145,7 +145,8 @@ data:
   `InsecureSkipTLSVerify` on a Go HTTP client.
 - The `InsecureSkipTLSVerify` property will be valid for both `oci://` and
   `https://` protocols.
-- If no protocol is defined, HTTPS is used.
+- If a protocol is not specified or an incorrect one is provided, an error
+  state should be activated.
 - If no `SecretRef` is defined, access will be anonymous (no authentication).
 
 ## Status
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -41,9 +41,9 @@ type PlanResourceURL struct {
         // URL using any of the supported protocols (http, https, and oci).
         ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
 
-        // Insecure indicates whether certificates in the remote URL (if using TLS) can
-        // be ignored, aka InsecureSkipVerify.
-        Insecure bool  `json:"insecure,omitempty"`
+        // InsecureSkipTLSVerify indicates whether certificates in the remote URL (if using
+        // TLS) can be ignored.
+        InsecureSkipTLSVerify bool  `json:"insecureSkipTLSVerify,omitempty"`
 }
 ```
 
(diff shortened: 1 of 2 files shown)


jnummelin:
@ricardomaraschini you're missing sign-off for the commits

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,100 @@
...
+type ArtifactPullSecret struct {
+      // Namespace of the secret.
+      Namespace string `json:"namespace"`
+
+      // Name of the secret.
+      Name string `json:"name"`
+}
+```
+
+The secret pointed to by the provided `ArtifactPullSecret` property is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` (see below) or of type `kubernetes.io/basic-auth` if protocols `http://` or `https://` are used .
```
jnummelin, line 60: Why only basic auth? If we'd have the full `Authorization` header value as a secret we'd be supporting basically everything, right?

ricardomaraschini: I was attempting to use only known [secret types](https://kubernetes.io/docs/concepts/configuration/secret/#secret-types) and there isn't one regarding a different kind of authentication. 

It does not mean we can't introduce our Opaque thingy though.

ricardomaraschini: I can change this proposal and add that if the secret contains an `authorization` entry we will send its content in the `Authorization` http header.  What do you think ?

ricardomaraschini: I think I followed you. Let me know what you think of the updated proposal.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,127 @@
...
+        // URL is the URL of a downloadable resource.
+        URL string `json:"url"`
+
+        // Sha256 provides an optional SHA256 hash of the URL's content for verification.
+        Sha256 string `json:"sha256,omitempty"`
+
+        // ArtifactPullSecrets holds a reference to a secret where Docker or Basic Auth
+        // credentials are stored. We use these credentials when pulling the artifacts from
+        // the URL.
+        ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
```
jnummelin, line 42: Am I assuming correct that his would then apply to the existing http protocol too? If so, maybe write it up a bit

ricardomaraschini: Yes, you are correct. This is the change I have introduced, let me know if it suffices: https://github.com/k0sproject/k0s/pull/4524/commits/89335cb350296a3bff5aeccaa7316f7c63b793fc

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+        Sha256 string `json:"sha256,omitempty"`
+
+        // ArtifactPullSecrets holds a reference to a secret where the credentials are
+        // stored. We use these credentials when pulling the artifacts from the provided
+        // URL using any of the supported protocols (http, https, and oci).
+        ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
+
+        // Insecure indicates whether certificates in the remote URL (if using TLS) can
+        // be ignored, aka InsecureSkipVerify.
+        Insecure bool  `json:"insecure,omitempty"`
```
twz123, line 46: I wonder if we should actually name this `InsecureSkipTLSVerify`. There's a lot of precedence for this name in the existing Kubernetes APIs.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+        // URL is the URL of a downloadable resource.
+        URL string `json:"url"`
+
+        // Sha256 provides an optional SHA256 hash of the URL's content for verification.
+        Sha256 string `json:"sha256,omitempty"`
+
+        // ArtifactPullSecrets holds a reference to a secret where the credentials are
+        // stored. We use these credentials when pulling the artifacts from the provided
+        // URL using any of the supported protocols (http, https, and oci).
+        ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
```
twz123, line 42: Comparing this to e.g. [`ImagePullSecrets`](https://github.com/kubernetes/kubernetes/blob/v1.30.2/staging/src/k8s.io/api/core/v1/types.go#L3711). There, it's an array. Would it make sense to use an array here, as well?

```suggestion
        ArtifactPullSecrets []ArtifactPullSecret `json:"artifactPullSecrets,omitempty"`
```

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+        // Insecure indicates whether certificates in the remote URL (if using TLS) can
+        // be ignored, aka InsecureSkipVerify.
+        Insecure bool  `json:"insecure,omitempty"`
+}
+```
+
+The `ArtifactPullSecret` property will be added and its struct will be defined as follow:
+
+```go
+type ArtifactPullSecret struct {
```
twz123, line 53: There's [`corev1.SecretReference`](https://github.com/kubernetes/kubernetes/blob/v1.30.2/staging/src/k8s.io/api/core/v1/types.go#L1097) already. If we use that, we don't need to define our own struct.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
twz123, line 42: Struggling a bit with "pull" here. I mean, sure, this is OCI terminology, but it sounds a bit strange in the usual web / HTTP context. WDYT about alternatives, like:

* ClientCredentialsSecretRef
* ClientAuthSecretRef
* ArtifactAuthSecretRef
* AuthCredentialSecretRef

Something along those lines? Or just bluntly SecretRef?

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+
+```yaml
+url: http://my.file.server/binaries/k0s-v1.30.1+k0s.0
+sha256: e95603f167cce6e3cffef5594ef06785b3c1c00d3e27d8e4fc33824fe6c38a99
+artifactPullSecret:
+  namespace: kube-system
+  name: artifacts-token-based-auth
+```
+
+### Secrets Layout
```
twz123, line 94: There's also the [TLS Secrets](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets) secret type; mTLS could be a very interesting future addition.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+
+For secrets of type `kubernetes.io/dockerconfigjson` the format is the default for Docker authentications, equal to what is used in a Pod's pull secret. For further details you can refer to the [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
+
+When it comes to the `Opaque` secret layout (used for HTTP requests) Autopilot will accept the following entries:
+
+- `username` and `password`: if both are set then Autopilot will attempt to pull the artifacts using [Basic Authentication](https://www.ibm.com/docs/en/cics-ts/6.1?topic=concepts-http-basic-authentication).
```
twz123, line 100: There's already the well-defined [Basic authentication Secret](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret) type. I'd just use that.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+- `authorization`: if this property is set then the `Authorization` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) will be set to its value when pulling the artifacts.
+
+No other property will be parsed and used. For sake of defining the expected behaviour in case of conflicting configurations:
+
+> In the case where the three properties are set (`username`, `password`, and `authorization`) Autopilot will ignore `username` and `password`, i.e. `authorization`  takes precedence over username and password.
```
twz123, line 105: What about using proper secret types for that, so there's no ambiguities?

There's already the well-defined [Basic authentication Secret](https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret) type. I'd just use that. We could add some k0s-specifc types, say `auth.k0s.k0sproject.io/bearer-token` (desugars into `Authorization: Bearer ${data.token}`) or `auth.k0s.k0sproject.io/http-authorization-header` (desugars into `Authorization: ${data.authorization}`).

twz123 (commented):
Small personal formatting request: Would be cool if the markdown paragraphs were hard-wrapped to a fixed width.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+  namespace: kube-system
+data:
+  authorization: "Bearer abc123def456ghi789jkl0"
+```
+
+### Additional Details
+
+- The `Insecure` property is equivalent of defining `InsecureSkipVerify` on a Go HTTP client.
+- The `Insecure` property will be valid for both `oci://` and `https://` protocols.
+- If no protocol is defined, HTTP is used.
```
juanluisvaladas, line 123: Defaulting to a non secure protocol is quite concerning for a few reasons:
1- Since the SHA256 is optional, so if a user doesn't set either they're subject to MITM attacks delivering malicious images 
2- The authentication will be sent plain

We should probably error if the protocol isn't explicitly defined, or at least default to a secure protocol...

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### Secrets Layout
+
+For secrets of type `kubernetes.io/dockerconfigjson` the format is the default for Docker authentications, equal to what is used in a Pod's pull secret. For further details you can refer to the [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/).
+
+When it comes to the `Opaque` secret layout (used for HTTP requests) Autopilot will accept the following entries:
+
+- `username` and `password`: if both are set then Autopilot will attempt to pull the artifacts using [Basic Authentication](https://www.ibm.com/docs/en/cics-ts/6.1?topic=concepts-http-basic-authentication).
+- `authorization`: if this property is set then the `Authorization` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) will be set to its value when pulling the artifacts.
```
makhov, line 101: Is there any possibility, that different registries might use different headers, e.g. `X-Auth-Token`, `X-Access-Token`, etc.? Or accept some additional headers? 

Maybe we can have a secret type, that will put all the secret content directly into headers without any processing.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Which secret would we use for basic and token based authentications ?

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Oh, there it is! Thanks!

ricardomaraschini: Oh, there it is! Thanks!

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: I don't have a strong opinion, any of these seem good to me. Let's see if someone else chimes in.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: I am afraid this will make the change not backward compatible.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Looks good to me. I will incorporate this.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
twz123: Interesting. So a `http.Client` assumes a "http" scheme when there's no scheme in the URL? How would such an URL look like?

ricardomaraschini: I gonna circle back on this. IIRC we use the `http.Client` by means of `grab`.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Fixed.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Fixed.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: I was wrong, `grab` simply fails if no protocol is present in the URL. I have made the default to `https` instead. Thanks for bringing this up @juanluisvaladas .

```
unsupported protocol scheme ""
```

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: As we have two similar comments going on I will resolve this one. Answers can be found in https://github.com/k0sproject/k0s/pull/4524#discussion_r1652654306.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: In my opinion we don't need to worry about this for registry access. Registries, more or less, work in the same way and we have implementation of registry clients for go. The `authorization` property here will be used only for HTTP or HTTPS requests.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: That is a nice idea. We can always create an addendum for this in the future. I need to avoid scope creep here. This proposal started as "oci + basic auth" and it is already "oci + basic auth + token".

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: For sake of completeness, this seems to be the go http client's behaviour. TIL.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,137 @@
...
+        // Sha256 provides an optional SHA256 hash of the URL's content for verification.
+        Sha256 string `json:"sha256,omitempty"`
+
+        // ArtifactPullSecrets holds a reference to a secret where the credentials are
+        // stored. We use these credentials when pulling the artifacts from the provided
+        // URL using any of the supported protocols (http, https, and oci).
+        ArtifactPullSecret *ArtifactPullSecret `json:"artifactPullSecret,omitempty"`
+
+        // Insecure indicates whether certificates in the remote URL (if using TLS) can
+        // be ignored, aka InsecureSkipVerify.
```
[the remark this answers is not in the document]
ricardomaraschini: fixed.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: As no one else chimed in we are going to go with `SecretRef` as it looks excellent to me.

ricardomaraschini: fixed.

ricardomaraschini:
Can we somehow expedite this ? I can see that the automation is already tired of the silent here :-) I think I have already addressed all comments.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
twz123: I'm still in favour of requiring an explicit scheme in the URL, which we should verify and fail if no or an unknown scheme is used.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+`SecretRef` property is of type `SecretReference` as defined by `k8s.io/api/core/v1` package. The secret pointed by the provided `SecretRef` will be used for pulling artifacts using either HTTP[S] or OCI protocols and is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in use is `oci://` or of type `Opaque` if protocols `http://` or `https://` are used (see below for details on the Secret layout).
+
+Example configuration for OCI:
+
+```yaml
+url: oci://my.registry/binaries/k0s:v1.30.1+k0s.0
+sha256: e95603f167cce6e3cffef5594ef06785b3c1c00d3e27d8e4fc33824fe6c38a99
+artifactPullSecret:
```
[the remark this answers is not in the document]
ricardomaraschini: Fixed, thanks.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
ricardomaraschini: Sounds good to me. Adjusted.

jnummelin (approved):
@ricardomaraschini I see all the comments are resolved so I'm stamping this 👍 Thanks for your patience 😄 

We can iron out the remaining details during implementation and update the ADR if needed.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
twz123: Right, I think I understand why this is a list of secrets: The image pull secrets may contain authentication material for multiple registries, just like a pod can pull images from multiple registries. That's not what we need to support here, as there's only ever one endpoint that we connect to.

Issue #3304, opened 2023-07-20
Reboot of the first control plane with dynamicConfig breaks Calico & Helm Charts


CmdrSharp:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 5.15.117-flatcar #1 SMP Tue Jul 4 14:43:38 -00 2023 x86_64 GNU/Linux
NAME="Flatcar Container Linux by Kinvolk"
ID=flatcar
ID_LIKE=coreos
VERSION=3510.2.4
VERSION_ID=3510.2.4
BUILD_ID=2023-07-04-1508
SYSEXT_LEVEL=1.0
PRETTY_NAME="Flatcar Container Linux by Kinvolk 3510.2.4 (Oklo)"
ANSI_COLOR="38;5;75"
HOME_URL="https://flatcar.org/"
BUG_REPORT_URL="https://issues.flatcar.org"
FLATCAR_BOARD="amd64-usr"
CPE_NAME="cpe:2.3:o:flatcar-linux:flatcar_linux:3510.2.4:*:*:*:*:*:*:*"
```

### Version

1.27.3+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "fbb20c14cb7ccdef4c6b0bc754438a16a7410a2e5c418fc935574c5e2dbf6c8c" (from machine) (pass)
Total memory: 11.7 GiB (pass)
Disk space available for /var/lib/k0s: 40.5 GiB (pass)
Operating system: Linux (pass)
  Linux kernel release: 5.15.117-flatcar (pass)
  Max. file descriptors per process: current: 524288 / max: 524288 (pass)
  Executable in path: modprobe: /usr/sbin/modprobe (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (pass)
    cgroup controller "memory": available (pass)
    cgroup controller "devices": available (assumed) (pass)
    cgroup controller "freezer": available (assumed) (pass)
    cgroup controller "pids": available (pass)
[...]

PR #4559, opened 2024-06-06
Document constraints of dynamic config install

juanluisvaladas:
## Description

I got a report of someone who had issues becuase the `kubernetes.default.svc` `clusterIP` was missing from the SANs of some control plane nodes. This happened because the certificate for kube-apiserver is generated without dynamic config. [In the past we saw a similar issue with specifying `spec.network.provider`](https://github.com/k0sproject/k0s/issues/3304),  but because it was reported in k0sctl we only fixed it for k0sctl.

We don't document anywhere this behavior but we copy it in  k0sctl and in k0smotron we use the same configuration for every replica of the statefulset, we should tell .

Non-changeable fields can cause issues if not defined in the configuration file. This isn't immediately obvious by reading the docs so document it explicitly.

This was in the past 

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) (I consider this a documentation bug)
- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/dynamic-configuration.md
+++ b/docs/dynamic-configuration.md
@@ -16,6 +16,7 @@ In the [k0s configuration options](configuration.md) there are some options that
 
 - `spec.api` - these options configure how the local Kubernetes API server is setup
 - `spec.storage` - these options configure how the local storage (etcd or sqlite) is setup
+- `spec.network.controlPlaneLoadBalancing` - these options configure how [Control Plane Load Balancing](cplb.md) is setup.
 
 In case of HA control plane, all the controllers will need this part of the configuration as otherwise they will not be able to get the storage and Kubernetes API server running.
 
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
Tagged it to backport to 1.30 only because the older versions need a manual backport due to CPLB.

twz123 (approved):
I wonder if there's anything stopping us from actually supporting this to be dynamically reconciled? It currently _feels_ like all that is needed is to add a `Reconcile` method to the `Keepalived` struct.

juanluisvaladas:
> I wonder if there's anything stopping us from actually supporting this to be dynamically reconciled? It currently feels like all that is needed is to add a Reconcile method to the Keepalived struct.

I discussed this with Jussi back in the day and we agreed we didn't want to support this because:
1- kube-apiserver relies con CPLB, if something goes wrong this creates a chicken-egg situation.
2- Some fields such as the interface might be different on a per node basis, for instance the interface used for VRRP instances.

twz123:
> 1- kube-apiserver relies con CPLB, if something goes wrong this creates a chicken-egg situation.

I don't see why kube-apiserver is relying on that? If yes, then because users have actively configured it in that way.

> 2- Some fields such as the interface might be different on a per node basis, for instance the interface used for VRRP instances.

Right. Currently, k0s is only able to reconcile cluster-wide settings. Maybe that's something to be revisited in the future.

juanluisvaladas:
> I don't see why kube-apiserver is relying on that? If yes, then because users have actively configured it in that way.

This is used to fill the SANs of the kube-apiserver certificate. I think that was the only thing where it affected apiserver.

Issue #4558, opened 2024-06-06
Not accurate documentation for nvidia-container-runtime


nekwar:
### Before creating an issue, make sure you've checked the following:

- [ ] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.29.1+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
➡️ Please replace this text with the output of `k0s sysinfo`. ⬅️
</pre>
</details>

### What happened?

This part of a documentation for nvidia-container-runtime is not accurate: https://docs.k0sproject.io/head/runtime/#using-nvidia-container-runtime

Following this doc, we're getting this error message on some of the pods (that are using `nvidia` RuntimeClass):
```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to get sandbox runtime: no runtime for "nvidia" is configured
```
More details can be obtained here: https://github.com/NVIDIA/gpu-operator/issues/463

### Steps to reproduce

1. Add new .toml config in `/etc/k0s/containerd.d/` according to the documentation https://docs.k0sproject.io/head/runtime/#using-nvidia-container-runtime with nvidia runtime configuration
2. Deploy NVIDIA GPU operator following this article https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/getting-started.html#gpu-operator-helm-chart-options

### Additional context
**Root cause**
[...]

PR #4573, opened 2024-06-07
CRI documentation improvements

nekwar:
## Description

- Add clarification on dynamic containerd configuration
- Update nvidia-container-runtime section with up-to-date instructions

Fixes # (issue)
https://github.com/k0sproject/k0s/issues/4558

## Type of change

- [X] Documentation update

## Checklist:

### DIFF
--- a/docs/runtime.md
+++ b/docs/runtime.md
@@ -66,9 +66,9 @@ decision] to replace rather than merge individual plugin configuration sections
 from imported configuration files. However, this behavior [may][containerd#7347]
 [change][containerd#9982] in future releases of containerd.
 
-Please note, that in order for drop-ins in `/etc/k0s/containerd.d` to take effect on running configuration, `/etc/k0s/containerd.toml` needs to be k0s managed. 
+Please note, that in order for drop-ins in `/etc/k0s/containerd.d` to take effect on running configuration, `/etc/k0s/containerd.toml` needs to be k0s managed.
 
-If you change `/etc/k0s/containerd.toml` (by accident or on purpose), it automatically becomes "not k0s managed". To make it "k0s managed" again, remove it and restart k0s service on the node, it'll be recreated by k0s. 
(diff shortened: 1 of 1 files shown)


dvirassamy:
I had a similar confusion.

You start reading about changing containerd.toml and removing the magic flag.

However the next paragraph provides the better solution, as from 1.27, where you can use the reconciler. which is a better solution as containerd.toml is still managed by k0s but you can extend it or customize it with side configs.

If there is not a good reason to modify containerd.toml, I would suggest to push to the top of the documentation the recommendation solution which uses the reconciler.

docs/runtime.md
```diff
@@ -66,6 +66,12 @@ decision] to replace rather than merge individual plugin configuration sections
 from imported configuration files. However, this behavior [may][containerd#7347]
 [change][containerd#9982] in future releases of containerd.
 
+Please note, that in order for drop-ins in `/etc/k0s/containerd.d` to take effect on running configuration, `/etc/k0s/containerd.toml` needs to be k0s managed. 
+
+If you change `/etc/k0s/containerd.toml` (by accident or on purpose), it automatically becomes "not k0s managed". To make it "k0s managed" again, remove it and restart k0s service on the node, it'll be recreated by k0s.
```
juanluisvaladas, line 71: This statement doesn't sound correct:

https://github.com/k0sproject/k0s/blob/5fb4f1091ac0fda4017d7418e420ce13aeffa423/pkg/component/worker/containerd/component.go#L323-L328

Are you 100% certain that if the file is modified accidentally it's automatically not k0s managed in every case? If I understand correctly, this is only true if the first line is modified. Otherwise a restart of k0s should be enough to recreate the file.

nekwar: yes, you're right, only `# k0s_managed=true` matters, just tested it. 
I'll update the doc accordingly

juanluisvaladas (approved):

PR #4675, opened 2024-06-25
Unify view of the world for the fake clients

twz123:
## Description

In order for the typed clients to share the same view of the world as the dynamic client, they all must share the same (wrapped) object tracker. Create typed clients using the dynamic client's object tracker and transform between typed and unstructured objects.

Use the auto-generated runtime schemes in the fake discovery client. The client is usually auto-detecting the list types from the scheme. By not passing a new, empty scheme, but the default scheme, there's no longer the need to manually manage the item to list type map. Support API resource discovery by taking all the GVKs out of the scheme and convert those to some API resource lists to be added to the discovery client. The conversion is just a best guess, but will suffice for testing needs.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/component/controller/apiendpointreconciler_test.go
+++ b/pkg/component/controller/apiendpointreconciler_test.go
@@ -25,6 +25,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 
 	"github.com/k0sproject/k0s/internal/testutil"
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
@@ -68,7 +69,7 @@ func TestBasicReconcilerWithNoExistingEndpoint(t *testing.T) {
 	assert.NoError(t, r.Init(ctx))
 
 	assert.NoError(t, r.reconcileEndpoints(ctx))
-	verifyEndpointAddresses(t, expectedAddresses, fakeFactory)
+	verifyEndpointAddresses(t, expectedAddresses, fakeFactory.Client)
 }
 
 func TestBasicReconcilerWithEmptyEndpointSubset(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/component/controller/apiendpointreconciler_test.go
```diff
@@ -172,7 +172,7 @@ func TestReconcilerWithNeedForUpdate(t *testing.T) {
 	assert.Equal(t, "bar", e.ObjectMeta.Annotations["foo"])
 }
 
-func verifyEndpointAddresses(t *testing.T, expectedAddresses []string, fakeFactory testutil.FakeClientFactory) *corev1.Endpoints {
+func verifyEndpointAddresses(t *testing.T, expectedAddresses []string, fakeFactory *testutil.FakeClientFactory) *corev1.Endpoints {
```
kke, line 175: This fakeclientfactory business is somehow hard to follow.

Maybe there should be some kind of `ClientGetter` interface and functions like this one would take `clientFactory ClientGetter` instead of `fakeFactory *testutil.FakeClientFactory`).

kke: It's `clientFactory kubeutil.ClientFactoryInterface` when it is ultimately used.

kke: Accept interfaces, return concrete types :)

pkg/component/controller/workerconfig/reconciler_test.go
```diff
@@ -52,7 +52,7 @@ type kubeletConfig = kubeletv1beta1.KubeletConfiguration
 
 // TODO: simplify it somehow, it is hard to read and to modify, both tests and implementation
 func TestReconciler_Lifecycle(t *testing.T) {
-	createdReconciler := func(t *testing.T) (*Reconciler, testutil.FakeClientFactory) {
+	createdReconciler := func(t *testing.T) (*Reconciler, *testutil.FakeClientFactory) {
```
kke, line 55: The same goes for these.

Maybe they don't need to be paired like this? Perhaps there could be something like:

```go
func createReconciler(t *testing.T, clientFactory ClientGetter) *Reconciler {
	k0sVars, err := config.NewCfgVars(nil, t.TempDir())
	require.NoError(t, err)
	reconciler, err := NewReconciler(
		k0sVars,
		&v1beta1.ClusterSpec{
			API: &v1beta1.APISpec{},
			Network: &v1beta1.Network{
				ClusterDomain: "test.local",
				ServiceCIDR:   "99.99.99.0/24",
			},
		},
		clients,
		&leaderelector.Dummy{Leader: true},
		true,
	)
	require.NoError(t, err)
	reconciler.log = newTestLogger(t)
	return reconciler
}
```

and:

```go
		clients := createFakeClientFactory()
		reconciler := createReconciler(t, clients)
```

twz123: Anyways, done in 2c790e52779951665636facdb31b39231a10cc5c. :sweat_smile:

jnummelin (approved):

PR #4744, opened 2024-07-10
Add controller lease counter component after API server

twz123:
## Description

The controller lease counter requires the API server to be up and running in order to work properly. It cannot acquire or count leases before the API server is up, nor can it release the lease after the API server is shut down.

Move the counter component down in the controller start method, so that it's started after the API server. Decouple the actual controller count value from the counter component by putting it directly into the start method. Let the counter component update that value, and let the konnectivity components consume those updates.

Introduce `internal/sync.Latest` as a small abstraction for values that may change over time and will be consumed by multiple observers.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -35,6 +35,7 @@ import (
 	"github.com/k0sproject/k0s/internal/pkg/file"
 	k0slog "github.com/k0sproject/k0s/internal/pkg/log"
 	"github.com/k0sproject/k0s/internal/pkg/sysinfo"
+	"github.com/k0sproject/k0s/internal/sync/value"
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 	"github.com/k0sproject/k0s/pkg/applier"
 	apclient "github.com/k0sproject/k0s/pkg/autopilot/client"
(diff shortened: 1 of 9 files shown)


cmd/controller/controller.go
```diff
@@ -215,15 +216,8 @@ func (c *command) start(ctx context.Context) error {
...
-		InvocationID:      c.K0sVars.InvocationID,
-		ClusterConfig:     nodeConfig,
-		KubeClientFactory: adminClientFactory,
-	}
-
-	if !c.SingleNode {
-		nodeComponents.Add(ctx, controllerLeaseCounter)
-	}
+	// Assume a single active controller during startup
+	numActiveControllers := value.NewLatest[uint](1)
```
jnummelin, line 220: Doesn't this mean that on the second controller, konnectivity would start once with `1` and then get restarted with `2` right away?

internal/sync/value/latest.go
```diff
@@ -0,0 +1,109 @@
...
+//		}()
+//
+//		// Set a new value, which will expire the previous value
+//		time.Sleep(1 * time.Second) // Simulate some delay
+//		l.Set(84)
+//		fmt.Println("New value set to 84")
+//
+//		wg.Wait() // Wait for the watcher goroutine to finish
+//	}
+type Latest[T any] struct {
```
jnummelin, line 72: An alternative name could be `Observable`. The channel provides a way to "observe" the changes, right. 😄

twz123: Yeah, but that's resembling too much the ReactiveX naming for my taste.

twz123: Maybe we could use RxGo then 🙈

twz123: Other ideas:
* `Now`
* `Currently`

jnummelin (approved):

internal/sync/value/latest.go
jnummelin: I'm also good with `Latest`, just wanted to throw out some suggestion. This is internal only stuff so naming is not _that_ important, we can change that later on too if needed/wanted.

PR #4792, opened 2024-07-24
bug: add support for image tag and digest

emosbaugh:
## Description

When adding an envoy image with tag and digest I get a failure:

>  invalid node config: spec: network: nodeLocalLoadBalancing.envoyProxy.image.version: Invalid value: "1.29.5-r0@sha256:7856e4d51f6631b1121c79ec93958897f1d90a52ce8d1d3513cff73df6821a2b": must match regular expression: ^[\w][\w.-]{0,127}$

This change adds support for an optional digested image reference.

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s/v1beta1/clusterconfig_types_test.go
+++ b/pkg/apis/k0s/v1beta1/clusterconfig_types_test.go
@@ -18,10 +18,12 @@ package v1beta1
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/k0sproject/k0s/internal/pkg/iface"
 	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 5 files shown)


twz123:
Darn! That was definitely an oversight. Surprising that there's no ready-made constant for this in `github.com/distribution/reference`. I wonder if it makes sense to simply use a single string instead of the image/version struct in a potential config version v2. This is how it's done in upstream Kubernetes anyways. Then we could use [`github.com/distribution/reference.ReferenceRegexp`](https://pkg.go.dev/github.com/distribution/reference@v0.6.0#ReferenceRegexp).

Fixes #2428.

twz123 (approved):
Thanks a lot!

sgalsaleh:
Is it possible to backport this fix to 1.28 as well?

twz123:
> Is it possible to backport this fix to 1.28 as well?

This is going into all supported releases, i.e. down to 1.27.

emosbaugh:
@twz123 I wanted to call out that `ClusterImages.Validate` [does not](https://github.com/k0sproject/k0s/blob/25941d8b934c776ce27f7b31b09e361e885fff38/pkg/apis/k0s/v1beta1/images.go#L100) call `ImageSpec.Validate`. Validate is only called for the envoy image. I unintentionally pushed [this test](https://github.com/k0sproject/k0s/pull/4792/files#diff-df8b5f1d36967d976cd305146646a7119907b0c8104eb4d54862a07e2a130778R85) which would have failed had it been called. Is there any follow up action you would like me to take here like to remove the test? Thanks for the quick turnaround!

twz123:
>I wanted to call out that `ClusterImages.Validate` [does not](https://github.com/k0sproject/k0s/blob/25941d8b934c776ce27f7b31b09e361e885fff38/pkg/apis/k0s/v1beta1/images.go#L100) call `ImageSpec.Validate`.

That explains why this didn't crop up earlier.

> Is there any follow up action you would like me to take here like to remove the test? 

Actually validating the images in `ClusterImages.Validate` seems reasonable to me. PRs are welcome, as always :upside_down_face:

PR #4832, opened 2024-08-02
Use Kubernetes APIs to construct user kubeconfigs

twz123:
## Description

Replace the Go template code with the typed structs and APIs provided by upstream Kubernetes. This avoids manual cert loading and provides type safety.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


cmd/kubeconfig/create.go
```diff
@@ -89,54 +66,57 @@ Note: A certificate once signed cannot be revoked for a particular user`,
...
+	userCert, err := certManager.EnsureCertificate(userReq, "root")
+	if err != nil {
+		return nil, fmt.Errorf("failed generate user certificate: %w, check if the control plane is initialized on this node", err)
+	}
+
+	const k0sContextName = "k0s"
+	kubeconfig := clientcmdapi.Config{
+		Clusters: map[string]*clientcmdapi.Cluster{k0sContextName: {
+			Server:               clusterAPIURL,
+			CertificateAuthority: userReq.CACert,
```
jnummelin, line 105: hmm, doesn't this just make it a path to the file? I think we'd want to use `CertificateAuthorityData` instead

jnummelin (commented):
I wonder if we should have some e2e test covering this too. I mean some test that would create new kubeconfig for a user and verify that it actually works

cmd/kubeconfig/create.go
jnummelin: ah, right.

twz123:
> I wonder if we should have some e2e test covering this too. I mean some test that would create new kubeconfig for a user and verify that it actually works

I think we can get away with unit tests here.  I've added some asserts to make sure the certs are properly embedded in 41cb3e6471638e60e956ef76b2efa02611657cc9.

I've scanned the existing inttests, and there's indeed the PSP inttest which invokes `k0s kubeconfig create test` and uses the generated kubeconfig to connect to the cluster. We need to do something about PSPs _eventually_ anyways, as noted in #1420. I reckon we want to add some real RBAC tests when doing the migration away from PSPs, then.

jnummelin (approved):

PR #4868, opened 2024-08-16
fix url.parse error

fruworg:
## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?
- [x] Manual test

This is just an example of problem solving for https://github.com/k0sproject/k0s/issues/4867.

### DIFF
--- a/pkg/component/controller/kine.go
+++ b/pkg/component/controller/kine.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+	"strings"
 
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
Hi @fruworg, thanks for the contribution. I'll wait for the tests to pass and if everything is OK I'll approve.

For the record: `url.Parse` parses URLs, DSNs aren't necessarily legal URLs: https://github.com/golang/go/issues/12200

juanluisvaladas (changes requested):
Linter fails becuase gofmt. Most certainly can be fixed by running`gofmt -s -w <file>`

fruworg:
Hi, I think the fix in https://github.com/k0sproject/k0s/pull/4879 is a better solution.

PR #4895, opened 2024-08-26
Replace go-bindata with go:embed

twz123:
## Description

Use Go's built-in way of embedding static assets in binaries, instead of relying on the somewhat legacy go-bindata tool. This allows the embedding to be handled transparently by the Go compiler, eliminating some extra Makefile steps.

Hide the path prefixes used to embed the data behind a sub-fs.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


pkg/autopilot/checks/checks.go
```diff
@@ -30,6 +30,9 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+//go:embed removed-gvks.yaml
```
kke, line 33: just a thought, how about writing that list as go consts?

twz123: Yeah, was thinking the same, but I decided to leave this as a TODO for the future us, and focus on the mere technical transition away from go-bindata here.

twz123: See #4916.

jnummelin (approved):

PR #4950, opened 2024-09-10
Concurrency safe client factory

twz123:
## Description

* Remove `GetRESTClient`
  This was accessing the client field without locking the mutex and was potentially missing initialization. The dynamic client is actually exposing its REST client, so that can be used instead.

* Use k0s's clientset in the k0s client factory
  Keep the old CRD specific versions for backwards compatibility.

* Lazy initialization for metadata client in Autopilot update checks
  This allows `FakeClientFactory.GetRESTConfig` method to panic without breaking tests. Code constructed with an empty rest.Config is not backed by a fake client at all, so it's better to have tests fail up front than to have strange errors when such constructed clients are actually used.

* Lazy initialization in client factory's `GetRESTClient`
  Previously, the `GetRESTClient` method was not synchronized and not concurrency safe. Callers could get a nil pointer under certain circumstances. Change this by adding mutexes and making it use lazy initialization like all other getters. This means that this method needs to be fallible.

* Introduce `LoadRESTConfig` function pointer
  This way, the way the REST config is loaded is up to the client factory users. The special "admin" settings are now part of the controller command, freeing the client factory from any particular way of loading and setting up the config.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -561,7 +561,11 @@ func (c *command) start(ctx context.Context) error {
 		EnableWorker:       c.EnableWorker,
 	})
 
-	apClientFactory, err := apclient.NewClientFactory(adminClientFactory.GetRESTConfig())
+	restConfig, err := adminClientFactory.GetRESTConfig()
+	if err != nil {
+		return err
+	}
+	apClientFactory, err := apclient.NewClientFactory(restConfig)
 	if err != nil {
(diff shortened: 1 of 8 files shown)


jnummelin (approved):
LGTM

PR #4951, opened 2024-09-10
[doc] Remove non-existant 'targets' field in Autopilot airgapupdate description

SebPlv:
## Description

Documentation describe field 'spec.commands[].airgapupdate.targets.workers' but crd doesn't have 'targets' level.

Fixes # (issue)

Just removing it in the documentation.


twz123 (approved):
Thanks!

PR #4978, opened 2024-09-16
Bump bootloose to v0.9.0

juanluisvaladas:
I don't know why dependabot isn't catching these, we're not ignoring them in the yaml.


juanluisvaladas:
Actually it had opened the PR: https://github.com/k0sproject/k0s/pull/4686

PR #5020, opened 2024-09-20
Replace grab with internal download function

twz123:
## Description

This makes the download process use a temporary file and atomic file replacements on success. Allows existing files to be re-downloaded, which was not possible before. On the contrary, it will re-download files even if they didn't change. Add a very long overall timeout for Autopilot downloads. This will ensure that the download will fail at some point, even if the remote server is artificially slow. Remove logger and httpResponse fields from the downloader struct.

Introduce `internal/http.Download`: A simple HTTP download function that can be used to download stuff from the Internet. It will detect and abort stale downloads as soon as the data transfer stalls for a certain amount of time. Special care is taken to detect and sanitize server-suggested file names.

Add `k0scontext.WithInactivityTimeout`, which returns a context that times out after a specified period of inactivity. It provides a "keep alive" mechanism to reset the timeout based on recent activity, ensuring that the context will remain valid for as long as there is activity. This is used to implement the cancellation of stale downloads.

Fixes:
* #4296

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/internal/http/download.go
+++ b/internal/http/download.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2024 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package http
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/k0sproject/k0s/pkg/build"
+)
+
+type DownloadOption func(*downloadOptions)
+
+// Downloads the contents of the given URL. Writes the HTTP response body to writer.
+func Download(ctx context.Context, url string, target io.Writer, options ...DownloadOption) (err error) {
+	opts := downloadOptions{}
+	for _, opt := range options {
+		opt(&opts)
+	}
+
+	// Prepare the client and the request.
+	client := http.Client{
+		Transport: &http.Transport{
+			// This is a one-shot HTTP client which should release resources immediately.
+			DisableKeepAlives: true,
+		},
+	}
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return fmt.Errorf("invalid download request: %w", err)
+	}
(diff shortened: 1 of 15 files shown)


internal/http/download.go
```diff
@@ -0,0 +1,90 @@
...
+		opt(&opts)
+	}
+
+	// Prepare the client and the request.
+	client := http.Client{
+		Transport: &http.Transport{
+			// This is a one-shot HTTP client which should release resources immediately.
+			DisableKeepAlives: true,
+		},
+	}
```
kke, line 44: Maybe connect/read timeouts would be good?

Something like

```go
client := http.Client{
    Transport: &http.Transport{
        // Timeout for establishing a connection
        DialContext: (&net.Dialer{
            Timeout: opts.requestTimeout,
        }).DialContext,
        ResponseHeaderTimeout: opts.requestTimeout,
        DisableKeepAlives:     true,
        IdleConnTimeout: opts.requestTimeout,
    },
}
```

twz123: I thought we could get away with a global external download timeout via the context. I've thought about cancelling stale downloads, but decided to do it in a follow-up PR: #5025.

kke: The full context timeout has to be quite long because of the file sizes for upgrades, so it would be nice to cancel early if the initial connection or response takes forever. We can do without, but it's so easy to add that it could be nice to add.

twz123: The staleness timeout in #5025 would address this, I suppose. It's also sort of a time-to-first-byte timeout. I put it in a separate PR to make this one a bit smaller, and to be able to backport this PR separately without adding timeouts that weren't present in the grab version of the downloader. However, if you think it makes sense to have timeouts from the beginning, I'll add the commits here.

pkg/autopilot/download/downloader.go
```diff
@@ -37,52 +40,58 @@ type Config struct {
...
+		expectedHash, err = hex.DecodeString(d.config.ExpectedHash)
 		if err != nil {
 			return fmt.Errorf("invalid update hash: %w", err)
 		}
+		targets = append(targets, d.config.Hasher)
+	}
 
-		dlreq.SetChecksum(d.config.Hasher, expectedHash, true)
+	// Set up target file for download.
+	target, err := file.AtomicWithTarget(filepath.Join(d.config.DownloadDir, "download")).Open()
```
jnummelin, line 69: hmm, doesn't this mean that `Download` func can only download one file at a time as the tmp file is always named `download`?

twz123: No, it's using this as the base name only.

jnummelin (approved):

jnummelin (approved):

PR #5025, opened 2024-09-23
Stale download cancellation

twz123:
## Description

Cancel downloads whenever there's no data flowing for a certain amount of time. Add a very long overall timeout for Autopilot downloads. This will ensure that the download will fail at some point, even if the remote server is artificially slow.

Add `k0scontext.WithInactivityTimeout`, which returns a context that times out after a specified period of inactivity. It provides a "keep alive" mechanism to reset the timeout based on recent activity, ensuring that the context will remain valid for as long as there is activity. This is used to implement the cancellation of stale downloads.

Move the private `writerFunc` implementation from the integration tests into the regular code base. This adapter is useful from time to time.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/internal/http/download.go
+++ b/internal/http/download.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2024 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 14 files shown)


internal/http/download.go
```diff
@@ -0,0 +1,115 @@
...
+	writeMonitor := internalio.WriterFunc(func(p []byte) (int, error) {
+		len := len(p)
+		if len > 0 {
+			keepAlive()
+		}
+		return len, nil
+	})
+
+	// Run the actual data transfer.
+	if _, err := io.Copy(io.MultiWriter(writeMonitor, target), resp.Body); err != nil {
```
kke, line 93: (I'm not sure but this may need to be chunked and done without io.Copy, checking context as it goes. Or maybe use bufio or whatever)

twz123:
As we now have #5042 that can be backported, this PR can now be part of #5020 again.

PR #5042, opened 2024-09-26
Disable grab download resume

jnummelin:
## Description

This is to mitigate cases like #4296

By default grab tries to resume the download if the file name determined from either the url or from content-type headers already exists. This makes things go side ways, if the existing file is smaller than the new one, the old content would still be there and only the "extra" new bytes would get written. I.e. the download would be "resumed". :facepalm:

So this change disables resuming altogether from grab.

To be able to download and replace k0s without resuming, i.e. "truncate" the file in-between. We need to force the download to download it as k0s.tmp first. Only after that we can safely move the file as truncating a running binary is not allowed and will error with `text file busy`.

This is a minimal possible fix that we can easily backport. @twz123 is already working on bigger refactoring of autopilot download functionality in #5020  that gets rid of grab. Grab seems to bring more (bad) surprises than real benefits. In the end, we just download files and we should pretty much always just replace them. No need for full library dependecy for that.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/autopilot/controller/signal/k0s/download.go
+++ b/pkg/autopilot/controller/signal/k0s/download.go
@@ -16,8 +16,10 @@ package k0s
 
 import (
 	"crypto/sha256"
+	"path/filepath"
 
 	apcomm "github.com/k0sproject/k0s/pkg/autopilot/common"
+	apconst "github.com/k0sproject/k0s/pkg/autopilot/constant"
 	apdel "github.com/k0sproject/k0s/pkg/autopilot/controller/delegate"
 	apsigcomm "github.com/k0sproject/k0s/pkg/autopilot/controller/signal/common"
 	apsigpred "github.com/k0sproject/k0s/pkg/autopilot/controller/signal/common/predicate"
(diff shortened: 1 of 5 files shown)


pkg/autopilot/controller/signal/k0s/download.go
```diff
@@ -88,6 +90,7 @@ func (b downloadManifestBuilderK0s) Build(signalNode crcli.Object, signalData ap
 			ExpectedHash: signalData.Command.K0sUpdate.Sha256,
 			Hasher:       sha256.New(),
 			DownloadDir:  b.k0sBinaryDir,
+			Filename:     filepath.Join(b.k0sBinaryDir, apconst.K0sTempFilename),
```
twz123, line 93: Hmm, shouldn't this be:

```suggestion
			Filename:     apconst.K0sTempFilename,
```

??? Otherwise, this would make the `DownloadDir` somehow superfluous.

jnummelin: In grab, if the `Filename` is set, it is used as absolute path when writing the file. So I think it would be correct to actually set the `DownloadDir` as empty in this case

twz123 (approved):

PR #5061, opened 2024-10-01
feat: add support for basic auth, insecure tls and headers

ricardomaraschini:
## Description

add suport for basic auth, insecure tls and headers to the download package. this PR is part of an [adr](https://github.com/k0sproject/k0s/blob/main/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md) implementation that i plan to introduce gradually.

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/internal/http/download.go
+++ b/internal/http/download.go
@@ -18,6 +18,7 @@ package http
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


internal/http/download.go
```diff
@@ -36,23 +37,35 @@ type DownloadOption func(*downloadOptions)
...
 		opt(&opts)
 	}
 
-	// Prepare the client and the request.
-	client := http.Client{
-		Transport: &http.Transport{
-			// This is a one-shot HTTP client which should release resources immediately.
-			DisableKeepAlives: true,
-		},
+	transport := http.DefaultTransport.(*http.Transport).Clone()
```
twz123, line 46: It was a deliberate decision not to use the default transport here, as it defines a bunch of timeouts that make configuration difficult. I think for this simple use case the staleness timeout is all that's needed.

The only thing that's probably good to set here is `ProxyFromEnvironment`. I'd even say setting this here should actually be backported.

internal/http/download_test.go
```diff
@@ -153,7 +205,17 @@ func startFakeDownloadServer(t *testing.T, handler http.Handler) string {
...
+			serverError <- server.Serve(listener)
+			return
+		}
+
+		cert, key := generateSelfSignedCert(t)
+		t.Cleanup(func() {
+			os.Remove(key)
+			os.Remove(cert)
+		})
+		serverError <- server.ServeTLS(listener, cert, key)
```
twz123, line 218: There's also a shorter solution for this in [joinclient_test.go](https://github.com/k0sproject/k0s/blob/v1.31.1%2Bk0s.0/pkg/token/joinclient_test.go#L147-L177). No need to write the certs to disk, so there's no need for cleanups.

internal/http/download_test.go
```diff
@@ -143,7 +150,52 @@ func TestDownload_RedirectLoop(t *testing.T) {
...
+	assert.ErrorContains(t, err, "bad status: 401 Unauthorized")
+	err = internalhttp.Download(context.TODO(), baseURL, io.Discard, internalhttp.WithBasicAuth("myuser", "mypassword"))
+	assert.NoError(t, err)
+}
+
+func TestDownload_InsecureSkipTLSVerify(t *testing.T) {
+	baseURL := startFakeDownloadServer(t, true, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte("content"))
+		require.NoError(t, err)
+		w.WriteHeader(http.StatusOK)
```
twz123, line 170: We can also just leave the handler body empty, as the content is discarded later on, and the status defaults to OK anyways.

(The call to WriteHeader has no effect anyways, as it's happening after the first write.)

```suggestion
```

internal/http/download.go
ricardomaraschini: https://github.com/k0sproject/k0s/pull/5064

internal/http/download_test.go
ricardomaraschini: That makes sense. I have incorporated the proposed changes on my branch.

internal/http/download_test.go
ricardomaraschini: Way better. This is fixed here.

twz123 (approved):

PR #5062, opened 2024-10-02
Applier manager improvements

twz123:
## Description

* The stacks don't need to be stored in the manager struct
  The map is only ever used in the loop to create and remove stacks, so it doesn't need to be stored in the struct. This ensures that there can't be any racy concurrent accesses to it.

* Don't check for closed watch channels
  The only reason these channels get closed is if the watcher itself gets closed. This happens only when the method returns, which in turn only happens when the context is done. In this case, the loop has already exited without a select on a potentially closed channel. So the branches that checked for closed channels were effectively unreachable during runtime.

* Wait for goroutines to exit
  Rename cancelWatcher to stop and wait until the newly added stopped channel is closed. Also, add a stopped channel to each stack to do the same for each stack-specific goroutine.

* Restart watch loop on errors
  Exit the loop on error and restart it after a one-minute delay to allow it to recover in a new run. Also replace the bespoke retry loop for stacks with the Kubernetes client's wait package.

* Improve logging
  Cancel the contexts with a cause. Add this cause to the log statements when exiting loops. Rename bundlePath to bundleDir to reflect the fact that it is a directory, not a file.

* Remove unused applier field
  Seems to be a remnant from the past.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -41,7 +41,6 @@ type Manager struct {
 	K0sVars           *config.CfgVars
 	KubeClientFactory kubeutil.ClientFactoryInterface
 
-	applier   Applier
 	bundleDir string
 	stop      func(reason string)
 	log       *logrus.Entry
@@ -65,8 +64,6 @@ func (m *Manager) Init(ctx context.Context) error {
 	}
 	m.log = logrus.WithField("component", constant.ApplierManagerComponentName)
 	m.bundleDir = m.K0sVars.ManifestsDir
-
-	m.applier = NewApplier(m.K0sVars.ManifestsDir, m.KubeClientFactory)
 
 	m.LeaderElector.AddAcquiredLeaseCallback(func() {
 		ctx, cancel := context.WithCancelCause(ctx)
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -18,6 +18,7 @@ package applier
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path"
 	"time"
@@ -38,10 +39,10 @@ type Manager struct {
 	K0sVars           *config.CfgVars
 	KubeClientFactory kubeutil.ClientFactoryInterface
 
-	applier    Applier
-	bundlePath string
-	stop       func()
-	log        *logrus.Entry
+	applier   Applier
+	bundleDir string
+	stop      func(reason string)
+	log       *logrus.Entry
 
 	LeaderElector leaderelector.Interface
 }
@@ -49,7 +50,7 @@ type Manager struct {
 var _ manager.Component = (*Manager)(nil)
 
 type stack = struct {
-	cancel  context.CancelFunc
+	cancel  context.CancelCauseFunc
 	stopped <-chan struct{}
 	*StackApplier
 }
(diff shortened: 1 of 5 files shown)


juanluisvaladas (dismissed):

emosbaugh:
I've created a pull request into the fork with a test https://github.com/twz123/k0s/pull/134

twz123:
Thx @emosbaugh! I added the commit here, but it's not signed-off...

emosbaugh:
> Thx @emosbaugh! I added the commit here, but it's not signed-off. Can you maybe sign it and do a force push? You should be able to do it directly on the branch in my fork, as "allow edits by maintainers" is checked.

Done. Sorry about that

emosbaugh:
@twz123 is the intention to backport this fix and to what version? thanks!

twz123:
We can check if it's easy to backport. If yes, all good, If not, we can maybe re-target your patch to the release-1.31 branch and backport that to 1.30 - 1.28 instead. Or we do it the other way round and merge your patch into main, and I'll rebase this one on top of yours.

I'll check tomorrow in detail...

emosbaugh:
> We can check if it's easy to backport. If yes, all good, If not, we can maybe re-target your patch to the release-1.31 branch and backport that to 1.30 - 1.28 instead. Or we do it the other way round and merge your patch into main, and I'll rebase this one on top of yours.
> 
> I'll check tomorrow in detail...

Thanks!

emosbaugh:
> Alright, the code changes themselves can be backported with just a few small merge conflicts that are straight forward to resolve. The test case, on the other hand, doesn't work at all in 1.28-1.30. This is supposedly due to some non-trivial improvements that have been implemented for the fake clients quite recently.
> 
> 
> 
> We can do a backport, excluding the tests, or we could try to backport the fake client improvements, as well, which might be quite a bit of work.

The code change is what is most important. Thanks

pkg/applier/manager.go
```diff
@@ -62,21 +63,21 @@ func (m *Manager) Init(ctx context.Context) error {
...
-		watcherCtx, cancel := context.WithCancel(ctx)
-		m.cancelWatcher = cancel
+		ctx, cancel := context.WithCancelCause(ctx)
+		stopped := make(chan struct{})
+
+		m.stop = func(reason string) { cancel(errors.New(reason)); <-stopped }
 		go func() {
-			_ = m.runWatchers(watcherCtx)
+			defer close(stopped)
+			wait.UntilWithContext(ctx, m.runWatchers, 1*time.Minute)
```
jnummelin, line 75: 1 minute seems bit abstract here, any reasoning why that time is used?

jnummelin (approved):
Left one minor Q on the timeout used

pkg/applier/manager.go
jnummelin: oh, didn't realize auto-merge was set. oh well, it was minor anyways 😂

twz123: It's a trade-off between busy-loop/log-spam and a reasonable self-healing delay. I think everything between, say, 10 secs and a couple of minutes would be fine here, so one minute was just the thing I came up with when writing that code :see_no_evil:

PR #5063, opened 2024-10-02
Inline manifestSsaver

twz123:
## Description

This makes the components more self-contained and gives them more control over the contents of the directories. Also, the `FsManifestsSaver' did not use the `ManifestsDir' configuration field.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


pkg/component/controller/crd.go
```diff
@@ -78,13 +86,17 @@ func (c CRD) Start(context.Context) error {
...
+
+		content, err := fs.ReadFile(static.CRDs, src)
 		if err != nil {
-			return fmt.Errorf("failed to fetch crd `%s`: %w", filename, err)
+			return fmt.Errorf("failed to fetch CRD %s manifest %s: %w", c.bundle, filename, err)
 		}
-		if err := c.saver.Save(manifestName, content); err != nil {
-			return fmt.Errorf("failed to save CRD `%s` manifest `%s` to FS: %w", c.bundle, manifestName, err)
+		if err := file.AtomicWithTarget(dst).
+			WithPermissions(constant.CertMode).
```
juanluisvaladas, line 97: I'm not so sure about using `constant.CertMode` here, 0644 is a good value but probably this deserve its own cosntant. Maybe introduce a new `constant.ManifestsMode` right under `ManifestsDirMode`?

twz123: We could consider to introduce constants for regular files, regular directories and regular executables.

It'd probably be best if we would rely on the umask here, and simply specify regular files as 666 😈, and regular directories or executables as 777. This requires non-trivial changes to the atomic file writer, though, as it's using the Go stdlib's temp file facility, which creates temp files using 600 in any case, and getting the umask from the kernel requires some racy syscalls, or reading and parsing a file from the proc file system.

PR #5067, opened 2024-10-02
Let ClusterConfigInitializer apply the api-config stack

twz123:
## Description

The k0s managed stacks don't have to live on disk. They can be used directly from in-memory. Let the api-config stack be the first. Include its application in the Start method of the ClusterConfigInitializer. Have the applier manager ignore this stack and write an ignore note in the stack directory.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:


internal/testutil/fakeclient/tracker.go
```diff
@@ -128,6 +130,16 @@ func (t *TransformingObjectTracker) List(gvr schema.GroupVersionResource, gvk sc
...
 
+	// Set a fake resource version, so that watches work.
+	if external, ok := external.(interface {
+		GetResourceVersion() string
+		SetResourceVersion(string)
+	}); ok {
+		if external.GetResourceVersion() == "" {
+			external.SetResourceVersion(strconv.FormatInt(time.Now().UnixMilli(), 10))
+		}
+	}
```
kke, line 141: Slightly confusing shadowing

```suggestion
	if versionedResource, ok := external.(interface {
		GetResourceVersion() string
		SetResourceVersion(string)
	}); ok && versionedResource != nil {
		if versionedResource.GetResourceVersion() == "" {
			versionedResource.SetResourceVersion(strconv.FormatInt(time.Now().UnixMilli(), 10))
		}
	}
```

Added a nil-check for good measure. I guess `external` could theoretically be a `runtime.Object` typed nil pointer 🤔

cmd/controller/controller.go
```diff
@@ -300,7 +300,10 @@ func (c *command) start(ctx context.Context) error {
 		nodeComponents.Add(ctx, &applier.Manager{
 			K0sVars:           c.K0sVars,
 			KubeClientFactory: adminClientFactory,
-			LeaderElector:     leaderElector,
+			IgnoredStacks: map[string]string{
+				controller.ClusterConfigStackName: "v1.32",
```
kke, line 304: Maybe the `v1.32` should come from some constant

internal/testutil/fakeclient/tracker.go
twz123: Good point about the nil interface value, that goes under the radar way too quickly. However, I think it's fair not to add the nil check: An `Externalize` implementation shouldn't return such a bogus interface in the first place. I think this would be a programming error in that function, and that it would be fair to panic because of it in a test case. If `Externalize` cannot produce a meaningful externalized object, then it should fail with an error instead. Moreover, maybe the backing methods are fine with a nil receiver, e.g. treat the calls as no-ops.

cmd/controller/controller.go
twz123: We'd have to add a constant for each version where we add an ignored stack, and the value of that constant would never change. I don't see the immediate benefit. Can you explain this in more detail?

kke: Magic string. Where does it come from? Is there a matching "v1.32" string somewhere else in the code? Does someone have to change this string sometime? Why is it this single version that is ignored? Does or will v1.33 exist? Was v1.31 different? These are the questions that come to mind when seeing such string.

kke: I don't think `controller.ClusterConfigStackName` will change either? Maybe `controller.ClusterConfigStackVersion`?

pkg/applier/manager.go
```diff
@@ -167,6 +170,19 @@ func (m *Manager) createStack(ctx context.Context, stacks map[string]stack, name
 		return
 	}
 
+	log := m.log.WithField("stack", name)
+
+	stackName := filepath.Base(name)
+	if sinceVersion, ignored := m.IgnoredStacks[stackName]; ignored {
```
kke, line 176: This is a bit misleading as any later versions will not match.

The k0s/version package offers version constraints 🤔

twz123: Which is the intended behavior. K0s 1.33 should still produce the ignored.txt file referring to k0s 1.32.

cmd/controller/controller.go
twz123: `controller.ClusterConfigStackName` is used in different parts of the code and has actual impact on k0s's behavior. This "v1.32" string is not really used at all in that sense. Maybe there's a misunderstanding in the meaning of  it? `controller.ClusterConfigStackVersion` wouldn't be appropriate, as "v1.32" is not referring the stack's version (it doesn't have any version in and of itself), it's referring to the first k0s version that ignored that stack folder. The only usage of "v1.32" is to be written into the "ignored.txt" file, to give admins a hint about the first k0s version that started to ignore this stack. We can also go without the values, and change the `IgnoredStacks` field from a map to a slice that only contains the stack names, and not put any version hint in that "ignored.txt" file.

> Where does it come from?

It's the k0s version in which k0s started to ignore the cluster config stack in the manifests applier.

> Is there a matching "v1.32" string somewhere else in the code?

No.

> Does someone have to change this string sometime?

Only if this PR doesn't make it into k0s 1.32, but into another k0s version.

> Why is it this single version that is ignored?

It's not a single version that's ignored, it's the k0s version at which k0s started to ignore a stack. It's not anticipated that this will change anymore.

> Does or will v1.33 exist?

Maybe, if we decide to ignore other stacks starting with k0s 1.33. However, this "v1.32" should remain unaffected, as we ignore this specific stack since version 1.32, which will not change.

> Was v1.31 different?

Yes, as we are relying on the cluster config stack to be applied by the manifest applier in that k0s version.

pkg/applier/manager.go
twz123: I'll try to add comments and use better variable names. Otherwise, I'd just remove the version hint.

cmd/controller/controller.go
kke: In any case, I think here we see that some kind of clarification is necessary there :)

twz123: I have removed the version hint. I couldn't think of a good way to express the intent without over-engineering :see_no_evil: (not that I'm not guilty of over-engineering from time to time :angel:).

kke (approved):
Whatever I noticed was addressed.

PR #5073, opened 2024-10-03
Bump calico to v3.28.2

juanluisvaladas:
Also includes a fix in nftables detection due to the lack of ip6tables binaries.


juanluisvaladas:
Rerunning all the tests as the new image with iptables 1.8.9 is already available.

jnummelin (approved):

PR #5110, opened 2024-10-10
ADR 1: Use secret types to select the authentication scheme

twz123:
## Description

This is a follow-up to https://github.com/k0sproject/k0s/pull/4524#discussion_r1622310448 and https://github.com/k0sproject/k0s/pull/4524#discussion_r1652654306. I think using secret types more could help in making the API better. WDYT?

The current version of this ADR proposes to use one secret type per protocol, which is somewhat limiting. It would be reasonable to consider the use of a basic auth secret for authentication to an OCI registry, for instance. Also, the use of the secret type to select the authentication scheme would help to reduce any potential ambiguity.

In the case of OCI registries, the ADR already explicitly mentions the dockerconfigjson secret type. For HTTP[S], it falls back on the Opaque type and attempts to figure out the appropriate authentication method based on the contents of the secret. This is inconsistent with the way it works for OCI, and can be made consistent by adding a custom secret type for the case that k0s wants to cover, but for which there's no predefined type available upstream.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -66,10 +66,7 @@ type PlanResourceURL struct {
 
 `SecretRef` property is of type `SecretReference` as defined by
 `k8s.io/api/core/v1` package. The secret pointed by the provided `SecretRef`
-will be used for pulling artifacts using either HTTP\[S\] or OCI protocols and
-is expected to by of type `kubernetes.io/dockerconfigjson` if the protocol in
-use is `oci://` or of type `Opaque` if protocols `http://` or `https://` are
-used (see below for details on the Secret layout).
+will be used for pulling artifacts using either HTTP\[S\] or OCI protocols.
 
 ### Example Configurations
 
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -103,48 +100,47 @@ secretRef:
...
+  metadata:
+    name: creds
+    namespace: kube-system
+  data:
+    authorization: "Bearer abc123def456ghi789jkl0"
+  ```
+
+  The `authorization` entry is used as is, with its content placed directly into
+  the `Authorization` header. For example, a secret like the following will make
+  Autopilot set the `Authorization` header to `Bearer abc123def456ghi789jkl0`.
```
ricardomaraschini, line 136: Somehow we lost the example so the "following" part here makes no sense anymore (maybe it should be "previous" now).

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -103,48 +100,47 @@ secretRef:
...
 
 ### Additional Details
 
-- The `InsecureSkipTLSVerify` property is equivalent of defining
+- The `InsecureSkipTLSVerify` property is equivalent to defining
   `InsecureSkipTLSVerify` on a Go HTTP client.
 - The `InsecureSkipTLSVerify` property will be valid for both `oci://` and
-  `https://` protocols.
+  `https://` protocols. It has no effect for the `http://` protocol.
```
ricardomaraschini, line 143: You may want to add some notes here regarding the new `oci+http`  introduced by https://github.com/k0sproject/k0s/pull/5107 as well.

ricardomaraschini (changes requested):

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
twz123: Oh, it has to be reversed :see_no_evil:

ricardomaraschini (dismissed):

ricardomaraschini (approved):

PR #5223, opened 2024-11-12
Enable nosprintfhostport linter

twz123:
## Description

... and fix lints on the way.

Inline the IsIPV6String function, as it had only one use left, which was inlined as a one-liner.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -216,11 +216,10 @@ func (n *Network) BuildServiceCIDR(addr string) string {
 	if !n.DualStack.Enabled {
 		return n.ServiceCIDR
 	}
-	// because in the dual-stack mode k8s
-	// relies on the ordering of the given CIDRs
-	// we need to first give family on which
-	// api server listens
-	if IsIPv6String(addr) {
+	// Because Kubernetes relies on the order of the given CIDRs in dual-stack
+	// mode, the CIDR whose version matches the version of the IP address the
+	// API server is listening on must be specified first.
+	if ip := net.ParseIP(addr); ip != nil && ip.To4() == nil {
 		return n.DualStack.IPv6ServiceCIDR + "," + n.ServiceCIDR
 	}
 	return n.ServiceCIDR + "," + n.DualStack.IPv6ServiceCIDR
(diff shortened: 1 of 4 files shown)


pkg/apis/k0s/v1beta1/network.go
```diff
@@ -136,7 +136,7 @@ func (n *Network) DNSAddress() (string, error) {
 	}
 
 	address := ipnet.IP.To4()
-	if IsIPv6String(ipnet.IP.String()) {
+	if address == nil {
 		address = ipnet.IP.To16()
 	}
```
twz123, line 141: I wonder if this could just become

```suggestion
	address := ipnet.IP
```

? Not sure if would make any difference if a v4 address might be passed on in its 16 byte representation.

ncopa: It should not make any difference at all. We return a string.

I support this change.

twz123: After a second look, I realized that the code actually relies on the 4 byte representation a few lines below, and produces a nonsense DNS address when ServiceCIDR describes a v6 network: E.g. given the CIDR `fd00:abcd:1234::/64`, the calculated address will be `fd00:abcf:1234::`, which is outside of the network to begin with. The `ipnet.Contains` check doesn't work as expected since `address` and `ipnet.IP` point to the same slice, i.e. the modifications to `address` will also affect `ipnet`.

That said, `n.ServiceCIDR` shouldn't be a v6 network in the first place, that's what `n.DualStack.IPv6ServiceCIDR` is for. This is all a bit weird: the `IsIPv6String` check has been added in #1292, to support "IPv6 only". Are we actually supporting this? If yes, then this function is definitely broken.

What do we want to do?

1. Error out if ServiceCIDR is a v6 network?
2. Fix the "IPv6 only" mode by calculating a meaningful v6 DNS address?

In any case, the `ipnet.Contains` check needs fixing, i.e. we need to properly copy `address` by value.

twz123: Proposed fix: #5236

kke (dismissed):

jnummelin (approved):

PR #5224, opened 2024-11-12
Enable perfsprint linter

twz123:
## Description

... and fix lints on the way.

Most of the fixes are straightforward replacements of `fmt.Errorf` with `errors.New` and string concatenations instead of `fmt.Sprintf` with a single leading or trailing `%s`. Single integer conversions have been replaced with the corresponding functions from the `strconv` package. Used `path.Join`, `filepath.Join` and `url.URL` where it made sense.

Only add reset and status subcommands on Linux. They were erroring out on Windows, so remove the OS checks from the
commands and only add them on OSes on which they are supported.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


cmd/airgap/listimages.go
```diff
@@ -39,7 +40,7 @@ func NewAirgapListImagesCmd() *cobra.Command {
 			}
 
 			if opts.EnableDynamicConfig {
-				return fmt.Errorf("dynamic config is not supported for airgap list-images")
+				return errors.New("dynamic config is not supported for airgap list-images")
```
kke, line 43: (The `err113` linter would complain about these dynamic errors with a static message)

cmd/api/api.go
```diff
@@ -244,19 +245,19 @@ func (c *command) authMiddleware(next http.Handler, usage string) http.Handler {
...
 		if len(parts) == 2 {
 			token := parts[1]
 			if !c.isValidToken(r.Context(), token, usage) {
-				sendError(fmt.Errorf("go away"), w, http.StatusUnauthorized)
+				sendError(errors.New("go away"), w, http.StatusUnauthorized)
 				return
 			}
 		} else {
-			sendError(fmt.Errorf("go away"), w, http.StatusUnauthorized)
+			sendError(errors.New("go away"), w, http.StatusUnauthorized)
```
kke, line 260: Maybe this `go away` error warrants a `var ErrUnauthorized = errors.New("go away")`.

cmd/controller/certificates.go
```diff
@@ -66,7 +68,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 	c.CACert = string(cert)
 	// Changing the URL here also requires changes in the "k0s kubeconfig admin" subcommand.
 	apiAddress := net.JoinHostPort(c.ClusterSpec.API.Address, strconv.Itoa(c.ClusterSpec.API.Port))
-	kubeConfigAPIUrl := fmt.Sprintf("https://%s", apiAddress)
+	kubeConfigAPIUrl := (&url.URL{Scheme: "https", Host: apiAddress}).String()
```
kke, line 71: There's probably a constant for `https` that the `usestdlib` linter will complain about :)

pkg/supervisor/supervisor.go
```diff
@@ -357,7 +357,7 @@ func (s *Supervisor) shouldKillProcess(ph procHandle) (bool, error) {
 // - inject k0s embedded bins into path
 func getEnv(dataDir, component string, keepEnvPrefix bool) []string {
```
kke, line 358: This function seems strange.

- The `i` is not incremented when an override is encountered. Doesn't this make the `env[i]=x` later modify the wrong entry?
- The `k0sManaged` is added even if it already exists. 
- The `i++` + `return env[:i]` is kind of suspicious. 

Even if it works, in the sake of clarity, I think this should be simplified.

kke: I don't understand what it does in general.

Any `COMPONENT_XXX` encountered will be converted to `XXX` once. After that any `COMPONENT_YYY` will be skipped?

kke: Hmm, no. It will skip any following `XXX` after `COMPONENT_XXX` has been encountered.

kke: So, `COMPONENT_HTTP_PROXY=abcd` will always become `HTTP_PROXY` and any following `HTTP_PROXY` will be skipped.

`COMPONENT_FOO=z` will go through as-is unless `KeepEnvPrefix` is false, in which case it will become `FOO=z` and any following `FOO=x` will be skipped.

kke: Ok, so maybe it does what it is supposed to. Maybe there isn't a one-off bug.

But it is very weird.

twz123: > Ok, so maybe it does what it is supposed to. Maybe there isn't a one-off bug.

There are tests for it, so I think the implementation should be "correct".

> But it is very weird.

It _is_ weird :see_no_evil:

cmd/controller/certificates.go
twz123: Funny enough, there isn't!

kke (approved):

PR #5238, opened 2024-11-13
add renovate

PrashantR30:
## Description

Adding Renovate  so to automate the checks and updates packages dependencies in the  embedded-bins/Makefiles.variables

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

The version of the packages version were modified to an earlier version and renovate was run https://github.com/PrashantR30/go_app_3/pulls and we could see the checks were done and PR created for new versions of the depencencies

## Checklist:

### DIFF
--- a/.github/workflows/renovate.json
+++ b/.github/workflows/renovate.json
@@ -1,0 +1,34 @@
+{
+    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+    "dependencyDashboard": true,
+    "onboarding": false,
+    "automerge": false,
+    "separateMinorPatch": true,
+    "platform": "github",
+    "extends": ["config:base"],
+    "labels": ["version_bump"],
+    "forkProcessing": "enabled",
+    "semanticCommits": "enabled",
+    "prConcurrentLimit": 20,
+    "prHourlyLimit": 20,
+    "dockerfile": {
+        "enabled": false
+    },
+    "customManagers": [
+        {
+        "customType": "regex",
+        "fileMatch": ["embedded-bins/Makefile.variables"],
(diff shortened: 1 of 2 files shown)


twz123:
Thanks for looking into this! Unfortunately, it's not that easy to generate proper version bumps for most of the stuff in the Makefiles. I'd like to start with something really simple first, and then add more and more stuff to it. How about starting with just Alpine (as I think this is the only one that can be done by solely incrementing stuff in the Makefile), so that we have the skeleton and a simple example, and then add other stuff to it in individual PRs. I imagine that we need to restructure things a bit on a case by case basis for automatic version bumps to be working reliably.

PrashantR30:
Thanks @twz123, I'll update the PR to focus on Renovate on-boarding and Alpine only. Once that's in place, we can add other dependencies gradually in separate PRs. Appreciate the guidance!

.github/workflows/renovate.json
```diff
@@ -0,0 +1,34 @@
...
+    "prConcurrentLimit": 20,
+    "prHourlyLimit": 20,
+    "dockerfile": {
+        "enabled": false
+    },
+    "customManagers": [
+        {
+        "customType": "regex",
+        "fileMatch": ["embedded-bins/Makefile.variables"],
+        "matchStrings": ["alpine_version\\s*=\\s*(?<currentValue>\\S+)"],
```
twz123, line 21: Just remembering that we distinguish between the full Alpine version (x.y.z) and only the major/minor part, so I think we need to augment this?

PR #5250, opened 2024-11-14
Remove stack when directory is renamed

twz123:
## Description

A directory rename is signaled as a rename with the old name followed by a create with the new name. The applier manager didn't honor this, effectively running the old stack along with the new one.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist:

### DIFF
--- a/pkg/applier/manager.go
+++ b/pkg/applier/manager.go
@@ -29,7 +29,6 @@ import (
 	"github.com/k0sproject/k0s/internal/pkg/file"
 	"github.com/k0sproject/k0s/pkg/component/controller/leaderelector"
 	"github.com/k0sproject/k0s/pkg/component/manager"
-	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
 	kubeutil "github.com/k0sproject/k0s/pkg/kubernetes"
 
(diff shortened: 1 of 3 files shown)


pkg/applier/manager.go
```diff
@@ -151,7 +150,7 @@ func (m *Manager) runWatchers(ctx context.Context) {
 				if dir.IsDirectory(event.Name) {
 					m.createStack(ctx, stacks, event.Name)
 				}
-			case fsnotify.Remove:
```
twz123, line 154: Interesting. The unit test fails on Windows, whereas it works well on Linux. I suspect this is due to the implicit event order which is probably different between Windows and Linux. I can imagine how the Create event for the new name happening before the Rename event for the old name will screw things up. Maybe the fix is more involved than simply reacting on Rename events, as well.

pkg/applier/manager.go
```diff
@@ -151,7 +150,7 @@ func (m *Manager) runWatchers(ctx context.Context) {
 				if dir.IsDirectory(event.Name) {
 					m.createStack(ctx, stacks, event.Name)
```
twz123, line 151: In the Windows unit test logs, I see paths in the logs like this: `stack="C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\TestManager_Rename2408267215\\001/lorem". Note the last slash, which should IMO be a backslash. IIUC the logged value is coming from fsnotify, so something is not normalizing path names there. Note that the slash is a valid separator on Windows, too, albeit not the canonical one.

PR #5293, opened 2024-11-25
[Backport release-1.30] Correctly detect DNS address computation errors

twz123:
Backport to `release-1.30`:

* #5270

See:

* #5236


juanluisvaladas:
The CI error looks legit, maybe it's worth running these in armv7 with -race?

juanluisvaladas:
1.30 is EOL and won't be getting further releases.

PR #5307, opened 2024-12-03
doc: adds expectations with regards to pulling using OCI

### LINKED ISSUE #5143: ADR 1: Spec out OCI artifact format for downloads
_Based on https://github.com/k0sproject/k0s/pull/5075#discussion_r1796840093_

> When having a look at our implementation, I realize that the ADR _just_ says: "We'll download via OCI", but not at all what's meant by this. As far as I understand, there's no common understanding of what it means to download an artifact via OCI, i.e. it is very much implementation dependent.
>
> I think we should augment the ADR with a spec what k0s means/expects when it tries to download files from OCI registries. WDYT?

ricardomaraschini:
## Description

this commit adds a new section to the adr on oci pulls where we lay down the expectations when it comes to pulling artifacts through oci.

such expectations are written as an example workflow of what should be supported.

Fixes #5143

## Type of change

- [x] Documentation update

### DIFF
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -116,14 +116,14 @@ This feature supports pulling both the k0s binary and the k0s air-gap bundle.
 Each repository tag should host only one artifact. To remotely host an artifact
 and later retrieve it using this feature, the following command is sufficient:
 
-```
+```sh
 oras push <registry-address>/<namespace>/<repository>:<tag> <artifact path>
 ```
 
 For example, to push the k0s binary and its corresponding airgap bundle, use
 the following commands:
 
-```
+```sh
 # pushing the k0s binary.
 oras push registry.company.com/k0s/binary:v1.31.1 k0s
 
--- a/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
+++ b/docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
@@ -110,6 +110,39 @@ secretRef:
   name: artifacts-token-based-auth
 ```
 
+#### Expected workflow
+
+This feature supports pulling both the k0s binary and the k0s air-gap bundle.
+Each repository tag should host only one artifact. To remotely host an artifact
+and later retrieve it using this feature, the following command is sufficient:
+
+```
+oras push <registry-address>/<namespace>/<repository>:<tag> <artifact path>
+```
+
+For example, to push the k0s binary and its corresponding airgap bundle, use
+the following commands:
+
+```
+# pushing the k0s binary.
+oras push registry.company.com/k0s/binary:v1.31.1 k0s
+
+# pushing the k0s airgap bundle.
+oras push registry.company.com/k0s/bundle:v1.31.1 k0s-airgap-bundle
+```
+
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -110,6 +110,39 @@ secretRef:
   name: artifacts-token-based-auth
 ```
 
+#### Expected workflow
+
+This feature supports pulling both the k0s binary and the k0s air-gap bundle.
+Each repository tag should host only one artifact. To remotely host an artifact
```
twz123, line 116: What should happen if there are multiple artifacts?

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -110,6 +110,39 @@ secretRef:
   name: artifacts-token-based-auth
 ```
 
+#### Expected workflow
+
+This feature supports pulling both the k0s binary and the k0s air-gap bundle.
```
twz123, line 115: Downloading executables raises the question of how to manage the executable bit of the downloaded files.

docs/architecture/adr-001-autopilot-oci-basic-auth-support.md
```diff
@@ -110,6 +110,39 @@ secretRef:
...
+# pushing the k0s airgap bundle.
+oras push registry.company.com/k0s/bundle:v1.31.1 k0s-airgap-bundle
+```
+
+Once the artifacts are hosted in the OCI registry, the following configuration
+is sufficient to enable the pull through OCI:
+
+```yaml
+url: oci://registry.company.com/k0s/binary:v1.31.1
+sha256: <sha-for-the-binary>
```
twz123, line 139: Big question: Is this the digest of the image or the digest of the embedded binary?

When dealing with OCI artifacts, it would seem more natural to be the digest of the image (or image index) itself.

If it is the latter, this means that as long as the published artifact also uses SHA-256 as its digest algorithm (I have yet to see an image that doesn't), we would have an accurate way of figuring out which artifact to pick out of the image index. The downside would be that this would make multi-platform downloads with the same name impossible.

twz123 (approved):
Thanks, that's a good starting point. I was thinking about some sort of "specification" though. It's not clear how k0s will select the artifact to download, given only the OCI name and tag, nor how k0s will figure out the file name to use. If we specify these things explicitly, there's less room for ambiguity and "specification by implementation". It will also help when discussing certain design aspects:

For example, regarding artifact selection, I think it would make perfect sense to support platform filtering, so that you can distribute different artifacts for different platforms under the same name, just as you can with multi-arch images.

Regarding file name determination, ORAS chooses to set the `org.opencontainers.image.title` annotation to the file name specified on the command line. This is a pragmatic choice, but an implementation detail of ORAS, since this annotation is for the "human-readable title of the image", according to the OCI spec.

It wouldn't be too unreasonable to not use the file name from the OCI manifests, but just require it to be provided in the k0s config, depending on whether that file name is something that should be decided at "k0s config time" or at "artifact push time". This way, we could completely eliminate the annotation processing mentioned above. This extends to other file system properties as well, such as permissions, ownership, maybe even file times.

PR #5360, opened 2024-12-13
Introduce k0s airgap bundle-artifacts

twz123:
## Description

This sub-command introduces a "k0s native" way to create airgap artifact bundles. The current way to do this requires either a Docker or containerd daemon. Even then, there are certain scenarios where the created bundles don't contain all the necessary tags, i.e. when images are referenced by both their tag and their digest.

Replace image-bundler with k0s airgap bundle-artifacts. Dog-food the new sub-command in the k0s build itself.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist:


cmd/airgap/listimages.go
```diff
@@ -25,12 +25,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func NewAirgapListImagesCmd() *cobra.Command {
+func newAirgapListImagesCmd() *cobra.Command {
```
kke, line 28: Other commands are exposed.

cmd/airgap/bundleartifacts.go
```diff
@@ -0,0 +1,192 @@
...
+			} else {
+				refs, err = parseArtifactRefsFromReader(cmd.InOrStdin())
+			}
+			if err != nil {
+				return err
+			}
+
+			buffered := bufio.NewWriter(out)
+			if err := bundler.Run(ctx, refs, out); err != nil {
+				return err
```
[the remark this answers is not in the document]
twz123: I could change this to `return fmt.Errorf("failed to bundle artifacts: %w", err)`, but I'm not sure if that adds any additional insights :thinking:

cmd/airgap/listimages.go
twz123: Yes, they are. But why?

cmd/airgap/bundleartifacts.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+	})
+	if err := errors.Join(err, words.Err()); err != nil {
+		return nil, err
+	}
+
+	return refs, nil
+}
+
+func parseArtifactRefsFromSeq(refs iter.Seq[string]) (collected []reference.Named, _ error) {
```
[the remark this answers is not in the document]
twz123: I'm not sure if I can follow. What would you suggest?

cmd/airgap/airgap.go
```diff
@@ -19,16 +19,24 @@ package airgap
...
-		Short: "Manage airgap setup",
-		Args:  cobra.NoArgs,
-		RunE:  func(*cobra.Command, []string) error { return pflag.ErrHelp }, // Enforce arg validation
+		Short: "Tooling for airgapped installations",
+		Long: `Tooling for airgapped installations.
+
+For example, to create an image bundle that contains the images required for
+the current configuration, use the following command:
+
+    k0s airgap list-images | k0s airgap bundle-artifacts -v -o image-bundle.tar
```
makhov, line 36: Does it make sense to run only `k0s airgap list-images | k0s airgap bundle-artifacts -v -o image-bundle.tar` for the default images?

cmd/airgap/listimages.go
kke: I guess it could be used in testing but isn't.

cmd/airgap/bundleartifacts.go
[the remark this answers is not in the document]
makhov: Not using named returns, I suppose, they are terrible :)

cmd/airgap/airgap.go
jnummelin: I think the idea having the input is for the users to be able to add their own images too. But that said, it _could_ work also in the way that if the list is not given, then it could use the default images. But would that need config loading etc.?

makhov: Yeah, that's what I mean, if the input list is not provided, use default images. If I remember correctly, existing k0s airgap list-images reads config already, so the user could use pipes.

cmd/airgap/bundleartifacts.go
```diff
@@ -0,0 +1,191 @@
...
+	cmd.Flags().Var((*platformFlag)(&platform), "platform", "the platform to export")
+	cmd.Flags().StringArrayVar(&bundler.RegistriesConfigPaths, "registries-config", nil, "paths to the authentication files for OCI registries (uses the standard Docker config if omitted)")
+
+	return cmd
+}
+
+func parseArtifactRefsFromReader(in io.Reader) ([]reference.Named, error) {
+	words := bufio.NewScanner(in)
+	words.Split(bufio.ScanWords)
+	refs, err := parseArtifactRefsFromSeq(func(yield func(string) bool) {
```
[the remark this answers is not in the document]
juanluisvaladas: I'm usually not a fan of these and this is no exception (seems unnecessarily complex) but given the deadline I'm OK merging this...

pkg/airgap/ociartifactsbundler.go
```diff
@@ -0,0 +1,452 @@
...
+			},
+			Reference: registry.Reference{
+				Registry:   reference.Domain(ref),
+				Repository: reference.Path(ref),
+			},
+			PlainHTTP: b.InsecureRegistries == PlainHTTPOCIRegistry,
+		}
+	}
+
+	copyOpts := oras.CopyOptions{
```
juanluisvaladas, line 103: IMO also worth extracting it to a separate function

juanluisvaladas (approved):

cmd/airgap/airgap.go
twz123: Yes, that was kinda the idea: Not to add any config loading here at all and use pipes. Otherwise this would be duplicating quite a bit of the logic in `list-images`. This is also advertised in the long help text of `k0s airgap`:

> For example, to create an image bundle that contains the images required for
> the current configuration, use the following command:
>
>    k0s airgap list-images | k0s airgap bundle-artifacts -v -o image-bundle.tar

cmd/airgap/bundleartifacts.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		collected = append(collected, parsed)
+	}
+	return collected, nil
+}
+
+type insecureRegistryFlag airgap.InsecureOCIRegistryKind
+
+var insecureRegistryFlagValues = [...]string{
+	airgap.NoInsecureOCIRegistry:    "no",
```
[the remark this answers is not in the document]
twz123: We could go for string constants, but that would only make sense if `airgap.InsecureOCIRegistryKind` was also stringly typed, and if the constants lived in pkg/airgap.

What I like about having this separate and number typed is that the current zero value is actually a good default (`NoInsecureOCIRegistry`). To get the same effect using strings, `NoInsecureOCIRegistry` would have to be the empty string, which doesn't feel right. In turn, not defining the empty string as a constant would mean that the zero value would live in an "undefined limbo", i.e. it wouldn't be valid. Also, the string value is only important for the command line flag, not for the actual implementation in pkg/airgap, so why not scope it to the command? We would need a way to validate raw input strings and provide a list of all available strings for command-line flag parsing anyway, so the code below wouldn't really go away unless we hard-coded it. But since it's just in this file, and there are probably no more options to come, that's fine.

cmd/airgap/bundleartifacts.go
```diff
@@ -0,0 +1,191 @@
...
+	idx := slices.Index(insecureRegistryFlagValues[:], value)
+	if idx >= 0 {
+		*(*airgap.InsecureOCIRegistryKind)(i) = airgap.InsecureOCIRegistryKind(idx)
+		return nil
+	}
+
+	return errors.New("must be one of " + strings.Join(insecureRegistryFlagValues[:], ", "))
+}
+
+type platformFlag imagespecv1.Platform
```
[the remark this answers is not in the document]
twz123: This is a newtype over the OCI image spec's `Platform` type that implements `pflag.Value` so that it can be added as a flag: `cmd.Flags().Var((*platformFlag)(&platform), "platform", "the platform to export")`

pkg/airgap/ociartifactsbundler.go
```diff
@@ -0,0 +1,452 @@
...
+
+type OCIArtifactsBundler struct {
+	Log                   logrus.FieldLogger
+	InsecureRegistries    InsecureOCIRegistryKind
+	RegistriesConfigPaths []string // uses the standard Docker config if empty
+	PlatformMatcher       platforms.MatchComparer
+	RewriteTarget         RewriteRefFunc
+}
+
+func (b *OCIArtifactsBundler) Run(ctx context.Context, refs []reference.Named, out io.Writer) error {
```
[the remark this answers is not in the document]
twz123: > This should be probably split into a few functions

I think the most interesting part to split up would be the loop body, which I tried to do. Added a function for the call to oras.Copy, and a method for creating the manifests for the copied ref.

> is it mostly copied from an example or something?

Not that I remember. It was more like tweaking the call to oras.Copy and the contents of the archive index long enough to get the desired outcome.

pkg/airgap/ociartifactsbundler.go
```diff
@@ -0,0 +1,452 @@
...
+		client, close = newHttpClient(b.InsecureRegistries == SkipTLSVerifyOCIRegistry)
+		defer close()
+	}
+
+	creds, err := newOCICredentials(b.RegistriesConfigPaths)
+	if err != nil {
+		return err
+	}
+
+	newSource := func(ref reference.Named) oras.ReadOnlyTarget {
```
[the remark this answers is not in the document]
twz123: I've moved it into the for loop, as this is basically just a struct literal. I think it's fine there after having factored out other functions/methods.

pkg/airgap/ociartifactsbundler.go
twz123: I think this is fine, as this is just a struct literal? Converting this into a function with a flat argument list would somehow feel wrong to me.

juanluisvaladas (dismissed):

pkg/airgap/ociartifactsbundler.go
juanluisvaladas: That's fair enough

juanluisvaladas (approved):

PR #5365, opened 2024-12-16
Tweak and document disabling Docker build integration

twz123:
## Description

Don't default `EMBEDDED_BINS_BUILDMODE` to none when disabling Docker integration. This makes it clear that the binaries can't be built without it, at the moment. Inline the `K0S_BUILD_IMAGE_FILE` variable and replace it with the more generic `GO_ENV_REQUISITES`. Use the empty string to disable Docker integration: The previous sentinel value of "false" may have given the impression that "true" is the correct value to enable Docker build integration. Also, "false" is actually an executable provided by coreutils. An empty string can never point to a potentially executable file, and won't give the impression that DOCKER is some sort of boolean variable.

Document the new build variables in the README. 

Follow-up to:

* #5096

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,23 +138,34 @@ The requirements for building k0s from source are as follows:
 All of the compilation steps are performed inside Docker containers, no
 installation of Go is required.
 
-The k0s binary can be built in two different ways:
+The k0s binary can be built in different ways:
 
 The "k0s" way, self-contained, all binaries compiled from source, statically
-linked and embedded:
+linked, including embedded binaries:
 
 ```shell
 make
 ```
 
-The "package maintainer" way, without any embedded binaries (requires that the
-required binaries are provided separately at runtime):
+The "package maintainer" way, without building and embedding the required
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -147,14 +147,23 @@ linked and embedded:
...
+The "package maintainer" way, without building and embedding the required
+binaries. This assumes necessary binaries are provided separately at runtime:
 
 ```shell
 make EMBEDDED_BINS_BUILDMODE=none
 ```
 
-The embedded binaries can be built on their own:
+As mentioned previously, Docker build integration is enabled by default.
+Building k0s without Docker using the system toolchain can be done as follows:
```
jnummelin, line 158: What do you mean by "system toolchain", maybe give an example what this means in this context

twz123: PTAL. I tried to clarify this sentence.

Issue #5350, opened 2024-12-11
k0s kubectl connection fails after upgrading to K0s v1.31.2


jsalgado78:
### Before creating an issue, make sure you've checked the following:

- [x] You are running the latest released version of k0s
- [x] Make sure you've searched for existing issues, both open and closed
- [x] Make sure you've searched for PRs too, a fix might've been merged already
- [x] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 4.18.0-553.22.1.el8_10.x86_64 #1 SMP Wed Sep 11 18:02:00 EDT 2024 x86_64 GNU/Linux
NAME="Red Hat Enterprise Linux"
VERSION="8.10 (Ootpa)"
ID="rhel"
ID_LIKE="fedora"
VERSION_ID="8.10"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Red Hat Enterprise Linux 8.10 (Ootpa)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:8::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8"
BUG_REPORT_URL="https://issues.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_BUGZILLA_PRODUCT_VERSION=8.10
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="8.10"
```

### Version

v1.31.2+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Total memory: 7.7 GiB (pass)
File system of /var/lib/k0s: xfs (pass)
Disk space available for /var/lib/k0s: 3.6 GiB (pass)
Relative disk space available for /var/lib/k0s: 30% (pass)
Name resolution: localhost: [127.0.0.1 ::1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 4.18.0-553.22.1.el8_10.x86_64 (pass)
  Max. file descriptors per process: current: 65536 / max: 65536 (pass)
  AppArmor: unavailable (pass)
  Executable in PATH: modprobe: /sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
[...]

PR #5426, opened 2025-01-09
Use localhost in admin kubeconfig again, if possible

twz123:
## Description

If the API server is binding to all addresses, it's better to use the loopback interface to communicate with the API server locally. This avoids e.g. weird HTTP proxy problems.

See:

* https://github.com/k0sproject/k0s/issues/5350#issuecomment-2580777985
* #3824

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/pkg/apis/k0s/v1beta1/api.go
+++ b/pkg/apis/k0s/v1beta1/api.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"encoding/json"
+	"fmt"
 	"net"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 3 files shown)


pkg/apis/k0s/v1beta1/api.go
```diff
@@ -72,6 +73,17 @@ func DefaultAPISpec() *APISpec {
 	return a
 }
 
+func (a *APISpec) LocalURL() *url.URL {
+	var host string
+	if a.OnlyBindToAddress {
+		host = net.JoinHostPort(a.Address, strconv.Itoa(a.Port))
+	} else {
+		host = fmt.Sprintf("localhost:%d", a.Port)
```
jnummelin, line 81: Why not use `net.JoinHostPort` in this case too?

twz123: We could, but we don't need to. JoinHostPort is only necessary if the host may be an IPv6 address, which, in this case, it isn't. Would you prefer to use it anyways?

pkg/apis/k0s/v1beta1/api.go
```diff
@@ -72,6 +73,17 @@ func DefaultAPISpec() *APISpec {
...
+	}
+
+	return &url.URL{Scheme: "https", Host: host}
```
ncopa, line 84: It doesn't matter, but....
```suggestion
	host := "localhost"
	if a.OnlyBindToAddress {
		host = a.Address
	}
	return &url.URL{Scheme: "https", Host: net.JoinHostPort(host, strconv.Itoa(a.Port))}
```

ncopa (approved):

PR #5436, opened 2025-01-14
Bump kubernetes to 1.32.1

juanluisvaladas:
## Description

Bump kubernetes to 1.32.1

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:


juanluisvaladas:
Unit test failure can be ignored, it's because I defined an image in `pkg/constant/constant.go` which breaks the test, that will be fixed once we release the kube-router 1.32.1 image.

Additional tests:
- network conformance (kube router): https://github.com/k0sproject/k0s/actions/runs/12829614009 
- Will manually launch a network conformance test with calico.

pkg/constant/constant_test.go
```diff
@@ -132,25 +132,9 @@ func TestContainerdModuleVersions(t *testing.T) {
 
 func TestRuncModuleVersions(t *testing.T) {
```
juanluisvaladas, line 133: The runc dependency is gone, I kept this to just check that vars.sh returns a runc_version number that makes sense. I'm not too sure what to do about this, maybe we should entirely remove the test or leave it as is because it's not hurting anyway...

Makefile
```diff
@@ -146,10 +146,15 @@ $(controller_gen_targets): $(GO_ENV_REQUISITES) hack/tools/boilerplate.go.txt ha
 	touch -- '$@'
 
 # Run register-gen for each API group version.
+# Usually register-gen should should be:
```
juanluisvaladas, line 149: Please check this. I think this is a reasonably good option given the current circumstances.

twz123: Urgh, so this is what we get for not evaluating the upstream alpha and beta releases. Pretty sure we could've avoided this to land in the final Kubernetes release. WDYT concerning https://github.com/kubernetes/kubernetes/issues/129290#issuecomment-2552233989? Should we maybe set those GODEBUG flags, as well?

twz123: I wonder if we should have this as a separate commit, just for reference, and to maybe revert it whenever the issue is fixed upstream.

go.mod
```diff
@@ -63,31 +63,33 @@ require (
...
+	k8s.io/kubectl v0.32.1
+	k8s.io/kubelet v0.32.1
+	k8s.io/kubernetes v1.32.1
+	k8s.io/mount-utils v0.32.1
+	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	sigs.k8s.io/controller-runtime v0.19.4
 	sigs.k8s.io/yaml v1.4.0
 )
 
+require k8s.io/cluster-bootstrap v0.0.0
```
twz123, line 85: Probably a bad merge conflict resolution?

pkg/constant/constant_test.go
twz123: Oh, this is already the second time the dependency vanished from upstream. I'd be comfortable with removing the test entirely. I think that's what I did last time (need to check the git history).

go.mod
juanluisvaladas: Most certainly, it used to be in line 72. I moved it back to lien 72 and set it to v0.32.1

pkg/constant/constant_test.go
juanluisvaladas: Removed the test entirely.

twz123 (approved):

PR #5445, opened 2025-01-17
Use full patch version for go directives in go.mod

twz123:
## Description

It turns out that since Go 1.21, in which Go started using fully qualified versions and not omitting zeros (i.e. "1.21.0" instead of just "1.21"), just stating the minor version without the patch version is actually considered a development version of Go. Who would have thought? The correct way to use the Go directive is to set it to "go 1.23.0".

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/k0sproject/k0s
 
-go 1.23
+go 1.23.0
 
 // k0s
 require (
--- a/hack/tool/go.mod
+++ b/hack/tool/go.mod
@@ -1,6 +1,6 @@
 module tool
 
-go 1.23
+go 1.23.0
 
 require (
 	github.com/hashicorp/terraform-exec v0.21.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/k0sproject/k0s
 
-go 1.23
+go 1.23.0
```
juanluisvaladas, line 3: Shouldn't it be 1.23.5 instead of 1.23.0?

twz123: No, this is more about the minimum required Go version. I'd rather avoid to update this for every Go patch version bump.

juanluisvaladas (approved):

go.mod
juanluisvaladas: Ah fair enough

PR #5476, opened 2025-01-24
Validate removed apis

juanluisvaladas:
## Description

Prior implementation checked that there were no objects present in the removed APIs, however, it failed if the object was available on both the removed API and the present API. This behavior is incorrect and now for each object present in the removed API, we verify it's present in the new version (if available).

This was tested on both 1.31 and 1.32

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:


pkg/autopilot/checks/checks.go
```diff
@@ -17,12 +17,14 @@ package checks
...
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/k0sproject/k0s/pkg/kubernetes"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/metadata"
+	"k8s.io/cri-api/pkg/errors"
```
twz123, line 27: That looks like the wrong import.

juanluisvaladas: Good catch, fixed.

twz123:
Did we have a look in how other projects check for the existence of old API versions in a live cluster?

juanluisvaladas:
> Did we have a look in how other projects check for the existence of old API versions in a live cluster?

OpenShift has a thing called admin-acks. Basically between some versions the admin has to do some manual checks and then patch a configmap to let openshift know they verified it.

For instance between 4.15 and 4.16 (1.28 to 1.28) the system administrator must check the objects manually and once the admin has checked it manually they have to execute: `oc -n openshift-config patch cm admin-acks --patch '{"data":{"ack-4.15-route-config-not-supported-in-4.16":"true"}}' --type=merge`

They don't do an automatic check. I haven't checked other projects.

juanluisvaladas:
@twz123 I'm going on vacation soon, could we move this forward?

pkg/autopilot/checks/removedapis.go
```diff
@@ -23,10 +23,14 @@ import (
 
 type removedAPI struct {
 	group, version, kind, removedInVersion string
+	// currentVersion declares a version that is still supported for the Group Kind.
+	// If it's empty, it means that the Group Kind is removed in the removedInVersion.
+	currentVersion string
```
makhov, line 28: Since the both return values mention version, but different ones, probably makes sense to mention it in names, like `removedInK8sVersion` and `currentAPIVersion` or smth

juanluisvaladas: Changed it.

makhov (approved):

PR #5479, opened 2025-01-27
Revert "Remove default-config and validate sub-commands"

twz123:
Reverts k0sproject/k0s#5376

There's a bug in k0sproject/version that causes k0sctl's version-based detection to make false assumptions about the supported sub-commands in k0s. That needs to be addressed first before we can remove these deprecated commands for realz.

### DIFF
--- a/cmd/validate/validate.go
+++ b/cmd/validate/validate.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2025 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


cmd/validate/validate.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2021 k0s authors
```
juanluisvaladas, line 2: The copyright linter complains about this https://github.com/k0sproject/k0s/blob/main/hack/copyright.sh#L23-L25

We could either remove the -1 flag from git log in the copyright script  and then it works, or we can update the copyright to 2025.

twz123: Damn. Did a force push and forgot to add the local change first ... that's what you get when you're in a hurry ... :disappointed:

jnummelin (approved):

Issue #5481, opened 2025-01-28
Manifest Deployer doesn't work


pepsi1k:
### Before creating an issue, make sure you've checked the following:

- [x] You are running the latest released version of k0s
- [x] Make sure you've searched for existing issues, both open and closed
- [x] Make sure you've searched for PRs too, a fix might've been merged already
- [x] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 5.15.0-130-generic #140-Ubuntu SMP Wed Dec 18 17:59:53 UTC 2024 x86_64 GNU/Linux
NAME="Linux Mint"
VERSION="21.1 (Vera)"
ID=linuxmint
ID_LIKE="ubuntu debian"
PRETTY_NAME="Linux Mint 21.1"
VERSION_ID="21.1"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.linuxmint.com/"
BUG_REPORT_URL="http://linuxmint-troubleshooting-guide.readthedocs.io/en/latest/"
PRIVACY_POLICY_URL="https://www.linuxmint.com/"
VERSION_CODENAME=vera
UBUNTU_CODENAME=jammy
```

### Version

v1.31.5+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
File system of /var/lib/k0s: ext4 (pass)
Disk space available for /var/lib/k0s: 117.8 GiB (pass)
Relative disk space available for /var/lib/k0s: 80% (pass)
Name resolution: localhost: [::1 127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 5.15.0-130-generic (pass)
  Max. file descriptors per process: current: 1048576 / max: 1048576 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
    cgroup controller "memory": available (is a listed root controller) (pass)
    cgroup controller "devices": available (device filters attachable) (pass)
[...]

PR #5482, opened 2025-01-28
Add note to the example of Manifest Deployer

pepsi1k:
## Description
It’s unclear that the applier-manager picks up only files with the `.yaml` file type.

Fixes: https://github.com/k0sproject/k0s/issues/5481

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

### DIFF
--- a/docs/manifests.md
+++ b/docs/manifests.md
@@ -20,8 +20,6 @@ The use of Manifest Deployer is quite similar to the use the `kubectl apply` com
 
 ## Example
 
-**Note:** The file type must be `.yaml` not `.yml`.
-
 To try Manifest Deployer, create a new folder under `/var/lib/k0s/manifests` and then create a manifest file (such as `nginx.yaml`) with the following content:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/manifests.md
```diff
@@ -20,6 +20,8 @@ The use of Manifest Deployer is quite similar to the use the `kubectl apply` com
 
 ## Example
 
+**Note:** The file type must be `.yaml` not `.yml`.
```
juanluisvaladas, line 23: Hi @pepsi1k, thanks for the contribution!

We should have this documented that we only match the .yaml and not .yml files, but I think this shouldn't be in the Example section. Can you move it to the Overview category? Maybe something like:

Maybe something like:

By default, k0s reads all manifest files under /var/lib/k0s/manifests that end with .yaml and ensures that their state matches the cluster state. It does not read files with other extensions, such as .yml. Moreover, on removal of a manifest file, k0s will automatically prune all of its associated resources.

juanluisvaladas (changes requested):
By default, k0s reads all manifest files under /var/lib/k0s/manifests that end with .yaml and ensures that their state matches the cluster state. It does not read files with other extensions, such as .yml. Moreover, on removal of a manifest file, k0s will automatically prune all of its associated resources.

juanluisvaladas (approved):

PR #5644, opened 2025-03-04
Adjust systemd service configuration based on `image-builder`

s3rj1k:
## Description

Syncs systemd service config with values that are used in vanilla `image-builder` images.

- https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/containerd/templates/etc/systemd/system/containerd.service.d/memory-pressure.conf#L8
- https://github.com/kubernetes-sigs/image-builder/blob/main/images/capi/ansible/roles/containerd/templates/etc/systemd/system/containerd.service.d/limit-nofile.conf#L4

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

Manual systemd service test of config validity.

## Checklist

### DIFF
--- a/pkg/install/linux_systemd.go
+++ b/pkg/install/linux_systemd.go
@@ -38,6 +38,7 @@ KillMode=process
 LimitCORE=infinity
 TasksMax=infinity
 TimeoutStartSec=0
+OOMScoreAdjust=-999
 
 {{- if .ChRoot}}RootDirectory={{.ChRoot|cmd}}{{- end}}
 
--- a/pkg/install/service_linux.go
+++ b/pkg/install/service_linux.go
@@ -37,7 +37,7 @@ func configureServicePlatform(s service.Service, svcConfig *service.Config) {
 		svcConfig.Dependencies = []string{"After=network-online.target", "Wants=network-online.target"}
 		svcConfig.Option = map[string]interface{}{
 			"SystemdScript": systemdScript,
-			"LimitNOFILE":   999999,
+			"LimitNOFILE":   1048576,
 		}
 	}
 }


makhov (approved):

twz123:
Just out of curiosity, have you run into OOM killer issues without the OOM score adjustment? The main difference with the upstream reference you provided is that in upstream these settings only apply to the container runtime, whereas here this change applies to the k0s process itself and all its child processes, including api-server, etcd/kine, and others, not just containerd.

s3rj1k:
> have you run into OOM killer issues without the OOM score adjustment

I did not, but I don not have any kind of load in dev clusters that could actually create that memory load

s3rj1k:
> The main difference with the upstream reference you provided is that in upstream these settings only apply to the container runtime, whereas here this change applies to the k0s process itself and all its child processes, including api-server, etcd/kine, and others, not just containerd.

I've just adjusted value to proper infinity, solution to actually do this in they way you are mentioning was already there.

pkg/install/linux_systemd.go
```diff
@@ -38,6 +38,7 @@ KillMode=process
 LimitCORE=infinity
 TasksMax=infinity
 TimeoutStartSec=0
+OOMScoreAdjust=-999
```
twz123, line 41: Can we add a condition around this directive, so it only kicks in if k0s is running the worker components? We could add a comment line with the link to the upstream source, too.

s3rj1k: I think you are overengeneering this. If I run k0s as system service I would want it to be not OOMed in any case, not for dev cluster and definitely not for prod cluster.

jnummelin: The referenced upstream code put this only on containerd, where it _kinda_ makes sense. Throwing it as a blanket over ALL k0s&k8s components like this doesn't make sense IMO. For example in the case of running a controller+worker node, this would mean that even your precious workloads will probable be thrown under the bus before k0s _stuff_. Is that what users always want?

But if we throw it like as a blanket like this we're making the decision on behalf of all the users that k0s is THE most important thing on the node, right? Are we sure it always is, I'm not really sure at all.

s3rj1k: From ChatGPT quoting

```
Setting OOMScoreAdjust to -999 gives a process the highest priority to remain running during memory shortages. This means the process is almost entirely protected from being terminated by the OOM killer.
```

s3rj1k: FYI: Fedora 42 uses `systemd-oomd`

s3rj1k: I wonder what will happen to `etcd` process when `oomd` decides that systemd controller unit has to go...

s3rj1k:
> The PR is marked as stale since no activity has been recorded in 30 days

👎🏻

PR #5725, opened 2025-04-07
Add defaultRoute strategy for konnectivity

shuichiro-makigaki:
To avoid breaking connectivity in the network set up when workers' networks are partitioned.

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/konnectivity.go
+++ b/pkg/component/controller/konnectivity.go
@@ -163,7 +163,7 @@ func (k *Konnectivity) serverArgs(count uint) []string {
 		"--delete-existing-uds-file": "true",
 		"--server-count":             strconv.FormatUint(uint64(count), 10),
 		"--server-id":                k.K0sVars.InvocationID,
-		"--proxy-strategies":         "destHost,default",
+		"--proxy-strategies":         "destHost,defaultRoute,default",
 		"--cipher-suites":            constant.AllowedTLS12CipherSuiteNames(),
 	}.ToArgs()
 }


shuichiro-makigaki:
@twz123 Sorry for long comments!

We have about 200 IoT devices managed by K0s as worker nodes. They are generally working well, and we are very grateful.

However, in our use case, in addition to the IoT devices, we have other worker nodes on the cloud side. On these cloud worker nodes, Webhook, Operator, etc., are running. They are separated from the IoT device network, which is a security specification for our business.

Since the control plane is separated, when the Admission Webhook is issued from the K8s API, it connects to the worker node via connectivity. However, at this time, if an IoT device is selected as the backend, the webhook will not work...! Because it can't communicate with the worker node on the cloud side.

> k0s will not add the "default-route" identifier to any of the agents. 

This is true. So we will create a separate DaemonSet for the worker nodes on the cloud.
The IoT side will use the DaemonSet manifest prepared by k0s. I will have to change the node selector to avoid placing them on the cloud side, though.

> adding the defaultRoute strategy will have no effect at all 

Actually, this is done intentionally. This pull request will not affect "normal" k0s use. In other words, it is backward compatible. But when you use k0s in a geeky way like us, it will work powerfully.

shuichiro-makigaki:
Actually, your question is quite valid. I also thought it would work out-of-the-box.

To be honest, I haven't delved deeply into the implementation of konnectivity, so there might be some incorrect assumptions.

First, our setup is as follows: the control plane (where the `k0s controller` command runs) is on Google Kubernetes Engine. It's a bit confusing that the Kubernetes API runs on Kubernetes, but this isn't an issue.

The worker nodes are either IoT devices or Google Compute Engine instances. On the IoT devices, only the minimum necessary Pods are running, while other system-required Pods (such as those processing webhooks or metrics) are placed on cloud instances.

In this state, here's what I've observed:
`kubectl logs` and `kubectl exec` work "perfectly".
However, only when the Kubernetes API tries to access the Admission Webhook, it fails with an IO Timeout.

Looking at the k0s controller logs, the connectivity server receives requests from the Kubernetes API via a Unix domain socket and tries to access the webhook. For example:
- https://cert-manager-webhook.cert-manager.svc
- https://opentelemetry-operator-webhook.otel-system.svc

After multiple attempts, I observed from the konnectivity server logs that the AgentID changes randomly each time.

At the point when the Kubernetes API tries to access the webhook, the node hasn't been determined yet, so it seems to be federated to a different backend, but I'm not sure if this is correct.

twz123 (approved):
Alright, that makes sense. In addition, k0s should get support for customizing konnectivity agent identifiers one way or another so that this is actually useful without without having to tinker that much.

Thanks for digging into this!

shuichiro-makigaki:
@twz123 I'd like to backport this fix to 1.30, 1.31 and 1.32! What should I do for that?

PR #5759, opened 2025-04-15
Add more complex Docker Compose example

twz123:
## Description

This shows how traefik can be used as a load balancer in front of the control plane, along with a post-start script that sets up join tokens in a way that doesn't require etcd to be operational.

Credits go to @tdesaules for the initial idea!

Also noteworthy: I had to adjust the security flags for `/dev/kmsg` to get the examples working again. I'm pretty sure this wasn't necessary on older kernels, but I couldn't find any reference and didn't bother downgrading to check.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docs/compose-cluster.yaml
+++ b/docs/compose-cluster.yaml
@@ -1,0 +1,177 @@
+name: compose-cluster
+
+x-k0s-controller: &k0s-controller
+  image: docker.io/k0sproject/k0s:v{{{extra.k8s_version}}}-k0s.0
+  networks:
+    - k0s-net
+  tmpfs:
+    - /run # this is where k0s stores runtime data
+    - /tmp
+  configs:
+    - source: k0s.yaml
+      target: /etc/k0s/k0s.yaml
+  labels:
+    - traefik.enable=true
+    - traefik.tcp.routers.kube-api.service=kube-api
+    - traefik.tcp.routers.kube-api.rule=HostSNI(`*`)
+    - traefik.tcp.routers.kube-api.entrypoints=kube-api
+    - traefik.tcp.services.kube-api.loadbalancer.server.port=6443
+    - traefik.tcp.routers.k0s-api.service=k0s-api
+    - traefik.tcp.routers.k0s-api.rule=HostSNI(`*`)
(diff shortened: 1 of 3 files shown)


docs/compose-cluster.yaml
```diff
@@ -0,0 +1,179 @@
...
+    post_start:
+      - command:
+          - /bin/sh
+          - -euc
+          - |
+            bootstrap() {
+              # this works even if etcd isn't up
+
+              rm /run/secrets/k0sproject.io/controller-token/token* || true
+              rm /run/secrets/k0sproject.io/controller-token/*.yaml || true
```
juanluisvaladas, line 95: Why `|| true`instead of `rm -f`? I'd guess we'd want to exit with error if the fails cannot be removed.

docs/compose-cluster.yaml
```diff
@@ -0,0 +1,179 @@
...
+              k0s token pre-shared --role controller \
+                --cert /var/lib/k0s/pki/ca.crt \
+                --url https://k0s-lb:9443 \
+                --out /run/secrets/k0sproject.io/controller-token/
+
+              rm /run/secrets/k0sproject.io/worker-token/token_* || true
+              rm /run/secrets/k0sproject.io/worker-token/*.yaml || true
```
juanluisvaladas, line 102: same here

docs/compose-cluster.yaml
twz123: Yeah, I guess `rm -f ` is also a valid option. When I wrote this, I somehow felt that it might be better to print something to stderr if there were no matching files, but now I don't feel strongly about it anymore. I didn't think so much about the possibility that deleting an existing file might fail. Not sure if we should exit the loop or try again in that case, tho. In the end it doesn't really matter, I suppose?

twz123: PTAL

juanluisvaladas (approved):

docs/compose-cluster.yaml
juanluisvaladas: Fair enough, approved

PR #5778, opened 2025-04-24
Bump go to v1.24.2

juanluisvaladas:
https://groups.google.com/g/golang-announce/c/_G2hEiKx8SE/m/5uEc3vwVAgAJ https://groups.google.com/g/golang-announce/c/Y2uBTVKjBQk/m/cs_6qIK5BAAJ https://go.dev/doc/devel/release#go1.24.2
https://github.com/golang/go/issues?q=milestone%3AGo1.24.2+label%3ACherryPickApproved

PR #5793, opened 2025-04-29
 Bump Kubernetes to v1.33.0

juanluisvaladas:
* Bump kubernetes to 1.33
* Remove unnecessary cleanup 
* Fix new linter issues that appeared after bumping go.mod.


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/k0sproject/k0s
 
-go 1.23.0
+go 1.24.0
+
+toolchain go1.24.2
```
twz123, line 5:
```suggestion
```

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/k0sproject/k0s
 
-go 1.23.0
+go 1.24.0
```
twz123, line 3: Can we separate this into its own Pull Request, along with the linter fixes, since it's belonging to #5778?

jnummelin (approved):

PR #5800, opened 2025-04-30
Support custom CoreDNS configuration

### LINKED ISSUE #4021: Feature: coredns config editing
I have run into an issue with [CoreDNS having a random loadbalancing policy applied to all /etc/resolve.conf entries by default](https://github.com/k3s-io/k3s/discussions/7822).
For reasons, we need to set the coreDNS policy to sequential, so it steps through the internal upstream servers first, then goes on to the upstream external servers. Without this change the cluster is unusable for us.
I could patch this manually, but I think it would be nice to have a way to specify coredns config in k0sctl config syntax.

iawia002:
## Description

Supports customizing the `Corefile` content for CoreDNS (default values remain unchanged). eg:

```yaml
kind: ClusterConfig
metadata:
  name: test
spec:
  network:
    coredns:
      corefile: |
        .:53 {
          errors
          health
          ready
          kubernetes cluster.local in-addr.arpa ip6.arpa {
            pods insecure
            ttl 30
            fallthrough in-addr.arpa ip6.arpa
          }
          prometheus :9153
          cache 30
          reload
          loadbalance
        }
```

cc @twz123 

close #4021

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist


twz123:
Thanks for this! I'm a bit reluctant to add specific customization options for k0s-managed manifests like this. A better (but much more elaborate) approach would be to allow generic patch options, something along the lines of [kubevirt's component customzation](https://kubevirt.io/user-guide/cluster_admin/customize_components/#customize-components-using-patches).

Especially for the CoreDNS config, there'd be another option that bypasses the k0s config: We could offer a way for users to just take ownership of the ConfigMap, for example. Say adding a label or annotation to it that tells k0s to not overwrite it.

dee-kryvenko:
My 2 cents: I do not see how that is a good idea. Sure, for k0s it may sound appealing to push responsibility for the whole configmap over to the user. But now as a user I have to keep track of the configmap as a whole. What if at some point there will be a major change to the configmap i.e. maybe a new feature implemented to the kubernetes plugin? Now, because all I wanted was to add a delegation to another internal zone, or maybe add a `k8s_external` plugin to the mix, I will have to forever remain responsible for watching the upstream changes and reconciling. And because how rare these changes happen in the Corefile - this will certainly become a big disaster one day.

PR #5827, opened 2025-05-08
[Backport release-1.32] Bump konnectivity to v0.32.0

makhov:
Backport #5741 to release-1.32


twz123:
I'm wondering what might be the requirement to have this in v1.32? It's not a simple patch version bump.

jnummelin:
@twz123 There's two "drivers" for this.

First is the fact that 0.32 has the logging fixes @makhov contributed so now konnectivity is much less noisy. Some k0smotron users have been complaining about this.

Second is that the konnectivity versioning scheme says:
> Releases and tags align with the Kubernetes minor release versions (the y in x.y.z). For instance, if Kubernetes releases version 1.99.0, the corresponding release and tag for apiserver-network-proxy will be 0.99.0.

IMO that means we _should_ really use 0.32 for our Kube 1.32 branch

twz123:
Hmm, still not entirely convinced. The logging changes broke the k0s OS tests. I admit that you shouldn't count the log format as a stable API, but it's still quite an unexpected change for a point release, going from chatty to almost dead. If we really want to have the version parity between konnectivity and Kubernetes, I'd add a unit test for that, as we have for all the other version constraints, so this won't go unnoticed in the future.

PR #5834, opened 2025-05-12
fix(network_config): don't set kuberouter config if calico config provided

JGAntunes:
## Description

When the network configuration has `provider` set to `calico` and a `calico` config is provided, a default `kuberouter` config is also being initialised in the struct. The following example illustrates the problem:

```go
package main

import (
	"fmt"

	k0sconfig "github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
	k8syaml "sigs.k8s.io/yaml"
)

var resultYAML = `
kind: ClusterConfig
metadata:
  creationTimestamp: null
  name: embedded-cluster
spec:
  network:
    calico:
      flexVolumeDriverPath: /usr/libexec/k0s/kubelet-plugins/volume/exec/nodeagent~uds
      mode: vxlan
      mtu: 1450
      overlay: Always
      vxlanPort: 4789
      vxlanVNI: 4096
    provider: calico
`

func main() {
	var patched k0sconfig.ClusterConfig

	if err := k8syaml.Unmarshal([]byte(resultYAML), &patched); err != nil {
		fmt.Errorf("unable to unmarshal patched config: %w", err)
	}
	fmt.Printf("Provider: %s\n", patched.Spec.Network.Provider)
	fmt.Printf("Calico: %v\n", patched.Spec.Network.Calico)
	fmt.Printf("KubeRouter: %v\n", patched.Spec.Network.KubeRouter)
}

```

Renders:

```sh
$> go run main.go
Provider: calico
Calico: &{false map[] /usr/libexec/k0s/kubelet-plugins/volume/exec/nodeagent~uds   1450 vxlan Always 4789 4096}
KubeRouter: &{0x140004aab44 0 8080 Enabled false false   map[]}
```

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -196,12 +196,13 @@ func (n *Network) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	if n.Provider == "calico" {
+	switch n.Provider {
+	case "calico":
 		if n.Calico == nil {
 			n.Calico = DefaultCalico()
 		}
 		n.KubeRouter = nil
-	} else if n.Provider == "kuberouter" {
+	case "kuberouter":
 		if n.KubeRouter == nil {
 			n.KubeRouter = DefaultKubeRouter()
 		}
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -196,11 +196,15 @@ func (n *Network) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	if n.Provider == "calico" && n.Calico == nil {
-		n.Calico = DefaultCalico()
+	if n.Provider == "calico" {
+		if n.Calico == nil {
+			n.Calico = DefaultCalico()
+		}
 		n.KubeRouter = nil
-	} else if n.Provider == "kuberouter" && n.KubeRouter == nil {
-		n.KubeRouter = DefaultKubeRouter()
+	} else if n.Provider == "kuberouter" {
+		if n.KubeRouter == nil {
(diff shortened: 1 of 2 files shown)


jnummelin:
@JGAntunes Do I get it right, you'd expect `patched.Spec.Network.KubeRouter` to be nil in the case calico config is provided?

JGAntunes:
Thanks for the reply @jnummelin.

> @JGAntunes Do I get it right, you'd expect patched.Spec.Network.KubeRouter to be nil in the case calico config is provided?

Correct, that is my interpretation from looking through the code, given the provider is set and a config is set I don't see why we would need to have the default config for kube router also present. Unless this is expected/by design for some specific reason I'm missing.

> What is the use case where this causes any problems?

This started tripping our unit tests in embedded cluster when we started looking through supporting k0s v1.31 - https://github.com/replicatedhq/embedded-cluster/pull/2148/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5L72 - the reason we're only seeing this now seems to be related to the introduction of the `omitempty` tag which triggers this on our side due to the marshaling/unmarshaling logic in place. My understanding is that the config will be disregarded if the selected `provider` isn't `kuberouter`(?) so I don't think this will lead to any issues in the resulting cluster. Nevertheless, I don't think it makes much sense to keep this redundant config in the resulting cluster config if there's no actual need for it, specially since it wasn't even set by the user.

twz123 (approved):

JGAntunes:
Forgot to signoff my last commit. Had to ammend and force push again, apologies.

twz123:
Thanks!

PR #5861, opened 2025-05-20
[bootloose image] Infer helm version from go mod

juanluisvaladas:
## Description

Fix the mismatch between `go.mod` and `embedded-bins/Makefile.variables`

### DIFF
--- a/inttest/Makefile
+++ b/inttest/Makefile
@@ -19,6 +19,7 @@ bin:
 bin/sonobuoy: | bin
 	$(curl) $(sonobuoy_url) | tar -C bin/ -zxv $(notdir $@)
 
+helm_version = $(strip $(subst v,,$(shell ../vars.sh FROM=../ helm_version)))
 bootloose_alpine_build_cmdline := \
 	--build-arg GOLANG_IMAGE=$(golang_buildimage) \
 	--build-arg ALPINE_VERSION=$(alpine_patch_version) \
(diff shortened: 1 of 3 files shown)


twz123:
I'd either do the version detection as we do it for k0sctl, so that we can rely on dependabot, or remove the interest completely.

inttest/Makefile
```diff
@@ -19,6 +19,7 @@ bin:
 bin/sonobuoy: | bin
 	$(curl) $(sonobuoy_url) | tar -C bin/ -zxv $(notdir $@)
 
+helm_version = $(strip $(subst v,,$(shell cd .. && go list -m -f '{{.Version}}' helm.sh/helm/v3 2>/dev/null)))
```
twz123, line 22: Hmm, usually this is handled via vars.sh, but that's in the repo root and not part of the inttest folder ... :thinking: But since we have a `cd ..` in here anyways, I'd prefer sth. like `../vars.sh FROM=.. helm_version`.

Or, since we're already looking into Renovate, we could just leave the version here around and add a regex pattern that bundles the updates :thinking:

juanluisvaladas: That was my initial intention but I think this is saner.

Anyway I changed it to use `vars.sh`

twz123:
This is now managed via Renovate. See #6163.

PR #5862, opened 2025-05-20
Introduce renovate bot

juanluisvaladas:
## Description

Dependabot has limited scope and it cannot update things based on regex or things that we define in Makefiles.

This is a very basic first configuration file for renovate bot which can update etcd and kine as a proof of concept so that we decide if we want to implement this and invest more time updating every component.

After this is merged we need an extra step which is enabling it in developer.mend.io.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,71 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:recommended",
+    ":configMigration",
+    ":gitSignOff"
+  ],
+  "includeForks": true,
+  "enabledManagers": [
+    "gomod",
+    "custom.regex"
+  ],
+  "packageRules": [
+    {
+      "description": "Ignore all Go dependencies by default",
+      "enabled": false,
+      "matchManagers": [
+        "gomod"
+      ]
+    },
+    {
+      "description": "Group all etcd updates (go.mod and Makefile) into a single PR",
+      "matchPackageNames": [
+        "go.etcd.io/etcd/**/v3",
+        "etcd-io/etcd"
+      ],
+      "groupName": "etcd dependencies",
+      "groupSlug": "etcd-all"
+    },
+    {
+      "description": "Ensure etcd in Makefile uses github-releases and is enabled",
+      "matchManagers": [
+        "custom.regex"
+      ],
+      "matchPackageNames": [
+        "etcd-io/etcd"
+      ],
+      "enabled": true
+    }
+  ],
+  "customManagers": [
+    {
+      "customType": "regex",
+      "description": "Update etcd version in embedded-bins/Makefile.variables",
+      "managerFilePatterns": [
+        "embedded-bins/Makefile.variables"
+      ],
+      "matchStrings": [
+        "etcd_version = (?<currentValue>v?[\\d\\.]+)"
+      ],
+      "depNameTemplate": "etcd-io/etcd",
+      "datasourceTemplate": "github-releases",
+      "versioningTemplate": "semver"
+    },
+    {
+      "customType": "regex",
+      "description": "Update kine versions across the codebase",
+      "managerFilePatterns": [
+        "/(^|/)[Mm]akefile(\\.[^/]+)?$/",
+        "*.go"
+      ],
+      "matchStrings": [
+        "kine_version = (?<currentValue>[\\d\\.]+)",
+        "\\bkine/blob/(v?<currentValue>[\\d\\.]+)\\b"
+      ],
+      "depNameTemplate": "k3s-io/kine",
+      "datasourceTemplate": "github-releases",
+      "versioningTemplate": "semver"
+    }
+  ]
+}
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -1,0 +1,22 @@
+name: Renovate
+
+on:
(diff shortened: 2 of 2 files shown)


juanluisvaladas:
To add support for this we'd need to merge this file and enable it in developer.mend.io (doesn't have additional costs)

I tested this on my own repository and I'm extremely happy with the results:
- etcd: https://github.com/juanluisvaladas/k0s/pull/6
- kine: https://github.com/juanluisvaladas/k0s/pull/7

I think this has insane value/effort and we can automate a lot of things in embedded-bins.

twz123:
I'm _so_ glad that you've put your time into this. Something that we should've done way earlier. Did you consider to use the GitHub Action for running Renovate ourselves, instead of relying on the external integration?

twz123:
> I like the external integration though and so far I haven't seen any reason not to use it (assuming that it's reliable over time). I think it's much easier to test on our personal forks, we don't have to worry about tokens and permissions and I don't know how the checkbox to trigger rebuilding a PR would work if we're self hosted. But certainly it's something we can look into but so far I prefer the external integration.

What I don't like that much is that I need to sign up with another external service. So from my PoV, testing on forks is even easier using the GitHub Action, as it will "just work" out of the box, just as any other CI build, including retries (just retrigger the Action run) and logs and so on. The action is self-contained, i.e. it can just use the GITHUB_TOKEN and request the right set of permissions using the "permissions" directive in the actions file. It's "debuggable" too, in a way, as the action itself is open source, and running the bot just boils down to run the Renovate Docker image once on the repo.

renovate.json
```diff
@@ -0,0 +1,62 @@
...
+      "depNameTemplate": "etcd-io/etcd",
+      "datasourceTemplate": "github-releases",
+      "versioningTemplate": "semver"
+    },
+    {
+      "description": "Update kine version in embedded-bins/Makefile.variables",
+      "fileMatch": ["^embedded-bins/Makefile\\.variables$"],
+      "matchStrings": [
+        "kine_version = (?<currentValue>v?[\\d\\.]+)"
+      ],
```
twz123, line 53: We can very well use this to replace all the kine versions:

```suggestion
      "description": "Update kine versions across the codebase",
      "managerFilePatterns": [
        "/(^|/)[Mm]akefile(\\.[^/]+)?$/",
        "*.go"
      ],
      "matchStrings": [
        "kine_version = (?<currentValue>[\\d\\.]+)",
        "\\bkine/blob/(v?<currentValue>[\\d\\.]+)\\b"
      ],
```

renovate.json
```diff
@@ -0,0 +1,62 @@
...
+      "description": "Update kine version in Makefile.variables from k3s-io/kine releases",
+      "matchManagers": ["regex"],
+      "matchPackageNames": ["k3s-io/kine"],
+      "enabled": true
+    }
+  ],
+  "regexManagers": [
+    {
+      "description": "Update etcd version in embedded-bins/Makefile.variables",
+      "fileMatch": ["^embedded-bins/Makefile\\.variables$"],
```
twz123, line 40: New syntax (supports globs):

```suggestion
      "managerFilePatterns": ["embedded-bins/Makefile.variables"],
```

renovate.json
```diff
@@ -0,0 +1,62 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base",
+    ":gitSignOff"
+  ],
+  "packageRules": [
```
twz123, line 7: I'd say start small and only enable the managers that we've tested, and only update selected dependencies. That gives us a way to migrate stuff from dependabot to Renovate if we want to, without having both of them doing the same stuff:

```suggestion
  "enabledManagers": [
    "gomod",
    "custom.regex"
  ],
  "packageRules": [
    {
      "description": "Ignore all Go dependencies by default",
      "enabled": false,
      "matchManagers": [
        "gomod"
      ]
    },
```

twz123 (commented):
:robot: :rocket:

renovate.json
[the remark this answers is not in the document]
juanluisvaladas: I'm using whatever VS does by default but I don't have it configured it by default so it's not correct according to it. Since I don't have it properly configured I'll set up my editor to whatever you're using and add a step on the linting job to make sure it's properly formatted.

renovate.json
```diff
@@ -0,0 +1,62 @@
...
+      "matchManagers": ["regex"],
+      "matchPackageNames": ["etcd-io/etcd"],
+      "enabled": true
+    },
+    {
+      "description": "Update kine version in Makefile.variables from k3s-io/kine releases",
+      "matchManagers": ["regex"],
+      "matchPackageNames": ["k3s-io/kine"],
+      "enabled": true
+    }
```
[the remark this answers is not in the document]
juanluisvaladas: Let's remove it, kine needs to be updated so if the PR isn't created automatically we'll add it again.

juanluisvaladas:
I added a basic github action, I'm struggling with the permissions for the pull requests.

There are two ways of doing this, through a github app or through a github user which is a bot account. I believe github apps are the prefered way soI made an app and I'm testing it on my local branch. Currently I'm struggling with acquiring the token for said app...

twz123 (approved):

twz123:
Eager to test this out!

PR #5868, opened 2025-05-22
Fix renovate etcd matcher

juanluisvaladas:
## Description

Because we disable the gomod matchManager the etcd packages were only enabled with custom.regex. We need to enable it explicitly in the packageRule.


juanluisvaladas:
Resulting PR: https://github.com/juanluisvaladas/k0s/pull/21/files

twz123 (approved):

PR #5873, opened 2025-05-23
Remove timeout from dynamic cluster config initializer

twz123:
## Description

For some nodes, initializing the cluster configuration can take quite a while, especially during the initial bootstrap phase, if many user-supplied stacks are applied. Moreover, if the cluster has a stale leader lease, it will take at least 60 seconds until it gets reacquired by Kubernetes leader election, hence the existing 20 second timeout will lead to a crash loop.

Remove the timeout completely and rely on external cancellation. Increase the log level for the error in the retry loop once every minute from debug to info so the retry loop is visible in non-debug logs without flooding them every second.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/clusterconfig.go
+++ b/pkg/component/controller/clusterconfig.go
@@ -17,7 +17,6 @@ limitations under the License.
 package controller
 
 import (
-	"cmp"
 	"context"
 	"errors"
 	"fmt"
@@ -39,7 +38,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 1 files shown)


pkg/component/controller/clusterconfig.go
```diff
@@ -204,36 +203,39 @@ func (i *ClusterConfigInitializer) ensureClusterConfigExistence(ctx context.Cont
...
 
-		i.log.WithError(err).Debug("Failed to ensure the existence of the cluster configuration")
-		return false, nil
-	})
-
-	if pollErr != nil {
-		pollErr = cmp.Or(context.Cause(ctx), pollErr)
-		if err != nil {
-			return fmt.Errorf("%w (%w)", pollErr, err)
+		var log func(...any)
```
juanluisvaladas, line 222: Maybe we should consider exiting with an error at some point.
Knowing leader elector takes at least 60 seconds to deal with a stale leader, maybe it makes sense to actually crash after 5 minutes.

twz123: Every value we put here seems kinda arbitrary. In the past years, I've hit so many cases of arbitrary timeouts breaking stuff that I I'm genuinely convinced that if unsure which timeout to pick, it's better to pick none at all (given that your caller can pick one). What I mean by that is that it's better to force users to hit Ctrl+C, or issue a `systemctl restart ...` if they think something takes too long than hard-coding a non-overridable value that will break stuff even when users know that it'll succeed eventually.

Issue #5877, opened 2025-05-23
Allow providing CA certificate for helm extensions that use OCI registries


byDimasik:
### Is your feature request related to a problem? Please describe.

Helm extension controller is unable to pull helm charts from an OCI registry that uses a self-signed CA certificate.

I create my helm extension as follows:
```yaml
      extensions:
        helm:
          charts:
          - chartname: oci://registry.mke4dev.mirantis.net/mke/mirrored-chart-tigera-operator
            name: tigera-operator
            namespace: tigera-operator
            order: 4
            timeout: 10m0s
```

Here, `oci://registry.mke4dev.mirantis.net` is my private registry that uses TLS certificates issued by a private self-signed CA. The helm extension controller fails to pull this chart with an error
```
May 23 14:22:23 node0 k0s[454]: time="2025-05-23 14:22:23" level=error msg="Reconciler error" Chart="{k0s-addon-chart-tigera-operator kube-system}" component=extensions_controller controller=chart controllerGroup=helm.k0sproject.io controllerKind=Chart error="can't update or install chart: can't reconcile installation for \"k0s-addon-chart-tigera-operator\": can't locate chart `oci://registry.mke4dev.mirantis.net/mke/mirrored-chart-tigera-operator-v3.29.3`: failed to do request: Head \"https://registry.mke4dev.mirantis.net/v2/mke/mirrored-chart-tigera-operator/manifests/v3.29.3\": tls: failed to verify certificate: x509: certificate signed by unknown authority" name=k0s-addon-chart-tigera-operator namespace=kube-system reconcileID="\"435a14d2-1632-4faf-b519-18fee6103c03\""
```

### Describe the solution you would like

I would like to be able to provide the CA certificate my registry uses as part of the helm chart spec. 

### Describe alternatives you've considered

I tried adding a helm repository, because repositories already allow setting CA cert
```yaml
      repositories:
      - caFile: /etc/containerd/certs.d/registry.mke4dev.mirantis.net/ca.crt
        name: mke-repository
        url: oci://registry.mke4dev.mirantis.net
```

However, these repositories are considered classic helm repositories that expect an index.yaml file to be present. OCI registries are not classic helm repositories and therefore cannot be added this way. The helm extension controller throws an error
```
[...]

PR #5901, opened 2025-05-29
Support extensions from OCI registries with self-signed CA

byDimasik:
## Description

Allow providing TLS certificates for OCI registries in Helm commands. Reuse existing v1beta1.Repository type to avoid changing CRDs.

While k0s already supports custom TLS transport for classic Helm repositories, it is lacking the same functionality for OCI registries. This makes it impossible to install Helm extensions from OCI registries that use private/self-signed CA for TLS.

With this change, users can add their OCI registry as a Helm repository and provide TLS certificates. This WILL NOT modify the repositories.yaml file, but instead create a custom registry client for the provided OCI registry. This client will be used with charts that reference the same registry in the chartName.

Fixes https://github.com/k0sproject/k0s/issues/5877

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


inttest/addons/addons_test.go
```diff
@@ -47,8 +56,79 @@ import (
...
+	echoServerTgzPath = "./testdata/chart/echo-server-0.5.0.tgz"
+)
+
 type AddonsSuite struct {
 	common.BootlooseSuite
+	localRegistryPort int
+}
+
+// initRegistry starts a local docker registry on a free random port with a self-signed CA certificate
+func (as *AddonsSuite) initRegistry() {
```
juanluisvaladas: The test is well written but I'm concerned with the dependencies. I know these aren't dependencies of the actual actual binary, they are in the same go.mod and I'm concerned that at some point there could be a conflict.

I think it's worth debating with the rest of the team if we prefer this or if we want to add a new field `WithRegistry` or `WithOCIRegistry` [similar to `WithExternalEtcd` and `WithLB`](https://github.com/k0sproject/k0s/blob/v1.33.1%2Bk0s.0/inttest/common/bootloosesuite.go#L100-L101). And yes, I'm aware BootlooseSuite is already fairly bloated.

And then creating the container [should be fairly easy](https://github.com/k0sproject/k0s/blob/v1.33.1%2Bk0s.0/inttest/common/bootloosesuite.go#L1203-L1226) except I believe distribution TLS configuration [can only be configured via environment variables](https://distribution.github.io/distribution/about/deploying/#get-a-certificate) and I believe that's not an option in bootloose at the moment. But other than that I don't see any other problems with this approach, we already rely on dockerhub images in our test suite.

byDimasik: @juanluisvaladas I've implemented the suggested change. Registry now is a bootloose container, and no new deps are added to the `go.mod`

The only concern I have is that I use the upstream `registry` image, which technically isn't bootloose-compliant. It doesn't run systemd as PID 1 and doesn't have SSH, but nevertheless it works fine, and there is no need for SSH or systemd in the registry container, so I hope this is fine for test needs.

inttest/addons/addons_test.go
```diff
@@ -47,18 +51,74 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
+const (
+	registryTLSPath = "./testdata/tls"
+	// registryCACertPath is pre-generated and valid until May 23, 2045
```
twz123: Generating a certificate on the fly would also be an option. We're doing it in other tests already: https://github.com/k0sproject/k0s/blob/v1.33.1%2Bk0s.1/internal/http/download_test.go#L205-L215

pkg/helm/oci_test.go
```diff
@@ -0,0 +1,210 @@
...
+	testOCIRegistryURL         = "oci://" + testRegistryHost
+	testOCIRegistryURLWithPort = "oci://" + testRegistryHostPort
+	testOCIRegistryURLWithRepo = testOCIRegistryURL + "/my-repo"
+
+	testCAPath   = "./testdata/tls/ca.crt"
+	testCertPath = "./testdata/tls/tls.crt"
+	testKeyPath  = "./testdata/tls/tls.key"
+)
+
+func TestOciRegistryManager_AddRegistry(t *testing.T) {
```
twz123: OCI should probably be uppercase: `TestOCIRegistryManager_AddRegistry`. Same for the other test methods.

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+	return &ociRegistryManager{
+		knownClients: make(map[string]*registry.Client),
+	}
+}
+
+// AddRegistry adds a new registry client for the given repository configuration
+func (m *ociRegistryManager) AddRegistry(repoCfg v1beta1.Repository) error {
+	registryURL, err := url.Parse(repoCfg.URL)
+	if err != nil {
+		return fmt.Errorf("can't parse registry URL %s: %w", repoCfg.URL, err)
```
twz123: The err returned from url.Parse will already include the URL itself, so no need to print it twice.

```suggestion
		return fmt.Errorf("can't parse repository URL: %w", err)
```

docs/extensions.md
```diff
@@ -17,6 +17,9 @@ helm:
     url: https://charts.helm.sh/stable
   - name: prometheus-community
     url: https://prometheus-community.github.io/helm-charts
+  - name: oci-registry-with-private-ca
+    url: oci://registry-with-private-ca.com:8080
+    caFile: /path/to/ca.crt
```
twz123, line 25: As we seem to support mTLS (certFile, keyFile), would it make sense to mention it here?

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	if registryURL.Scheme != "oci" {
```
twz123: Found a constant for this in helm: https://pkg.go.dev/helm.sh/helm/v3@v3.18.2/pkg/registry#OCIScheme

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+		return fmt.Errorf("registry URL %s is not an OCI registry", repoCfg.URL)
+	}
```
twz123: Helm has [`registry.IsOCI`](https://pkg.go.dev/helm.sh/helm/v3@v3.18.2/pkg/registry#IsOCI), so we could write it like this:

```suggestion
	if !registry.IsOCI(repoCfg.URL) {
		return fmt.Errorf("%w: not an OCI registry", errors.ErrUnsupported)
	}

	registryURL, err := url.Parse(repoCfg.URL)
	if err != nil {
		return fmt.Errorf("can't parse repository URL: %w", err)
	}
```

And then leverage the error type at the callsite and get rid of our own version of IsOCI completely:

```go
if err := hc.registryManager.AddRegistry(repoCfg); !errors.Is(err, errors.ErrUnsupported) {
	return err
}
```

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+	if repoCfg.CertFile != "" && repoCfg.KeyFile != "" {
+		keyPair, err := tls.LoadX509KeyPair(repoCfg.CertFile, repoCfg.KeyFile)
+		if err != nil {
+			return fmt.Errorf("can't load certificate and key files: %w", err)
+		}
+		tlsConfig.Certificates = []tls.Certificate{keyPair}
+	}
+
+	registryClient, err := registry.NewClient(
+		registry.ClientOptWriter(os.Stdout),
```
twz123, line 99: That should go to the logger. There's some precedence for this in k0s's supervisor. However, the `logWriter` struct that's used over there is not exported. Maybe time to pull this out of the supervisor, and put it into, say, `internal/log`, and reuse it here? Same for the other usages of `os.Stdout`.

I can do the code shuffling in a preliminary PR if you don't want to do it in this one.

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+		registry.ClientOptEnableCache(true),
```
twz123, line 100: Worth a comment why that gets enabled here.

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+		registry.ClientOptBasicAuth(repoCfg.Username, repoCfg.Password),
+		registry.ClientOptHTTPClient(&http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: tlsConfig,
+				Proxy:           http.ProxyFromEnvironment,
+			},
+		}),
+	)
+	if err != nil {
+		return fmt.Errorf("can't create registry client for registry %s: %w", repoCfg.URL, err)
```
twz123: I wonder if we should phrase it like this:

```suggestion
		return fmt.Errorf("can't create registry client for repository %s: %w", repoCfg.Name, err)
```

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"helm.sh/helm/v3/pkg/registry"
+	"k8s.io/client-go/util/cert"
+
+	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+)
+
+// ociRegistryManager is a helper struct for managing OCI registry clients
+type ociRegistryManager struct {
+	knownClients map[string]*registry.Client
```
twz123: We need to be prepared for concurrent usage here. The repo addition and the chart reconciliation happen from different goroutines. So we'd either need a `sync.RWMutex` or a `sync.Map` here.

inttest/addons/addons_test.go
```diff
@@ -47,18 +51,74 @@ import (
...
 	addonName := "test-addon"
 	ociAddonName := "oci-addon"
 	fileAddonName := "tgz-addon"
-	as.PutFile(as.ControllerNode(0), "/tmp/k0s.yaml", fmt.Sprintf(k0sConfigWithAddon, addonName))
+	selfSignedOCIAddonName := "self-signed-oci-addon"
+
+	p := k0sConfigParams{
+		BasicAddonName:      addonName,
+		LocalRegistryCAPath: registryCACertContainerPath,
+		LocalRegistryHost:   "host.docker.internal",
```
twz123, line 166: The inttest doesn't work on my machine :thinking:

> Test addon release doesn't meet criteria yet (version "935"): no release name; expected revision to be 1, but was 0; expected error to be empty, but was "can't locate chart `oci://host.docker.internal:33074/charts/echo-server-0.5.0`: failed to perform \"FetchReference\" on source: Get \"https://host.docker.internal:33074/v2/charts/echo-server/manifests/0.5.0\": dial tcp 172.17.0.1:33074: connect: connection timed out"

inttest/common/bootloosesuite.go
```diff
@@ -1244,6 +1265,13 @@ func (s *BootlooseSuite) initializeBootlooseClusterInDir(dir string) error {
 		})
 	}
 
+	if s.WithRegistry {
+		cfg.Machines = append(cfg.Machines, config.MachineReplicas{
```
twz123, line 1272: The test suite tries to download k0s logs from the registry:

> Failed to ssh into registry0 to get logs: hostport: failed to inspect container: [ template parsing error: template: :1:4: executing "" at <index (index .NetworkSettings.Ports "22/tcp") 0>: error calling index: index of untyped nil]: exit status 1

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+
+// GetRegistryClient returns a registry client for a previously added registry URL or
+// a default registry client if the URL is not known.
+func (m *ociRegistryManager) GetRegistryClient(rawRegistryURL string) (*registry.Client, error) {
+	registryURL, err := url.Parse(rawRegistryURL)
+	if err != nil {
+		return m.getDefaultRegistryClient()
+	}
+
+	client, exists := m.knownClients[registryURL.Host]
```
twz123: Since we support mTLS here, I'd say we should be more restrictive for the client selection. Maybe we should match the host plus the whole base path. Context: https://github.com/k0sproject/k0s/issues/5877#issuecomment-2915394044

pkg/helm/helm.go
```diff
@@ -220,12 +230,19 @@ func (hc *Commands) InstallChart(ctx context.Context, chartName string, version
...
 		return nil, fmt.Errorf("can't create action configuration: %w", err)
 	}
+
+	registryClient, err := hc.registryManager.GetRegistryClient(chartName)
+	if err != nil {
+		return nil, fmt.Errorf("can't get registry client for chart `%s`: %w", chartName, err)
+	}
+
 	install := action.NewInstall(cfg)
+	install.SetRegistryClient(registryClient)
```
twz123: I wonder if we should only set the registry client for OCI charts? So maybe return `nil, nil` instead of returning a default client?

pkg/helm/oci.go
```diff
@@ -0,0 +1,111 @@
...
+func (m *ociRegistryManager) getDefaultRegistryClient() (*registry.Client, error) {
+	return registry.NewClient(
+		registry.ClientOptWriter(os.Stdout),
+		registry.ClientOptEnableCache(true),
+	)
+}
+
+// GetRegistryClient returns a registry client for a previously added registry URL or
+// a default registry client if the URL is not known.
+func (m *ociRegistryManager) GetRegistryClient(rawRegistryURL string) (*registry.Client, error) {
```
twz123: Maybe we should only return a client for OCI charts, instead of returning a client for every chart?

pkg/helm/oci.go
twz123: Did you consider to create all clients on the fly instead of caching them? We don't cache the default client, so why should we do it for the others?

pkg/helm/oci.go
byDimasik: @twz123 

> I can do the code shuffling in a preliminary PR if you don't want to do it in this one.

That would be very nice of you. I just checked the code, and it looks to me that it won't be a simple move from one folder to another; it actually needs some refactoring. 

Also, I wasn't reinventing the wheel here, `os.Stdout` was used in multiple places before my changes, e.g.
- https://github.com/k0sproject/k0s/blob/e401b131f397670e5de5fc62e22fa7556ae664a0/pkg/helm/helm.go#L165
- https://github.com/k0sproject/k0s/blob/e401b131f397670e5de5fc62e22fa7556ae664a0/pkg/helm/helm.go#L196

So I believe to avoid the scope creep for this PR, it should be ok to use stdout, and do the whole logging refactoring as a separate effort. I can file an issue if you would like me to.

pkg/helm/oci.go
byDimasik: The only reason I enabled this is because it's enabled by default in helm - https://github.com/helm/helm/blob/main/pkg/cmd/root.go#L362

I'll add a comment

docs/extensions.md
byDimasik: @twz123 

I don't have tests for it because #5877 only mentions the CA certificate. Technically, `certFile` and `keyFile` are out of scope of this change, but included in the `oci.go` as a side effect. 

If you think it's necessary to have it here, I can spend some time modifying the tests to cover mTLS as well, and once it's confirmed to be working, I'll modify the docs. Wdyt?

inttest/addons/addons_test.go
byDimasik: Hmm, maybe you have some firewall rules that break it? Seems to be a common issue
- https://github.com/moby/moby/issues/46892#issuecomment-1855799760
- https://forums.docker.com/t/curl-on-host-docker-internal-on-a-specific-port-runs-in-timeout/142989/2

The tests passes on my machine and [in CI](https://github.com/k0sproject/k0s/actions/runs/15428486369/job/43452628826?pr=5901) so most likely there is something about your setup

byDimasik:
@twz123 

Thank you for the thorough review. You've mentioned mTLS support multiple times
- https://github.com/k0sproject/k0s/pull/5901#discussion_r2126304450
- https://github.com/k0sproject/k0s/pull/5901#discussion_r2126495627
and this was not part of the scope for #5877. 

The issue I'm trying to solve is related only to self-signed CA. It looks like proper mTLS support requires changes in `ociRegistryManager`, tests, and docs. I'll be off next week, and the CA issue is a blocker for MKE that must be included in the next release. If I make all the changes necessary for mTLS, I may not be able to get this reviewed and merged in 2 days before my vacation. I don't want to delay this PR for too long because of mTLS. mTLS is currently not requested by any k0s or MKE users, while self-signed CA support is a must-have.

So, how about I'll add a code to throw an error if `certFile` and/or `keyFile` is provided for OCI registry, saying that mTLS for OCI registries is not currently supported. We will proceed with this PR focusing purely on the CA cert. I'll create a new k0s issue for mTLS, and can work on this once I'm back from vacation. 

We don't need mTLS in the next k0s release, but self-signed CA support is a must-have, so I want to tightly limit the scope of this change to CA support only to make sure I merge it asap and it makes its way into the next release. How does this sound to you?

inttest/addons/addons_test.go
byDimasik: done

pkg/helm/oci_test.go
byDimasik: done

pkg/helm/oci.go
byDimasik: done

docs/extensions.md
byDimasik: please see my comment below - https://github.com/k0sproject/k0s/pull/5901#issuecomment-2941637572

pkg/helm/oci.go
byDimasik: done

pkg/helm/oci.go
byDimasik: changed as suggested

pkg/helm/oci.go
byDimasik: done

pkg/helm/oci.go
byDimasik: done

pkg/helm/oci.go
byDimasik: done

inttest/common/bootloosesuite.go
byDimasik: fixed - https://github.com/k0sproject/k0s/pull/5901/files#diff-5ce55b33fdea68d04798d1490596b0293a68fc70375cbf254de6f3880dcbf95eR391

pkg/helm/oci.go
byDimasik: please see my comment below - https://github.com/k0sproject/k0s/pull/5901#issuecomment-2941637572

pkg/helm/helm.go
byDimasik: done

pkg/helm/oci.go
byDimasik: done

pkg/helm/oci.go
byDimasik: done

byDimasik:
@twz123 I addressed all your comments.

I also removed the cert and key file handling and replaced it with an error that mTLS is not supported for OCI registries. I will proceed with filing a separate enhancement proposal for this.

Now, this change is entirely focused on CA cert

twz123:
I'm fine with not testing or documenting mTLS. However, I'm concerned that we're only matching on the host name and disregarding the path. While this is sufficient for selecting the right CA chain, it could affect future mTLS support. If we leave it as it is now, adding base path matching in the future would be a breaking change.

As a compromise, we could allow repository entries only for OCI registries with the host set and no path elements. This would allow us to opt to support path matching in the future without compromising backwards compatibility.

inttest/addons/addons_test.go
juanluisvaladas: > The only concern I have is that I use the upstream registry image, which technically isn't bootloose-compliant. It doesn't run systemd as PID 1 and doesn't have SSH, but nevertheless it works fine[...]

Since it's something self-contained I don't think that's a big issue at all to be honest. I'm happy to have this in it's current state.

byDimasik:
> I'm fine with not testing or documenting mTLS. However, I'm concerned that we're only matching on the host name and disregarding the path. While this is sufficient for selecting the right CA chain, it could affect future mTLS support. If we leave it as it is now, adding base path matching in the future would be a breaking change.
> 
> As a compromise, we could allow repository entries only for OCI registries with the host set and no path elements. This would allow us to opt to support path matching in the future without compromising backwards compatibility.

@twz123 I've added a path check - https://github.com/k0sproject/k0s/pull/5901/files#diff-bd56e3ef5245c767ecb2d396b681f2d1ea34e530d60151456e1764161984a577R54-R58

Please take another look

inttest/addons/addons_test.go
twz123: Thanks for digging into it! Will check that out!

twz123 (dismissed):
Looks good! Kudos for the integration test!

byDimasik:
> Looks good! Kudos for the integration test!

@twz123 thank you for the review! 

Now, how do I make sure it's backported to 1.32?

twz123:
Like so. We'll backport it to all actively maintained releases, as long as there's no reason not to do it, starting with 1.33.

byDimasik:
> Like so. We'll backport it to all actively maintained releases, as long as there's no reason not to do it, starting with 1.33.

@twz123 
Sounds good! Thank you! Do I get this right that you'll take care of it from here?
I'll keep an eye on it in case you need my help with anything, but as I mentioned earlier, I'll be off next week, and I hope this will be finalized by the release time.

Btw, as promised, here is an issue for potential mTLS support in the future - https://github.com/k0sproject/k0s/issues/5932

twz123:
> Do I get this right that you'll take care of it from here?

Yep! :saluting_face: :upside_down_face:

> Btw, as promised, here is an issue for potential mTLS support in the future - #5932

Nice!

byDimasik:
@twz123 Recent changes in the main branch created some merge conflicts that I just resolved, but now the PR needs approval again. Could you take a look?

twz123 (approved):

tppolkow:
@twz123 I noticed the backport failed, is there any manual action needed by us? Checking in on this for Dmitrii as he's out on PTO this week and we need this PR to make the next release.

twz123:
@tppolkow No action required, as said in https://github.com/k0sproject/k0s/pull/5901#issuecomment-2948285919. It's planned to be included in the next release.

PR #5855, opened 2025-05-19
Use the correct address for BuildServiceCIDR

juanluisvaladas:
## Description

We were using always API.Spec.Address as a source of information for BuildServiceCIDR which resulted in an incorrect service CIDR if spec.API.ExternalAddress was IPv6 and spec.API.Address was IPv4.

Instead use Spec.API.Address() which results in a correct value.

Fixes #5841
## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #5912, opened 2025-06-02
Improve API endpoint reconciliation on dual stack

juanluisvaladas:
## Description

After https://github.com/k0sproject/k0s/pull/5855 we had a discussion regarding what would happen if the API address is a hostname in slack: https://kubernetes.slack.com/archives/C07VAPJUECS/p1748427050289409

After some deliberation, I think we simply cannot rely on DNS for determining whether we should prioritize IPv4 or IPv6 as it's not reliable enough and therefore this PR introduces two changes:

1- Allow the end user to chose explicitly the cluster address family while maintaining the older behavior so that we don't break existing installations.
2- In the API endpoint reconciler only lookup addresses for the primary address family. 

I'm not 100% happy with this solution, but I don't think we can come up with something much better until config v2 because we need to preserve existing behavior.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist


docs/configuration.md
```diff
@@ -163,6 +163,7 @@ k0s can also work with externally managed Etcd cluster. If this is configured, k
...
 | `serviceCIDR`   | Network CIDR to use for cluster VIP services. Defaults to `10.96.0.0/12`.                                                                                                                                                                                                                                                                                                                                                                                                         |
+| `addressFamily` | AddressFamily defines the primary family for the cluster. Valid values are empty, `IPv4`, `IPv6` and `Auto`. If empty or `Auto`, K0s determines it based on `.spec.API.ExternalAddress`, if this isn't present it will use `.spec.API.Address.`. If both addresses are empty or the chosen address is a hostname, defaults to `IPv4`. |
```
twz123: Maybe `primaryAddressFamily`?

pkg/apis/k0s/v1beta1/clusterconfig_types.go
```diff
@@ -427,6 +427,7 @@ func (c *ClusterConfig) GetClusterWideConfig() *ClusterConfig {
 			c.Spec.Network.ServiceCIDR = ""
 			c.Spec.Network.ClusterDomain = ""
 			c.Spec.Network.ControlPlaneLoadBalancing = nil
+			c.Spec.Network.AddressFamily = ""
```
twz123: Okay, so we're excluding it here because we don't support this to be changed dynamically? We're probably excluding the domain and service CIDR for the same reason. We should do better in config v2. From my PoV, those settings are really cluster wide settings. Maybe we can't change them after the cluster has been created, but we could at least reflect their values in the cluster.

twz123: We probably want to add this to the list in the method comment.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,6 +61,15 @@ type Network struct {
...
 	// +kubebuilder:default="cluster.local"
 	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// AddressFamily defines the primary family for the cluster.
+	// If empty or `Auto`, K0s determines it based on `.spec.API.ExternalAddress`,
+	// if this isn't present it will use `.spec.API.Address.`.
+	// If both addresses are empty or the chosen address is a hostname, defaults to `IPv4`.
+	// +Kubebuilder:default="Auto"
+	// +Kubebuilder:validation:Enum=Auto;IPv4;IPv6
+	// +optional
```
twz123:
```suggestion
```

IIRC this is not required, as we have the omitempty JSON tag, which automaticall makes this an optional field.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,6 +61,15 @@ type Network struct {
...
+	AddressFamily string `json:"addressFamily,omitempty"`
```
twz123: Can we add constants for those, and maybe its own type? I see we have different upper/lowercase variants for this in this PR already.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,6 +61,15 @@ type Network struct {
 	// Cluster Domain
 	// +kubebuilder:default="cluster.local"
 	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// AddressFamily defines the primary family for the cluster.
+	// If empty or `Auto`, K0s determines it based on `.spec.API.ExternalAddress`,
```
twz123:
```suggestion
	// If empty or `Auto`, k0s determines it based on `.spec.API.ExternalAddress`,
```

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,6 +61,15 @@ type Network struct {
...
+	// if this isn't present it will use `.spec.API.Address.`.
+	// If both addresses are empty or the chosen address is a hostname, defaults to `IPv4`.
+	// +Kubebuilder:default="Auto"
```
twz123: I wonder if we need to add this as its own enum variant. Maybe it's easier to say "autodetect if unspecified". WDYT?

Background: We basically have two ways to express the same thing. The "Auto" variant would especially make sense if we'd add the address family as a as required field, but that seems to be impossible for backwards compatibility.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -119,6 +128,11 @@ func (n *Network) Validate() []error {
 		}
 	}
 
+	if !slices.Contains([]string{"", "Auto", "IPv4", "IPv6"}, n.AddressFamily) {
+		errors = append(errors, field.Invalid(field.NewPath("addressFamily"), n.AddressFamily,
+			fmt.Sprintf("invalid address family `%s`. Valid values are: empty, `Auto`, `IPv4` and `IPv6`", n.AddressFamily)))
+	}
+
```
twz123: I'd separate the empty case, as this has a separate meaning, in a way. Then there's a special "NotSupported" error type for enumerated values:

```suggestion
	if n.AddressFamily != "" {
		if allowed := []string{"Auto", "IPv4", "IPv6"}; !slices.Contains(allowed, n.AddressFamily) {
			err := field.NotSupported(field.NewPath("addressFamily"), n.AddressFamily, allowed)
			errors = append(errors, err)
		}
	}
```

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -216,18 +230,39 @@ func (n *Network) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// PrimaryAddressFamilyIPv4 returns true if the primary address family is IPv4.
+// If the address family is not defined, it will be determined based on the provided address.
+// If the address is not provided, it will assume it's IPv4
+func (n *Network) PrimaryAddressFamilyIPv4(addr string) bool {
```
twz123: Feels odd to have this on the network struct. Did you consider to have the second part on the API struct? Then one could use it like so (given that we don't support the "Auto" variant):

```go
addressFamily := spec.Network.AddressFamily
if addressFamily == "" {
	addressFamily = spec.API.DetectPrimaryAddressFamily()
}
```

or even as a oneliner, if one wouldn't mind to execute the family detection if it might be unnecessary:

```go
addressFamily := cmp.Or(spec.Network.AddressFamily, spec.API.DetectPrimaryAddressFamily())
```

If we'd have the `AddressFamily` as its own type, we could write the above like that, no matter if we support "Auto" or not:

```go
addressFamily := spec.Network.AddressFamily
if !addressFamily.IsSpecified() { // this would check if it's "" or maybe "Auto"
	addressFamily = spec.API.DetectPrimaryAddressFamily()
}
```

Or if we define a constant like this: `const AddressFamilyAuto AddressFamily = ""`, then we could write the above like that:

```go
addressFamily := spec.Network.AddressFamily
if addressFamily == v1beta1.AddressFamilyAuto {
	addressFamily = spec.API.DetectPrimaryAddressFamily()
}
```

Lastly, we could wrap up the above code into a method on the ClusterSpec itself, as this is the place where we have both the network info and the API info.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -216,18 +230,39 @@ func (n *Network) UnmarshalJSON(data []byte) error {
...
+		// cache the result for following executions
+		n.AddressFamily = "IPv6"
+		return false
+	}
+	n.AddressFamily = "IPv4"
+	return true
+}
+
 // BuildServiceCIDR returns actual argument value for service cidr
 func (n *Network) BuildServiceCIDR(addr string) string {
```
twz123: Maybe we could change the parameter to be an address family? Also, maybe we should move this method up one level to the ClusterSpec, as we have both the API and Network information? Related to my longer comment about the PrimaryAddressFamilyIPv4 method.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -216,18 +230,39 @@ func (n *Network) UnmarshalJSON(data []byte) error {
...
+	// Execute anyway so that we have the correct value in the struct
 	if !n.DualStack.Enabled {
 		return n.ServiceCIDR
 	}
-	// Because Kubernetes relies on the order of the given CIDRs in dual-stack
-	// mode, the CIDR whose version matches the version of the IP address the
-	// API server is listening on must be specified first.
```
twz123: I think that comment would still be helpful, WDYT?

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -216,18 +230,39 @@ func (n *Network) UnmarshalJSON(data []byte) error {
...
+		n.AddressFamily = "IPv6"
+		return false
+	}
+	n.AddressFamily = "IPv4"
+	return true
+}
+
 // BuildServiceCIDR returns actual argument value for service cidr
 func (n *Network) BuildServiceCIDR(addr string) string {
+	// Execute anyway so that we have the correct value in the struct
```
twz123: Not sure what's meant by this.

pkg/component/controller/apiendpointreconciler_test.go
```diff
@@ -183,12 +202,16 @@ func verifyEndpointAddresses(t *testing.T, expectedAddresses []string, clients k
...
-		{IP: net.ParseIP("240.0.0.2")},
-		{IP: net.ParseIP("240.0.0.3")},
+func (fr fakeResolver) LookupIP(ctx context.Context, afnet string, host string) ([]net.IP, error) {
+	if afnet == "ip4" {
+		return []net.IP{
+			net.ParseIP("240.0.0.2"),
+			net.ParseIP("240.0.0.3"),
+		}, nil
+	}
+	return []net.IP{
```
twz123: Maybe we should check for afnet == "ipv6" here?

pkg/component/controller/apiendpointreconciler.go
```diff
@@ -111,15 +113,15 @@ func (a *APIEndpointReconciler) reconcileEndpoints(ctx context.Context) error {
 
 	externalAddress := a.externalAddress
 
-	ipAddrs, err := a.resolver.LookupIPAddr(ctx, externalAddress)
+	ipAddrs, err := a.resolver.LookupIP(ctx, a.afnet, externalAddress)
```
twz123, line 124: In the light of this: Maybe we should use the address family type for the afnet field here. And maybe we should call the empty address family "Unspecified".

We could either add a switch here to get the right network value for the lookup call right here:

```go
network := "ip"
switch n.AddressFamily {
case AddressFamilyIPv4:
	network = "ipv4"
case AddressFamilyIPv6:
	network = "ipv6"
}
```

or add the above as a method to the AddressFamily type:

```go
ipAddrs, err := a.resolver.LookupIP(ctx, a.addressFamily.ipNetwork(), externalAddress)
```

pkg/apis/k0s/v1beta1/network.go
twz123: After the full review, I think we should ditch the "Auto" value and simply rely on the field being empty or not. For the empty value, we can add a constant named "Unspecified" in the code.

twz123 (commented):
Good stuff. I have some discussion points around the address family field. IMO it makes sense to have it as its own type. During review, I came up with this as a starting point:

```go
// +kubebuilder:validation:Enum=IPv4;IPv6
type IPAddressFamily string

const (
	IPAddressFamilyUnspecified IPAddressFamily = ""
	IPv4AddressFamily                          = "IPv4"
	IPv6AddressFamily                          = "IPv6"
)

func (af IPAddressFamily) ipNetwork() string {
	switch af {
	case IPv4AddressFamily:
		return "ipv4"
	case IPv6AddressFamily:
		return "ipv6"
	default:
		return "ip"
	}
}
```

Anyhow, this PR is good as is already.

twz123 (approved):
(Bummer! Wanted to approve it in the previous review already :upside_down_face:)

pkg/apis/k0s/v1beta1/network.go
twz123: Nevermind about the different casings. Those were related, but used for different things.

docs/configuration.md
juanluisvaladas: I was going to call it that initally but I felt it was a bit too long. I've never been very convinced about this name so I'm changing it.

pkg/apis/k0s/v1beta1/clusterconfig_types.go
juanluisvaladas: In config v2 we should probably have a bunch of well defined immutable fields, anything messing with the ServiceCIDR should probably immutable for the lifetime of the cluster

pkg/apis/k0s/v1beta1/network.go
juanluisvaladas: I checked and the CRD generation didn't change so you must be correct.

We have 7 instances of having the +optional with omitempty. I'll make a PR later to clean it up a bit as it makes me happy to make PRs without additions :)

pkg/apis/k0s/v1beta1/network.go
juanluisvaladas: Good one!

pkg/apis/k0s/v1beta1/network.go
juanluisvaladas: Brought it back

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,8 +61,23 @@ type Network struct {
 	// Cluster Domain
 	// +kubebuilder:default="cluster.local"
 	ClusterDomain string `json:"clusterDomain,omitempty"`
+
+	// PrimaryAddressFamily defines the primary family for the cluster.
+	// If empty, k0s determines it based on `.spec.API.ExternalAddress`,
+	// if this isn't present it will use `.spec.API.Address.`.
+	// If both addresses are empty or the chosen address is a hostname, defaults to `IPv4`.
+	// +Kubebuilder:validation:Enum=IPv4;IPv6
```
twz123, line 69: We could move this annotation to the type itself, so all usages will get this automatically.  Also wondering if we should use a lowercase k.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -61,8 +61,23 @@ type Network struct {
...
+	PrimaryAddressFamily PrimaryAddressFamilyType `json:"primaryAddressFamily,omitempty"`
 }
 
+type PrimaryAddressFamilyType string
```
twz123, line 73: FamilyType feels a bit like saying the same thing twice? I'd rather call this `PrimaryAddressType` or `PrimaryAddressFamily`?

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -217,17 +239,22 @@ func (n *Network) UnmarshalJSON(data []byte) error {
...
 	// mode, the CIDR whose version matches the version of the IP address the
 	// API server is listening on must be specified first.
-	if ip := net.ParseIP(addr); ip != nil && ip.To4() == nil {
+	switch primaryAddressFamily {
+	case PrimaryFamilyIPv4:
+		return n.ServiceCIDR + "," + n.DualStack.IPv6ServiceCIDR
+	case PrimaryFamilyIPv6:
 		return n.DualStack.IPv6ServiceCIDR + "," + n.ServiceCIDR
+	default:
+		panic(fmt.Sprintf("BuildServiceCIDR called invalid PrimaryAddressFamily %q family. This is theoretically impossible", primaryAddressFamily))
```
twz123, line 256: We could also just default to IPv4 in this case.

pkg/apis/k0s/v1beta1/network_test.go
```diff
@@ -367,6 +367,20 @@ func (s *NetworkSuite) TestValidation() {
...
 	})
+
+	s.Run("invalid_address_family", func() {
+		n := DefaultNetwork()
+		for _, af := range []PrimaryAddressFamilyType{PrimaryFamilyUnknown, PrimaryFamilyIPv4, PrimaryFamilyIPv6} {
+			n.PrimaryAddressFamily = af
+			errors := n.Validate()
+			s.Nil(errors)
+		}
+		n.PrimaryAddressFamily = PrimaryAddressFamilyType("IPv5")
```
twz123, line 378: haha :D I know there was a gap between four and six :rofl:

pkg/component/controller/apiendpointreconciler.go
```diff
@@ -52,12 +52,21 @@ type APIEndpointReconciler struct {
...
 	resolver          resolver
+	afnet             string
 
 	stopCh chan struct{}
 }
 
 // NewEndpointReconciler creates new endpoint reconciler
-func NewEndpointReconciler(nodeConfig *v1beta1.ClusterConfig, leaderElector leaderelector.Interface, kubeClientFactory kubeutil.ClientFactoryInterface, resolver resolver) *APIEndpointReconciler {
+func NewEndpointReconciler(nodeConfig *v1beta1.ClusterConfig, leaderElector leaderelector.Interface, kubeClientFactory kubeutil.ClientFactoryInterface, resolver resolver, primaryAddressFamily v1beta1.PrimaryAddressFamilyType) *APIEndpointReconciler {
+	var afnet string
```
twz123, line 62: no default to "ip"?

pkg/component/controller/apiendpointreconciler_test.go
```diff
@@ -18,12 +18,13 @@ package controller
...
 import (
 	"context"
+	"fmt"
 	"net"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
```
twz123, line 27: We call this apierrors in other places of the code. Basically extending to the next path element on the left.

PR #5263, opened 2024-11-15
Overhaul container entrypoint script and Docker docs

twz123:
## Description

Add some extra smarts to the k0s container entrypoint script. By adding cgroup setup capabilities, it's no longer necessary to use the host's cgroups when running k0s in a container. Track down and document the necessary security-related container flags, so there's a slightly more audited alternative to just using `--privileged`. Re-add the `/run` directory as tmpfs, as the data in it is not meant to be persistent and could potentially cause problems when containers are restarted.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist:

- [x] My code follows the style [guidelines](https://github.com/k0sproject/k0s/blob/main/docs/contributors/overview.md) of this project 
- [x] My commit messages are [signed-off](https://github.com/k0sproject/k0s/blob/main/docs/contributors/github_workflow.md)
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #5958, opened 2025-06-11
Remount sysctl paths in Docker entrypoint

twz123:
## Description

Kubelet attempts to enforce certain sysctl tunables in /proc/sys. If the host configuration differs from what kubelet expects, errors may occur, such as "Failed to start ContainerManager: open /proc/sys/kernel/panic: read-only file system" or similar. Mitigate this by remounting /proc/sys in read-write mode if necessary.

The tunables only need to be set once and will stay in place until the next time the host reboots. If the values match the kubelet's expectations, the containers could get away with a read-only /proc/sys. This is inconvenient. Therefore, perform an automatic remount. Users may opt out of this via an environment variable.

Reference: https://github.com/kubernetes/kubernetes/blob/v1.33.1/pkg/kubelet/cm/container_manager_linux.go#L412

Note that if kubelet is running in a user namespace and the KubeletInUserNamespace alpha feature gate is enabled, kubelet will ignore any errors when setting tunables.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,6 +20,11 @@ Environment variables:
   K0S_CONFIG
     Optional configuration for k0s, written to /etc/k0s/config.yaml if set.
 
+  K0S_ENTRYPOINT_REMOUNT_SYSCTLS
+    Set to 1 to force remounting of /proc/sys in read-write mode.
+    Set to 0 to disable remounting.
+    The default is automatic detection.
+
   K0S_ENTRYPOINT_REMOUNT_CGROUP2FS
     Set to 1 to force remounting of the cgroup2 filesystem in read-write mode.
     Set to 0 to disable remounting.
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -98,6 +96,44 @@ is_cgroupfs_rw() {
...
+        echo "$0: won't remount /proc/sys without CAP_SYS_ADMIN (disable with K0S_ENTRYPOINT_REMOUNT_SYSCTLS=0)" >&2
+        return
+      fi
+
+      echo "$0: failed to determine capabilities (disable with K0S_ENTRYPOINT_REMOUNT_SYSCTLS=0)" >&2
+    }
+    ;;
+  esac
+
+  mount --make-rslave / # don't propagate mount events to other namespaces
```
ncopa, line 119: Why do we need this? does docker mount / as shared?

ncopa: And do we need `mount --make-rslave /` twice? we do it again in `remount_cgroup2fs`

twz123: This has been copied over from how we [remount cgroup2fs]. IIRC I copied the cgroup2fs remount over from some kernel manpage example, but, of course, I fail to find the original source. The cgroup2fs comment says "don't propagate mount events to other namespaces".

[remount cgroup2fs]: https://github.com/k0sproject/k0s/pull/5263/files#diff-79738685a656fe6b25061bb14181442210b599f746faeaba408a2401de45038aR121

twz123: @ncopa any opinions about this?

Issue #5875, opened 2025-05-23
Single stack IPv6 Support


juanluisvaladas:
Currently k0s doesn't support IPv6 single stack, IPv6 only works on dualStack installations.
This was requested in the past and there are few issues and pull requests related to this such as https://github.com/k0sproject/k0s/issues/1726

Known issues to be solved:
* [ ] Have a proper single stack test (might need to be airgapped due to github limitations)
* [ ] CoreDNS is known to have some issues with its configuration
* [ ] Sane defaults (including but not limited to https://github.com/k0sproject/k0s/issues/5844)
* [ ] Configuration validation (including but not limited to https://github.com/k0sproject/k0s/pull/5840)
* [ ] Documentation

Known unknowns:
* [ ] Calico
* [ ] Kube router
* [ ] Kube proxy
* [ ] CPLB
    * [ ] IPVS
    * [ ] Userspace
* [ ] NLLB

juanluisvaladas:
I managed to run locally the network conformance tests on calico IPv6 single stack. Didn't manage so with kube-router.

I'm just going to leave here some quick notes:
1- The image list is acquired programatically. The source of truth is here: https://github.com/kubernetes/kubernetes/blob/v1.33.3/test/utils/image/manifest.go#L70
2- I need to double check but I believe sonobuoy doesn't have anything like an image list command
3-If we can't use anything off the shelf from sonobuoy, the ideal thing to do is to implement it on sonobuoy, the fast thing to do is to do something like we do with `hack/gen-bindata/gen_bindata.go`

twz123:
Currently, the only pending PRs are:

* #5840
* #6262

Let's not make those release blockers and move this to 1.35. Everything else has already landed and will be part of k0s 1.34.

juanluisvaladas:
Closing as https://github.com/k0sproject/k0s/pull/6263 got merged

PR #6010, opened 2025-06-19
Fix CPLB userspace on ipv6 only clusters by using address labels

juanluisvaladas:
## Description

Fix IPv6 for CPLB with reverse userspace proxy. Changes include

1- Fix routing priority using labels
2- Fix the iptables rule creation
3- Fix the DialProxy route creation for IPv6
4- Introduce tests

Part of #5875

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -281,7 +281,6 @@ func (c *command) start(ctx context.Context, flags *config.ControllerOptions, de
 			LogConfig:       debug,
 			KubeConfigPath:  c.K0sVars.AdminKubeConfigPath,
 			APIPort:         nodeConfig.Spec.API.Port,
-			PrimaryAFIPv4:   nodeConfig.PrimaryAddressFamily() == v1beta1.PrimaryFamilyIPv4,
 		})
 	}
 
--- a/inttest/common/bootloosesuite.go
+++ b/inttest/common/bootloosesuite.go
@@ -1392,16 +1392,6 @@ func (s *BootlooseSuite) GetIPAddress(nodeName string) string {
 	return ipAddress
 }
 
-func (s *BootlooseSuite) GetIPv6Address(nodeName string) string {
-	ssh, err := s.SSH(s.Context(), nodeName)
-	s.Require().NoError(err)
-	defer ssh.Disconnect()
-
-	ipAddress, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 | awk '{print $4}' | grep -v '^fe80' | cut -d/ -f1")
-	s.Require().NoError(err)
-	return ipAddress
-}
-
 func (s *BootlooseSuite) GetRegistryHostPort() int {
 	machine, err := s.MachineForName(s.RegistryNode())
 	s.Require().NoError(err)
--- a/inttest/cplb-userspace/cplbuserspace_test.go
+++ b/inttest/cplb-userspace/cplbuserspace_test.go
@@ -15,11 +15,13 @@
 package keepalived
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"html/template"
 	"net"
 	"net/http"
 	"net/url"
@@ -58,11 +60,12 @@ spec:
 const extAddrControllerConfig = `
 spec:
   images:
-    # Set to "IfNotPresent" to allow using the airgap bundle on ipv6
-    # environments and to pull on ipv4 ones.
+    # GitHub Actions don't support IPv6, for this reason in the test IPv6-only
+    # we air gap the cluster but in IPv4 we don't. Setting the default pull
+    # policy to "IfNotPresent" works on both environments.
     default_pull_policy: IfNotPresent
   api:
-    externalAddress: %s
+    externalAddress: {{ .ExtAddr }}
   network:
     provider: calico
     controlPlaneLoadBalancing:
@@ -70,46 +73,58 @@ spec:
       type: Keepalived
       keepalived:
         vrrpInstances:
-        - virtualIPs: ["%s"]
+        - virtualIPs: ["{{ .VIP }}"]
           authPass: "123456"
-`
-
-const ipv6CIDRs = `
+{{ if .IsIPv6Only }}
     podCIDR: fd00::/108
     serviceCIDR: fd01::/108
+{{ end }}
 `
+
+func (s *CPLBUserSpaceSuite) getK0sCfg(useExtAddr bool, lb string) string {
+
+	if !useExtAddr {
+		return fmt.Sprintf(nllbControllerConfig, s.lbCIDR(lb))
+	}
+
+	k0sCfg := bytes.NewBuffer([]byte{})
+	data := map[string]interface{}{
+		"ExtAddr":    lb,
+		"VIP":        s.lbCIDR(lb),
+		"IsIPv6Only": s.isIPv6Only,
+	}
+	s.Require().NoError(template.Must(template.New("k0s.yaml").
+		Parse(extAddrControllerConfig)).
+		Execute(k0sCfg, data), "can't execute k0s.yaml template")
+	return k0sCfg.String()
+}
 
 // SetupTest prepares the controller and filesystem, getting it into a consistent
 // state which we can run tests against.
 func (s *CPLBUserSpaceSuite) TestK0sGetsUp() {
 	useExtAddr := os.Getenv("K0S_USE_EXTERNAL_ADDRESS") == "yes"
-	if useExtAddr {
-		s.T().Log("Using external address")
+	if useExtAddr && s.isIPv6Only {
+		s.T().Log("Using external address in IPv6 mode")
+	} else if useExtAddr {
+		s.T().Log("Using external address in IPv4 mode")
 	} else {
 		s.T().Log("Using CPLB + NLLB")
 	}
-	if s.isIPv6Only {
-		s.T().Log("Running in IPv6-only mode")
-	}
 
 	lb := s.getLBAddress()
+	k0sCfg := s.getK0sCfg(useExtAddr, lb)
+
 	ctx := s.Context()
 	var joinToken string
-
 	for idx := range s.ControllerCount {
 		s.Require().NoError(s.WaitForSSH(s.ControllerNode(idx), 2*time.Minute, 1*time.Second))
-		if useExtAddr {
-			k0sCfg := fmt.Sprintf(extAddrControllerConfig, lb, s.lbCIDR(lb))
-			if s.isIPv6Only {
-				k0sCfg += ipv6CIDRs
-			}
-
-			s.PutFile(s.ControllerNode(idx), "/tmp/k0s.yaml", k0sCfg)
+		s.PutFile(s.ControllerNode(idx), "/tmp/k0s.yaml", k0sCfg)
+		if s.isIPv6Only {
 			// Note that the token is intentionally empty for the first controller
-			// Disable coreDNS to prevent crashloop backoff due to IPv6 and loop
+			// Disable coreDNS to prevent crashloop backoff in github actions. This happens because
+			// there is no IPv6 connectivity to the outside world in the CI environment and sets the DNS to ::1.
 			s.Require().NoError(s.InitController(idx, "--config=/tmp/k0s.yaml", "--disable-components=endpoint-reconciler,coredns", "--enable-worker", joinToken))
 		} else {
-			s.PutFile(s.ControllerNode(idx), "/tmp/k0s.yaml", fmt.Sprintf(nllbControllerConfig, s.lbCIDR(lb)))
 			// Note that the token is intentionally empty for the first controller
 			s.Require().NoError(s.InitController(idx, "--config=/tmp/k0s.yaml", "--enable-worker", joinToken))
 		}
@@ -193,7 +208,7 @@ func (s *CPLBUserSpaceSuite) TestK0sGetsUp() {
 func (s *CPLBUserSpaceSuite) getLBAddress() string {
 	var ip string
 	if s.isIPv6Only {
-		ip = s.GetIPv6Address(s.ControllerNode(0))
+		ip = s.firstPublicIPv6Address(s.ControllerNode(0))
 	} else {
 		ip = s.GetIPAddress(s.ControllerNode(0))
 	}
@@ -284,6 +299,29 @@ func getServerCertSignature(ctx context.Context, url string) (string, error) {
 
 	// Return the signature as a hex string
 	return hex.EncodeToString(signature), nil
+}
+
+func (s *CPLBUserSpaceSuite) firstPublicIPv6Address(nodeName string) string {
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 scope global")
+	s.Require().NoError(err)
+
+	// Parse the output line by line
+	for line := range strings.SplitSeq(strings.TrimSpace(output), "\n") {
+		fields := strings.Fields(line)
+		s.Require().GreaterOrEqual(len(fields), 4, "Expected at least 4 fields in the output line")
+
+		ip, _, err := net.ParseCIDR(fields[3])
+		s.Require().NoError(err, "Failed to parse IP address from output line")
+
+		return ip.String()
+	}
+
+	s.Require().Fail("No IPv6 address found on eth0")
+	return ""
 }
 
 // TestKeepAlivedSuite runs the keepalived test suite. It verifies that the
--- a/inttest/ipv6/ipv6_test.go
+++ b/inttest/ipv6/ipv6_test.go
@@ -18,8 +18,10 @@ package ipv6
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 
 	"github.com/k0sproject/k0s/inttest/common"
@@ -70,11 +72,11 @@ func (s *IPv6Suite) TestK0sGetsUp() {
 
 	if os.Getenv("K0S_NETWORK") == "kube-router" {
 		s.T().Log("Using kube-router network")
-		k0sConfig = fmt.Sprintf(k0sConfigWithKuberouter, s.GetIPv6Address(s.ControllerNode(0)))
+		k0sConfig = fmt.Sprintf(k0sConfigWithKuberouter, s.firstPublicIPv6Address(s.ControllerNode(0)))
 		cniDS = "kube-router"
 	} else {
 		s.T().Log("Using calico network")
-		k0sConfig = fmt.Sprintf(k0sConfigWithCalico, s.GetIPv6Address(s.ControllerNode(0)))
+		k0sConfig = fmt.Sprintf(k0sConfigWithCalico, s.firstPublicIPv6Address(s.ControllerNode(0)))
 		cniDS = "calico-node"
 	}
 
@@ -122,6 +124,29 @@ func (s *IPv6Suite) validateDockerBridge() {
 	output, err := cmd.CombinedOutput()
 	s.Require().NoError(err)
 	s.Require().Containsf(string(output), "false,true", "expected bridge-ipv6 network to have IPv6 enabled")
+}
+
+func (s *IPv6Suite) firstPublicIPv6Address(nodeName string) string {
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 scope global")
+	s.Require().NoError(err)
+
+	// Parse the output line by line
+	for line := range strings.SplitSeq(strings.TrimSpace(output), "\n") {
+		fields := strings.Fields(line)
+		s.Require().GreaterOrEqual(len(fields), 4, "Expected at least 4 fields in the output line")
+
+		ip, _, err := net.ParseCIDR(fields[3])
+		s.Require().NoError(err, "Failed to parse IP address from output line")
+
+		return ip.String()
+	}
+
+	s.Require().Fail("No IPv6 address found on eth0")
+	return ""
 }
 
 func TestAirgapIPv6Suite(t *testing.T) {
--- a/pkg/component/controller/cplb/cplb_label_linux.go
+++ b/pkg/component/controller/cplb/cplb_label_linux.go
@@ -25,34 +25,9 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// maybeSetAddrLabels sets the labels for the vips to vrrp.AddressLabel
-// so that the real IP is preferred.
-func (k *Keepalived) maybeSetAddrLabels() error {
-	for _, vrrp := range k.Config.VRRPInstances {
-		for _, vip := range vrrp.VirtualIPs {
-			// Only set labels for IPv6 addresses
-			ipAddr, _, err := net.ParseCIDR(vip)
-			if err != nil {
-				return fmt.Errorf("failed to parse CIDR %s: %w", vip, err)
-			}
-
-			// Only set labels for IPv6 addresses
-			if ipAddr.To4() != nil {
-				continue
-			}
-
-			// Set address label for IPv6 VIP
-			if err := setAddressLabel(ipAddr, 128, vrrp.AddressLabel); err != nil {
-				return fmt.Errorf("failed to set address label for %s: %w", ipAddr, err)
-			}
-		}
-	}
-	return nil
-}
-
-func setAddressLabel(ip net.IP, bits int, label uint32) error {
+func setAddressLabel(ip net.IP, label uint32) error {
 	// Use iproute2 as reference for setting the address label.
-	// https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2/+/refs/tags/v6.15.0/ip/ipaddrlabel.c#176
+	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n178
 	if ip.To4() != nil {
 		return fmt.Errorf("cannot set address label for IPv4 address %s", ip)
 	}
@@ -61,13 +36,13 @@ func setAddressLabel(ip net.IP, bits int, label uint32) error {
 
 	msg := &IfAddrLabelmsg{
 		Family:    uint8(unix.AF_INET6),
-		PrefixLen: uint8(bits),
+		PrefixLen: uint8(128), // The netmask for IPv6. In CPLB we always use /128.
 	}
 
 	req.AddData(msg)
 
 	// RFC 3484 and 6724 don't specify label size, but iproute2 uses 32 bit labels:
-	// https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2/+/refs/tags/v6.15.0/ip/ipaddrlabel.c#176
+	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n178
 	labelBytes := make([]byte, 4)
 	nl.NativeEndian().PutUint32(labelBytes, label)
 
@@ -88,12 +63,12 @@ func setAddressLabel(ip net.IP, bits int, label uint32) error {
 // define it here.
 
 type IfAddrLabelmsg struct {
-	Family    uint8
-	_         uint8
-	PrefixLen uint8
-	Flags     uint8
-	Index     uint32
-	Seq       uint32
+	Family    uint8  // Address family
+	_         uint8  // Reserved
+	PrefixLen uint8  // Prefix length
+	Flags     uint8  // Flags
+	Index     uint32 // Link index
+	Seq       uint32 // Sequence number
 }
 
 const (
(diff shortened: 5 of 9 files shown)


inttest/common/bootloosesuite.go
```diff
@@ -1392,6 +1392,16 @@ func (s *BootlooseSuite) GetIPAddress(nodeName string) string {
 	return ipAddress
 }
 
+func (s *BootlooseSuite) GetIPv6Address(nodeName string) string {
```
twz123: I'd be glad if we could avoid adding new methods to `BootlooseSuite`. It's a melting pot for many different topics, and it'd be nice if we could separate concerns a bit. From my PoV, the suite should only manage the test lifecycle. All the other utility methods would be better off living somewhere else.

A long time ago, I started working on this, but, like so many other things, didn't follow through ... :shrug:

inttest/common/bootloosesuite.go
```diff
@@ -1392,6 +1392,16 @@ func (s *BootlooseSuite) GetIPAddress(nodeName string) string {
...
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	ipAddress, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 | awk '{print $4}' | grep -v '^fe80' | cut -d/ -f1")
```
twz123: Did you consider to do the parsing Go? It might be better to not use pipes, as you wouldn't get an error if any of `ip`, `awk` and `grep` would fail. Also, there's methods like `net.IP.IsLinkLocalUnicast` that may be easier to understand as `grep -v '^fe80'`.

inttest/common/bootloosesuite.go
twz123: Maybe return `net.IP` instead of `string`?

inttest/common/bootloosesuite.go
twz123: IIUC `ipAddress` might be empty if no appropriate address has been found. Is it worth checking?

inttest/common/bootloosesuite.go
twz123: Could you add a comment how the IP is chosen? Right now, it disregards link-local addresses. Maybe rename to `FirstPublicIPv6Address` or something along those lines?

inttest/cplb-userspace/cplbuserspace_test.go
```diff
@@ -56,6 +57,10 @@ spec:
 
 const extAddrControllerConfig = `
 spec:
+  images:
+    # Set to "IfNotPresent" to allow using the airgap bundle on ipv6
+    # environments and to pull on ipv4 ones.
+    default_pull_policy: IfNotPresent
```
twz123, line 66: I'm afraid I don't get this comment :sweat_smile: ... How does this relate to airgap?

inttest/cplb-userspace/cplbuserspace_test.go
```diff
@@ -78,18 +88,28 @@ func (s *CPLBUserSpaceSuite) TestK0sGetsUp() {
...
 	ctx := s.Context()
 	var joinToken string
 
 	for idx := range s.ControllerCount {
 		s.Require().NoError(s.WaitForSSH(s.ControllerNode(idx), 2*time.Minute, 1*time.Second))
 		if useExtAddr {
-			s.PutFile(s.ControllerNode(idx), "/tmp/k0s.yaml", fmt.Sprintf(extAddrControllerConfig, lb, lb))
+			k0sCfg := fmt.Sprintf(extAddrControllerConfig, lb, s.lbCIDR(lb))
+			if s.isIPv6Only {
+				k0sCfg += ipv6CIDRs
```
twz123: That's really wonky :see_no_evil:

inttest/cplb-userspace/cplbuserspace_test.go
```diff
@@ -78,18 +88,28 @@ func (s *CPLBUserSpaceSuite) TestK0sGetsUp() {
...
+			}
+
+			s.PutFile(s.ControllerNode(idx), "/tmp/k0s.yaml", k0sCfg)
 			// Note that the token is intentionally empty for the first controller
-			s.Require().NoError(s.InitController(idx, "--config=/tmp/k0s.yaml", "--disable-components=endpoint-reconciler", "--enable-worker", joinToken))
+			// Disable coreDNS to prevent crashloop backoff due to IPv6 and loop
```
twz123: That sounds like something that needs to be fixed?

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+
+			// Only set labels for IPv6 addresses
+			if ipAddr.To4() != nil {
+				continue
+			}
+
+			// Set address label for IPv6 VIP
+			if err := setAddressLabel(ipAddr, 128, vrrp.AddressLabel); err != nil {
+				return fmt.Errorf("failed to set address label for %s: %w", ipAddr, err)
```
twz123: Would it be better to collect the errors and try to set the addresses for all the virtual IPs before returning?

I'm thinking of a scenario where one VIP is "broken" and this "blocks" a successful setup of the other ones, but maybe that's not something that we should be concerned with?

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+func setAddressLabel(ip net.IP, bits int, label uint32) error {
+	// Use iproute2 as reference for setting the address label.
+	// https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2/+/refs/tags/v6.15.0/ip/ipaddrlabel.c#176
```
twz123:
```suggestion
	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v6.15.0#n176
```

:sweat_smile:

twz123: I wonder if we should pick an older tag as reference. This behavior is present since Kernel 2.6, so having a link to 6.15 might give the impression that it's something rather new. Not sure which version to pick, though. Something 4.x maybe?

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	if ip.To4() != nil {
+		return fmt.Errorf("cannot set address label for IPv4 address %s", ip)
+	}
+	req := nl.NewNetlinkRequest(unix.RTM_NEWADDRLABEL,
+		unix.NLM_F_REQUEST|unix.NLM_F_ACK|unix.NLM_F_REPLACE|unix.NLM_F_CREATE)
```
twz123, line 35: Low-level APIs 4tw! You can probably drop `NLM_F_REQUEST` as this gets automatically added by `NewNetlinkRequest`. AFAICT upstream specifies `NLM_F_CREATE|NLM_F_EXCL`, not `NLM_F_ACK|NLM_F_REPLACE|NLM_F_CREATE`. Any reason not to do the same here?

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+
+	msg := &IfAddrLabelmsg{
+		Family:    uint8(unix.AF_INET6),
+		PrefixLen: uint8(bits),
+	}
+
+	req.AddData(msg)
+
+	// RFC 3484 and 6724 don't specify label size, but iproute2 uses 32 bit labels:
+	// https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2/+/refs/tags/v6.15.0/ip/ipaddrlabel.c#176
```
twz123: Repetition from above.

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	_, err := req.Execute(unix.NETLINK_ROUTE, 0)
+	return err
+}
+
+// Everything below this line should be defined in "golang.org/x/sys/unix"
+// and "github.com/vishvananda/netlink/nl" packages.
+// Ideally we'll get it merged into those packages in the future, but for now
+// define it here.
+
+type IfAddrLabelmsg struct {
```
twz123, line 65: Worth adding a reference to the upstream sources: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/linux/if_addrlabel.h

twz123, line 65: I don't think this and all the following names in this file need to be exported.

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (msg *IfAddrLabelmsg) Len() int {
+	return SizeOfIfAddrLabelMsg
+}
+
+func (msg *IfAddrLabelmsg) Serialize() []byte {
+	return (*(*[SizeOfIfAddrLabelMsg]byte)(unsafe.Pointer(msg)))[:]
+}
+
+// https://github.com/torvalds/linux/blob/v6.15/include/uapi/linux/if_addrlabel.h
```
twz123: I'd be adding this directly to the Go struct and the const block. Then we don't need to repeat the sources here:

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/linux/if_addrlabel.h#n16

and

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/linux/if_addrlabel.h#n25

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+// struct ifaddrlblmsg {
+//	__u8		ifal_family;		/* Address family */
```
twz123, line 91: Worth copying over the comments to the Go struct.

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+			// Set address label for IPv6 VIP
+			if err := setAddressLabel(ipAddr, 128, vrrp.AddressLabel); err != nil {
+				return fmt.Errorf("failed to set address label for %s: %w", ipAddr, err)
+			}
+		}
+	}
+	return nil
+}
+
+func setAddressLabel(ip net.IP, bits int, label uint32) error {
```
twz123: Upstream disallows 0 and MAX_UINT32 as a label value. We should probably do the same?

twz123: IIUC `ip` and `bits` form a network, right? So why not use something like this:

```suggestion
func setAddressLabel(prefix net.IPNet, label uint32) error {
```

Granted, `net.IPMask` is a bit clunky to use, but the `bits` parameter puzzled me for quite some time, as its usage in the netfilter request is kinda bogus and made its intent not very obvious. I'd at least change the bits parameter to `uint8`, rename it to `prefixLen` and add a short doc string to the function.

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	// Use iproute2 as reference for setting the address label.
+	// https://kernel.googlesource.com/pub/scm/network/iproute2/iproute2/+/refs/tags/v6.15.0/ip/ipaddrlabel.c#176
+	if ip.To4() != nil {
+		return fmt.Errorf("cannot set address label for IPv4 address %s", ip)
```
twz123, line 32: Just curious: Upstream code seems to support IPv4 just fine, doesn't it?

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	Index     uint32
+	Seq       uint32
+}
+
+const (
+	IFAL_ADDRESS = 1
+	IFAL_LABEL   = 2
+)
+
+const SizeOfIfAddrLabelMsg = 12 // Size of IfAddrLabelMsg in bytes
```
twz123:
```suggestion
const sizeOfIfAddrLabelMsg = int(unsafe.Sizeof(IfAddrLabelmsg{}))
```

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,128 @@
...
+	"net"
+	"unsafe"
+
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+// maybeSetAddrLabels sets the labels for the vips to vrrp.AddressLabel
+// so that the real IP is preferred.
+func (k *Keepalived) maybeSetAddrLabels() error {
```
twz123: What about moving this method into `cplb_linux.go` and rename it to `setVirtualIPAddressLabels`?

pkg/component/controller/cplb/cplb_linux.go
```diff
@@ -398,7 +404,11 @@ func (k *Keepalived) watchReconcilerUpdatesReverseProxy(ctx context.Context) err
 func (k *Keepalived) setProxyRoutes() {
 	routes := []tcpproxy.Route{}
 	for _, addr := range k.reconciler.GetIPs() {
-		routes = append(routes, tcpproxy.To(fmt.Sprintf("%s:%d", addr, k.APIPort)))
+		if k.PrimaryAFIPv4 {
+			routes = append(routes, tcpproxy.To(fmt.Sprintf("%s:%d", addr, k.APIPort)))
+		} else {
+			routes = append(routes, tcpproxy.To(fmt.Sprintf("[%s]:%d", addr, k.APIPort)))
+		}
```
twz123: This might be better handled via `net.JoinHostPort`

pkg/component/controller/cplb/cplb_linux.go
```diff
@@ -426,7 +436,11 @@ func (k *Keepalived) redirectToProxyIPTables(op string) error {
 				k.log.Infof("Deleting iptables rule to redirect %s", vip)
 			}
 
-			cmd := exec.Command(filepath.Join(k.K0sVars.BinDir, "iptables"), cmdArgs...)
+			iptablesBin := "iptables"
+			if !k.PrimaryAFIPv4 {
```
twz123: Wouldn't it make more sense to check if `vip` is an IPv4 or an IPv6 address?

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: You could if you converted the IPv4 to IPv6, but I'm not sure how that works.

If you read [RFC 6724 section 2.1](https://datatracker.ietf.org/doc/html/rfc6724#section-2.1) you'll see a reference to `::ffff:0:0/96` which are remapped versions of IPv4. Then in [section 10.3](https://datatracker.ietf.org/doc/html/rfc6724#section-10.3) it tells you can use this to prefer IPv4 over IPv6. I would expect it tothe preference of the IPv4 routing policy, but I haven't tested and I'm not too sure.

Regarding sending an IPv4, the kernel [will reject it if the family isn't IPv6](https://github.com/torvalds/linux/blob/78f4e737a53e1163ded2687a922fce138aee73f5/net/ipv6/addrlabel.c#L402-L404) and [ip-addrlabel(8)](https://man7.org/linux/man-pages/man8/ip-addrlabel.8.html) says it's IPv6 only.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: > I added NLM_F_REQUEST because I thought it was a bit easier to read, but I can take that away.

> AFAICT upstream specifies NLM_F_CREATE|NLM_F_EXCL, not NLM_F_ACK|NLM_F_REPLACE|NLM_F_CREATE. Any reason not to do the same here?

`NLM_F_EXCL`will throw an error if the entry exists, `NLM_F_REPLACE`will replace it. This saves us a lot of logic for label changes, if a user has `::1/128`and label `10000` but then sets the label to `20000`and restarts this will handle all for us.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: We can do that, I did it this way because the code of [sys/unix](https://github.com/golang/sys/blob/master/unix/ztypes_linux.go#L688C1-L688C26) just uses a constant value. But again, not a big deal if you want I can change that.

Anyway I'm working on adding it to upstream go once I figure out the sys/unix code generation.

inttest/cplb-userspace/cplbuserspace_test.go
juanluisvaladas: This is due to github actions not supporting IPv6. Because this is an IPv6 only network, the DNS points to `::1`.

In the [IPv6 test](https://github.com/k0sproject/k0s/blob/main/inttest/ipv6/ipv6_test.go#L32-L35) we write a new `/etc/resolv.conf` which prevents it from breaking in github actions and that also works if you have working IPv6.

I could replace resolv.conf, but since we don't need coreDNS on this test I decided to just disable it.

inttest/common/bootloosesuite.go
juanluisvaladas: We use this function in `inttest/ipv6`, `inttest/cplb-userspace` and I also intend to use it in both `inttest/cplb-ipvs` and `inttest/nllb` so probably it should be in some package.

Anyway I'll duplicate it while I think of a better solution.

inttest/cplb-userspace/cplbuserspace_test.go
juanluisvaladas: The IPv6 tests are IPv6 only, which means no internet access in github actions. When setting up the suite we do:

```go
	if cplbSuite.isIPv6Only {
		cplbSuite.Networks = []string{"bridge-ipv6"}
		cplbSuite.AirgapImageBundleMountPoints = []string{"/var/lib/k0s/images/bundle-ipv6.tar"}
	}
```

So IPv4 clusters have internet access but no airgap bundle, and IPv6 clusters have airgap bundle but no internet. This works with both and is static.

inttest/cplb-userspace/cplbuserspace_test.go
juanluisvaladas: It is, I converted it into a template. I thought of doing it a template before but I was lazy because this is a test.

inttest/common/bootloosesuite.go
juanluisvaladas: We get it from the IP command, which gives a string and we want to consume it as a string. I may be missing something but I don't see a good reason to do that...

inttest/common/bootloosesuite.go
juanluisvaladas: My thought was precisely that by having it less controlled if it had anything unexpected it would break the test which is probably something that we want. But anyway I did a more robust function.

juanluisvaladas: Added this too.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: That's a better name, done.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: I think that scenario cannot happen because we already verify it here:
https://github.com/k0sproject/k0s/blob/v1.33.2%2Bk0s.0/pkg/apis/k0s/v1beta1/cplb.go#L170-L175

And CPLB cannot be configured with dynamic config.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: Good catch! the MAX_UINT32 is a bug

As for value 0 we currently use it to set it to the default value of 10000.

Regarding in this function, since we validated it previously in `Keepalived.Validate()` and the kernel will return EINVAL if we pass an invalid value, I don't think we really need to care.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: Done

juanluisvaladas: I set it to 4.0.0

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: They don't, I did this in order to be consistent with `sys/unix` and because my intention is to get this merged in `sys/unix`and eventually get rid of this code. But if you want we can change it, it's not a big deal anyway.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: Did that, but I added `?h=v4.0` pointing to the branch 4.0 so that we're a bit more consistent with the other comments.

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: Done

pkg/component/controller/cplb/cplb_linux.go
juanluisvaladas: Not only it's cleaner, this also solves a bug I hadn't considered in dual stack clusters.

pkg/component/controller/cplb/cplb_linux.go
juanluisvaladas: Indeed, you found a bug in dual stack clusters

pkg/component/controller/cplb/cplb_label_linux.go
juanluisvaladas: Since we have it hardcoded to 128 bits in the function call I decided to remove the parameter.

I can't think of a scenario in the future where we want to add a different prefix length, and if that happens we can think about that in the future.

juanluisvaladas:
> A better PR title might be "Address label support for CPLB IPv6 virtual IPs".

I don't really like that title because we're fixing more things here, not just adding address labels.

juanluisvaladas:
I believed all the concerns should be addressed now, I added a new commit for easier review, so that you don't need to diff the whole thing.

juanluisvaladas:
For reference: https://github.com/golang/sys/pull/252

inttest/cplb-userspace/cplbuserspace_test.go
```diff
@@ -258,13 +300,48 @@ func getServerCertSignature(ctx context.Context, url string) (string, error) {
 	return hex.EncodeToString(signature), nil
 }
 
+func (s *CPLBUserSpaceSuite) firstPublicIPv6Address(nodeName string) string {
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0")
```
ncopa:
```suggestion
	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 scope global")
```
We can let `ip` filter out the link local by using `scope global`.

We could also output it as json by replacing `-oneline` with `-json`, but It looks simpler to parse it as text.

inttest/cplb-userspace/cplbuserspace_test.go
```diff
@@ -258,13 +300,48 @@ func getServerCertSignature(ctx context.Context, url string) (string, error) {
...
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		s.Require().GreaterOrEqual(len(fields), 4, "Expected at least 4 fields in the output line")
+
+		ip, _, err := net.ParseCIDR(fields[3])
+		s.Require().NoError(err, "Failed to parse IP address from output line")
+
+		if ip.IsGlobalUnicast() {
+			return ip.String()
+		}
```
ncopa:
```suggestion
```

We can remove those lines if we use `ip -6 ... scope global` as mentioned above.

cmd/controller/controller.go
```diff
@@ -281,6 +281,7 @@ func (c *command) start(ctx context.Context, flags *config.ControllerOptions, de
 			LogConfig:       debug,
 			KubeConfigPath:  c.K0sVars.AdminKubeConfigPath,
 			APIPort:         nodeConfig.Spec.API.Port,
+			PrimaryAFIPv4:   nodeConfig.PrimaryAddressFamily() == v1beta1.PrimaryFamilyIPv4,
```
ncopa:
```suggestion
```

This is not used anywhere anymore and can be removed.

pkg/component/controller/cplb/cplb_linux.go
```diff
@@ -57,6 +57,7 @@ type Keepalived struct {
 	LogConfig              bool
 	APIPort                int
 	KubeConfigPath         string
+	PrimaryAFIPv4          bool
```
ncopa:
```suggestion
```

This is not used anymore and can be removed.

pkg/component/controller/cplb/cplb_other.go
```diff
@@ -37,6 +37,7 @@ type Keepalived struct {
 	LogConfig       bool
 	APIPort         int
 	KubeConfigPath  string
+	PrimaryAFIPv4   bool
```
ncopa:
```suggestion
```
This can be removed.

inttest/common/bootloosesuite.go
twz123: Just because it is an IP and by returning a real type, we document what that returned string should be, enforce invariants and make it better distinguishable from all the other strings. What about badly behaving scripts that return garbage instead of real IPs, for example?

inttest/cplb-userspace/cplbuserspace_test.go
twz123: > We could also output it as json by replacing -oneline with -json, but It looks simpler to parse it as text.

I think we'd then need to install another `ip` binary. We only have the BusyBox version in place right now.

inttest/cplb-userspace/cplbuserspace_test.go
twz123: I see your point, albeit I found the Go version easier to understand for the uninitiated. I'd need to look up what "scope global" does first :see_no_evil:

inttest/ipv6/ipv6_test.go
```diff
@@ -134,6 +126,32 @@ func (s *IPv6Suite) validateDockerBridge() {
...
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0")
+	s.Require().NoError(err)
+
+	// Parse the output line by line
+	lines := strings.Split(strings.TrimSpace(output), "\n")
+	for _, line := range lines {
```
twz123:
```suggestion
	for line := range strings.SplitSeq(strings.TrimSpace(output), "\n") {
```

pkg/component/controller/cplb/cplb_label_linux.go
twz123: I really think we should use the unsafe.Sizeof version. Otherwise, we should add a unit test that asserts that asserts `sizeOfIfAddrLabelMsg = int(unsafe.Sizeof(IfAddrLabelmsg{}))`. Even if this code seems to be intermediary until upstream changes land, you'll never know how long it will stay, and whenever the struct size changes for whichever reason, you'd be on your own to remember that this constant needs to be changed.

I don't think the constant needs to be public, too.

inttest/cplb-userspace/cplbuserspace_test.go
twz123: Oh, I see. Thanks for changing the comment!

pkg/component/controller/cplb/cplb_label_linux.go
twz123: Alright. Although I guess `setAddressLabel` can fail for many reasons despite `vrrp.AddressLabel` being invalid.

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,104 @@
...
+	"net"
+	"unsafe"
+
+	"github.com/vishvananda/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+func setAddressLabel(ip net.IP, label uint32) error {
+	// Use iproute2 as reference for setting the address label.
+	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n176
```
twz123:
```suggestion
	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n178
```

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,104 @@
...
+
+	msg := &IfAddrLabelmsg{
+		Family:    uint8(unix.AF_INET6),
+		PrefixLen: uint8(128), // The netmask for IPv6. In CPLB we always use /128.
+	}
+
+	req.AddData(msg)
+
+	// RFC 3484 and 6724 don't specify label size, but iproute2 uses 32 bit labels:
+	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n176
```
twz123:
```suggestion
	// https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/ip/ipaddrlabel.c?h=v4.0.0#n178
```

pkg/component/controller/cplb/cplb_label_linux.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (msg *IfAddrLabelmsg) Len() int {
+	return SizeOfIfAddrLabelMsg
+}
+
+func (msg *IfAddrLabelmsg) Serialize() []byte {
+	return (*(*[SizeOfIfAddrLabelMsg]byte)(unsafe.Pointer(msg)))[:]
+}
+
+// https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/linux/if_addrlabel.h?h=v4.0#n16
```
twz123:
```suggestion
// https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/uapi/linux/if_addrlabel.h?h=v4.0#n15
```

twz123: If you include the upstream source reference in the Go struct, I think you can safely delete those comments.

pkg/component/controller/cplb/cplb_label_linux.go
twz123: > the MAX_UINT32 is a bug

What do you mean exactly? A bug in iproute2?

> As for value 0 we currently use it to set it to the default value of 10000.

Yeah, but that doesn't happen in this function.

> since we validated it previously in Keepalived.Validate()

While the `UserSpaceProxyPort` is being validated there, it's not obvious that only this value will end up being used with this function. Defense in depth might catch some bugs here.

> the kernel will return EINVAL if we pass an invalid value, I don't think we really need to care.

Did you check if EINVAL gets returned for 0 and MAX_UINT32? I think it's worth adding a comment why we do it differently from upstream and don't do this validation before the syscall. I have a feeling those checks are there for a reason.

inttest/ipv6/ipv6_test.go
juanluisvaladas: Modified this in both instances of `firstPublicIPv6Address`

juanluisvaladas:
I think everything is addressed now and ready to merge.

ncopa (approved):

PR #6023, opened 2025-06-20
[backport release-1.33] Bump kubernetes to 1.33.2

juanluisvaladas:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.33.md#changelog-since-v1331


juanluisvaladas:
I just noticed we are still hardcoding the version in a few places. Since that is only changing the docs which has a fast CI, I'll fix that on main and backport it on a separate PR.

twz123 (approved):

PR #6053, opened 2025-06-26
[docs] make security docs always point to main

juanluisvaladas:
## Description

This prevents having supported branches showing incorrect documentation such as: https://docs.k0sproject.io/v1.32.6+k0s.0/security/

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -1,1 +1,3 @@
-{% include-markdown "../SECURITY.md" %}
+# Security policy
+
+For the security policy, please refer to the [upstream k0s documentation](https://github.com/k0sproject/k0s/blob/main/SECURITY.md).


docs/security.md
```diff
@@ -1 +1 @@
-{% include-markdown "../SECURITY.md" %}
+{% include-markdown "https://raw.githubusercontent.com/k0sproject/k0s/refs/heads/main/SECURITY.md" %}
```
twz123: This will only reflect the state of the main SECURITY.md at docs build time.

twz123 (approved):
Now I start to understand. You're concerned with the documentation. Maybe it's best not to include SECURITY.md at all in the docs, and just add an explainer here that the git repo's main branch is the only up-to-date source of truth (including a link, of course.)

docs/security.md
juanluisvaladas: Correct, I think it's worth discussing not including security.md at all and having a link. Should we go that way instead?

twz123: Sounds like a plan.

docs/security.md
```diff
@@ -1 +1,3 @@
-{% include-markdown "../SECURITY.md" %}
+# Security policy
+
+For the security policy, please refer to the [upstream k0s documentation](https://github.com/k0sproject/k0s/blob/main/SECURITY.md).
```
twz123, line 3: Hmm, not entirely sure about the wording, as this file is actually part of the k0s documentation.

PR #6054, opened 2025-06-26
Introduce ExecutableSuffix

twz123:
## Description

Eases the handling of the .exe suffix for Windows, removes the need for `runtime.GOOS` conditions.

Rename `assets.Stage` to `assets.StageExecutable`. Makes it obvious that it's intended for executable files, as it adds the executable bit by default. Leave a non-exported stage function in place that accepts the permission bits as a parameter.

Inline `EmbeddedBinaryNeedsUpdate`, because it's only used once in the stage function. Having it as its own freestanding function doesn't improve readability.

Use conditional compilation for containerd component. No more branching on `runtime.GOOS`. Extend the precision for the "running executable" workaround.

Remove `assets.BinPath` and let `assets.StageExecutable` return the path instead. Refactor `StageExecutable` to return the resolved binary path. This allows for the `BinPath` function to be integrated into `StageExecutable`, since its sole purpose was to support PATH lookups of binaries that weren't bundled with k0s. As a consequence, the containerd component had to be refactored to pick up the correct path from the appropriate call to `StageExecutable`.

Write the executables atomically. Use the atomic file writer when extracting embedded assets to the local file system. To support this use case, the file modification time must be settable. Add an appropriate method for that. Add the file path as a structured logging argument. Also change the mode from 0550 to 0750.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/assets/stage.go
+++ b/pkg/assets/stage.go
@@ -14,6 +14,7 @@ import (
 	"syscall"
 
 	"github.com/k0sproject/k0s/internal/pkg/file"
+	"github.com/k0sproject/k0s/pkg/constant"
 
 	"github.com/sirupsen/logrus"
 )
@@ -26,8 +27,9 @@ func StageExecutable(dir, name string) (string, error) {
 	// Always returning the path, even under error conditions, is kind of a hack
 	// to work around the "running executable" problem on Windows.
 
-	path := filepath.Join(dir, name)
-	err := stage(name, path, 0750)
+	executableName := name + constant.ExecutableSuffix
+	path := filepath.Join(dir, executableName)
+	err := stage(executableName, path, 0750)
 	if err == nil {
 		return path, nil
 	}
@@ -50,6 +52,8 @@ func StageExecutable(dir, name string) (string, error) {
 	}
 
 	// If we still haven't found the executable, look for it in the PATH.
+	// Don't pass in the executable suffix here, so that the PathExt environment
+	// variable works as expected on Windows.
 	lookedUpPath, lookErr := exec.LookPath(name)
 	if lookErr == nil {
 		logrus.WithField("path", lookedUpPath).WithError(err).Debug("Executable found in PATH")
--- a/pkg/component/worker/containerd/component.go
+++ b/pkg/component/worker/containerd/component.go
@@ -72,7 +72,7 @@ func (c *Component) Init(ctx context.Context) error {
 	g, _ := errgroup.WithContext(ctx)
 
 	g.Go(func() (err error) {
-		c.executablePath, err = stageExecutable(c.K0sVars.BinDir, executableName)
+		c.executablePath, err = stageExecutable(c.K0sVars.BinDir, "containerd")
 		return err
 	})
 	for _, bin := range additionalExecutableNames {
--- a/pkg/component/worker/containerd/component_other.go
+++ b/pkg/component/worker/containerd/component_other.go
@@ -8,7 +8,6 @@ package containerd
 import "github.com/k0sproject/k0s/pkg/assets"
 
 const (
-	executableName     = "containerd"
 	defaultConfPath    = "/etc/k0s/containerd.toml"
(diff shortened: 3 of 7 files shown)


pkg/assets/stage.go
```diff
@@ -73,65 +76,47 @@ func Stage(dataDir string, name string) error {
...
-		return fmt.Errorf("failed to write to %s: %w", p, err)
-	}
-	return nil
+	log.Debug("Writing static file")
+
+	return file.AtomicWithTarget(p).
+		WithPermissions(0750).
+		// In order to properly determine if an update of an embedded binary
+		// file is needed, the staged embedded binary needs to have the same
+		// modification time as the `k0s` executable.
```
juanluisvaladas, line 122: This is smart, I like it.

pkg/component/worker/containerd/component.go
```diff
@@ -69,26 +64,17 @@ type Component struct {
...
-		c.importsPath = importsPathWindows
-	} else {
-		c.binaries = []string{"containerd", "containerd-shim", "containerd-shim-runc-v1", "containerd-shim-runc-v2", "runc"}
-		c.confPath = confPathPosix
-		c.importsPath = importsPathPosix
+		LogLevel:    logLevel,
+		K0sVars:     vars,
+		Profile:     profile,
+		confPath:    defaultConfPath,
+		importsPath: defaultImportsPath,
```
juanluisvaladas, line 63: I think both the fields confPath and ImportsPath can be removed and they can just be replaced by `defaultConfPath` `defaultImportsPath`

pkg/assets/stage.go
```diff
@@ -18,53 +18,67 @@ package assets
...
+// executable is not embedded in the k0s executable, this function first checks
+// if an executable exists at the desired path. If not, it falls back to a PATH
+// lookup. Returns the path to the executable, even if an error occurs.
+func StageExecutable(dir, name string) (string, error) {
+	// Always returning the path, even under error conditions, is kind of a hack
+	// to work around the "running executable" problem on Windows.
+
+	executableName := name + constant.ExecutableSuffix
+	path := filepath.Join(dir, executableName)
+	err := stage(executableName, path, 0750)
```
juanluisvaladas, line 32: 0750 should probably be a constant in pkg/constant/constant.go.
Anyway that's just a nit.

pkg/component/controller/apiserver.go
```diff
@@ -52,9 +52,9 @@ type APIServer struct {
 	Storage                   manager.Component
 	EnableKonnectivity        bool
 	DisableEndpointReconciler bool
-	gid                       int
```
ncopa, line 42: So we never set the gid? This means that we have always run the processes as gid 'root'?

I think we should set the gid instead?

pkg/component/worker/containerd/component.go
twz123: I had it that way. Then I realized that they were overridden in the unit tests.

pkg/component/controller/apiserver.go
twz123: K0s was never doing this. It only deals with users so far. What group would you pick here?

pkg/component/worker/containerd/component.go
```diff
@@ -97,19 +84,14 @@ var _ manager.Component = (*Component)(nil)
 // Init extracts the needed binaries
 func (c *Component) Init(ctx context.Context) error {
 	g, _ := errgroup.WithContext(ctx)
-	for _, bin := range c.binaries {
+
+	g.Go(func() (err error) {
+		c.executablePath, err = stageExecutable(c.K0sVars.BinDir, "containerd")
+		return err
+	})
+	for _, bin := range additionalExecutableNames {
```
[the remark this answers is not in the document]
twz123: That used to be true for older Go versions. I think this was changed in Go 1.22 already.

juanluisvaladas (approved):

PR #6082, opened 2025-07-03
Refactor CPLB tests and implement CPLB IPVS IPv6 test

juanluisvaladas:
## Description

This PR has 7 commits, but most of them are very short.
The first three commits improve the code quality of CPLB tests, the next 3 commits extract common functions to the `inttest/common` package. The last commit actually implements the test

This is part of #5875

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,7 +250,12 @@ airgap-image-bundle-linux-riscv64.tar: k0s airgap-images.txt
 	./k0s airgap bundle-artifacts -v --platform='$(TARGET_PLATFORM)' -o '$@' <airgap-images.txt
 
 .PHONY: $(smoketests)
-check-airgap check-ap-airgap check-ipv6-calico check-ipv6-kuberouter check-cplb-userspace-ipv6: airgap-image-bundle-linux-$(HOST_ARCH).tar
+check-airgap \
+check-ap-airgap \
+check-ipv6-calico \
+check-ipv6-kuberouter \
+check-cplb-ipvs-ipv6 \
+check-cplb-userspace-ipv6: airgap-image-bundle-linux-$(HOST_ARCH).tar
 $(smoketests): k0s
 	$(MAKE) -C inttest K0S_IMAGES_BUNDLE='$(CURDIR)/airgap-image-bundle-linux-$(HOST_ARCH).tar' $@
 
--- a/inttest/Makefile
+++ b/inttest/Makefile
@@ -87,6 +87,10 @@ check-customports-dynamicconfig: TEST_PACKAGE=customports
 
 check-kubeletcertrotate: TIMEOUT=15m
 
+check-cplb-ipvs-ipv6: export K0S_IPV6_ONLY=yes
+check-cplb-ipvs-ipv6: TEST_PACKAGE=cplb-ipvs
+
+check-cplb-userspace-ipv6: export K0S_IPV6_ONLY=yes
 check-cplb-userspace-extaddr: export K0S_USE_EXTERNAL_ADDRESS=yes
 check-cplb-userspace-extaddr: TEST_PACKAGE=cplb-userspace
 
--- a/inttest/common/network_util.go
+++ b/inttest/common/network_util.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2025 k0s authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+import (
+	"net"
+	"strings"
+)
+
+// FirstPublicIPv6Address retrieves the first public IPv6 address from the eth0 interface of a node.
+func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string) string {
+	ssh, err := s.SSH(s.Context(), nodeName)
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+
+	output, err := ssh.ExecWithOutput(s.Context(), "ip -6 -oneline addr show eth0 scope global")
+	s.Require().NoError(err)
+
+	// Parse the output line by line
+	for line := range strings.SplitSeq(strings.TrimSpace(output), "\n") {
+		fields := strings.Fields(line)
+		s.Require().GreaterOrEqual(len(fields), 4, "Expected at least 4 fields in the output line")
+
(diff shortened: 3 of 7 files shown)


inttest/cplb-ipvs/cplbipvs_test.go
```diff
@@ -31,25 +33,42 @@ type cplbIPVSSuite struct {
 	common.BootlooseSuite
 }
 
-const haControllerConfig = `
+const cplbCfgTemplate = `
```
twz123, line 37: Sometimes I wonder if we'd better use a `ClusterConfig` struct in the code and then use `yaml.Marshal`, instead of doing lots of string templates.

inttest/common/network_util.go
```diff
@@ -44,3 +44,25 @@ func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string) string {
...
+// This is so that get a virtual IP in the same subnet as the controller.
+func GetCPLBVIP(s *BootlooseSuite, isIPv6Only bool) string {
+	var ip string
+	if isIPv6Only {
+		ip = FirstPublicIPv6Address(s, s.ControllerNode(0))
+	} else {
+		ip = s.GetIPAddress(s.ControllerNode(0))
+	}
+
+	addr := net.ParseIP(ip)
```
twz123, line 64: What about checking for `nil` here?

inttest/common/network_util.go
```diff
@@ -44,3 +44,25 @@ func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string) string {
...
+	if addr[15] >= 154 {
+		addr[15] -= 100
+	} else {
+		addr[15] += 100
+	}
+
+	return addr.String()
```
twz123, line 72: We've already parsed it. Why not returning it directly as `net.IP`?

Makefile
```diff
@@ -250,7 +250,12 @@ airgap-image-bundle-linux-riscv64.tar: k0s airgap-images.txt
 	./k0s airgap bundle-artifacts -v --platform='$(TARGET_PLATFORM)' -o '$@' <airgap-images.txt
 
 .PHONY: $(smoketests)
-check-airgap check-ap-airgap check-ipv6-calico check-ipv6-kuberouter check-cplb-userspace-ipv6: airgap-image-bundle-linux-$(HOST_ARCH).tar
+check-airgap \
+check-ap-airgap \
+check-ipv6-calico \
+check-ipv6-kuberouter \
+check-cplb-ipvs-ipv6 \
+check-cplb-userspace-ipv6: airgap-image-bundle-linux-$(HOST_ARCH).tar
```
twz123, line 258: Maybe it's time to introduce a new Make variable in `inttest/Makefile.variables`, e.g. sth. like `air_gapped_smoketests`. Then this could be replaced with:

```Makefile
$(air_gapped_smoketests): airgap-image-bundle-linux-$(HOST_ARCH).tar
```

inttest/common/network_util.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 // FirstPublicIPv6Address retrieves the first public IPv6 address from the eth0 interface of a node.
-func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string) string {
+// A cplbVIP may be provided so that it can be ignored.
+func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string, cplbVip string) string {
```
twz123: What about a signature like this:

```go
func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string, excludedIPs ...string) string {
```

Then you could change the check a few lines below to:

```go
if ip := ip.String(); !slices.Contains(excludedIPs, ip) {
	return ip
}
```

This wouldn't require you to pass an empty string in case you don't want to filter anything out.

---

Albeit the flexibility is not required right now, this comes to my mind, as well:

```go
func FirstPublicIPv6Address(s *BootlooseSuite, nodeName string, exclude func(string) bool) string {
```

I somehow still feel that we should be using proper types (i.e. `net.IP`) instead of strings.

inttest/Makefile
```diff
@@ -87,6 +87,10 @@ check-customports-dynamicconfig: TEST_PACKAGE=customports
 
 check-kubeletcertrotate: TIMEOUT=15m
 
+check-cplb-ipvs-ipv6: export K0S_IPV6_ONLY=yes
+check-cplb-ipvs-ipv6: TEST_PACKAGE=cplb-ipvs
+
+check-cplb-userspace-ipv6: export K0S_IPV6_ONLY=yes
```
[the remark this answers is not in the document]
twz123: See line 99.

twz123 (approved):

Makefile
juanluisvaladas: That's a good point. I just relaized we'red missing check-nllb-ipv6 so I'll address this in a new PR.

inttest/Makefile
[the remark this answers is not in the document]
juanluisvaladas: this line is duplicate of line 98. I removed it in a new commit.

inttest/common/network_util.go
juanluisvaladas: Using slices.Contains is much cleaner. I did that.

As for using the type net.IP I think it complicates the code a bit more than necessary. We do most of comparisons checking strings from IP commands and it works well, it's a very contained environment. At the end of the day we're always running the same alpine image with the same ip command and the same predictable output.

Using net.IP is a bit annoying because net.IP is not `comparable`, which means things like `slices.Contains` won't work. It can be done, but it's not that easy and the code looks a bit uglier. If it were for the actual program yes, but for the test I'm not sure it's worth doing it.

inttest/common/network_util.go
juanluisvaladas: Good point, it'd fail anyway but that would make things easier to debug.

inttest/common/network_util.go
juanluisvaladas: We use strings everywhere, I think it's just easier and it's for a test. Not quite sure it's worth working with net.IP here and convert back and forth. hasVip checkDummy and validateRealServers get the output from the ip command and compare the strings. It would be converting a lot more.

If it was for the actual program it'd be 100% worth doing it, but for this I'm not sure the benefit is worth making the code longer, with more type conversions and harder to read and all

inttest/cplb-ipvs/cplbipvs_test.go
juanluisvaladas: I considered doing that, in fact before the template I copied the code from the NLLB test, but for this test which is a bit longer than usual the struct got pretty big and I think this is a bit easier to read.

I think both are valid, I preferred this one but if you. want I can change it.

juanluisvaladas:
@twz123  I did a couple changes based on the code review if you could have another look:

https://github.com/k0sproject/k0s/compare/e97f1e280e72db038c50b526e908a39a4c21cc53..6b526fc738bcc5037c2caf0576dd02c682b2e91b

PR #6085, opened 2025-07-07
Use Go's os.Process API when cleaning up PID files

twz123:
## Description

In recent Go versions, Go's process API gained the ability to open processes using pidfds on kernels that support it. Support was added in Linux 5.3. This addresses a possible race condition that may occur if the kernel reuses recycled PIDs. Keeping an open file descriptor guarantees that the PID won't be recycled. This ensures that the signal is sent to the intended process and not to another process that was assigned the same recycled PID. On older kernels, it simply falls back to the old, racy implementation that uses PIDs.

Rename some `procHandle` methods:

* `newProcHandle` -> `openPID`  
  Indicates that the handle needs to be closed now.
* `terminateGracefully` -> `requestGracefulTermination`  
  This reflects better the fact that this is an asynchronous method, i.e. it won't block until the process actually terminated.
* `terminateForcefully` -> `kill`
 
## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/supervisor/prochandle.go
+++ b/pkg/supervisor/prochandle.go
@@ -25,9 +25,9 @@ type procHandle interface {
 	// Reads and returns the process's environment.
 	environ() ([]string, error)
 
-	// Terminates the process gracefully.
-	terminateGracefully() error
+	// Requests graceful process termination.
+	requestGracefulShutdown() error
 
-	// Terminates the process forcibly.
-	terminateForcibly() error
+	// Kills the process.
+	kill() error
 }
--- a/pkg/supervisor/prochandle.go
+++ b/pkg/supervisor/prochandle.go
@@ -16,9 +16,15 @@ limitations under the License.
 
 package supervisor
 
+import (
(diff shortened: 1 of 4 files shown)


juanluisvaladas (approved):
Just for reference (because I was looking into this myself while doing the review): This requires kernel 5.3 because it's when `pidfd_open` is introduced, which is used by `os.FindProcess`, however pidfds themselves are introduced earlier in the kernel.

twz123:
> Just for reference (because I was looking into this myself while doing the review): This requires kernel 5.3 because it's when pidfd_open is introduced, which is used by os.FindProcess, however pidfds themselves are introduced earlier in the kernel.

Right. You can use directory file descriptors from the proc file system, as well, but those support a bit of a different set of capabilities, but there's overlap. Those aren't used by the Go process API, though. I've some PoCs for that stuff, as well.

PR #6086, opened 2025-07-07
[Backport release-1.32] Replace hardcoded k0s versions in docs

juanluisvaladas:
Backport to `release-1.32`:

* #5996
* #5983
* #5995
* #6016
* #5992
* #6024

See:

* #6018


juanluisvaladas:
I'm going this so that we can use the same logic in renovate for 1.33 and 1.32 when bumping kubernetes versions.

PR #6130, opened 2025-07-11
Implement k0s feature gates

### LINKED ISSUE #6084: Introduce an ipv6 single stack feature gate.
Known unknowns:

juanluisvaladas:
## Description

Introduces a feature gates system to k0s, inspired by Kubernetes feature gates, enabling controlled rollout of experimental and new features.

These are heavily inspired but kubernetes' feature gates, but we don't need something that sophisticated and hence we're doing a much simpler implementation for them.
    
Unlike in kubernetes, we've decided not to implement AllAlpha and AllBeta at least for now.
    
Implements #608

Fixes #6084

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/docs/feature-gates.md
+++ b/docs/feature-gates.md
@@ -1,0 +1,51 @@
+<!--
+SPDX-FileCopyrightText: 2025 k0s authors
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Feature Gates
+
+Feature gates are a mechanism to enable or disable experimental features in k0s.
+They allow you to control which features are active in your cluster, providing
+a way to safely test new functionality before it becomes generally available.
+
+## Overview
+
+Feature gates in k0s follow a lifecycle model similar to Kubernetes:
+
+- **Alpha**: Experimental features that are disabled by default and may be unstable or removed
+- **Beta**: Features that are more stable but still being tested, enabled by default
+- **GA (Generally Available)**: Stable features that cannot be disabled. If explicitly disabled
+k0s will fail to start
+
+Once a feature reaches GA status, the feature gate cannot be disabled anymore and it will be
+eventually removed.
+
+## Using Feature Gates
+
+Feature gates can be configured using the `--feature-gates` flag when starting the k0s controllers:
+
+```shell
+k0s controller --feature-gates="ExampleFeature=true,AnotherFeature=false"
+```
+
+The feature gates configuration must be consistent across all controllers in the cluster.
+
+**Important notes:**
+
+- Currently feature gates apply only to the controllers. They may be added to workers in future releases
+- Setting a GA feature gate to false will cause k0s startup to fail
+- Unknown feature gates will cause k0s start up to fail
+- Feature gates use case-sensitive CamelCase names
+
+## Available Features
+
+The following table is a summary of the feature gates that you can set. Each feature is designed for a specific stage of stability:
+
+| Feature           | Default | Stage | Since | Until |
+|-------------------|---------|-------|-------|-------|
+| `IPv6SingleStack` | `false` | Alpha | 1.34  |       |
+
+### Feature Details
+
+- `IPv6SingleStack`: Enables single-stack IPv6 support in k0s clusters.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -46,6 +46,7 @@ nav:
       - System Components Monitoring: system-monitoring.md
       - Manifest Deployer: manifests.md
       - Helm Charts: helm-charts.md
+      - Feature Gates: feature-gates.md
       - Cloud Providers: cloud-providers.md
       - IPv4/IPv6 dual-stack networking: dual-stack.md
       - Control Plane High Availability: high-availability.md
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"slices"
 
+	"github.com/k0sproject/k0s/pkg/featuregate"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	utilnet "k8s.io/utils/net"
 
@@ -106,11 +107,17 @@ func (n *Network) Validate() []error {
 		validCIDRs = false
 	}
 
-	// podCIDR and service CIDR must both be either ipv4 or ipv6, but we should only verify this
-	// if they have a valid IP to begin with because otherwise the error can be confusing
 	if validCIDRs {
+		// podCIDR and service CIDR must both be either ipv4 or ipv6, but we should only verify this
+		// if they have a valid IP to begin with because otherwise the error can be confusing
 		if (podNetIP.To4() != nil) != (serviceNetIP.To4() != nil) {
 			errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "podCIDR and serviceCIDR must be both IPv4 or IPv6"))
+		}
+
+		// Single stack IPv6 is an alpha feature, so we need to check if the feature gate is enabled, but only report it if both
+		// podCIDR and serviceCIDR are IPv6 have been validated so that the error is easier to understand.
+		if !n.DualStack.Enabled && podNetIP.To4() == nil && !featuregate.IsEnabled(featuregate.IPv6SingleStack) {
+			errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "feature gate IPv6SingleStack must be explicitly enabled to use IPv6 single stack"))
 		}
 	}
 
--- a/pkg/config/cli.go
+++ b/pkg/config/cli.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/constant"
+	"github.com/k0sproject/k0s/pkg/featuregate"
 	"github.com/k0sproject/k0s/pkg/k0scloudprovider"
 
 	cliflag "k8s.io/component-base/cli/flag"
@@ -54,6 +55,7 @@ type ControllerOptions struct {
 	EnableDynamicConfig             bool
 	EnableMetricsScraper            bool
 	KubeControllerManagerExtraArgs  string
+	FeatureGates                    featuregate.FeatureGates
 
 	enableWorker, singleNode bool
 }
@@ -285,6 +287,7 @@ func GetControllerFlags(controllerOpts *ControllerOptions) *pflag.FlagSet {
 	flagset.BoolVar(&controllerOpts.EnableMetricsScraper, "enable-metrics-scraper", false, "enable scraping metrics from the controller components (kube-scheduler, kube-controller-manager)")
 	flagset.StringVar(&controllerOpts.KubeControllerManagerExtraArgs, "kube-controller-manager-extra-args", "", "extra args for kube-controller-manager")
 	flagset.BoolVar(&controllerOpts.InitOnly, "init-only", false, "only initialize controller and exit")
+	flagset.Var(&controllerOpts.FeatureGates, "feature-gates", "feature gates to enable (comma separated list of key=value pairs)")
 	return flagset
 }
 
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -1,0 +1,172 @@
+// SPDX-FileCopyrightText: 2025 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package featuregate
+
+import (
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+type Feature string
+
+type FeatureGates struct {
+	featureMap map[Feature]FeatureGate
+}
+
+type stage string
+
+const (
+	Alpha stage = "ALPHA"
+	Beta  stage = "BETA"
+	GA    stage = "GA"
+
+	Deprecated stage = "DEPRECATED"
+)
+
+type FeatureGate struct {
+	// Enabled determines if the feature is enabled.
+	// When declared, it should set to its default value, which may be
+	// overwritten based on CLI flags.
+	Enabled bool
+	// PreRelease indicates the stage of the feature
+	Stage stage
+}
+
+// isEnabled checks if a feature is enabled.
+func (fg *FeatureGates) isEnabled(feature Feature) bool {
+	if f, ok := fg.featureMap[feature]; ok {
+		return f.Enabled
+	}
+	return false
+}
+
+// ConfigureFeatureGates parses the provided flag string and configures the
+// feature gates accordingly. This function is not thread-safe and is only
+// expected to be called once during the initialization phase.
+func (fg *FeatureGates) configureFeatureGates(flag string) error {
+	featureMap, err := parseFeatureGateFlags(flag)
+	if err != nil {
+		return err
+	}
+
+	fg.setDefaults()
+
+	// Apply the parsed feature gates
+	for feature, enabled := range featureMap {
+		gate, exists := fg.featureMap[Feature(feature)]
+		if !exists {
+			return fmt.Errorf("unknown feature gate %q", feature)
+		}
+		if gate.Stage == GA && !enabled {
+			// GA features cannot be disabled, so revert the change before returning an error
+			return fmt.Errorf("feature gate %q is GA and cannot be disabled", feature)
+		}
+		gate.Enabled = enabled
+		fg.featureMap[Feature(feature)] = gate
+	}
+
+	return nil
+}
+
+func (fg *FeatureGates) setDefaults() {
+	for key, featureGate := range fg.featureMap {
+		featureGate.Enabled = featureGate.Stage != Alpha
+		fg.featureMap[key] = featureGate
+	}
+}
+
+// parseFeatureGateFlags parses a comma-separated string of key=value pairs
+// and returns a map of feature names to their enabled state.
+func parseFeatureGateFlags(flag string) (map[string]bool, error) {
+	if flag == "" {
+		return nil, nil
+	}
+
+	featureMap := make(map[string]bool)
+
+	for _, pair := range strings.Split(flag, ",") {
+		key, value, found := strings.Cut(pair, "=")
+		if !found {
+			return nil, fmt.Errorf("invalid feature gate format %q, expected key=value", pair)
+		}
+
+		key = strings.TrimSpace(key)
+		value = strings.TrimSpace(value)
+
+		if key == "" {
+			return nil, fmt.Errorf("empty feature gate key in %q", pair)
+		}
+
+		enabled, err := strconv.ParseBool(value)
+		if err != nil {
+			return nil, fmt.Errorf("invalid feature gate value %q for key %q, expected true or false", value, key)
+		}
+		featureMap[key] = enabled
+	}
+
+	return featureMap, nil
+}
+
+// IsEnabled checks if a feature is enabled in the defaultFeatureGates singleton
+// This function expects that defaultFeatureGates has been initialized and otherwise
+// will panic.
+func IsEnabled(feature Feature) bool {
+	return defaultFeatureGates.isEnabled(feature)
+}
+
+// Set initializes the default feature gates with the provided val string.
+// This function must be called only once during the initialization phase
+// by spf13/pflag while parsing command line flags.
+func (fg *FeatureGates) Set(val string) error {
+	if defaultFeatureGates != nil {
+		return errors.New("feature gates already initialized")
+	}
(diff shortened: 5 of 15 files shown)


pkg/apis/k0s/v1beta1/network.go
```diff
@@ -118,12 +119,18 @@ func (n *Network) Validate() []error {
...
 	if validCIDRs {
+		// podCIDR and service CIDR must both be either ipv4 or ipv6, but we should only verify this
+		// if they have a valid IP to begin with because otherwise the error can be confusing
 		if (podNetIP.To4() != nil) != (serviceNetIP.To4() != nil) {
 			errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "podCIDR and serviceCIDR must be both IPv4 or IPv6"))
 		}
+
+		// Single stack IPv6 is an alpha feature, so we need to check if the feature gate is enabled, but only report it if both
+		// podCIDR and serviceCIDR are IPv6 have been validated so that the error is easier to understand.
+		if !n.DualStack.Enabled && podNetIP.To4() == nil && !featuregate.DefaultFeatureGates.IsEnabled(featuregate.IPv6SingleStack) {
```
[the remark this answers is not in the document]
juanluisvaladas: I'm going to ignore this deliberately because we already check `(podNetIP.To4() == nil) == (serviceNetIP.To4() == nil )` so it's redundant. Otherwise would be a good suggestion.

pkg/featuregate/featuregate_test.go
```diff
@@ -0,0 +1,170 @@
...
+			Feature("Feature3"): {Enabled: false, Prerelease: Beta},
+			Feature("Feature4"): {Enabled: true, Prerelease: Beta},
+		},
+	}
+	testFG.configureGlobalFeatureGates(false, false)
+	for feature, gate := range testFG.featureMap {
+		s.Falsef(gate.Enabled, "%q should be disabled", feature)
+	}
+	for feature, gate := range testFG.featureMap {
+		s.Falsef(gate.Enabled, "%q should be enabled", feature)
```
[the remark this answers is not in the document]
juanluisvaladas: The suggestion isn't entirely correct, but it caught a problem I just fixed.

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+			featureMap[key] = false
+		default:
+			return nil, fmt.Errorf("invalid feature gate value %q for key %q, expected true or false", value, key)
+		}
+	}
+
+	return featureMap, nil
+}
+
+func (fg *FeatureGates) configureGlobalFeatureGates(alphaEnabled bool, betaEnabled bool) {
```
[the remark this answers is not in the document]
juanluisvaladas: I think that merits some discussion but 1.34 is only a month and a half away and I'd prefer to have this discussion when the 1.34 deadline isn't a thing... I don't know if we want to throw an error if the user tries to disable them or not, how we want to handle this.

At this point I don't think we really need to take care of those details and I'd rather delay it a bit.

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+They allow you to control which features are active in your cluster, providing
+a way to safely test new functionality before it becomes generally available.
+
+## Overview
+
+Feature gates in k0s follow a lifecycle model similar to Kubernetes:
+
+- **Alpha**: Experimental features that are disabled by default and may be unstable
+- **Beta**: Features that are more stable but still being tested, disabled by default
+- **GA (Generally Available)**: Stable features that are enabled by default
```
twz123: This feels a bit duplicate with the lifecycle section at the end, where it's also mentioned that GA feature flags are going to be removed and always on.

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Using Feature Gates
+
+Feature gates can be configured using the `--feature-gates` flag when starting the k0s controller:
```
twz123: Can't the workers have feature gates as well?

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+
+```shell
+k0s controller --feature-gates="FeatureName=true,AnotherFeature=false"
+```
+
+Feature gates are case sensitive.
```
twz123: My spell checker says `case-sensitive` 🤔

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Global Feature Gates
+
+k0s supports global toggles for feature maturity levels:
+
+- `AllAlpha=true`: Enables all alpha features
+- `AllBeta=true`: Enables all beta features
```
twz123: What's the rationale behind those flags, and why could they be useful? I mean, we just have one feature gate right now, so this doesn't add much value, I think? What about removing this right now and add it on demand, if we have, say, three feature gates?

Thinking longer about this, modelling those flags as feature gates feels wrong. Those might be extra flags, like `--alpha-feature-gates-enabled-by-default` and `--beta-feature-gates-disabled-by-default` or similar. They don't fit into the Feature/Prerelease pattern.

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+- **Alpha**: Experimental features that are disabled by default and may be unstable
+- **Beta**: Features that are more stable but still being tested, disabled by default
+- **GA (Generally Available)**: Stable features that are enabled by default
+
+## Using Feature Gates
+
+Feature gates can be configured using the `--feature-gates` flag when starting the k0s controller:
+
+```shell
+k0s controller --feature-gates="FeatureName=true,AnotherFeature=false"
```
twz123: Maybe we could mention that feature gates are PascalCase/UpperCamelCase somewhere?

docs/feature-gates.md
```diff
@@ -0,0 +1,63 @@
...
+Feature gates are a mechanism to enable or disable experimental features in k0s.
+They allow you to control which features are active in your cluster, providing
+a way to safely test new functionality before it becomes generally available.
+
+## Overview
+
+Feature gates in k0s follow a lifecycle model similar to Kubernetes:
+
+- **Alpha**: Experimental features that are disabled by default and may be unstable
+- **Beta**: Features that are more stable but still being tested, disabled by default
```
twz123: Shouldn't beta features be on by default? I thought this is what upstream is doing.

I understand the different maturity levels as follows:

Alpha=Opt-In
Beta=Opt-Out
GA=Always-on

mkdocs.yml
```diff
@@ -56,6 +56,7 @@ nav:
       - OpenID Connect: ./examples/oidc/oidc-cluster-configuration.md
       - SELinux: selinux.md
       - Pod Security Standards: podsecurity.md
+      - Feature Gates: feature-gates.md
```
twz123: Maybe rank it a bit higher in the docs?

pkg/config/cli.go
```diff
@@ -65,6 +65,7 @@ type ControllerOptions struct {
 	EnableDynamicConfig             bool
 	EnableMetricsScraper            bool
 	KubeControllerManagerExtraArgs  string
+	FeatureGates                    string
```
twz123: I'd rather use a real type like `FeatureGates` directly here, and implement the flag parsing via Cobra and an adapter type. Similar to what we do for, say, log levels.

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+	Deprecated = prerelease("DEPRECATED")
+)
+
+type FeatureGate struct {
+	// Enabled determines if the feature is enabled.
+	// When declared, it should set to its default value, which may be
+	// overwritten based on CLI flags.
+	Enabled bool
+	// PreRelease indicates the stage of the feature
+	Prerelease prerelease
```
twz123: I wonder if that is a property of the feature, rather than the feature gate? I mean, the feature is in ALPHA/BETA/GA. The gate just says if it's enabled or not?

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+	"strings"
+)
+
+type Feature string
+
+type FeatureGates struct {
+	featureMap map[Feature]FeatureGate
+}
+
+type prerelease string
```
twz123: Is this the name that's used in upstream? Feels like a slight misnomer? I'd rather call it "maturity", "stage", or "status" or sth. along those lines; "prerelease" sounds so much like a boolean to me 🤷

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+
+const (
+	Alpha = prerelease("ALPHA")
+	Beta  = prerelease("BETA")
+	GA    = prerelease("")
+
+	Deprecated = prerelease("DEPRECATED")
+)
```
twz123: Nit about formatting. Wouldn't this be more idiomatic?

```suggestion
const (
	Alpha prerelease = "ALPHA"
	Beta  prerelease = "BETA"
	GA    prerelease = ""

	Deprecated prerelease = "DEPRECATED"
)
```

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+// and returns a map of feature names to their enabled state.
+func parseFeatureGateFlags(flag string) (map[string]bool, error) {
+	if flag == "" {
+		return nil, nil
+	}
+
+	featureMap := make(map[string]bool)
+	pairs := strings.Split(flag, ",")
+
+	for _, pair := range pairs {
```
twz123: This is slightly more efficient (in a code path where it doesn't matter at all 😆):

```suggestion
	for pair := range strings.SplitSeq(flag, ",") {
```

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+		pair = strings.TrimSpace(pair)
```
twz123: All the TrimSpace calls don't have test coverage. Do we want to keep them?

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+			return nil, fmt.Errorf("empty feature gate entry in %q", flag)
+		}
+
+		parts := strings.SplitN(pair, "=", 2)
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("invalid feature gate format %q, expected key=value", pair)
+		}
+
+		key := strings.TrimSpace(parts[0])
+		value := strings.TrimSpace(parts[1])
```
twz123: This is a first class use case for `strings.Cut`:

```suggestion
		key, value, ok := strings.Cut(pair, "=")
		if !ok {
			return nil, fmt.Errorf("invalid feature gate format %q, expected key=value", pair)
		}

		key = strings.TrimSpace(key)
		value = strings.TrimSpace(value)
```

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		switch strings.ToLower(value) {
+		case "true":
+			featureMap[key] = true
+		case "false":
+			featureMap[key] = false
+		default:
+			return nil, fmt.Errorf("invalid feature gate value %q for key %q, expected true or false", value, key)
+		}
```
twz123: We could use `strconv.ParseBool` here. This allows for more patterns to be recognized, like 1 and 0 and so on. We could remove the ToLower call on value. This way we would recognize TRUE, True, true, but not tRuE...

```suggestion
		enabled, err := strconv.ParseBool(value)
		if err != nil {
			return nil, fmt.Errorf("invalid feature gate value %q for key %q: %w", value, key, err)
		}
		featureMap[key] = enabled
```

pkg/featuregate/featuregate.go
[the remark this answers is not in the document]
twz123: TBH, I think this is a very important design decision that should be agreed on up front. I've made some comments in the docs section about this already.

pkg/featuregate/featuregate_test.go
```diff
@@ -0,0 +1,170 @@
...
+		s.Truef(gate.Enabled, "%q should be enabled", feature)
+	}
+}
+
+func (s *FeatureGateSuite) TestParseFeatureGateFlags() {
+	tests := []struct {
+		name        string
+		input       string
+		expected    map[string]bool
+		expectError bool
```
twz123, line 83: you could make it a string, so you can actually assert the correct error info via `assert.ErrorContains(t, err, ...)`

pkg/featuregate/featuregate_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+package featuregate
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type FeatureGateSuite struct {
```
twz123: Not sure if using a test suite will gain us much here. Did you consider to just use plain old test functions instead?

pkg/featuregate/featuregate.go
twz123: What's meant by "Global" in the method's name? This looks rather like some sort of "setDefaults"?

pkg/featuregate/features.go
```diff
@@ -0,0 +1,42 @@
...
+	// values override the default set by allBetaGate. Examples:
+	//   AllBeta=false,NewFeature=true  will result in NewFeature=true
+	//   AllBeta=true,NewFeature=false  will result in NewFeature=false
+	allBetaGate Feature = "AllBeta"
+
+	// IPv6SingleStack enables single stack IPv6 support in k0s
+	IPv6SingleStack Feature = "IPv6SingleStack"
+)
+
+var DefaultFeatureGates FeatureGates = FeatureGates{
```
twz123: I'm not fond of having another mutable global variable. Personally, I'd like the feature gates to be injected into the places where they're needed, either explicitly, or via, say, a context value. I've seen the latter pattern being used in containerd for logging, and I think it's kinda nice middle ground to make "ambient" stuff available throughout the application without introducing new parameters virtually everywhere and not depending on global state at the same time.

docs/feature-gates.md
juanluisvaladas: We only have one feature gate right now and it would only affect the controllers. We could add it to the worker flags too but it wouldn't be doing anything at all, so I think it'd make more sense to add it in the future if we need it at some point.

docs/feature-gates.md
juanluisvaladas: I'll address this once we are clear on the rest of things

docs/feature-gates.md
juanluisvaladas: I did that to resemble the kubernetes feature gates:

https://github.com/kubernetes/kubernetes/blob/v1.33.3/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L42-L56

It can definitely be removed or changed, but I thought it'd be a nice addition. Definitely worth more discussion.

docs/feature-gates.md
juanluisvaladas: You're right. Once we're clear on what we want to do with the `AllAlpha` and `AllBeta` feature gates (if we want to keep them) I'll fix this.

mkdocs.yml
juanluisvaladas: Where do you think would be appropriate? Maybe between Helm Charts and Cloud Providers?

pkg/featuregate/featuregate.go
juanluisvaladas: Yes, that's the name used in upstream. I'm not very convinced with it either. I think `stage` is a much better name. I'll change that

pkg/featuregate/featuregate.go
juanluisvaladas: Kubernetes refers to `AllAlpha` and `AllBeta` feature gates as global feature gates, but `setDefaults` is indeed snaer.

juanluisvaladas:
These are all solid points, let's have a proper discussion on whether we want to do. The AllAlpha and AllBeta feature gates  before https://github.com/kubernetes/kubernetes/blob/v1.33.3/staging/src/k8s.io/component-base/featuregate/feature_gate.go#L95-L106

Once we're clear on what we want in that regard we'll evaluate the rest of the choices.

And definitely we can inject the specific feature gate value in the places where it's required.

docs/feature-gates.md
twz123: I see. Maybe change the wording a bit, then? I.e. mentioning that, currently, all the feature gates only affect k0s nodes that run control plane components, (controller and controller --enable-worker)?

juanluisvaladas:
As discussed in the call, we're removing `AllAlpha` and `AllBeta` features for now. We'll revisit this in the future if need be.  I'll mark all the discussions regarding these special flags as resolved and will remove them entirely.

juanluisvaladas:
I did some experiments trying to remove the singleton but it really looks pretty bad. It requires modifying the `v1beta1.Validateable` interface which eventually means we have to modify all these files:
- cmd/config/validate.go
- cmd/controller/controller.go
- cmd/install/controller.go
- pkg/apis/k0s/v1beta1/api.go
- pkg/apis/k0s/v1beta1/api_test.go
- pkg/apis/k0s/v1beta1/clusterconfig_types.go
- pkg/apis/k0s/v1beta1/clusterconfig_types_test.go
- pkg/apis/k0s/v1beta1/cplb.go
- pkg/apis/k0s/v1beta1/extensions.go
- pkg/apis/k0s/v1beta1/feature_gates.go
- pkg/apis/k0s/v1beta1/konnectivity.go
- pkg/apis/k0s/v1beta1/kubeproxy.go
- pkg/apis/k0s/v1beta1/network.go
- pkg/apis/k0s/v1beta1/network_test.go
- pkg/apis/k0s/v1beta1/nllb_test.go
- pkg/apis/k0s/v1beta1/storage.go
- pkg/apis/k0s/v1beta1/storage_test.go
- pkg/apis/k0s/v1beta1/telemetry.go
- pkg/apis/k0s/v1beta1/workerprofile.go
- pkg/component/controller/clusterconfig.go
- pkg/featuregate/features.go

And it's not the prettiest thing. I think in this case the singleton really makes sense, however instead of doing a global variable which may be overwritten I'll do a proper singleton that cannot be overwritten after its initialization

pkg/featuregate/featuregate.go
juanluisvaladas: Removed as agreed in the call

docs/feature-gates.md
juanluisvaladas: Indeed, removed

docs/feature-gates.md
juanluisvaladas: I modified the docs, please review.

docs/feature-gates.md
juanluisvaladas: I think this should be taken care of now

pkg/config/cli.go
juanluisvaladas: Done, if you want we can add an intermediary adapter type like in the log levels.

juanluisvaladas:
I believe all the concerns are addressed except the using a global variable, which isn't entirely addressed but at least now isn't mutable: https://github.com/k0sproject/k0s/pull/6130#issuecomment-3133135253 and deciding where we should put it in [mkdocs.yml](https://github.com/k0sproject/k0s/pull/6130/files#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2)

I don't think it's worth adding a mutex in the package featuregate because `FeatureGate.Set` shouldn't be called more than once and everything else (after `Set` is finished) should be thread safe, but if anyone is concerned about it, I'll add it.

The PR is. pretty much a full rewrite at this point.

pkg/config/cli.go
```diff
@@ -54,6 +55,7 @@ type ControllerOptions struct {
 	EnableDynamicConfig             bool
 	EnableMetricsScraper            bool
 	KubeControllerManagerExtraArgs  string
+	FeatureGates                    featuregate.FeatureGates
```
juanluisvaladas, line 58: I decided to add it here instead of adding it as yet another global variable such as `CfgFile`or `K0sVars`, this limits its capability to run as part of `k0s worker` but I'm not overly concerned about this at the moment.

I think this is cleaner and if we need to add it in the future we can worry about it later.

pkg/featuregate/featuregate.go
```diff
@@ -0,0 +1,172 @@
...
+// This function expects that defaultFeatureGates has been initialized and otherwise
+// will panic.
+func IsEnabled(feature Feature) bool {
+	return defaultFeatureGates.isEnabled(feature)
+}
+
+// Set initializes the default feature gates with the provided val string.
+// This function must be called only once during the initialization phase
+// by spf13/pflag while parsing command line flags.
+func (fg *FeatureGates) Set(val string) error {
```
juanluisvaladas, line 126: Should we lock this? I don't think it's an issue but theoretically if someone did something wrong it could be a problem

juanluisvaladas:
Rebasing for having the license check clean, no actual changes made

PR #6220, opened 2025-07-31
[WIP] Bump bootloose and remove dependency docker/docker

juanluisvaladas:
This is PR is just to run CI and verify that nothing breaks. I did a quick test on my machine


juanluisvaladas:
https://github.com/k0sproject/bootloose/pull/179 is merged so there is no need for this. Dependabot will bump it soon.

Issue #3978, opened 2024-01-28
extraArgs is a dictionary


mxrss2:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 5.15.0-92-generic #102-Ubuntu SMP Wed Jan 10 09:33:48 UTC 2024 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.28.5+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "fd7a81e4083cbd180875b2e3c1156717f7e016f6406efc4460fead22a9875783" (from machine) (pass)
Total memory: 3.7 GiB (pass)
Disk space available for /var/lib/k0s: 14.1 GiB (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 5.15.0-92-generic (pass)
  Max. file descriptors per process: current: 1048576 / max: 1048576 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (pass)
    cgroup controller "memory": available (pass)
[...]

PR #6235, opened 2025-08-06
[etcd] Implement rawArgs

juanluisvaladas:
## Description

As discussed in #3978 we need a better way to deal with the arguments. This is a first implementation for etcd.

This isn't a perfect solution, but it solves the problem until configV2.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/storage.go
+++ b/pkg/apis/k0s/v1beta1/storage.go
@@ -128,6 +128,10 @@ type EtcdConfig struct {
 	// Map of key-values (strings) for any extra arguments you want to pass down to the etcd process
 	ExtraArgs map[string]string `json:"extraArgs,omitempty"`
 
+	// Slice of strings with raw arguments to pass to the etcd process
+	// These arguments will be appended to the `ExtraArgs` and aren't validated at all.
+	RawArgs []string `json:"rawArgs,omitempty"`
+
 	// Custom config for CA certificates.
 	CA *CA `json:"ca,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


kke:
I think a type switch as discussed in https://github.com/k0sproject/k0s/issues/3978#issuecomment-1914121701 and `ExtraArgs any 'json:'extraArgs''` would work just as well (except in the example the type of the field in the struct needs to be `any` and the typeswitch needs to be nested/recursive or have some kind of reflect thing to differentiate between `string`, `map` and a slice).

kke:
Any of these solutions will be "api breaking" in the sense that a config written for a newer version will not parse or may parse incorrectly on an older version.

Having the two fields might be easier from k0sctl's perspective 🤔

pkg/component/controller/etcd.go
```diff
@@ -225,14 +225,20 @@ func (e *Etcd) Start(ctx context.Context) error {
 		args["--cipher-suites"] = constant.AllowedTLS12CipherSuiteNames()
 	}
 
+	// Remove duplicates between extraArgs and rawArgs
+	for _, arg := range e.Config.RawArgs {
+		key, _, _ := strings.Cut(arg, "=")
+		delete(args, key)
+	}
+
```
kke: I think removing duplicates could be against the wild spirit of `RawArgs`.

juanluisvaladas: Fair enough, removed this behavior in a second commit.

kke (approved):

pkg/apis/k0s/v1beta1/storage.go
```diff
@@ -128,6 +128,10 @@ type EtcdConfig struct {
 	// Map of key-values (strings) for any extra arguments you want to pass down to the etcd process
 	ExtraArgs map[string]string `json:"extraArgs,omitempty"`
 
+	// Slice of strings with raw arguments to pass to the etcd process
+	// These arguments will be appended to the `ExtraArgs` and aren't validated at all.
+	RawArgs []string `json:"rawArgs,omitempty"`
```
kke, line 133: (Maybe this could be marked as deprecated from the get go 🤔 )

PR #6241, opened 2025-08-07
Improve worker profile validation

juanluisvaladas:
## Description

An invalid worker profile that was a syntactically correct yaml would fail in an unexpected way. The validation step wouldn't throw an error and the configmaps wouldn't be created. Detect the problem early and exit early.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/workerprofile.go
+++ b/pkg/apis/k0s/v1beta1/workerprofile.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	kubeletv1beta1 "k8s.io/kubelet/config/v1beta1"
 )
 
 var _ Validateable = (*WorkerProfiles)(nil)
@@ -22,7 +23,7 @@ func (wps WorkerProfiles) Validate() []error {
 	var errors []error
 	for _, p := range wps {
 		if err := p.Validate(); err != nil {
-			errors = append(errors, err)
+			errors = append(errors, err...)
 		}
 	}
 	return errors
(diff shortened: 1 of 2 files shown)


pkg/apis/k0s/v1beta1/workerprofile.go
```diff
@@ -37,26 +38,29 @@ type WorkerProfile struct {
...
 
-	for field := range parsed {
-		if _, found := lockedFields[field]; found {
-			return fmt.Errorf("field `%s` is prohibited to override in worker profile", field)
-		}
+	var errs []error
+
+	kubeletCfg := &kubeletv1beta1.KubeletConfiguration{}
+	if err := json.Unmarshal(wp.Config.Raw, kubeletCfg); err != nil {
+		errs = append(errs, fmt.Errorf("failed to decode worker profile %q: %w", wp.Name, err))
```
twz123, line 48: We can exit early if this can't be parsed.

```suggestion
		errs = append(errs, fmt.Errorf("failed to decode worker profile %q: %w", wp.Name, err))
		return errs
```

pkg/apis/k0s/v1beta1/workerprofile.go
```diff
@@ -37,26 +38,29 @@ type WorkerProfile struct {
...
-	}
+func (wp *WorkerProfile) Validate() []error {
 
-	for field := range parsed {
-		if _, found := lockedFields[field]; found {
-			return fmt.Errorf("field `%s` is prohibited to override in worker profile", field)
-		}
+	var errs []error
+
+	kubeletCfg := &kubeletv1beta1.KubeletConfiguration{}
```
twz123, line 46: Might be nice to check if we can also perform semantic config validation like the kubelet does. This is probably worth another PR, though.

pkg/apis/k0s/v1beta1/workerprofile.go
```diff
@@ -37,26 +38,29 @@ type WorkerProfile struct {
...
-}
-
 // Validate validates instance
-func (wp *WorkerProfile) Validate() error {
-	var parsed map[string]any
-	err := json.Unmarshal(wp.Config.Raw, &parsed)
-	if err != nil {
-		return err
-	}
+func (wp *WorkerProfile) Validate() []error {
```
twz123, line 42: We should probably also validate that `wp.Name` is not empty.

twz123: In fact, as the name is used as a template for the ConfigMap name, we should check if it's a valid Kubernetes object name, and that it doesn't exceed a certain length.

twz123:
Created a follow-up for the semantic validation: #6264

juanluisvaladas:
Since there is a follow up PR I'll just go ahead and merge this without more modifications.

PR #6264, opened 2025-08-12
Semantic kubelet config validation for worker profiles

twz123:
## Description

Follow-up to #6241.

Use the kubelet's config validation code to verify the semantic correctness of the worker profiles.

Validate worker profile names. Those are used as part of ConfigMap names and used as label values, so validate them accordingly. Ignore nil configs, so that you don't need to specify an empty object as an empty kubelet config. Use field paths instead of the possibly empty profile names in error messages.

Allow apiVersion and kind in the KubeletConfiguration, so that they don't need to be stripped out when copying otherwise valid KubeletConfigurations into worker profiles. Fail fast and skip field validations if JSON decoding fails, since those validations would be irrelevant or misleading in that case.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,6 +68,7 @@ require (
 	k8s.io/api v0.34.0-beta.0
 	k8s.io/apiextensions-apiserver v0.34.0-beta.0
 	k8s.io/apimachinery v0.34.0-beta.0
+	k8s.io/apiserver v0.34.0-beta.0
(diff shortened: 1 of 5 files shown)


juanluisvaladas:
The failing tests happen because of: https://github.com/k0sproject/k0s/pull/6291

PR #6303, opened 2025-08-20
Add `--context-name` flag to kubeconfig create command

### LINKED ISSUE #6302: Support custom context names in kubeconfig create command

I'm always frustrated when managing multiple k0s clusters because the `k0s kubeconfig create` command always generates kubeconfigs with the same context name "k0s". This causes two main problems:

1. **Naming conflicts**: When trying to merge multiple cluster configurations into a single kubeconfig file, all contexts have the same name
2. **Poor visibility**: When working with multiple k0s clusters, it's difficult to identify which cluster you're accessing since all contexts are named "k0s"

Currently, to work around this limitation, I need to manually edit the generated kubeconfig files using sed commands like:

```bash
sed -i 's/: k0s/: wl1/g' ./.kube/wl1.yaml
sed -i 's/: k0s/: wl2/g' ./.kube/wl2.yaml
KUBECONFIG=./.kube/wl1.yaml:./.kube/wl2.yaml kubectl config view --flatten > ./.kube/merged.yaml
```

This manual step is error-prone and makes automation difficult.

It would be awesome if the `k0s kubeconfig create` command supported a `--context-name` flag that allows customizing the context name and cluster name in the generated kubeconfig.

For example:
```bash
k0s kubeconfig create admin --context-name my-cluster --groups system:masters
```

This should generate a kubeconfig with:
- Cluster name: `my-cluster`
- Context name: `my-cluster`
- User name: `admin`

The flag should be optional, maintaining backward compatibility by defaulting to the current behavior ("k0s") when not specified.

1. **Manual editing with sed/yq**: Current workaround, but error-prone and not automation-friendly
2. **Separate `--cluster-name` and `--context-name` flags**: More granular control but adds unnecessary complexity
3. **Post-processing scripts**: Custom tooling to modify generated kubeconfigs, but requires additional maintenance
4. **Using different usernames to differentiate**: Doesn't solve the core issue of context/cluster naming

### Additional context

This feature would significantly improve multi-cluster management workflows and eliminate the need for manual kubeconfig manipulation. The implementation would need to modify the `createUserKubeconfig` function in `cmd/kubeconfig/create.go` where the `k0sContextName` constant is currently hardcoded.

Benefits include:
- Better cluster identification when using `kubectl config get-contexts`
- Simplified multi-cluster kubeconfig merging
- Improved operational clarity in multi-cluster environments

kahirokunn:
## Description

This PR adds a `--context-name` flag to the `k0s kubeconfig create` command to support custom context and cluster names in generated kubeconfigs. This enhancement enables better multi-cluster management workflows by allowing users to avoid naming conflicts when merging multiple k0s cluster configurations.

The implementation modifies the `createUserKubeconfig` function in `cmd/kubeconfig/create.go` to accept a configurable context name instead of using the hardcoded "k0s" constant. The username remains unchanged from the command argument, preserving existing behavior while adding the requested customization capability.

Fixes #6302

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/cmd/kubeconfig/create.go
+++ b/cmd/kubeconfig/create.go
@@ -23,6 +23,7 @@ func kubeconfigCreateCmd() *cobra.Command {
 	var (
 		groups                  string
 		certificateExpiresAfter time.Duration
+		contextName             string
 	)
 
 	cmd := &cobra.Command{
@@ -38,12 +39,19 @@ Note: A certificate once signed cannot be revoked for a particular user`,
 	$ k0s kubeconfig create username --groups [groups]
 
 	customize the expiration duration of the certificate:
-	$ k0s kubeconfig create username --certificate-expires-after 8760h`,
+	$ k0s kubeconfig create username --certificate-expires-after 8760h
+
+	set custom context name:
+	$ k0s kubeconfig create username --context-name my-cluster`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			username := args[0]
 			if username == "" {
 				return errors.New("username cannot be empty")
+			}
+
+			if contextName == "" {
+				return errors.New("context-name cannot be empty")
 			}
 
 			opts, err := config.GetCmdOpts(cmd)
@@ -56,7 +64,7 @@ Note: A certificate once signed cannot be revoked for a particular user`,
 			}
 			clusterAPIURL := nodeConfig.Spec.API.APIAddressURL()
 
-			kubeconfig, err := createUserKubeconfig(opts.K0sVars, clusterAPIURL, username, groups, certificateExpiresAfter)
+			kubeconfig, err := createUserKubeconfig(opts.K0sVars, clusterAPIURL, username, groups, certificateExpiresAfter, contextName)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


kke (dismissed):
lgtm

kahirokunn:
Thanks for you review 🙏

cmd/kubeconfig/create_test.go
```diff
@@ -44,6 +44,7 @@ func TestKubeconfigCreate(t *testing.T) {
 		"--config", "-",
 		"--data-dir", k0sVars.DataDir,
 		"kubeconfig", "create", "test-user",
+		"--context-name", "my-cluster",
```
twz123, line 47: Can we add an assertion that this is the current context in the generated kubeconfig?

cmd/kubeconfig/create.go
```diff
@@ -23,6 +23,7 @@ func kubeconfigCreateCmd() *cobra.Command {
 	var (
 		groups                  string
 		certificateExpiresAfter time.Duration
+		contextName             string
```
twz123, line 26: Are there any constraints on the context name that need to be checked, such as length or forbidden characters? At the very least, we should error out on an empty context name.

kahirokunn: No forbidden characters. I’ve added validation to require at least one character (same as username), so an empty context name now errors.

cmd/kubeconfig/create_test.go
kahirokunn: Added. Thanks 🙏

kahirokunn:
@twz123 Thank you! When you have a moment, I’d really appreciate another review.

juanluisvaladas:
LGTM too

twz123 (approved):
Thanks 🙏

PR #6311, opened 2025-08-20
Don't kill processes when cleaning up PID files

twz123:
## Description

It's better to leave the decision what to do about a lingering process to the users. Try to terminate the process gracefully once and poll for process termination. If the process didn't terminate in time, bail out with an error.

Re-implement the Supervisor proc handle without the stdlib. The os/exec API is not a good fit for implementing the PID file cleanup. Instead, use a combination of dirfds to `/proc/<pid>` directories, along with `pidfd_send_signal` (if available) to implement a race-free process termination.

Don't misuse `shouldKillProcess` to check if a process has terminated or not. Rename it to `isK0sManaged` and add appropriate docs. Instead, introduce a purpose-built `hasTerminated` function.

Split up Linux and macOS process handling. There is no proc filesystem on macOS. Split things up and leave a minimal implementation in place to reflect this fact. This allows most of the unit tests to run on macOS while making it clear that proc stuff doesn't exist on macOS.

Also enable the bogus PID file test unconditionally. This is just about the PID file parsing, which doesn't require platform-specific APIs.

See:

* https://github.com/k0sproject/k0s/pull/6203#issuecomment-3206939981

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/supervisor/prochandle_linux.go
+++ b/pkg/supervisor/prochandle_linux.go
@@ -4,146 +4,103 @@
 package supervisor
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
-	"strconv"
-	"strings"
 	"syscall"
+
+	"github.com/k0sproject/k0s/internal/os/linux"
+	"github.com/k0sproject/k0s/internal/os/linux/procfs"
+	osunix "github.com/k0sproject/k0s/internal/os/unix"
 )
 
 type unixProcess struct {
-	// The PID that was used when opening the process.
-	// Note: Don't rely on [os.Process.Pid] here, as it's not thread safe.
-	pid     int
-	process *os.Process
+	pid    int
+	pidDir *osunix.Dir
 }
 
-func openPID(pid int) (_ procHandle, err error) {
-	var process *os.Process
-	process, err = os.FindProcess(pid)
+func openPID(pid int) (_ *unixProcess, err error) {
+	p := &unixProcess{pid: pid}
+	p.pidDir, err = procfs.OpenPID(pid)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, syscall.ESRCH
+		}
 		return nil, err
 	}
 	defer func() {
 		if err != nil {
-			err = errors.Join(err, process.Release())
+			err = errors.Join(err, p.Close())
 		}
 	}()
 
-	// FindProcess won't return an error if no such process exists.
-	// Be fail-fast and check it directly by sending "the null signal".
-	// https://www.man7.org/linux/man-pages/man3/kill.3p.html
-	if err := process.Signal(syscall.Signal(0)); err != nil {
-		if errors.Is(err, os.ErrProcessDone) {
-			return nil, syscall.ESRCH
-		}
+	// The dir is open. It might refer to a thread, though.
+	// Check if the thread group ID is the process ID.
+	if status, err := p.dir().Status(); err != nil {
+		return nil, err
+	} else if tgid, err := status.ThreadGroupID(); err != nil {
+		return nil, fmt.Errorf("failed to get thread group ID: %w", err)
+	} else if tgid != pid {
+		return nil, fmt.Errorf("%w (thread group ID is %d)", syscall.ESRCH, tgid)
 	}
 
-	return &unixProcess{pid, process}, nil
+	return p, nil
 }
 
 func (p *unixProcess) Close() error {
-	return p.process.Release()
+	return p.pidDir.Close()
 }
 
 func (p *unixProcess) hasTerminated() (bool, error) {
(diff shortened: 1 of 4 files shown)


pkg/supervisor/prochandle_linux.go
```diff
@@ -83,3 +123,27 @@ func (p *unixProcess) requestGracefulTermination() error {
 func (p *unixProcess) kill() error {
 	return p.process.Kill()
 }
+
+func isZombie(pid int) (bool, error) {
+	// https://man7.org/linux/man-pages/man5/proc_pid_stat.5.html
```
juanluisvaladas: Nit: Maybe it's worth adding another comment with the relevant kernel source (I had to search it to understand the fields and all)

https://github.com/torvalds/linux/blob/v4.19/fs/proc/array.c#L152

I picked 4.19 so that it's not a very recent version and people reading it don't think it's a new feature but it doesn't really matter.

juanluisvaladas (commented):
I'm fine with the first 5 commits in their current state but it's not clear to me why it's better to bail out instead of just terminating the process.

I can't think of a scenario where killing the process is a bad idea, I guess I'm missing something?

pkg/supervisor/prochandle_linux.go
juanluisvaladas: It's also worth noting that we already depend indirectly on `github.com/prometheus/procfs`, which handles this, but actually I prefer this.

twz123:
> I'm fine with the first 5 commits in their current state but it's not clear to me why it's better to bail out instead of just terminating the process.
> 
> I can't think of a scenario where killing the process is a bad idea, I guess I'm missing something?

During review of #6203 @ncopa raised concerns about killing processes. He would prefer to leave that decision to the users. There's always the chance to break something when killing processes, e.g., say, etcd while it's doing stuff on disk.

pkg/supervisor/prochandle_linux.go
twz123: > Nit: Maybe it's worth adding another comment with the relevant kernel source (I had to search it to understand the fields and all)

TBH, I found the man pages to be very clear and approachable for this. They mention field names, their order, their sprintf format directive, and a comprehensive description. I tried to pick that up in the comments below: "pid" and "comm" are the first two fields, "state" is the third one. Maybe we can make those comments more clear?

twz123: > we already depend indirectly on github.com/prometheus/procfs

Yeah, I think there's quite a few indirect dependencies dealing with procfs. I think containerd might have something into that direction, too. But I think they're all coming from a different angle to the problem: Prometheus is about monitoring, i.e. gathering data in a bunch, containerd, again, might be more concerned about security aspects when querying procfs data.

pkg/supervisor/prochandle_unix.go
```diff
@@ -22,11 +22,26 @@ type unixProcess struct {
...
+	}()
+
+	// FindProcess won't return an error if no such process exists.
+	// Be fail-fast and check it directly by sending "the null signal".
+	// https://www.man7.org/linux/man-pages/man3/kill.3p.html
+	if err := process.Signal(syscall.Signal(0)); err != nil {
+		if errors.Is(err, os.ErrProcessDone) {
+			return nil, syscall.ESRCH
+		}
+	}
```
ncopa: This test should be removed. The process can disappear right after the test is performed and before the return of function, so we need to handle the case of missing process late. No need to handle it twice.

pkg/supervisor/prochandle_linux.go
```diff
@@ -48,6 +49,45 @@ func (p *unixProcess) Close() error {
...
+
+	// Instead, rely on the /proc filesystem once again, and check if the
+	// process is a zombie.
+	zombie, err := isZombie(p.pid)
+	if err != nil {
+		return false, err
+	}
+
+	// Do a last TOCTOU check: We now know the state of the process with the
+	// referenced PID. We go back to os.Process, once again, to ensure that
```
ncopa: We actually don't know the state of the process at this point. The state may have changed after we did the test(s). 
For that reason I think the entire `hasTerminated` function should be removed.

ncopa:
> Don't misuse shouldKillProcess to check if a process has terminated or not. Rename it to isK0sManaged and add appropriate docs. Instead, introduce a purpose-built hasTerminated function.

This does not work. We can not rely on a `hasTerminated` function because the process state can change at any time. The process can disappear while we check if it is k0s managed. I also think that the `isZombie` is meaningless, as the process state can change after the test is done.

ncopa:
> The os package uses ErrProcessDone ("os: process already finished"), not syscall.ESRCH ("no such process").

Good catch! Seems to be since go 1.4.

ncopa (approved):
LGTM

PR #6325, opened 2025-08-26
Allow configuring "docker" and additional container runtime args in Makefiles

ashjones-replicated:
## Description
This enables contributors to lint, test and build their changes when using container runtimes other than docker (i.e podman).

To allow the containerised sections of this Makefile to operate under a podman system with selinux enabled (i.e RHEL / fedora) the additional options can be added by a developer to $DOCKER_OPTS before running `make`:

```bash
export DOCKER=podman
export DOCKER_RUN_OPTS="--security-opt label=disable --userns keep-id"
```

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -91,7 +91,7 @@ clean: clean-k0s
 	rm -rf ../site
 	for i in .docker-image.*.stamp; do \
 	  if [ -f "$$i" ]; then \
-	    docker rmi -f "k0sdocs$$(basename "$$i" .stamp)" && rm -- "$$i";\
+	    $(DOCKER) rmi -f "k0sdocs$$(basename "$$i" .stamp)" && rm -- "$$i";\
 	  fi; \
 	done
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -35,6 +35,7 @@ ifeq ($(DOCKER),)
 else
   GO_ENV_REQUISITES ?= .k0sbuild.docker-image.k0s
   GO_ENV ?= $(DOCKER) run --rm \
+  	$(DOCKER_OPTS) \
```
twz123: What about calling this `DOCKER_RUN_OPTS` and move this to line 47, just before the run args, so that one can override all the hard-coded flags, if needed?

Also, I think it makes sense to add this to all "$(DOCKER) run" invocations in the Makefiles, so it works consistently across all targets?

hedge-sparrow: updated any hardcoded references to `docker` while I was at it. 

only thing I can foresee that might not work perfectly is the one call to `docker buildx`, podman _does_ have a `buildx` argument but I've not used it before and can't attest to it's feature parity.

twz123: Yeah, I mean, it's best effort. One thing at a time ^^

twz123 (approved):

PR #6343, opened 2025-09-01
Bump Kubernetes to v1.34.0

juanluisvaladas:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.34.md#changelog-since-v1330


juanluisvaladas:
Not quite sure what was fixed between 1.34.0-rc.1 and now that makes the etcdmember pass again, but everything passed first try.

PR #6363, opened 2025-09-04
Don't use arm64 self-hosted runners in go.yml workflows

juanluisvaladas:
## Description

We're still using self hosted arm64 runners for the release, and we still have armV7 self hosted runners. I want to try to run the armV7 in github hosted runners but for now this should ease the workload in our runners and improve the release speed and the overall CI speed.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/constant/constant.go
+++ b/pkg/constant/constant.go
@@ -52,7 +52,7 @@ const (
 	ApiserverUser = "kube-apiserver"
 	// SchedulerUser defines the user to use for running k8s scheduler
 	SchedulerUser = "kube-scheduler"
-	// KonnectivityServerUser deinfes the user to use for konnectivity-server
+	// KonnectivityServerUser defines the user to use for konnectivity-server
 	KonnectivityServerUser = "konnectivity-server"
 	// KeepalivedUser defines the user to use for running keepalived
 	KeepalivedUser = "keepalived"


.github/workflows/go.yml
```diff
@@ -222,8 +237,11 @@ jobs:
       fail-fast: false
       matrix:
         smoke-suite: ${{ fromJson(needs.prepare.outputs.smoketest-matrix) }}
+        target-arch:
+          - amd64
+          - arm64
 
-    name: "Smoke test :: ${{ matrix.smoke-suite }}"
+    name: "Smoke test :: ${{ matrix.smoke-suite }} :: ${{ matrix.target-arch }}"
```
juanluisvaladas: Does it sense to run EVERY smoke test on both arm64 and amd64? This nearly duplicates our checks. Perhaps it's more logical to leave just a small subset like we do for armV7.

jnummelin: Didn't we run some basic tests only on arm64 before?

IMO it does not make sense to run every single smoke on both, some subset feels fine to me

juanluisvaladas: On arm and arm64 we checked `check-basic`, `check-airgap` and `check-calico`.

I changed it for `check-basic`, `check-airgap`, `check-network-conformance-calico` and `check-network-conformance-kuberouter`, which I believe are more reasonable checks. It increases the amount of tests in one but I think that's OK.

jnummelin: SGTM

jnummelin (approved):

twz123:
One downside of this is that we won't notice any problems with our arm64 runners until we cut the RCs ...

twz123:
I was somehow thinking that GitHub would offer hosted arm64 runners now, but this is something else, actually. What about changing the release workflow to do the same? We should try to use GitHub runners there, as well, as good as we can, at least.

twz123:
@juanluisvaladas Are you actually sure that we're testing arm64 at all? AFAICT this just runs the usual amd64 build, but relabels it to arm64?

PR #6369, opened 2025-09-05
Allow to configure CPLB interfaces with mac addr

juanluisvaladas:
## Description

Allow to configure CPLB using interface mac address instead of interface name.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist


pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -131,12 +152,36 @@ func (k *KeepalivedSpec) validateVRRPInstances(getDefaultNICFn func() (string, e
...
+			if k.VRRPInstances[i].Interface == "" {
+				iface, err := getDefaultNICFn()
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to get default NIC: %w", err))
+				}
+				k.VRRPInstances[i].Interface = iface
+			}
+		case VRRPInterfaceTypeMAC:
+			if len(k.VRRPInstances[i].InterfaceMACAddresses) == 0 {
+				errs = append(errs, errors.New("InterfaceMACAddresses must be defined when InterfaceType is mac"))
```
juanluisvaladas: TODO: The PR is large already, so I want to deal with this in a separate PR,  but the validation could be better, we should use `k8s.io/apimachinery/pkg/util/validation/field` like we do in `workerprofile.go`

Also we could do better error checking in the test for this function, again, we should improve it like we do in `workerprofile.go`...

ncopa:
What is the reasoning to support multiple MAC addresses?

If we could get away with a single, we could maybe let the `interface` be either an interface name *or* a MAC address.
```yaml
  interface: 00:00:5E:00:53:00 
```

juanluisvaladas:
> What is the reasoning to support multiple MAC addresses?

It was requested by k0rdent people so that they could have the same configuration for all the control plane nodes. If they have 3 control plane nodes with MAC addresses X, Y and Z, they'll define X, Y and Z in the configuration file and only the correct value will be used. Apparently providing different configuration per control plane node is major change for them.

> If we could get away with a single, we could maybe let the interface be either an interface name or a MAC address.

That was actually the initial implementation.

docs/configuration.md
```diff
@@ -348,7 +348,9 @@ Configuration options required for using VRRP to configure VIPs in control plane
...
+| `interfaceType`         | The type of the interface. If not specified, k0s will use `interface`. Valid values are `interface` and `mac`. If `interfaceType` is `mac`, k0s will use the field `interfaceMACAddresses`, otherwise it will use the field `interface`. Default: `interface`.                                       |
 | `interface`             | The NIC used by the virtual router. If not specified, k0s will use the interface that owns the default route.                                                                                                                                                                                        |
+| `interfaceMACAddresses` | A list of MAC addresses for the interface. MAC addresses are evaluated in order and the first match will be chosen. If none of the MAC addresses match, an error is returned and k0s will not start. This field is only used when `interfaceType` is set to `mac`.                                   |
```
twz123: If we have two separate fields for this, we don't need the type field IMO in this case. We can make it either interface or interfaceMACAddresses. I reckon this will save some code/docs.

pkg/apis/k0s/v1beta1/cplb.go
twz123: I'd prefer to use those errors right away. We'll never get around to refactor this once this lands. We can just omit the proper field path propagation for now, and use something like this?


```suggestion
				path := field.NewPath("interfaceMACAddresses")
				err := field.Required(path, `must be defined when interfaceType is "mac"`)
				errs = append(errs, err)
```

Also, it's a good practice to use the JSON field names in the error messages, as this is what users will be facing.

pkg/apis/k0s/v1beta1/cplb.go
```diff
@@ -131,12 +152,36 @@ func (k *KeepalivedSpec) validateVRRPInstances(getDefaultNICFn func() (string, e
...
+			}
+			if k.VRRPInstances[i].Interface == "" {
+				iface, err := getDefaultNICFn()
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to get default NIC: %w", err))
+				}
+				k.VRRPInstances[i].Interface = iface
+			}
+		case VRRPInterfaceTypeMAC:
+			if len(k.VRRPInstances[i].InterfaceMACAddresses) == 0 {
```
twz123: We might want to distinguish `nil` from an empty slice. Would be different errors (required vs. invalid).

twz123:
> Apparently providing different configuration per control plane node is major change for them.

I don't think this is a good argument for allowing multiple interfaces. Can we come up with some reasoning that makes sense for k0s itself to support MAC lists for a single node?

twz123 (commented):
WDYT about removing the type field?

PR #6420, opened 2025-09-22
Don't block fsnotify watcher during stack application

twz123:
## Description

Decouple the stack application from the fsnotify watcher loop. This prevents the loop from stalling due to a long-running application and reduces the risk of the operating system buffer getting overflowed. Trigger applications via a separate channel with a buffer size of one item. This naturally implements debouncing.

Fixes a panic that could occur when the Run method exits before the initial apply event has been sent from the separate goroutine.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/applier/stackapplier.go
+++ b/pkg/applier/stackapplier.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/avast/retry-go"
-	"github.com/k0sproject/k0s/pkg/debounce"
 	"github.com/k0sproject/k0s/pkg/kubernetes"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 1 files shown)


pkg/applier/stackapplier.go
```diff
@@ -55,57 +54,63 @@ func (s *StackApplier) Run(ctx context.Context) error {
...
+		defer close(trigger)
+		defer func() {
+			if err := watcher.Close(); err != nil {
+				s.log.WithError(err).Error("Failed to close watcher")
 			}
-			s.log.WithError(err).Error("Error while watching stack")
-		}
+		}()
+		err = s.runWatcher(watcher, trigger, ctx.Done())
 	}()
```
jaitaiwan: This part seems like unnecessary complication; `s.log` is available inside `s.runWatcher` so why not just do the defer and log the error inside `s.runWatcher` and just `go s.runWatcher`?

twz123: Good question. Maybe I felt that the function which created the watcher should close it, instead of delegating the ownership to runWatcher. Anyhow, I've pushed the defers down the line. PTAL!

pkg/applier/stackapplier.go
```diff
@@ -55,57 +55,60 @@ func (s *StackApplier) Run(ctx context.Context) error {
...
-	// Consume and log any errors.
 	go func() {
-		for {
-			err, ok := <-watcher.Errors
-			if !ok {
-				return
-			}
-			s.log.WithError(err).Error("Error while watching stack")
-		}
+		err = s.runWatcher(watcher, trigger, ctx.Done())
```
jaitaiwan: I don’t think `err =` is needed now

twz123: It is, if we want to forward any errors that might occur when running the watcher to the caller. But: we have to close the trigger channel _after_ assigning to `err` here. Therefore, at least that defer must be pulled one level up from `runWatcher`.

jaitaiwan: So I did a bit of pontificating, I think this should still pass the unit test but will make it a bit clearer what's happening with errors:

```
// Run executes the initial apply and watches the stack for updates.
func (s *StackApplier) Run(ctx context.Context) error {
	if ctx.Err() != nil {
		return nil // The context is already done.
	}

	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		return fmt.Errorf("failed to create watcher: %w", err)
	}

	errCh := make(chan error)
	go s.runWatcher(watcher, errCh, ctx)

	if addErr := watcher.Add(s.path); addErr != nil {
		return fmt.Errorf("failed to watch %q: %w", s.path, addErr)
	}

	for err := range errCh {
		return err
	}

	return nil
}

func (s *StackApplier) runWatcher(watcher *fsnotify.Watcher, errCh chan error, ctx context.Context) (err error) {
	defer func() { errCh <- errors.Join(err, watcher.Close()) }()

	const timeout = 1 * time.Second // debounce events for one second
	timer := time.NewTimer(timeout)
	defer timer.Stop()

	for {
		select {
		case err := <-watcher.Errors:
			return fmt.Errorf("while watching stack: %w", err)

		case event := <-watcher.Events:
			// Only consider events on manifest files
			if match, _ := filepath.Match(manifestFilePattern, filepath.Base(event.Name)); !match {
				continue
			}
			timer.Reset(timeout)

		case <-timer.C:
			s.apply(ctx)

		case <-ctx.Done():
			return nil
		}
	}
}
```

jaitaiwan: Not sure if the errCh would need to be buffered and have it closed in the defer of the runWatcher function though.

twz123: The main purpose of this PR is to decouple the stack application from the watcher loop, to ensure that long-running stack applications do not create backpressure into the kernel inotify event queue. While I agree your proposed change is more straight forward, it's quite important to not block the event consumption.

jaitaiwan: That's a good point, I went a bit far in my simplification. I think if you used the channel approach but kept the triggers that would make the non-obvious coupling between when to close the triggers and the `err = ` being populated by the time the program returns.

jaitaiwan (approved):
LGTM

PR #6427, opened 2025-09-23
More Renovate version bumps

twz123:
## Description

* Enable Kubernetes updates, including pre-releases. Select the "right" version number to be included in the commit message and PR title.
* Enable containerd updates.
* Change the commit message action to "Bump". That's how Dependabot does it. It's shorter, too. Rename everything "update" to "bump" as well. Drop the "dependency" prefix from commit messages. Unify the groupSlugs, as well.
* Bump Go and kine versions via comments instead of custom regexes
* Have a blank update regex for all GitHub source code links, and add some ignored dependencies for them.
* Group Calico version bumps. We can't bump calico automatically yet, but this way we get at least a single PR reminding us about a missing update.
* Disable semantic commits. RenovateBot will otherwise try to auto-detect this.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -16,6 +16,8 @@
     "/^release-1\\.3[1-4]$/"
   ],
   "prTitleStrict": true,
+  "commitMessageAction": "Bump",
+  "commitMessageTopic": "{{depName}}",
   "enabledManagers": [
     "gomod",
     "custom.regex"
@@ -84,7 +86,7 @@
         "**/kube-proxy"
       ],
       "groupName": "Kubernetes",
-      "groupSlug": "kubernetes-all",
+      "groupSlug": "kubernetes",
       "commitMessageExtra": "{{~#each upgrades}}{{~#if (equals depName \"k8s.io/kubernetes\")}} to {{{newVersion}}}{{/if}}{{/each~}}",
       "ignoreUnstable": false,
       "respectLatest": false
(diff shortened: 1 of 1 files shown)


twz123:
Blocked by

* #6424 
* #6426

renovate.json
```diff
@@ -30,6 +30,48 @@
         "gomod"
       ]
     },
+    {
+      "description": "Group all Kubernetes updates",
+      "enabled": true,
+      "matchPackageNames": [
+        "k8s.io/api",
```
juanluisvaladas, line 47: I don't understand how both 0.X.Y and 1.X.Y work, but if it gets the job done it's welcome

twz123: From Renovate's point of view, these are different dependencies. The k8s.io/* dependencies from the go datasource have the version 0.xx, the k8s.io/kubernetes dependency has 1.xx. So Renovate will simply bump all of them to the newest version. This only works out because Kubernetes releases everything together at the same time. One disadvantage is that the PR title won't contain the Kubernrnetes version number, as Renovate will only include a version number in the title if it's unambiguous, but we have two.

I'm still investigating. I noticed this afternoon that some bits are still missing.

twz123: Alright. I've put more wort into this, and now I'm quite confident that it works. Can you please take another look @juanluisvaladas?

juanluisvaladas (approved):

PR #6464, opened 2025-10-02
Run containerd as a supervised process on Windows

twz123:
## Description

This simplifies the component code quite a bit, and removes the PowerShell dependency. This way, containerd on Windows is no different than on Linux, modulo the following:

- There are no signals on Windows, so the `SIGHUP` trick that tells containerd to reload its configuration won't work. In this case, fall back to a full stop/start cycle.

- There's a bug in containerd 1.x on Windows that prevents it from properly responding to Ctrl+C and Ctrl+Break events. Apply the 2.x one-liner patch to the 1.x codebase so the Supervisor can stop it.

- On Windows, containerd requires the special `SeAssignPrimaryTokenPrivilege`, which allows processes to logon any user without knowing the password. This privilege is not granted to administrator accounts, it is only granted to the built-in SYSTEM and LOCAL SERVICE accounts. Since containerd will inherit k0s's privileges, it can only function properly if k0s itself has this privilege. Therefore, k0s must be run as a service. Consequently, running a k0s worker in a terminal with a standard administrator account will result in containerd throwing permission errors when running containers. This is an acceptable trade-off IMO, as any real setup will involve using Windows services anyway. For quick manual testing, dealing with the container errors is acceptable.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist


makhov (approved):

pkg/supervisor/supervisor.go
```diff
@@ -82,15 +82,20 @@ func (s *Supervisor) processWaitQuit(ctx context.Context) bool {
...
+				}
 				return true
 			}
 		}
 	case err := <-waitresult:
 		if err != nil {
 			s.log.WithError(err).Warn("Failed to wait for process")
 		} else {
-			s.log.Warnf("Process exited: %s", s.cmd.ProcessState)
+			s.log.Warnf("Process exited: ", s.cmd.ProcessState)
```
twz123, line 98: Oops. This should have been:

```suggestion
			s.log.Warn("Process exited: ", s.cmd.ProcessState)
```

twz123: Don't we have some linters for this? :thinking:

PR #6472, opened 2025-10-03
Don't unmount kubelet-root-dir on k0s reset

dockerpac:
## Description

Fixes an issue where `k0s reset` was unmounting and deleting the kubelet-root-dir mount point when configured on a separate filesystem.

When `--kubelet-root-dir` is on a dedicated partition (common in production), `k0s reset` would unmount the filesystem and delete the mount point.

This PR applies the same mount point preservation logic to `kubelet-root-dir` that was already existing for `data-dir`.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

Tested scenarios:
- **Mounted filesystem**: Kubelet mount point preserved, only contents deleted
- **Default setup**: Directory completely removed as before

## Checklist

### DIFF
--- a/inttest/reset/reset_test.go
+++ b/inttest/reset/reset_test.go
@@ -13,23 +13,45 @@ import (
 
 	testifysuite "github.com/stretchr/testify/suite"
 
+	"github.com/k0sproject/bootloose/pkg/config"
 	"github.com/k0sproject/k0s/inttest/common"
 )
 
+const dataDir = "/var/lib/k0s"
+
 type suite struct {
 	common.BootlooseSuite
+
+	// kubeletRootDir is the directory passed to the worker and to k0s reset
+	// via --kubelet-root-dir.
+	kubeletRootDir string
 }
 
 //go:embed clutter-data-dir.sh
 var clutterScript []byte
 
+// kubeletRootDirSeparate reports whether the kubelet root dir lives outside the
+// data dir. A separate kubelet root dir is its own mount point that reset must
+// keep in place (only emptying it); one under the data dir gets removed
+// together with the data dir.
+func (s *suite) kubeletRootDirSeparate() bool {
+	return !strings.HasPrefix(s.kubeletRootDir, dataDir+"/")
+}
+
 func (s *suite) TestReset() {
 	ctx := s.Context()
 	workerNode := s.WorkerNode(0)
+	separate := s.kubeletRootDirSeparate()
+
+	// A mount nested under the kubelet root dir, and the source it's bound
+	// from. Reset must unmount and remove the nested mount, but leave the
+	// (out-of-tree) bind source alone.
+	nestedMount := s.kubeletRootDir + "/k0s_reset_inttest_nested_mount"
+	nestedMountSource := "/k0s_reset_inttest_nested_source"
 
 	if !s.Run("k0s gets up", func() {
 		s.Require().NoError(s.InitController(0, "--disable-components=konnectivity-server,metrics-server"))
-		s.Require().NoError(s.RunWorkers("--kubelet-root-dir=/var/lib/kubelet"))
+		s.Require().NoError(s.RunWorkers("--kubelet-root-dir=" + s.kubeletRootDir))
 
 		kc, err := s.KubeClient(s.ControllerNode(0))
 		s.Require().NoError(err)
@@ -44,9 +66,9 @@ func (s *suite) TestReset() {
 		s.Require().NoError(err)
 		defer ssh.Disconnect()
 
-		s.NoError(ssh.Exec(ctx, "test -d /var/lib/k0s", common.SSHStreams{}), "/var/lib/k0s is not a directory")
+		s.NoError(ssh.Exec(ctx, "test -d "+dataDir, common.SSHStreams{}), "%s is not a directory", dataDir)
 		s.NoError(ssh.Exec(ctx, "test -d /run/k0s", common.SSHStreams{}), "/run/k0s is not a directory")
-		s.NoError(ssh.Exec(ctx, "test -d /var/lib/kubelet", common.SSHStreams{}), "/var/lib/kubelet is not a directory")
+		s.NoError(ssh.Exec(ctx, "test -d "+s.kubeletRootDir, common.SSHStreams{}), "%s is not a directory", s.kubeletRootDir)
 
 		s.NoError(ssh.Exec(ctx, "pidof containerd-shim-runc-v2 >&2", common.SSHStreams{}), "Expected some running containerd shims")
 	}) {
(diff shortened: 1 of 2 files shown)


jnummelin:
@ncopa You were involved designing the `data-dir` reset/unmount logic, could you review this? 🙏

ncopa:
The `make check-reset` test fails:

```
        	Test:       	TestResetSuite/TestReset/k0s_reset
        	Messages:   	/var/lib/kubelet still exists
```

dockerpac:
> The `make check-reset` test fails:
> 
> ```
>         	Test:       	TestResetSuite/TestReset/k0s_reset
>         	Messages:   	/var/lib/kubelet still exists
> ```

I have done manual tests on a real linux box (Rocky Linux with systemd) and the reset works correctly: 
- if `/var/lib/kubelet` is just a directory --> it gets deleted
- if `/var/lib/kubelet` is a user-mounted filesystem --> contents are deleted but the mount point is preserved

When running the bootloose environment manually, I noticed that kubelet creates an expected mount: 
```
# before kubelet is started (this is expected)
worker0:/# df -h /var/lib/kubelet
df: /var/lib/kubelet: No such file or directory

# after kubelet is started (unexpected mount appears)
worker0:/# df -h /var/lib/kubelet
Filesystem      Size  Used Avail Use% Mounted on
overlay         119G   31G   88G  27% /var/lib/kubelet

worker0:/# cat /proc/self/mountinfo | grep /var/lib/kubelet
647 686 0:47 /var/lib/kubelet /var/lib/kubelet rw,relatime shared:272 - overlay overlay rw,seclabel,lowerdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/339/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/216/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/215/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/214/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/213/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/212/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/211/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/210/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/209/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/208/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/207/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/206/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/205/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/204/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/203/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/202/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/201/fs:/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/3/fs,upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/340/fs,workdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/340/work
```

This mount does not appear on a real Linux system, so it is probably something specific to the containerized environment that doesn't reflect real-world scenarios.

twz123:
> When running the bootloose environment manually, I noticed that kubelet creates an expected mount:

I wouldn't know that kubelet is creating a mount for its data directory. This definitely needs some investigation, and to me, the integration test failures look legit, still.

vikramhh:
@dockerpac @ncopa @twz123 

By using bpftrace, it can be confirmed that the mounting of the configured value of kubelet root directory is done by kubelet itself. The failure above looks legit and arises not due to containerized environment per-se but likely due to the following sequence of calls in kubelet:

```
Kubelet.setupDataDirs()
  -> hostutil.MakeRShared(rootDir)
    -> DoMakeRShared(rootDir, /proc/self/mountinfo)
      -> if not already shared:
           bind-mount rootDir onto itself
           make it recursively shared
```

The above is the kubelet:

1. Needing its root dir mount tree to support mount propagation
2. Checking whether the relevant mount is already shared
3. If not, it making the kubelet root dir a mountpoint by doing a self bind mount

By doing the following before installing k0s, you can force this behavior on a "real linux box" as well - it was confirmed on aws ec2 ubuntu[ the following assumes that `/var/lib/kubelet` is the configured value of kubelet root directory]:

```
sudo mkdir -p /var/lib/kubelet
sudo mount --make-private /

```

The reverse (making the root non-private/shared on bootloose) was not tried to make the failure go away because IMO in this instance it is the test that needs to be modified for reasons like scoping things tightly and because the "failure" actually reflects what should be happening here with this PR-s changes. Besides bootloose runs the node as a container and it would be somewhat unexpected for the container's root to have shared propagation.

If you concur, I can push the test changes otherwise could you suggest what else we could try?

jnummelin:
> Besides bootloose runs the node as a container and it would be somewhat unexpected for the container's root to have shared propagation

We mitigate this in bootloose based tests by mounting `/var/lib/k0s` as a volume. So the tests involved with custom kubelet dir could maybe do the same by attaching an extra volume for the given dir. Something like:

```
func TestXYZSuite(t *testing.T) {
	s := XYZSuite{
		common.BootlooseSuite{
			ControllerCount: 1,
			WorkerCount:     1,
			ExtraVolumes: []config.Volume{
				{
					Type:        "volume",
					Destination: "/var/lib/kubelet",
				},
			},
		},
	}
	suite.Run(t, &s)
}
```

vikramhh:
Hi @jnummelin - the CI workflows are awaiting your approval from the fork. Thanks!

vikramhh:
Hi @jnummelin - could you please approval a workflow run for the latest push. Thanks!

vikramhh:
@twz123 @makhov 

The two failing checks — `check-dualstack-kuberouter` and `check-dualstack-kuberouter-dynamicconfig` — fail on the same assertion, and it's unrelated to this PR:

`dualstack_test.go:252: pod metrics-server-d987bf784-xxxxx container metrics-server has 1 restarts. Should be 0
`

This is `VerifyNoRestartedPods` tripping because the metrics-server container restarted once during cluster bring-up — a transient flake. (The No agent available lines earlier in the log are just retries that subsequently succeed — "Connection to ipv4 address ... was successful" — not the cause.)

This PR only touches `pkg/cleanup/directories.go` (the k0s reset path) and `inttest/reset/reset_test.go`. The dualstack suite never runs k0s reset, so the changed code path isn't exercised by these tests at all.

The same `metrics-server-restart` failure has recurred on several unrelated PRs over the last few days — dependency bumps and networking changes that can't be affected by reset/cleanup logic — which rules this change out as the cause.

Could you please re-run the two failed jobs? Everything else (108 checks) is green. Thanks!

(There doesn't appear to be a tracking issue for this particular dualstack flake yet — happy to open one if that's useful.)

vikramhh:
Hey @twz123 @makhov @jnummelin - tests runs are all green, could you please take a look.

jnummelin (approved):

PR #6478, opened 2025-10-07
Dissolve Windows stack

twz123:
## Description

Instead of deploying the necessary manifests for Windows workers in one stack, incorporate them into the relevant stacks: calico and kubeproxy. This allows for reusing the template values of each stack without duplicating the logic. The old Windows stack component is now solely responsible for counting Windows nodes. The count is broadcasted to the kubeproxy and calico components, which update the manifests on disk if necessary.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist


static/manifests/windows/DaemonSet/kube-proxy-windows.yaml
```diff
@@ -21,11 +22,11 @@ spec:
           runAsUserName: "NT AUTHORITY\\system"
       hostNetwork: true
       containers:
-      - image: "{{ .KubeProxyImage }}:{{ .KubeProxyVersion }}"
+      - name: kube-proxy
```
juanluisvaladas, line 25: I know it wasn't introduced here, but I think the right the label should be `k8s-app: kube-proxy-windows`so that it matches with the pod labels.

twz123: That's actually coming in the follow-up PR :ok_hand:

twz123: Ah, you mean the container name, or the label? I think the container name should be "kube-proxy", actually

PR #6507, opened 2025-10-10
CIS bechmark 1.11: change file permissions to 0600 for admin.conf and kubelet config.yaml

sakshisharma84:
## Description

This PR updates file permissions to be more restrictive (0600) as required by the [CIS Kubernetes Benchmark (v1.11)](https://github.com/aquasecurity/kube-bench/tree/main/cfg/cis-1.11):
- /var/lib/k0s/pki/admin.conf (CIS 1.1.13)
- /run/k0s/kubelet/config.yaml (CIS 4.1.9)

Fixes # (issue)
CIS benchmark control IDs
- **1.1.13** – /var/lib/k0s/pki/admin.conf → currently 640 root:root (CIS requires 600)
- **4.1.9** – /run/k0s/kubelet/config.yaml → currently 644 root:root (CIS requires 600)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

### Testing details:

- Built k0s locally and installed with sudo k0s install controller --single.
- Verified k0scontroller service is running with all components
- Checked file permissions:

1) /var/lib/k0s/pki/admin.conf → 600 root:root
```
sudo stat -c "%a %U:%G %n" /var/lib/k0s/pki/admin.conf
600 root:root /var/lib/k0s/pki/admin.conf
```
2) /run/k0s/kubelet/config.yaml → 600 root:root
```
sudo stat -c "%a %U:%G %n" /run/k0s/kubelet/config.yaml
600 root:root /run/k0s/kubelet/config.yaml
```

## Checklist

### DIFF
--- a/cmd/controller/certificates.go
+++ b/cmd/controller/certificates.go
@@ -95,7 +95,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 			return err
 		}
 
-		if err := kubeConfig(c.K0sVars.AdminKubeConfigPath, kubeConfigAPIUrl, c.CACert, adminCert.Cert, adminCert.Key, users.RootUID); err != nil {
+		if err := kubeConfig(c.K0sVars.AdminKubeConfigPath, kubeConfigAPIUrl, c.CACert, adminCert.Cert, adminCert.Key, users.RootUID, constant.OwnerOnlyMode); err != nil {
 			return err
 		}
 
@@ -124,7 +124,7 @@ func (c *Certificates) Init(ctx context.Context) error {
 			return err
 		}
 
-		return kubeConfig(c.K0sVars.KonnectivityKubeConfigPath, kubeConfigAPIUrl, c.CACert, konnectivityCert.Cert, konnectivityCert.Key, uid)
+		return kubeConfig(c.K0sVars.KonnectivityKubeConfigPath, kubeConfigAPIUrl, c.CACert, konnectivityCert.Cert, konnectivityCert.Key, uid, constant.CertSecureMode)
 	})
 
 	eg.Go(func() error {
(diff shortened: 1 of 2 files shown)


cmd/controller/certificates.go
```diff
@@ -302,10 +302,16 @@ func kubeConfig(dest string, url *url.URL, caCert, clientCert, clientKey string,
...
+		fileMode = constant.OwnerOnlyMode
+	}
+
+	err = file.WriteContentAtomically(dest, kubeconfig, fileMode)
 	if err != nil {
 		return err
 	}
 
-	return file.Chown(dest, ownerID, constant.CertSecureMode)
+	return file.Chown(dest, ownerID, fileMode)
```
twz123: What about inlining `file.WriteContentAtomically`, which gives finer control over the permissions, and will allow for removing the call to `file.Chown` a few lines below:


```suggestion
	return file.AtomicWithTarget(dest).
		WithOwner(ownerID).
		WithPermissions(fileMode).
		Write(kubeconfig)
```

cmd/controller/certificates.go
```diff
@@ -302,10 +302,16 @@ func kubeConfig(dest string, url *url.URL, caCert, clientCert, clientKey string,
 		return err
 	}
 
-	err = file.WriteContentAtomically(dest, kubeconfig, constant.CertSecureMode)
+	// Use OwnerOnlyMode (0600) for admin.conf, CertSecureMode (0640) for other kubeconfigs
+	fileMode := os.FileMode(constant.CertSecureMode)
+	if filepath.Base(dest) == "admin.conf" {
```
twz123: What about adding the `fileMode` parameter to the `kubeConfig` function, and passing `constant.CertSecureMode` everywhere but in line 98?

sakshisharma84: @twz123 I actually tried that first. But I thought maybe it is better not to change the signature of the function. If you recommend, I can make this change.

sakshisharma84:
@twz123 Comments addressed! PTAL

cmd/controller/certificates.go
twz123: Yeah, I think so. AFAICT that function is only used in this file, so this change shouldn't be affecting anything else, and the string matching feels a bit brittle, although it's not too likely that we'll either rename a file from or to "admin.conf".

sakshisharma84: done!

twz123:
Thanks!

PR #6517, opened 2025-10-14
Bridge GRPC logs to logrus

twz123:
## Description

By default, grpc-go only logs errors. However, the Kubernetes API server installs a klog backend for GRPC logs. In k0s, this means that GRPC warnings will be printed to stderr. Install our own logrus backend instead. Set the default log level to "error," as vanilla grpc-go does. Support the same environment variables to change that behavior.

Note that to see GRPC info logs, one must set the GRPC log level to info and run k0s with either the --verbose or --debug flag.

See:

* https://github.com/kubernetes/kubernetes/blob/v1.34.1/staging/src/k8s.io/apiserver/pkg/storage/etcd3/logger.go#L27
* k0sproject/k0sctl#953

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/internal/pkg/log/k0s.go
+++ b/internal/pkg/log/k0s.go
@@ -7,6 +7,7 @@ import (
 	"github.com/bombsimon/logrusr/v4"
 	cfssllog "github.com/cloudflare/cfssl/log"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/grpclog"
 	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -24,6 +25,7 @@ func InitLogging() (Backend, ShutdownLoggingFunc) {
 
 	cfssllog.SetLogger((*cfsslAdapter)(logrus.WithField("component", "cfssl")))
 	crlog.SetLogger(logrusr.New(logrus.WithField("component", "controller-runtime")))
+	grpclog.SetLoggerV2(&grpcAdapter{logrus.WithField("component", "grpc")})
 
 	SetWarnLevel()
 
@@ -44,3 +46,7 @@ func SetWarnLevel() {
 	logrus.SetLevel(logrus.WarnLevel)
 	cfssllog.Level = cfssllog.LevelWarning
 }
+
+type grpcAdapter struct{ *logrus.Entry }
+
+func (*grpcAdapter) V(level int) bool { return false }


internal/pkg/log/grpc.go
```diff
@@ -0,0 +1,117 @@
...
+	"strconv"
+
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc/grpclog"
+)
+
+func newGRPCLogger(log logrus.FieldLogger) grpclog.LoggerV2 {
+	// https://github.com/grpc/grpc-go/blob/v1.76.0/grpclog/loggerv2.go#L65-L79
+
+	levelValue := os.Getenv("GRPC_GO_LOG_SEVERITY_LEVEL")
```
makhov: We have `--logging` flag, shouldn't it be there as well?

twz123: Hmm, right now, the logging flag is only for embedded binaries, not for the logging of stuff happening in the k0s executable itself.

makhov: Then why we need a separate log level for it? Can we just use the k0s log level here?

twz123: We could, but the warning would still be displayed because k0s logs warnings and errors by default. This cannot be turned off. You can only request info logs (--verbose) or debug logs (--debug). Logrus itself is not flexible enough to support different log levels for different components either.

Kimmo was trying to troubleshoot this and [found the environment variable that usually controls it](https://github.com/k0sproject/k0sctl/issues/953#issuecomment-3401015555). That's why I thought it wouldn't be a bad idea for k0s to honor those settings. However, it currently doesn't because the API server overwrites grpc-go's defaults (which is error) in one of its package's init functions.

So yes, we can remove the separate log level, but that wouldn't silence the warning. I'm trying to figure out if this can be fixed upstream because either etcd is doing something unusual or GRPC is logging at the wrong log level. However, I'm not convinced that this will lead anywhere.

twz123: Another option would be to shift the GRPC log levels by one: Error -> Warning, Warning -> Info, Info -> Debug. The GRPC logger doesn't have a Debug level, it has a V threshold instead.

makhov (dismissed):

internal/pkg/log/grpc.go
twz123: Hold on! Maybe there's really a chance to fix it upstream. In that case, it'd probably be best to just use k0s log levels.

twz123: @makhov I've removed all the custom handling and made this a simple adapter, just as we do it for cfssl. I'm trying my best to get rid of the warning upstream. However, if that won't work out, then we can still opt to add some means of changing the GRPC log level separately from the general k0s log level.

PR #6566, opened 2025-10-30
Update nvidia runtime setup info (gpu-operator Helm chart)

josef-hak:
## Description

Use up-to-date NVidia GPU operator Helm chart installation - updated parameters. Previous version doesn't work anymore as the chart was updated and the command didn't use `--version` parameter.

Link: https://docs.k0sproject.io/head/runtime/#using-nvidia-container-runtime

### DIFF
--- a/docs/runtime.md
+++ b/docs/runtime.md
@@ -212,14 +212,16 @@ helm repo add nvidia https://helm.ngc.nvidia.com/nvidia
 helm repo update
 helm install nvidia-gpu-operator -n nvidia-gpu-operator \
   --create-namespace \
-  --set operator.defaultRuntime=containerd \
-  --set toolkit.env[0].name=CONTAINERD_CONFIG \
-  --set toolkit.env[0].value=/etc/k0s/containerd.d/nvidia.toml \
-  --set toolkit.env[1].name=CONTAINERD_SOCKET \
-  --set toolkit.env[1].value=/run/k0s/containerd.sock \
-  --set toolkit.env[2].name=CONTAINERD_RUNTIME_CLASS \
-  --set toolkit.env[2].value=nvidia \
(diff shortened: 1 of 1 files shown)


docs/runtime.md
```diff
@@ -212,14 +212,16 @@ helm repo add nvidia https://helm.ngc.nvidia.com/nvidia
...
-  --set toolkit.env[0].name=CONTAINERD_CONFIG \
-  --set toolkit.env[0].value=/etc/k0s/containerd.d/nvidia.toml \
-  --set toolkit.env[1].name=CONTAINERD_SOCKET \
-  --set toolkit.env[1].value=/run/k0s/containerd.sock \
-  --set toolkit.env[2].name=CONTAINERD_RUNTIME_CLASS \
-  --set toolkit.env[2].value=nvidia \
-  nvidia/gpu-operator
+  --set "operator.defaultRuntime=containerd" \
+  --set "toolkit.env[0].name=CONTAINERD_CONFIG" \
+  --set "toolkit.env[0].value=/run/k0s/containerd-cri.toml" \
```
jnummelin, line 217: I wonder if these should point to the "real" containerd config instead of this k0s managed CRI merged one 🤔  I guess it boils down to what exactly the operator does with this.

This is what the docs say:
```
CONTAINERD_CONFIG
The path on the host to the top-level containerd config file. By default this will point to /etc/containerd/containerd.toml (the default location for containerd). It should be customized if your containerd installation is not in the default location.
```

Based on the above, I'd say this should point to `/etc/k0s/containerd.toml` which in turn imports the CRI config

josef-hak: I just tried the config and it worked out-of-the box.

makhov:
Could you please sign off the commit?

PR #6604, opened 2025-11-06
Implement CPLB reconciler healtcheck

juanluisvaladas:
## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/cplb/cplb_reconciler.go
+++ b/pkg/component/controller/cplb/cplb_reconciler.go
@@ -35,6 +35,7 @@ type CPLBReconciler struct {
 	log            *logrus.Entry
 	kubeconfigPath string
 	mu             sync.RWMutex
+	wg             sync.WaitGroup
 	updateCh       chan<- struct{}
 	stop           func()
 	healthCheckers map[string]*healthChecker
@@ -47,6 +48,7 @@ func NewCPLBReconciler(kubeconfigPath string, apiPort int, updateCh chan<- struc
 		kubeconfigPath: kubeconfigPath,
 		updateCh:       updateCh,
 		healthCheckers: make(map[string]*healthChecker),
+		wg:             sync.WaitGroup{},
 	}
 }
 
@@ -76,6 +78,7 @@ func (r *CPLBReconciler) Start() error {
 func (r *CPLBReconciler) Stop() {
 	r.log.Debug("Stopping")
 	r.stop()
+	r.wg.Wait()
 	r.log.Info("Stopped")
 }
 
@@ -234,7 +237,8 @@ func (r *CPLBReconciler) newHealthChecker(ctx context.Context, restConfig *rest.
 		updateCh:      r.updateCh,
 		healthCheckFn: healthCheckFn,
 	}
-	go hc.runHealthCheck(c, addr, nil)
+	r.wg.Add(1)
+	go hc.runHealthCheck(c, addr, nil, r.wg.Done)
 	return hc
 }
 
@@ -242,7 +246,8 @@ func (hc *healthChecker) Stop() {
 	hc.cancel()
 }
 
-func (hc *healthChecker) runHealthCheck(ctx context.Context, addr string, c <-chan time.Time) {
+func (hc *healthChecker) runHealthCheck(ctx context.Context, addr string, c <-chan time.Time, done func()) {
+	defer done()
 	if c == nil {
 		ticker := time.NewTicker(time.Second)
 		defer ticker.Stop()
--- a/pkg/component/controller/cplb/cplb_reconciler_test.go
+++ b/pkg/component/controller/cplb/cplb_reconciler_test.go
@@ -49,7 +49,7 @@ func TestHealthCheckerRunHealthCheck(t *testing.T) {
 	}
 	c := make(chan time.Time)
 
-	go hc.runHealthCheck(t.Context(), "192.168.1.1", c)
+	go hc.runHealthCheck(t.Context(), "192.168.1.1", c, func() {})
 	require.False(t, hc.healthy, "Expected the healthChecker to be unhealthy before the first tick")
 
 	c <- time.Now()


pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -18,40 +22,50 @@ import (
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
 // CPLBReconciler monitors the endpoints of the "kubernetes" service in the
 // "default" namespace. It notifies changes though the updateCh channel provided
 // in the constructor.
 type CPLBReconciler struct {
+	apiPort        int
```
twz123, line 34: This could be a `uint16`

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -18,40 +22,50 @@ import (
...
+		healthCheckers: make(map[string]*healthChecker),
 	}
 }
 
 func (r *CPLBReconciler) Start() error {
-	clientset, err := kubeutil.NewClientFromFile(r.kubeconfigPath)
+	clientSet, err := kubeutil.NewClientFromFile(r.kubeconfigPath)
 	if err != nil {
-		return fmt.Errorf("failed to get clientset: %w", err)
+		return fmt.Errorf("failed to get clienStet: %w", err)
```
twz123: There's a typo in the error string. I'd say, though, that we can pass on the error as is. That's what we're doing in the rest of the codebase, at least.

```suggestion
		return err
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+	shouldNotify := false
+	for addr := range r.healthCheckers {
+		if !slices.Contains(newAddresses, addr) {
+			r.healthCheckers[addr].Stop()
+			delete(r.healthCheckers, addr)
+			shouldNotify = true
 		}
 	}
+	if shouldNotify {
+		r.updateCh <- struct{}{}
```
twz123: Any reason not to keep the old way of doing it? Especially since we're holding the lock.


```suggestion
		select {
		case r.updateCh <- struct{}{}:
		default:
		}
```

There's some other places that are sending to the channel in a blocking manner. I don't think this is necessary, and sending in a non-blocking manner would avoid potential starvations/deadlocks.

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -18,40 +22,50 @@ import (
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/kubernetes/scheme"
```
twz123, line 23: Can you group those together with line 22?

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+			}
+			// If it's not an *http.Transport, return it as-is
+			return rt
+		}
+
+		rc, err = rest.RESTClientFor(cfg)
+		fmt.Println(rc.Client.Transport)
+	}
+
+	c, cancelFunc := context.WithCancel(ctx)
```
twz123: Would be more idiomatic to call `cancelFunc` `cancel`.

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+	// If something went wrong or we didn't crate a rest client, run without health checking
```
twz123:
```suggestion
	// If something went wrong or we didn't create a REST client, run without health checking
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+		healthCheckFn: healthCheckFn,
+	}
+	go hc.runHealthCheck(c, addr, nil)
+	return hc
+}
+
+func (hc *healthChecker) Stop() {
+	// if hc.cancelFunc != nil {
+	hc.cancelFunc()
+	// }
```
twz123:
```suggestion
	hc.cancelFunc()
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+	c, cancelFunc := context.WithCancel(ctx)
+	// If something went wrong or we didn't crate a rest client, run without health checking
+	if err != nil || rc == nil {
+		// I think this scenario is impossible outside testing.
+		r.log.WithError(err).Error(fmt.Sprintf("Unable to create restClient for %q healthCheck. Running without healthCheck", addr), err)
+		hc := &healthChecker{
+			healthy:    true,
+			cancelFunc: cancelFunc,
+		}
+		r.updateCh <- struct{}{}
```
twz123: Could be non-blocking:

```suggestion
		select {
		case r.updateCh <- struct{}{}:
		default:
		}
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+		select {
+		case <-ctx.Done():
+			hc.log.Infof("Stopping healthchecker for %q", addr)
+			hc.healthy = false
+			return
+		case <-c:
+			result := hc.healthCheckFn(ctx)
+			if err := result.Error(); err != nil {
+				if hc.healthy {
+					hc.log.Infof("Health check failed, %q became innactive", addr)
```
twz123:
```suggestion
					hc.log.Infof("Health check failed, %q became inactive", addr)
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+	return healthyAddrs
+}
+
+// healthChecker provides health checking functionality for Kubernetes API servers
+type healthChecker struct {
+	cancelFunc    context.CancelFunc
+	rc            rest.Interface
+	log           *logrus.Entry
+	healthy       bool
+	updateCh      chan<- struct{}
```
twz123, line 179: This somehow looks like it could be replaced by

```suggestion
	healthy       value.Latest[bool]
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+		case <-ctx.Done():
+			hc.log.Infof("Stopping healthchecker for %q", addr)
+			hc.healthy = false
+			return
+		case <-c:
+			result := hc.healthCheckFn(ctx)
+			if err := result.Error(); err != nil {
+				if hc.healthy {
+					hc.log.Infof("Health check failed, %q became innactive", addr)
+					hc.healthy = false
```
twz123, line 269: This and the other assignments to `hc.healthy` in this method are racing with the reads happening in `CPLBReconciler.GetIPs` on [line 152](https://github.com/k0sproject/k0s/pull/6604/files#diff-f51f898f31787ec23608ab3ddf4db0f22eab29b2b7f46b9cb34d0d5fa95a840dR152):

```go
	for addr := range r.healthCheckers {
		if r.healthCheckers[addr].healthy {
			healthyAddrs = append(healthyAddrs, addr)
		}
	}
```

pkg/component/controller/cplb/cplb_reconciler.go
```diff
@@ -109,22 +123,154 @@ func (r *CPLBReconciler) maybeUpdateIPs(endpoint *corev1.Endpoints) {
...
+		}
+	}
+	hc := &healthChecker{
+		rc:            rc,
+		log:           r.log,
+		cancelFunc:    cancelFunc,
+		updateCh:      r.updateCh,
+		healthCheckFn: healthCheckFn,
+	}
+	go hc.runHealthCheck(c, addr, nil)
```
twz123: I think there's a potential possibility for a panic due to a send to a closed channel here when `Keepalived.Stop()` is called:

The health checkers might still be running after `CPLBReconciler.Stop()` has returned. The update channel will be closed, and a health checker that didn't terminate yet might try to write to a closed channel.

It might be better to actually block in `CPLBReconciler.Stop()` until all health checkers fanned in. Could be e.g. done with a `sync.WaitGroup`. Moreover,  as a rule of thumb, channels should be closed by the sending party, not the receiving party. Hence the ownership of the update channel could be transferred from `Keepalived` into `CPLBReconciler`, so that the reconciler manages its lifecycle. It could then simply expose, say, an `Updates() <-chan struct{}` method that can be used by `Keepalived`.

Another possibility would be to never close that channel at all, as long as we are e.g. using "for select" loops, with an additional ctx.Done() case to break out of the loop, instead of "for range" loops. But then it becomes really important to use non-blocking sends.

pkg/component/controller/cplb/cplb_reconciler.go
juanluisvaladas: I prefer to keep it this way so that it matches the type of `APISpec.Port`. The port validated before reaching this point

pkg/component/controller/cplb/cplb_reconciler.go
juanluisvaladas: > The health checkers might still be running after CPLBReconciler.Stop() has returned. The update channel will be closed, and a health checker that didn't terminate yet might try to write to a closed channel.

I don't think this actually possible because we don't close the channel explicitly, when we cancel the context we later remove it from the `healthCheckers` map and eventually the garbage collector will take care of it. And we shouldn't close it at all because that channel is shared by all the `healthChecker` in the map.

pkg/component/controller/cplb/cplb_reconciler.go
juanluisvaladas: Good point, I added a mutex to help with this.

pkg/component/controller/cplb/cplb_reconciler.go
juanluisvaladas: Probably, but then we'd need to monitor N channels in parallel which would change the logic enitrely, not quite sure if that's worth it.

pkg/component/controller/cplb/cplb_reconciler.go
twz123: Alright. Although I think it's much nicer to push validation to the type system. Maybe something to keep an eye on for the new Config structs.

pkg/component/controller/cplb/cplb_reconciler.go
twz123: Yeah. Changing this here would mean to rethink/rewrite everything else, as well.

pkg/component/controller/cplb/cplb_reconciler.go
twz123: But is *is* closed by `Keepalived.Stop()`:

https://github.com/k0sproject/k0s/blob/f826531dae64e59ab36538c09c7f5d710cd0ad32/pkg/component/controller/cplb/cplb_linux.go#L181

While just removing this line would fix the possible panic, I think we should nonetheless add a `sync.WaitGroup` for these goroutines and fan those in when the reconciler stops. IMO that makes Start and Stop symmetric, i.e. the first starts goroutines, the latter actually stops them.

juanluisvaladas: You're right, I fixed it in a separate commit for easier review. Good catch!

PR #6627, opened 2025-11-08
Add documentation for Longhorn installation

shikanime:
## Description

Fixes # (issue)

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docs/examples/longhorn.md
+++ b/docs/examples/longhorn.md
@@ -1,0 +1,210 @@
+<!--
+SPDX-FileCopyrightText: 2025 k0s authors
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Longhorn
+
+This tutorial covers installing Longhorn as a Helm extension on k0s. Longhorn is a distributed block storage system for Kubernetes that provides enterprise-grade features including:
+
+- **Distributed storage**: Provides persistent storage across your cluster nodes
+- **Snapshots and backups**: Built-in snapshot and backup capabilities for data protection
+- **Live upgrades**: Zero-downtime upgrades of storage volumes
+- **High availability**: Automatic failover and replication for data durability
+- **Cross-cluster disaster recovery**: Backup to external storage systems
+
+Longhorn is particularly well-suited for stateful applications that require persistent storage, such as databases, message queues, and content management systems.
+
+For more information on Longhorn, refer to the [official documentation](https://longhorn.io/docs/).
+
+## Prerequisites
+
+Before installing Longhorn, ensure your k0s cluster meets these requirements:
+
+### System Requirements
+
+Please refer to the [official Longhorn v1.10.x Support Matrix](https://www.suse.com/suse-longhorn/support-matrix/all-supported-versions/longhorn-v1-10-x/) for detailed system requirements.
+
+### Required Software
+
+Longhorn requires the iSCSI initiator utilities to be installed on every node in your cluster:
+
+**Why iSCSI is required**: Longhorn uses iSCSI protocol to provide block storage to pods. The iSCSI initiator allows nodes to connect to Longhorn volumes as if they were local block devices.
+
+Ubuntu/Debian:
+
+```sh
+sudo apt-get install -y open-iscsi
+```
+
+RHEL/CentOS/Rocky:
+
+```sh
+sudo yum install -y iscsi-initiator-utils
+```
+
+SUSE:
+
+```sh
+sudo zypper install -y open-iscsi
+```
+
+Ensure the iSCSI daemon is enabled and running:
+
+```sh
+sudo systemctl enable --now iscsid
+```
+
+## Install via k0s Helm extension
+
+Add Longhorn as a Helm chart in your k0s ClusterConfig. Apply this configuration through your normal k0s configuration workflow.
+
+### Basic Installation
+
+The simplest way to install Longhorn is to add it to your k0s ClusterConfig:
+
+```yaml
+apiVersion: k0s.k0sproject.io/v1beta1
+kind: ClusterConfig
+metadata:
+  name: k0s
+spec:
+  extensions:
+    helm:
+      repositories:
+        - name: longhorn
+          url: https://charts.longhorn.io
+      charts:
+        - name: longhorn
+          chartname: longhorn/longhorn
+          namespace: longhorn-system
+          version: 1.10.1
+```
+
+### Verification Steps
+
+After applying the configuration, verify that Longhorn is properly installed:
+
+**Check pod status**: Wait for all Longhorn components to become ready:
+
+```sh
+k0s kubectl -n longhorn-system get pods
+```
+
+Expected output should show all pods in `Running` status. This may take 2-3 minutes.
+
+**Verify StorageClass**: Confirm that the `longhorn` StorageClass was created:
+
+```sh
+$ k0s kubectl get storageclass
+NAME                 PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
+longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   16m
+longhorn-static      driver.longhorn.io   Delete          Immediate           true                   16m
+```
+
+You should see `longhorn` listed as a storage class.
+
+### Single-Node Cluster Configuration
+
+Single-node clusters require setting the default replica count to 1 to prevent scheduling failures, as Longhorn's default of 3 replicas cannot be satisfied on a single node. This setting can be adjusted in the values or post-installation via the UI (Settings > General > Default Replica Count).
+
+```yaml
+spec:
+  extensions:
+    helm:
+      charts:
+        - name: longhorn
+          chartname: longhorn/longhorn
+          namespace: longhorn-system
+          version: "latest"
+          values: |
+            defaultSettings:
+              defaultReplicaCount: 1
+```
+
+## Access the Longhorn UI
+
+You can port-forward the frontend service:
+
+```sh
+k0s kubectl -n longhorn-system port-forward svc/longhorn-frontend 8080:80
+```
+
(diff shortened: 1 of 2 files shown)


twz123:
Cool! I'll check this out now! Would you mind changing the shell from "bash" to "sh"?

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
...
+## Prerequisites
+
+- Install `open-iscsi` on every Linux node (required by Longhorn).
+- Ensure the `iscsid` service is enabled and running on each node.
+- Make sure nodes have a stable hostname and reachable IPs.
+
+Ubuntu/Debian:
+
+```bash
+sudo apt-get update
```
twz123: Updating the package cache isn't mentioned for the other distros. Any special reason why to do it here?

If we wanna keep it, I'd merge it with the next shell block. Looks a bit more consistent when rendered.

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
...
+
+SUSE:
+
+```bash
+sudo zypper install -y open-iscsi
+```
+
+## Install via k0s Helm extension
+
+Add Longhorn as a Helm chart in your k0s ClusterConfig. Apply this configuration through your normal k0s configuration workflow.
```
twz123, line 60: Is there some link to the upstream chart docs that we could include for reference?

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
+<!--
+SPDX-FileCopyrightText: 2025 k0s authors
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Longhorn
+
+This tutorial covers installing Longhorn as a Helm extension on k0s. Longhorn provides distributed block storage for Kubernetes with features like snapshots, backup, and live upgrades.
```
twz123: Would be cool to have a link to the upstream project here.

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Access the Longhorn UI
+
+You can port-forward the frontend service:
+
+```bash
+k0s kubectl -n longhorn-system port-forward svc/longhorn-frontend 8080:80
+```
+
+Then open `http://localhost:8080` in your browser.
```
twz123: Let's make it clickable:

```suggestion
Then open <http://localhost:8080> in your browser.
```

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
...
+-->
+
+# Longhorn
+
+This tutorial covers installing Longhorn as a Helm extension on k0s. Longhorn provides distributed block storage for Kubernetes with features like snapshots, backup, and live upgrades.
+
+## Prerequisites
+
+- Install `open-iscsi` on every Linux node (required by Longhorn).
+- Ensure the `iscsid` service is enabled and running on each node.
```
twz123: We might be more precise and say that this is a requirement for every "storage node". So you e.g. don't need it on pure controller nodes, and, if you e.g. select only certain nodes via some node selector, then only on those. WDYT?

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
...
+
+You can port-forward the frontend service:
+
+```bash
+k0s kubectl -n longhorn-system port-forward svc/longhorn-frontend 8080:80
+```
+
+Then open `http://localhost:8080` in your browser.
+
+## Example: PVC and Deployment
```
twz123, line 135: Hehe, tried this out on a single node cluster, got this, of course 🤣 Maybe it's worth mentioning that you need at least three storage nodes? Or reduce the replica level here? IDK 🙃

Image

shikanime:
@twz123 No problem

docs/examples/longhorn.md
```diff
@@ -0,0 +1,134 @@
+<!--
+SPDX-FileCopyrightText: 2025 k0s authors
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Longhorn
+
+This tutorial covers installing Longhorn as a Helm extension on k0s. Longhorn provides distributed block storage for Kubernetes with features like snapshots, backup, and live upgrades.
+
+## Prerequisites
```
twz123, line 20: I've also tried to run this on Alpine workers:

```sh
doas apk add open-iscsi
doas rc-update add iscsid default
doas rc-service iscsid start
```

However, that failed completely:

```console
$ kubectl -n longhorn-system get events | grep Error | head -n1
102s        Warning   Failed              pod/longhorn-manager-4n888                     Error: failed to generate container "5e66ceb5e441db848f598691d81bf9bc8c46fa282783af0c6b8ff56d166485e7" spec: failed to generate spec: path "/var/lib/longhorn/" is mounted on "/" but it is not a shared mount
```

shikanime: Oh, that's interesting, let me check this.

docs/examples/longhorn.md
shikanime: You're right. By default, the Helm chart sets the minimum number of replicas to three. I usually set it to one for small clusters or when network bandwidth is constrained, completely doable.

docs/examples/longhorn.md
shikanime: You are right, it's such a common command to enter before 'apt-get' on Debian based that it didn't even cross my mind.

docs/examples/longhorn.md
shikanime: Fair enough.

twz123: Wait! Is this maybe required to mount PVs? Then it's required on every worker node, not every storage node. Not familiar with longhorn, so I'm hypothesizing a bit.

shikanime: Yes, the Longhorn Manager is responsible for the plumbing, which makes use of various node utilities, including ISCSI for RWO, NFS for RWX and other crypto kernel modules. However, if you require a non-homogeneous cluster, for example a Windows node or a node without the necessary kernel module, we will need to implement a `cattle.io/os=linux:NoSchedule` label. This will also prevent the scheduling of pods that use Longhorn PV, but the cluster will still be in a stable state.

docs/examples/longhorn.md
shikanime: @twz123 I think it's an issue with the host mount missing options `rshared`. It's probably similar to [Talos](https://longhorn.io/docs/1.10.1/advanced-resources/os-distro-specific/talos-linux-support/).

docs/examples/longhorn.md
```diff
@@ -0,0 +1,241 @@
...
+
+### System Requirements
+
+- **Operating System**: Ubuntu 24.04+, SLES 15 SP7+, SLE Micro 6.1+, RHEL 10.0+, Oracle Linux 9.6+, Rocky Linux 10.0+, Talos 1.10.6+, Container-Optimized OS 121+
+- **Kernel**: Linux kernel 3.10 or later
+- **Storage**: At least 10GB of free disk space per node for Longhorn data
+- **Memory**: Minimum 2GB RAM per node (4GB+ recommended for production)
+- **Network**: Stable network connectivity between all cluster nodes
+
+For more detailed system requirements, refer to the [official Longhorn v1.10.x Support Matrix](https://www.suse.com/suse-longhorn/support-matrix/all-supported-versions/longhorn-v1-10-x/).
```
twz123: Maybe the link to the upstream support matrix is enough, so we don't need to sync this list with upstream whenever they change something.

docs/examples/longhorn.md
```diff
@@ -0,0 +1,241 @@
...
+  extensions:
+    helm:
+      repositories:
+        - name: longhorn
+          url: https://charts.longhorn.io
+      charts:
+        - name: longhorn
+          chartname: longhorn/longhorn
+          namespace: longhorn-system
+          version: "latest"
```
twz123: That won't work. We need the concrete version, otherwise k0s won't find the chart.
```suggestion
          version: 1.10.1
```

docs/examples/longhorn.md
```diff
@@ -0,0 +1,241 @@
...
+k0s kubectl -n longhorn-system get pods
+```
+
+Expected output should show all pods in `Running` status. This may take 2-3 minutes.
+
+2. **Verify StorageClass**: Confirm that the `longhorn` StorageClass was created:
+
+```sh
+k0s kubectl get storageclass
+```
```
twz123, line 103: We could show the expected output here:

```suggestion
```console
$ k0s kubectl get storageclass
NAME                 PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
longhorn (default)   driver.longhorn.io   Delete          Immediate           true                   16m
longhorn-static      driver.longhorn.io   Delete          Immediate           true                   16m
\```
```

docs/examples/longhorn.md
```diff
@@ -0,0 +1,241 @@
...
+k0s kubectl apply -f nginx.yaml
+```
+
+Check the PVC and pod status:
+
+```sh
+k0s kubectl get pvc,pods
+```
+
+## Advanced Configuration
```
twz123: I wonder if that's a bit too much for this guide. I'd rather expect this in the [upstream chart docs](https://github.com/longhorn/charts/blob/v1.10.x/charts/longhorn/README.md) and then add a link to those here.

twz123 (commented):
Alright! I was able to deploy this on a three node cluster. Needed to change the chart version from latest, though. Moreover, on one node, all the DaemonSet Pods went into a crash loop. After wiping all of them with `kubectl delete po` they got recreated and became ready. The other nodes didn't have that problem (just one or two pods that got restarted once).

Left some nits in addition to the chart version change. Would you mind addressing the linter issues?

twz123:
Thanks! :slightly_smiling_face:

PR #6654, opened 2025-11-13
Add honorLabels to pushgateway ServiceMonitor example in the Docs

hille721:
## Description

Without `honorLabels: true` the labels of the scraped metrics from the Pushgateway will have a `exported_` prefix.

With that the query in the docs (`absent(apiserver_audit_event_total{job="kube-scheduler"})`) would still not work as it would be `absent(apiserver_audit_event_total{exported_job="kube-scheduler"})`.

With `honorLabels: true`, everything works as expected and also other queries are working in the same way as they are on other distributions without isolated CP.

## Type of change

- [X] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [ ] Auto test added

### DIFF
--- a/docs/system-monitoring.md
+++ b/docs/system-monitoring.md
@@ -46,6 +46,7 @@ metadata:
 spec:
   endpoints:
   - port: http
+    honorLabels: true 
   selector:
     matchLabels:
       app: k0s-observability


twz123 (approved):
Nice! Thank you!

PR #6669, opened 2025-11-14
Disable CGO for k0s

s3rj1k:
Disable CGO usage for k0s only, simplifies cross compiling

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,7 +207,7 @@ pkg/assets/zz_generated_offsets_linux.go pkg/assets/zz_generated_offsets_windows
 endif
 
 k0s: TARGET_OS = linux
-k0s: BUILD_GO_CGO_ENABLED = 1
+k0s: BUILD_GO_CGO_ENABLED = 0
 
 k0s.exe: TARGET_OS = windows
 k0s.exe: BUILD_GO_CGO_ENABLED = 0


embedded-bins/Makefile.variables
```diff
@@ -41,9 +41,9 @@ kubernetes_build_go_ldflags_extra = "-extldflags=-static"
 kine_version = 0.14.6
 kine_buildimage = $(golang_buildimage)
 kine_build_go_tags = nats
-#kine_build_go_cgo_enabled =
+kine_build_go_cgo_enabled = 0
```
twz123: Oh, are you sure that kine supports to be built without CGO at all? I mean, you just replaced the sqlite dependency for k0s, but not for kine.

s3rj1k: I just wanted to quick test, but looking at `kine` imports I can only quote classics of our time, "You are absolutely right!", it uses CGO only libs and will never work without CGO.

ncopa:
I would like some kind of explanation in the commit message why this change is needed or what benefit it brings. Something we can refer to in case someone want revert the change in future.

s3rj1k:
> I would like some kind of explanation in the commit message why this change is needed or what benefit it brings. Something we can refer to in case someone want revert the change in future.

added in PR message and updated commit message

juanluisvaladas:
@s3rj1k  I think this commit was added by accident? https://github.com/k0sproject/k0s/pull/6669/changes/ec136908a89a47afb8f269e74fec9ccd61bff228

s3rj1k:
> @s3rj1k I think this commit was added by accident? [ec13690](https://github.com/k0sproject/k0s/commit/ec136908a89a47afb8f269e74fec9ccd61bff228)

Tests where not working at the time of PR creation, feel free to drop this commit

juanluisvaladas:
Can you please rebase without it?

s3rj1k:
> Can you please rebase without it?

Done

PR #6719, opened 2025-11-20
Enable "k0s status" on Windows workers

### LINKED ISSUE #2887: k0s status on windows

It was just never implemented, there shouldn't be anything else to it than defining a proper default SocketPath like `\\.\pipe\k0sstatus` and maybe using winio.ListenPipe in place of net.Listen and winio.DialPipe instead of net.Dial in the status command.

_No response_

_No response_

### Additional context

_No response_

kke:
## Description

Uses pipe based transport on Windows to serve and query the status component.

Fixes #2887

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/status/status.go
+++ b/pkg/component/status/status.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
-	"net"
 	"net/http"
 	"strconv"
 	"time"
@@ -58,7 +57,6 @@ func (s *Status) Init(_ context.Context) error {
 			w.WriteHeader(http.StatusInternalServerError)
 		}
 	})
-	var err error
 	s.httpserver = http.Server{
 		Handler: mux,
 	}


kke:
```console
C:\Users\Administrator> k0s status
Version: v1.34.1+k0s.0-272-gb9f57feaf
Process ID: 4028
Role: worker
Workloads: true
SingleNode: false
Kube-api probing successful: true
Kube-api probing last error:
```

kke:
Autopilot always uses the default socket path 🤔 ref: #6750

cmd/worker/worker_windows.go
```diff
@@ -45,6 +48,23 @@ func initLogging(ctx context.Context, logDir string) error {
...
+			Role:       "worker",
+			Args:       os.Args,
+			Version:    build.Version,
+			Workloads:  true,
+			SingleNode: false,
+			K0sVars:    k0sVars,
+		},
+		CertManager: certManager,
+		Socket:      k0sVars.StatusSocketPath,
+	})
```
twz123: I'd rather move this out of `worker_unix.go` into `worker.go`.

pkg/component/status/defaults.go
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2025 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package status
+
+import "github.com/k0sproject/k0s/pkg/constant"
+
+// DefaultSocketPath is the platform-specific default socket path used by k0s
+// when no explicit status socket is provided.
+const DefaultSocketPath = constant.StatusSocketPathDefault
```
twz123, line 13: What about directly adding a "Deprecated" godoc, so it's clear that it's a remnant from the past that needs to be refactored and should be avoided?

pkg/component/status/transport_windows.go
```diff
@@ -0,0 +1,60 @@
...
+	path := pipePath(socketPath)
+	transport := &http.Transport{
+		DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+			return winio.DialPipeContext(ctx, path)
+		},
+	}
+	return &http.Client{Transport: transport}, nil
+}
+
+func pipePath(socketPath string) string {
```
twz123: Do you think we could just drop this function? We're already specifying the default socket path on Windows in the correct format (`\\.\pipe\k0s-status`). Would require less magic here. Moreover, we're not doing anything similar to post-process the npipe paths for external CRIs.

pkg/component/status/client.go
```diff
@@ -49,22 +45,18 @@ func GetStatusInfo(socketPath string) (*K0sStatus, error) {
...
-		Transport: &http.Transport{
-			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
-				var d net.Dialer
-				return d.DialContext(ctx, "unix", socketPath)
-			},
-		},
+func doStatusHTTPRequest(socketPath string, path string, tgt any) error {
+	httpc, err := newStatusHTTPClient(socketPath)
+	if err != nil {
+		return fmt.Errorf("status: can't prepare transport for %q: %w", socketPath, err)
```
twz123: What about doing to the OS specific split a bit deeper at the dialer level?

```suggestion
	httpc := http.Client{
		Transport: &http.Transport{
			DisableKeepAlives: true,
			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
				return dialSocket(ctx, socketPath)
			},
		},
	}
```

That `dialSocket` function could be implemented as follows:

```go
// unix
func dialSocket(ctx context.Context, socketPath string) (net.Conn, error) {
	var d net.Dialer
	return d.DialContext(ctx, "unix", socketPath)
}

// windows
func dialSocket(ctx context.Context, socketPath string) (net.Conn, error) {
	return winio.DialPipeContext(ctx, pipePath(socketPath))
}
```

I'd also specify `DisableKeepAlives: true`, as those are one-shot HTTP clients, and keeping connections around is counter-productive in this case, unless we want to reuse the clients for multiple requests.

pkg/component/status/status.go
```diff
@@ -67,8 +64,7 @@ func (s *Status) Init(_ context.Context) error {
 		Handler: mux,
 	}
 
-	removeLeftovers(s.Socket)
-	s.listener, err = net.Listen("unix", s.Socket)
+	s.listener, err = newStatusListener(s.Socket)
```
twz123: I know this has been here before, but I'd say this would be better located in the `Start` method, as `Init` is not supposed to open resources that need a cleanup (i.e. after `Init` returns, there's no obligation to call `Stop`). As a plus, we'd get rid of the listener field in the `Status` struct.

pkg/component/status/status.go
```diff
@@ -103,8 +91,10 @@ func (s *Status) Stop() error {
 	if err := s.httpserver.Shutdown(ctx); err != nil && !errors.Is(err, context.Canceled) {
 		return err
 	}
-	// Unix socket doesn't need to be explicitly removed because it's handled
-	// by httpserver.Shutdown
+	if s.listener != nil {
+		_ = s.listener.Close()
+	}
```
twz123: What the original comment wanted to say is this:


```suggestion
	// There is no need to explicitly close the listener because it is now owned
	// by the HTTP server, which will have already closed it during shutdown.
```

pkg/config/cfgvars.go
```diff
@@ -131,13 +132,29 @@ func NewCfgVars(cobraCmd command, dirs ...string) (*CfgVars, error) {
...
+		if val, err := cobraCmd.Flags().GetString("status-socket"); err == nil {
+			statusSocketPath = val
+		}
+	}
+	if statusSocketPath == "" {
+		if runtime.GOOS == "windows" || euid == 0 {
+			statusSocketPath = constant.StatusSocketPathDefault
+		} else {
+			statusSocketPath = filepath.Join(runDir, "status.sock")
+		}
```
twz123: The non-windows fallback is right for every user, as the correct runDir is set based on the euid already.

```suggestion
		if runtime.GOOS == "windows" {
			statusSocketPath = constant.StatusSocketPathDefault
		} else {
			statusSocketPath = filepath.Join(runDir, "status.sock")
		}
```

pkg/config/cfgvars.go
```diff
@@ -131,13 +132,29 @@ func NewCfgVars(cobraCmd command, dirs ...string) (*CfgVars, error) {
...
 	} else {
 		runDir = filepath.Join(dataDir, "run")
 	}
 
+	var statusSocketPath string
+	if cobraCmd != nil {
+		if val, err := cobraCmd.Flags().GetString("status-socket"); err == nil {
+			statusSocketPath = val
+		}
+	}
```
twz123, line 147: I _think_ this is not required, as it's done in `WithCommand`, which is being called later in this method?

```suggestion
```

kke: Yeah, looks like that, I followed what is done above with data-dir and kubelet-root-dir. I think the withCommand overwrites those too.

pkg/component/status/transport_windows.go
kke: It turns user supplied paths into pipes, but maybe the user will get some error if they give something odd.

pkg/config/cfgvars.go
kke: Looks like it may be a larger can of worms and it could be better to fix separately as it applies to other flags too.

pkg/component/status/transport_windows.go
kke: Threw it out, let it fail downstream

pkg/component/status/client.go
kke: Simplified

pkg/component/status/status.go
kke: Moved to Start

twz123 (approved):
Nice!

Issue #5253, opened 2024-11-14
Support getting join token from environment variable


twz123:
### Is your feature request related to a problem? Please describe.

There's currently two ways of passing the join token to `k0s controller` or `k0s worker`:

1. As an argument on the command line
2. As a file, by passing the `--token-file /path/to/tokenfile` flag on the command line.

The first approach is only feasible for development and testing purposes, since the command line is not confidential except on particularly hardened machines. The second approach doesn't suffer from this, as it relies on the token file's permissions and is therefore used by automation tools (e.g. `k0sctl`). However, this means that the token must be written to a place where it is potentially persistent and, if the permissions are inappropriate, can be leaked. While persistence may or may not be a desired property, it still means that the lifecycle of the token file must be managed.

### Describe the solution you would like

In addition to the above ways of passing the join token, have both subcommands accept it via an environment variable, say `K0S_JOIN_TOKEN`. This would be a fairly standard way of passing credentials. It doesn't require the file system or the command line, and is a good middle ground for tools.

In fact, I'm about to ask for a deprecation of support for join tokens as CLI arguments, since managing this via environment variables is potentially even more friendly to manual plumbing.

adlion:
Just to add about security.

When creating a join token there is an argument where you can define time expiration of the token.
Currently I am setting it in about 1 hour on my cluster.
This might help regarding security.

psaradhi:
@twz123  - please assign this to me.

@adlion  @twz123  - couple of questions reg implementation --

1.  If command line arg & token file path needs to be deprecated 
       a. print a warning to users that these features are deprecated and removed from next major version ?
     
2. if all 3 are specified  does the following order of priority sound reasonable?
      a. argument
      b. token file
      c. env variable 
    
      for eg:  if json token is specified as an argument (even though it's insecure), it will take the highest priority , followed by token file and then env variable

twz123:
Definitely option b including the deprecation process to give folks time to adapt.

twz123:
@psaradhi Are you still planning to work on this?

PR #6727, opened 2025-11-21
Add ability to use `ENVIRON` for passing bootstrap token

s3rj1k:
Adds two new flags to improve k0s installation and token management for both worker and controller:

  - `--start`: Automatically starts the k0s service immediately after installation
  - `--token-env`: Allows specifying the join token via an environment variable name instead of passing it via file

Example Usage

Install and start controller with token from environment
```
export K0S_TOKEN="your-token-here"
k0s install controller --token-env K0S_TOKEN --start
```

Install and start worker with token from environment
```
k0s install worker --token-env K0S_TOKEN --start
```

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -182,10 +182,31 @@ func (c *command) start(ctx context.Context, flags *config.ControllerOptions, de
 
 	var joinClient *token.JoinClient
 
-	if (c.TokenArg != "" || c.TokenFile != "") && c.needToJoin(nodeConfig) {
+	if (c.TokenArg != "" || c.TokenFile != "" || c.TokenEnv != "") && c.needToJoin(nodeConfig) {
+		tokenSources := 0
+		if c.TokenArg != "" {
+			tokenSources++
+		}
+		if c.TokenFile != "" {
+			tokenSources++
+		}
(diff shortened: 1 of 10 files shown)


twz123:
Can we split this up into two PRs, one for the start flag, and the other one for the environment var? /cc #5253

s3rj1k:
> Can we split this up into two PRs, one for the start flag, and the other one for the environment var? /cc #5253

Probably, but I need both to be in binary at same time

jnummelin:
Let's have these two things as separate PRs, and issues for them, so we can properly handle those as separate as they really are separate

s3rj1k:
> Let's have these two things as separate PRs, and issues for them, so we can properly handle those as separate as they really are separate

RN I still need this PR to test things, when I will finish testing related work against this, I'll prepare 2 separate PRs and move this one into Draft

s3rj1k:
I've extracted `--start` feature into separate PR https://github.com/k0sproject/k0s/pull/6753

and `--token-env` into https://github.com/k0sproject/k0s/pull/6755

s3rj1k:
Both got merged, safe to close
https://github.com/k0sproject/k0s/pull/6766
https://github.com/k0sproject/k0s/pull/6753

Issue #6733, opened 2025-11-21
Allow splitting k0s into embedded binaries and a stripped k0s executable


twz123:
### Is your feature request related to a problem? Please describe.

The k0s executables distributed via the GitHub Releases page include all their embedded binaries. These binaries are extracted by k0s at runtime. Over the years, there have been several requests for a way to extract these binaries _without_ actually starting k0s. This is useful when building filesystem images or overlays for downstream distribution. It may also be of interest to Kairos or other immutable/host-image systems.

Currently, there is no practical (let alone supported) way to derive these components from an official release artifact.

### Describe the solution you would like

Introduce a `k0s payload split` command that disassembles the current k0s executable its constituents:

- Extract all embedded binaries into a user-specified directory.
- Optionally write a stripped k0s executable with the embedded payload removed.

This is a proposal for the  help text:

```console
$ k0s payload split --help
Splits off the embedded binaries and writes them to the given directory.

Usage:
      k0s payload split [output-dir]

Flags:
      --stripped-executable   Path to write the stripped k0s executable to
```

Producing a filesystem overlay could look like this:

```sh
k0s payload split root/usr/libexec/k0s \
  --stripped-executable root/usr/bin/k0s
```

Resulting layout:

```text
root/
  usr/
    bin/
      k0s
    libexec/
      k0s/
        kube-apiserver
        kube-controller-manager
        ...
```

### Describe alternatives you've considered

Publish all artifacts as part of the release pipeline. This is a viable alternative. However, there are already a lot of release artifacts. We would need to produce two more per target triple: the stripped k0s executable and a tar file containing the embedded binaries. If folks start asking for additional signatures and hashes for those files, more artifacts will need to be added. This is getting out of hand.

### Additional context

Introducing a top-level `payload` subcommand creates space for additional functionality in the future, such as listing and inspecting or even modifying the embedded payload (see #6731).

See also:

- #5468
- #5853
- #1016

twz123:
> Currently, there is no practical (let alone supported) way to derive these components from an official release artifact.

I stand corrected: There's at least a way to prepare the whole data directory without actually starting a controller: `--init-only`. This doesn't work for workers (yet), and servers another use case, but it could at least partly be exploited to get hold of some of the executables.

PR #6737, opened 2025-11-25
Add `k0s payload extract` command

jnummelin:
## Description

Add command to pre-extract embedded binaries to <data-dir>/bin before starting k0s. The command extracts all embedded binaries (containerd, kubelet, etcd, etc.) to <data-dir>/bin, defaulting to /var/lib/k0s/bin. When k0s starts, it will use the pre-extracted binaries instead of extracting them again.

Part of #6733

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/cmd/payload/extract.go
+++ b/cmd/payload/extract.go
@@ -1,0 +1,90 @@
+// SPDX-FileCopyrightText: 2025 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package payload
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/k0sproject/k0s/cmd/internal"
+	"github.com/k0sproject/k0s/pkg/assets"
+	"github.com/k0sproject/k0s/pkg/constant"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func NewExtractCmd() *cobra.Command {
+	var dataDir string
+	debugFlags := (&internal.DebugFlags{}).LongRunning()
+
+	cmd := &cobra.Command{
+		Use:   "extract",
+		Short: "Extract embedded binaries",
+		Args:  cobra.NoArgs,
+		Long: `Extract all embedded binaries to the bin directory.
+
+This command extracts all binaries embedded in the k0s executable to the
+bin directory. By default, binaries are extracted to /var/lib/k0s/bin.
+
+When a custom data-dir is specified, binaries are extracted to <data-dir>/bin.
+
+This is useful for pre-staging binaries before starting k0s, especially in
+airgapped environments or when the k0s binary is stored on a read-only
+filesystem.
+
+Examples:
+  # Extract to default location (/var/lib/k0s/bin)
+  k0s payload extract
+
+  # Extract to custom location
+  k0s payload extract --data-dir /custom/path
+`,
+		PersistentPreRun: debugFlags.Run,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return extractPayload(dataDir)
+		},
+	}
+
+	cmd.Flags().StringVar(&dataDir, "data-dir", constant.DataDirDefault, "Data directory for k0s")
+	debugFlags.AddToFlagSet(cmd.Flags())
+
+	return cmd
+}
+
+func extractPayload(dataDir string) error {
+	// Resolve absolute path
+	absDataDir, err := filepath.Abs(dataDir)
+	if err != nil {
+		return fmt.Errorf("failed to resolve data directory path: %w", err)
+	}
+
+	binDir := filepath.Join(absDataDir, "bin")
+	logrus.Infof("Extracting embedded binaries to %s", binDir)
+
+	// Create bin directory if it doesn't exist
+	if err := os.MkdirAll(binDir, 0755); err != nil {
+		return fmt.Errorf("failed to create bin directory: %w", err)
+	}
+
+	// Get list of all embedded binaries
+	binaries := assets.GetEmbeddedBinaries()
+	if len(binaries) == 0 {
+		return errors.New("no payload found")
+	}
+
+	// Extract each binary
+	for _, binaryName := range binaries {
+		_, err := assets.StageExecutable(binDir, binaryName)
+		if err != nil {
+			return fmt.Errorf("failed to stage %s: %w", binaryName, err)
+		}
+
+		logrus.Infof("Extracted %s", binaryName)
+	}
+
+	return nil
+}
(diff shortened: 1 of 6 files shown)


twz123:
What about calling this `split`, as in the proposal? Another idea would be `disassemble`.

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+bin directory. By default, binaries are extracted to /var/lib/k0s/bin.
+
+When a custom data-dir is specified, binaries are extracted to <data-dir>/bin.
+
+This is useful for pre-staging binaries before starting k0s, especially in
+airgapped environments or when the k0s binary is stored on a read-only
+filesystem.
+
+Examples:
+  # Extract to default location (/var/lib/k0s/bin)
```
twz123, line 40: When would this default be useful? Personally, I wouldn't structure this subcommand this way and leave it completely independent from data-dir.

We already have `k0s controller --init-only` which does sort of this. Granted, it's not implemented for `k0s worker` yet, but seems to fit the use case better.

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+  k0s payload extract
+
+  # Extract to custom location
+  k0s payload extract --data-dir /custom/path
```
twz123, line 44: I had some proposals around the UX in #6733. You'd prefer this one?

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// Initialize logging
+			internallog.InitLogging()
+			debugFlags.Run(cmd, args)
+			return nil
+		},
```
twz123: The logging initialization happens in `main.go`. This is how other subcommands are doing it:

```suggestion
		PersistentPreRun: debugFlags.Run,
```

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// Get list of all embedded binaries
+	binaries := assets.GetEmbeddedBinaries()
+	if len(binaries) == 0 {
+		logrus.Warn("No embedded binaries found in this k0s build")
+		return nil
+	}
+
+	logrus.Infof("Found %d embedded binaries", len(binaries))
+
```
twz123:
```suggestion
```

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+	// Create bin directory if it doesn't exist
+	if err := os.MkdirAll(binDir, 0755); err != nil {
+		return fmt.Errorf("failed to create bin directory: %w", err)
+	}
+
+	// Get list of all embedded binaries
+	binaries := assets.GetEmbeddedBinaries()
+	if len(binaries) == 0 {
+		logrus.Warn("No embedded binaries found in this k0s build")
+		return nil
```
twz123:
```suggestion
		return errors.New("no payload found")
```

or sth. along these lines?

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	logrus.Infof("Found %d embedded binaries", len(binaries))
+
+	// Extract each binary
+	extracted := 0
+	failed := 0
+
+	for _, binaryName := range binaries {
+		_, err := assets.StageExecutable(binDir, binaryName)
```
twz123, line 81: I'm not convinced that we should reuse this function, as this is tailor-made for managing executables in k0s's data directory. It does quite some stuff that is not really of interest here, e.g. PATH lookups, applying some special permissions,  ...

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil
+	}
+
+	logrus.Infof("Found %d embedded binaries", len(binaries))
+
+	// Extract each binary
+	extracted := 0
+	failed := 0
+
+	for _, binaryName := range binaries {
```
twz123: Personally, I don't think this adds much value.

```suggestion
	// Extract each binary
	for _, binaryName := range binaries {
```

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+		_, err := assets.StageExecutable(binDir, binaryName)
+		if err != nil {
+			logrus.Warnf("Failed to extract %s: %v", binaryName, err)
+			failed++
+			continue
+		}
+
+		logrus.Infof("Extracted %s", binaryName)
```
twz123, line 86: I think we can just be fail-fast here:

```suggestion
		path, err := assets.StageExecutable(binDir, binaryName)
		if err != nil {
			return fmt.Errorf("failed to stage %s: %w", path, err)
		}

		logrus.Info(path)
```

cmd/payload/extract.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	// Summary
+	logrus.Infof("Extraction complete: %d extracted, %d failed", extracted, failed)
+
+	if failed > 0 {
+		return fmt.Errorf("failed to extract %d binaries", failed)
+	}
+
+	return nil
```
twz123:
```suggestion
	return nil
```

twz123 (commented):
I think this is besides the point of the actual issue. It was not about pre-populating the data directory (we have `--init-only` for that, despite it could be improved for workers, as well), it was about low-level plumbing with the k0s payload directly.

cmd/payload/extract.go
jnummelin: > We already have k0s controller --init-only which does sort of this.

That does A LOT more than just split/extract. It also initializes all the certs as it runs through all the components `Init()` funcs

> When would this default be useful? 

It is the default location for k0s, hence default also here. The two general use cases for this, in my head, are:
- baking some sort of OS images: actual full ISO/qcow/qht-ever or sysext
- baking OS packages like RPMs,DEBs etc.

With this one can pre-populate the FS layout with the expected "defaults" k0s uses in any other case

jnummelin:
> What about calling this split, as in the proposal?

Didn't wanna, yet, go for that since this does not really "split" anything. Split in my head mean that we'd actually also rip out the embedded payload from the original binary but since this is not covered, yet, I thought `extract` makes more sense.

cmd/payload/extract.go
jnummelin: I'll have a look at the lower level plumbing for this to see what's reusable. Don't wanna re-implement lot of that stuff. I agree the `StageExecutable` does bit more than needed here but then again, it does fullfill the need here too 🤷

cmd/payload/extract.go
jnummelin: I agree

PR #6741, opened 2025-11-26
host:port support for konnectivity externalAddress

makhov:
## Description

To make it consistent with `api.externalAddress` and to unblock proper k0smotron ingress support fix we need a way to decouple agentPort from proxyServerPort.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

- [xx] My code follows the style [guidelines](https://docs.k0sproject.io/head/contributors/) of this project
- [xx] My commit messages are [signed-off](https://docs.k0sproject.io/head/contributors/github_workflow/)

### DIFF
--- a/pkg/component/controller/konnectivityagent.go
+++ b/pkg/component/controller/konnectivityagent.go
@@ -6,8 +6,12 @@ package controller
 import (
 	"context"
 	"fmt"
+	"net"
 	"path/filepath"
+	"strconv"
 	"time"
+
+	"github.com/sirupsen/logrus"
 
 	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
@@ -16,7 +20,6 @@ import (
 	"github.com/k0sproject/k0s/pkg/component/prober"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
-	"github.com/sirupsen/logrus"
 )
 
 type KonnectivityAgent struct {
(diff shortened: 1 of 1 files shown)


pkg/component/controller/konnectivityagent.go
```diff
@@ -103,11 +106,26 @@ func (k *KonnectivityAgent) writeKonnectivityAgent(clusterConfig *v1beta1.Cluste
 	if err != nil {
 		return err
 	}
+
+	proxyServerHost := k.APIServerHost
```
twz123, line 110: Oops. Just realizing that we're not reconciling the external Konnectivity address.

pkg/component/controller/konnectivityagent.go
```diff
@@ -103,11 +106,26 @@ func (k *KonnectivityAgent) writeKonnectivityAgent(clusterConfig *v1beta1.Cluste
...
+	// We don't use k0snet.ParseHostPort here because the API server host might be an IP
+	// literal (IPv6). We don't want to change the current behavior and fail here. So we
+	// just use the standard library function and change default values only if we successfully parsed host and port.
+	host, port, _ := net.SplitHostPort(k.APIServerHost)
+	if host != "" {
+		proxyServerHost = host
+	}
+	if p, _ := strconv.Atoi(port); p != 0 {
+		proxyServerPort = int32(p)
+	}
```
twz123, line 122: I wonder if this is the same as:


```suggestion
	serverEndpoint, err := k0snet.ParseHostPortWithDefault(
		cmp.Or(clusterConfig.Spec.Konnectivity.ExternalAddress, k.APIServerHost),
		uint16(clusterConfig.Spec.Konnectivity.AgentPort),
	)
	if err != nil {
		return fmt.Errorf("failed to determine server endpoint: %w", err)
	}
```

twz123 (dismissed):

pkg/component/controller/konnectivityagent.go
makhov: No, it's not. k.APIServerHost is the ExternalAddress, we just do it in the func above. And I left a comment why we can't use ParseHostPort/ParseHostPortWithDefault

twz123: I did read the comment, but I didn't understand it at first, since it didn't occur to me that ParseHostPortWithDefault is not able to cope with IPv6 literals 🙈

twz123 (approved):

PR #6753, opened 2025-12-01
Add `--start` flag to install command

s3rj1k:
Adds new flag to improve k0s installation experience for both worker and controller:

`--start`: Automatically starts the k0s service immediately after installation

Related: https://github.com/k0sproject/k0s/pull/6727

### DIFF
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -11,7 +11,6 @@ import (
 	"github.com/k0sproject/k0s/cmd/internal"
 	"github.com/k0sproject/k0s/pkg/install"
 
-	"github.com/kardianos/service"
 	"github.com/spf13/cobra"
 )
 
@@ -27,15 +26,7 @@ func NewStartCmd() *cobra.Command {
 			if runtime.GOOS != "windows" && os.Geteuid() != 0 {
 				return errors.New("this command must be run as root")
 			}
-			svc, err := install.InstalledService()
-			if err != nil {
-				return err
-			}
-			status, _ := svc.Status()
-			if status == service.StatusRunning {
-				return errors.New("already running")
-			}
-			return svc.Start()
+			return install.StartInstalledService(false)
 		},
 	}
 


cmd/install/install.go
```diff
@@ -38,9 +41,15 @@ func NewInstallCmd() *cobra.Command {
...
+	pflags.BoolVar(&installFlags.start, "start", false, "start the service immediately after installation")
 
 	cmd.AddCommand(installWorkerCmd(&installFlags))
 	addPlatformSpecificCommands(cmd, &installFlags)
 
 	return cmd
 }
+
+// startInstalledService starts (or restarts with force) the installed k0s service.
+func startInstalledService(force bool) error {
```
twz123: Can we inline this function?

pkg/install/service.go
```diff
@@ -122,3 +123,22 @@ func GetServiceConfig(role string) *service.Config {
 		Description: k0sDescription,
 	}
 }
+
+// StartInstalledService starts (or restarts with force) the installed k0s service.
+func StartInstalledService(force bool) error {
```
twz123, line 128: I'd rather refactor the `k0s start` subcommand's `Run` function into something that can be called from here, so that both the `--start` flag and the `start` subcommand call the same code. We could add a `--force` flag to `start`, as well, so the functionality is truly identical.

cmd/install/install.go
s3rj1k: ~~No, look at https://github.com/k0sproject/k0s/pull/6727/files, it is being used more than once~~

Ah, pardon, yes, we can probably call `install.StartInstalledService(` directly

pkg/install/service.go
s3rj1k: How refactoring `start` would help me using `--start` during single install command?

makhov: While I agree, I'd prefer to keep refactoring out of scope of the PR that adds a new feature.

cmd/install/install.go
s3rj1k: Should hopefully be handled by https://github.com/s3rj1k/k0s/commit/bd92ba03304fece7498d016a8c7efa82323d7abe

pkg/install/service.go
twz123: New features sometimes require refactoring. Duplicating code means that there's the chance that `k0s install --start` will be different from `k0s install && k0s start`, which I'd rather like to avoid TBH.

twz123: This would already do the trick I guess:

```diff
diff --git a/cmd/start/start.go b/cmd/start/start.go
index 6e0425eda..c6fd82d68 100644
--- a/cmd/start/start.go
+++ b/cmd/start/start.go
@@ -11,7 +11,6 @@ import (
 	"github.com/k0sproject/k0s/cmd/internal"
 	"github.com/k0sproject/k0s/pkg/install"
 
-	"github.com/kardianos/service"
 	"github.com/spf13/cobra"
 )
 
@@ -27,15 +26,7 @@ func NewStartCmd() *cobra.Command {
 			if runtime.GOOS != "windows" && os.Geteuid() != 0 {
 				return errors.New("this command must be run as root")
 			}
-			svc, err := install.InstalledService()
-			if err != nil {
-				return err
-			}
-			status, _ := svc.Status()
-			if status == service.StatusRunning {
-				return errors.New("already running")
-			}
-			return svc.Start()
+			return install.StartInstalledService(false)
 		},
 	}
 
```

cmd/install/worker.go
```diff
@@ -37,6 +37,12 @@ All default values of worker command will be passed to the service stub unless o
 				return fmt.Errorf("failed to install worker service: %w", err)
 			}
 
+			if installFlags.start {
+				if err := install.StartInstalledService(installFlags.force); err != nil {
```
twz123, line 41: Not directly related to this PR, but just realizing: What happens if we are trying to `install --force` over a running service? Changing the service config while a service is running is probably not the best of ideas? Would it be better for `--force` to stop the service first, then install it, and start it afterwards? I could file an issue for that, if you think it's worth caring about.

s3rj1k: In case of systemd it will rewrite unit file and do daemon-read all of of that os perfectly fine to do on a running service.

pkg/install/service.go
s3rj1k: Isn't this changing a behavior?, lets create issue for that.

s3rj1k: ok, @makhov helped me to understand what this is about, i'll fixitup later today

cmd/install/worker.go
twz123: What about other init systems?

s3rj1k: pretty sure they don't care also, in Linux you can modify a running bash script in place and nothing will happen to your application

pkg/install/service.go
s3rj1k: should be handled by https://github.com/k0sproject/k0s/pull/6753/commits/61ed06795267ca99322faabc3aa801291b4b2072

makhov (approved):

twz123:
Can we have some doc changes for this?

> Could you add some examples to the docs, as well? For example, we could mention this in the quick start guide, à la "Alternatively, both steps can be combined with the `--start` flag ..." and so on. Maybe there's other places in the docs where it makes sense to drop a line, too.

s3rj1k:
> Can we have some doc changes for this?
> 
> > Could you add some examples to the docs, as well? For example, we could mention this in the quick start guide, à la "Alternatively, both steps can be combined with the `--start` flag ..." and so on. Maybe there's other places in the docs where it makes sense to drop a line, too.

I'll do it in a followup

twz123:
> I'll do it in a followup

Alright (although there's always the risk that it will slip through :see_no_evil:)

s3rj1k:
> > I'll do it in a followup
> 
> Alright (although there's always the risk that it will slip through 🙈)

I am on it, we will not loose that https://github.com/k0sproject/k0s/pull/6784

PR #6755, opened 2025-12-01
Add `--token-env` flag to install command

s3rj1k:
Adds new flags to improve k0s token management experience for both worker and controller:

`--token-env`: Allows specifying the join token via an environment variable name instead of passing it via file

Related: https://github.com/k0sproject/k0s/pull/6727

### DIFF
--- a/cmd/install/controller.go
+++ b/cmd/install/controller.go
@@ -49,9 +49,19 @@ With the controller subcommand you can setup a single node cluster by running:
 				return fmt.Errorf("invalid node config: %w", errors.Join(errs...))
 			}
 
+			// Convert --token-env to --token-file
+			tokenFilePath, err := handleTokenEnv(cmd, k0sVars.DataDir)
+			if err != nil {
+				return err
+			}
+
 			flagsAndVals, err := cmdFlagsToArgs(cmd)
 			if err != nil {
 				return err
+			}
+
+			if tokenFilePath != "" {
+				flagsAndVals = append(flagsAndVals, "--token-file="+tokenFilePath)
 			}
 
 			systemUsers := nodeConfig.Spec.Install.SystemUsers
--- a/cmd/worker/worker.go
+++ b/cmd/worker/worker.go
@@ -80,7 +80,7 @@ func NewWorkerCmd() *cobra.Command {
 				c.TokenArg = args[0]
 			}
 
-			getBootstrapKubeconfig, err := kubeconfigGetterFromJoinToken(c.TokenFile, c.TokenArg)
+			getBootstrapKubeconfig, err := kubeconfigGetterFromJoinToken(c.TokenFile, c.TokenEnv, c.TokenArg)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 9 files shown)


cmd/install/controller.go
```diff
@@ -49,11 +49,21 @@ With the controller subcommand you can setup a single node cluster by running:
 				return fmt.Errorf("invalid node config: %w", errors.Join(errs...))
 			}
 
+			// Convert --token-env to --token-file
```
twz123, line 52: Personally, I think this is a very unintuitive thing to do. By nature, an environment variable is ephemeral and belongs to the current process only. Capturing it in a file is not something I'd want to happen with it. Moreover, this creates a discrepancy between the arguments passed to the `install controller/worker` command and the arguments passed to `controller/worker` command, which we have managed to avoid so far. If folks need to persist the token, then they can just use `--token-file` without surprises.

cmd/worker/worker.go
```diff
@@ -193,7 +220,7 @@ func loadKubeconfigFromJoinToken(tokenData string) (*clientcmdapi.Config, error)
 	return kubeconfig, nil
 }
 
-func loadKubeconfigFromTokenFile(path string) (*clientcmdapi.Config, error) {
+func loadKubeconfigFromToken(path string) (*clientcmdapi.Config, error) {
```
twz123, line 223: It's still reading a token file, so we could keep the old name.

pkg/config/cli.go
```diff
@@ -251,6 +252,7 @@ func GetWorkerFlags() *pflag.FlagSet {
 	flagset.StringVar(&workerOpts.WorkerProfile, "profile", defaultWorkerProfile, "worker profile to use on the node")
 	flagset.BoolVar(&workerOpts.CloudProvider, "enable-cloud-provider", false, "Whether or not to enable cloud provider support in kubelet")
 	flagset.StringVar(&workerOpts.TokenFile, "token-file", "", "Path to the file containing join-token.")
+	flagset.StringVar(&workerOpts.TokenEnv, "token-env", "", "Environment variable name containing the join-token.")
```
twz123, line 255: I'd rather not add an additional flag for this. Instead, I'd simply check for `K0S_JOIN_TOKEN` if the token isn't provided as an argument or a file.

twz123 (commented):
Can you mention this somewhere in the docs, too? /cc #5253

pkg/config/cli.go
s3rj1k: Hiding flag means, extra document section about secret environ values

cmd/install/controller.go
s3rj1k: Are you fine with keeping temporary bootstrap token inside init system file instead of keeping pointing it to a file?

And no, there are use cases that do not allow just using "--token-file without surprises"

twz123: @makhov pointed out that there's a straight forward solution which wouldn't require any special-handling at all here: We could simply use `k0s install -e K0S_JOIN_TOKEN controller [...]` and be done with it. That would store the token somewhere in the init system config, and the behavior would be much more obvious. /cc #6763

pkg/config/cli.go
twz123: Yes, we need to document that.

cmd/install/controller.go
s3rj1k: yes, I am working on this RN, ETA PR today

s3rj1k: alternative PR https://github.com/k0sproject/k0s/pull/6766

s3rj1k:
Turning this to draft, https://github.com/k0sproject/k0s/pull/6766 should supersede this PR

s3rj1k:
https://github.com/k0sproject/k0s/pull/6766 got merged, we can abandon this.

PR #6766, opened 2025-12-03
Allow using K0S_TOKEN environ as source of Join token

s3rj1k:
Implements ability to pass Join token via K0S_TOKEN environ.

```
How to test (Needs 3 test nodes):

CP Node 1
    k0s install controller \
        --enable-dynamic-config \
        --disable-components=konnectivity-server \
        --enable-worker \
        --no-taints \
        --kubelet-root-dir=/var/lib/kubelet \
        --verbose
    systemctl enable --now k0scontroller

    k0s token create
    k0s token create --role controller

Worker Node 1
    k0s install worker --env K0S_TOKEN="***" --kubelet-root-dir=/var/lib/kubelet --verbose
    systemctl enable --now k0sworker

CP Node 2
    k0s install controller \
        --enable-dynamic-config \
        --disable-components=konnectivity-server \
        --enable-worker \
        --no-taints \
        --kubelet-root-dir=/var/lib/kubelet \
        --env K0S_TOKEN="***" \
        --verbose
    systemctl enable --now k0scontroller
```

This is alternative to https://github.com/k0sproject/k0s/pull/6755

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -93,6 +93,9 @@ func NewControllerCmd() *cobra.Command {
 			if len(args) > 0 {
 				c.TokenArg = args[0]
 			}
+			if err := internal.CheckSingleTokenSource(c.TokenArg, c.TokenFile); err != nil {
+				return err
+			}
 			if err := controllerFlags.Normalize(); err != nil {
 				return err
 			}
--- a/cmd/internal/tokendata.go
+++ b/cmd/internal/tokendata.go
@@ -12,13 +12,9 @@ import (
 // EnvVarToken is the environment variable name for the join token
 const EnvVarToken = "K0S_TOKEN"
 
-// GetTokenData resolves the join token from multiple possible sources:
-// CLI argument, token file, or K0S_TOKEN environment variable.
-// Returns the token data or an error if multiple sources are provided.
-// Returns empty string if no token source is available.
-func GetTokenData(tokenArg, tokenFile string) (string, error) {
-	tokenEnvValue := os.Getenv(EnvVarToken)
-
+// CheckSingleTokenSource verifies that at most one token source is provided.
+// Returns an error if multiple sources are specified.
+func CheckSingleTokenSource(tokenArg, tokenFile string) error {
 	tokenSources := 0
 	if tokenArg != "" {
 		tokenSources++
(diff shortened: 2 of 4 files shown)


pkg/token/tokendata.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	if tokenFile != "" {
+		tokenSources++
+	}
+	if tokenEnvValue != "" {
+		tokenSources++
+	}
+
+	if tokenSources > 1 {
+		return "", fmt.Errorf("you can only pass one token source: either as a CLI argument, via '--token-file [path]', or via the %s environment variable", constant.EnvVarToken)
```
twz123: I wouldn't make the presence of the environment variable a hard error. It should rather be a fallback if the token hasn't been passed as an arg or via the file flag.

cmd/internal/tokendata.go
twz123, line 1: Let's move this to `cmd/internal`.

pkg/constant/constant.go
```diff
@@ -127,6 +127,9 @@ const (
 	ClusterConfigObjectName = "k0s"
 
 	K0SNodeRoleLabel = "node.k0sproject.io/role"
+
+	// EnvVarToken is the environment variable name for the join token
+	EnvVarToken = "K0S_TOKEN"
```
twz123: I'd move this into `tokendata.go`.

cmd/controller/controller.go
```diff
@@ -89,9 +92,6 @@ func NewControllerCmd() *cobra.Command {
 			if len(args) > 0 {
 				c.TokenArg = args[0]
 			}
-			if c.TokenArg != "" && c.TokenFile != "" {
```
twz123, line 93: I'd prefer to leave this check in place (or refactor it out and reuse it here via a function, whichever makes more sense.) We should verify the command line's correctness before attempting to do anything.

twz123 (commented):
Can you add some sentences / an example about the environment variable to the documentation, e.g. to `k0s-multi-node.md`?

pkg/token/tokendata.go
s3rj1k: It would be unclear in that case what token was used, do you want to add ambiguous behavior and solve that by extending documentation and telling people why passing token via env won't take precedence over a flag?

twz123: I guess it's standard behavior, isn't it? We have a flag (although it's an optional arg in this case), and a related environment variable as a fallback. If the flag (arg) is present, it'll use that, if the flag is missing, it'll fall back to the env.

s3rj1k: no, it is not standard, what you are describing is optional reconfiguration with environ, PR is after having a clear way of setting token only with one option and if that fails report error instead of trying to run k0s cluster and fail expectation

cmd/controller/controller.go
s3rj1k: hopefully handled by `internal.CheckSingleTokenSource`

jnummelin (approved):

PR #6768, opened 2025-12-03
Enable "k0s reset" on Windows

kke:
## Description

Enable "k0s reset" subcommand on Windows workers.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [X] Auto test added

## Checklist

### DIFF
--- a/pkg/cleanup/cni_windows.go
+++ b/pkg/cleanup/cni_windows.go
@@ -6,30 +6,177 @@
 package cleanup
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
 )
 
 // Run removes Windows CNI artifacts.
 func (c *cni) Run() error {
+	removeCNIConfigFiles()
+	removeCalicoServices()
+	cleanupHNSArtifacts()
+	cleanupVethernetAdapters()
+	cleanupCalicoEnvVars()
+	cleanupCalicoFirewallRules()
+	return nil
+}
+
+// removeCNIConfigFiles deletes all known Windows CNI config artifacts from disk.
+func removeCNIConfigFiles() {
+	logrus.Debug("removing Windows CNI configuration files")
 	var errs []error
 
 	files := []string{
-		`C:\etc\cni\net.d\10-calico.conflist`,
-		`C:\etc\cni\net.d\calico-kubeconfig`,
-		`C:\etc\cni\net.d\10-kuberouter.conflist`,
+		`C:\\etc\\cni\\net.d\\10-calico.conflist`,
+		`C:\\etc\\cni\\net.d\\calico-kubeconfig`,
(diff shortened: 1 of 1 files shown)


kke:
Removing `C:\var\lib\k0s` seems to always fail because some containerd files are in use.

pkg/cleanup/directories_windows.go
```diff
@@ -0,0 +1,59 @@
...
+		if path == "" {
+			continue
+		}
+		removeDirectory(path)
+	}
+	return nil
+}
+
+func removeDirectory(path string) {
+	if out, err := exec.Command("powershell", "-NoProfile", "-Command",
```
kke, line 30: Maybe change back to `os.RemoveAll` as the PowerShell `RemoveAll -Force` doesn't work any better at removing the in-use files.

Alternatives are `takeown` and `icacls` to take ownership of them or maybe some [hcsshim](https://github.com/microsoft/hcsshim) trickery to flush the containers?

pkg/cleanup/steps_linux.go
```diff
@@ -0,0 +1,36 @@
...
+		&services{},
+		&directories{
+			dataDir:        k0sVars.DataDir,
+			kubeletRootDir: k0sVars.KubeletRootDir,
+			runDir:         k0sVars.RunDir,
+		},
+		&cni{},
+	}
+
+	if bridge := newBridgeStep(); bridge != nil {
```
juanluisvaladas, line 31: Doesn't calico leave any leftovers on windows?

pkg/cleanup/directories_windows.go
juanluisvaladas: Not too sure about this, I think it's OK for now, if in the future there is a better proposal I it's worth looking into it.

pkg/uninstall/privileges_windows.go
```diff
@@ -0,0 +1,49 @@
...
+package uninstall
+
+import (
+	"errors"
+	"fmt"
+
+	"golang.org/x/sys/windows"
+)
+
+func ensurePrivileges() error {
```
juanluisvaladas, line 15: I'm not too sure about this whole block, but it's passing CI so I guess it's fine.

kke: It's quite complicated for just verifying that user is admin.

kke: Maybe the powershell stuff could be written in native go and the token stuff could be used for that too. (it would probably be a lot more complicated)

pkg/cleanup/directories_windows.go
jnummelin: > Alternatives are takeown and icacls to take ownership of them or maybe some hcsshim trickery to flush the containers?

I think we can revisit this as @juanluisvaladas also mentioned. This area seems to be very error prone on Windows in general and I haven't found any clean way to do this yet.

pkg/uninstall/privileges_windows.go
jnummelin: IMO this is good enough

PR #6874, opened 2025-12-18
Add validation for IPv6 ServiceCIDR prefix size

### LINKED ISSUE #6545: k0s crashes when ipv6 prefix is too large
when setting `spec.k0s.config.spec.network.dualStack.IPv6serviceCIDR` to a prefix >108 and then running k0sctl apply, the k0scontroller service crashes with
```
 --service-cluster-ip-range[1] is too large; for 128-bit addresses, the mask must be >= 108
```

but because k0scontroller is not running, running k0sctl apply again results in

```
 - [ssh] x.x.x.x:22: k0s doesn't appear to be running but has been installed as a service at /etc/systemd/system/k0scontroller.service - please remove it or start the service 
```

To fix it you have to correct the rendered `/etc/k0s/k0s.yaml` on the host(s). Not a big deal, but it would be nice to validate the config before applying to avoid cluster downtime

DeeBi9:
## Description
Adds validation for IPv6 ServiceCIDR prefix length (<= /108) per k8s requirement.

Fixes #6545

## Type of change

## How Has This Been Tested?
- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -101,7 +101,7 @@ func (n *Network) Validate() []error {
 		validCIDRs = false
 	}
 
-	serviceNetIP, _, err := net.ParseCIDR(n.ServiceCIDR)
+	serviceNetIP, serviceNet, err := net.ParseCIDR(n.ServiceCIDR)
 	if err != nil {
 		errors = append(errors, field.Invalid(field.NewPath("serviceCIDR"), n.ServiceCIDR, "invalid CIDR address"))
 		validCIDRs = false
@@ -112,6 +112,15 @@ func (n *Network) Validate() []error {
 		// if they have a valid IP to begin with because otherwise the error can be confusing
 		if (podNetIP.To4() != nil) != (serviceNetIP.To4() != nil) {
 			errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "podCIDR and serviceCIDR must be both IPv4 or IPv6"))
+		}
+
+		// Validate IPv6 ServiceCIDR prefix length per Kubernetes requirements (<= /108).
+		// https://github.com/kubernetes/kubernetes/blob/v1.34.3/cmd/kube-apiserver/app/options/validation.go#L52-L58
+		if serviceNetIP.To4() == nil {
+			ones, bits := serviceNet.Mask.Size()
+			if bits == 128 && ones > 108 {
+				errors = append(errors, field.Invalid(field.NewPath("serviceCIDR"), n.ServiceCIDR, "IPv6 service CIDR prefix must be <= 108"))
+			}
 		}
 
 		// Single stack IPv6 is an alpha feature, so we need to check if the feature gate is enabled, but only report it if both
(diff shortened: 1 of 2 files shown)


pkg/apis/k0s/v1beta1/network.go
```diff
@@ -114,6 +114,15 @@ func (n *Network) Validate() []error {
 			errors = append(errors, field.Invalid(field.NewPath("podCIDR"), n.PodCIDR, "podCIDR and serviceCIDR must be both IPv4 or IPv6"))
 		}
 
+		// Validate IPv6 ServiceCIDR prefix length per Kubernetes requirements (<= /108).
+		// https://github.com/kubernetes/kubernetes/blob/v1.34.2/cmd/kubeadm/app/apis/kubeadm/validation/validation.go#L461-L478
```
twz123: Very good to provide an upstream reference! However, I think this one is more appropriate:

```suggestion
		// https://github.com/kubernetes/kubernetes/blob/v1.34.3/cmd/kube-apiserver/app/options/validation.go#L52-L58
```

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -133,9 +142,14 @@ func (n *Network) Validate() []error {
 		if err != nil {
 			errors = append(errors, field.Invalid(field.NewPath("dualStack", "IPv6podCIDR"), n.DualStack.IPv6PodCIDR, "invalid CIDR address"))
 		}
-		_, _, err = net.ParseCIDR(n.DualStack.IPv6ServiceCIDR)
+		ip, ipNet, err := net.ParseCIDR(n.DualStack.IPv6ServiceCIDR)
 		if err != nil {
 			errors = append(errors, field.Invalid(field.NewPath("dualStack", "IPv6serviceCIDR"), n.DualStack.IPv6ServiceCIDR, "invalid CIDR address"))
+		} else if ip.To4() == nil {
```
twz123, line 148: Ha! Not directly related to this fix, but I think we should generate another validation error if `ip.To4()` is non-nil, i.e. if an IPv4 CIDR has been provided in the IPv6 CIDR config field, for both the Pod CIDR and the Service CIDR. WDYT?

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -133,9 +142,14 @@ func (n *Network) Validate() []error {
...
+			ones, bits := ipNet.Mask.Size()
+			if bits == 128 && ones > 108 {
```
twz123: After I had a look at what kube-apiserver expects, I think we need to check both IPv4 and IPv6 CIDRs if they're not bigger than 20 bits.

https://github.com/kubernetes/kubernetes/blob/v1.34.3/cmd/kube-apiserver/app/options/validation.go#L52-L58

pkg/apis/k0s/v1beta1/network.go
DeeBi9: Yea, that makes sense. This looks like a separate validation check. If you r okay with it, I can open a follow-up issue and work on it in a separate PR.

pkg/apis/k0s/v1beta1/network.go
juanluisvaladas: I'd lile to have a comment here so that we remember to check it during updates:

```suggestion

			ones, bits := ipNet.Mask.Size()
			// https://github.com/kubernetes/kubernetes/blob/v1.34.3/cmd/kube-apiserver/app/options/validation.go#L39
			maxCIDRBits := 20			
			if bits - ones > maxCIDRBits {
```

juanluisvaladas:
The last merge was automated and it duplicated some lines, please rebase and squash it into a single commit.

pkg/apis/k0s/v1beta1/network.go
twz123: * #6887

juanluisvaladas:
Had a quick look at the unit tests and the failures are legit.

twz123 (approved):

PR #6925, opened 2026-01-07
Reclassify controller readiness failures in Autopilot

twz123:
## Description

Previously, problems with controller readiness would halt the execution of an Autopilot plan, requiring manual intervention. This was explicitly documented. However, it is anticipated that controllers won't always be ready, e.g. during restarts.

Don't let a readiness failure terminate plan execution. Instead, requeue the plan and retry later.

This renders the `InconsistentTargets` plan state unnecessary. The `IncompleteTargets` state serves the same purpose. Remove the `InconsistentTargets` state and keep it only as a legacy state in the docs. The controller delegate now reports `Incomplete` rather than `Inconsistent`, which maps to `IncompleteTargets`. At the same time, merge the `MissingSignalNode` into the `IncompleteTargets` target state, as they both represented "a resolved target no longer exists", which was confusing.

Let the ready prober categorize failures into the following categories:

- Target resolution failed (e.g. missing `ControlNode`)
- Readiness probe unsuccessful
- Everything else

The prober returns the worst error encountered and cancels in-flight probes as soon as a target resolution failure is detected. This is later used to determine if a plan needs to go into the `incompleteTarget` state.

While at it, refactor some other parts of the ready prober: Make it package private, removing the public interface, make it concurrency-safe by removing the targets from the struct, pass on the context properly, ...

Note that a missing `ControlNode`'s internal IP address is now considered a transient error. Therefore, a controller that temporarily loses its IP address will not fail the plan permanently. The quorumsafety integration test has been updated to better reflect the intended outcome. It stops a real controller, verifies that the plan remains in a schedulable state as long as not all controllers are ready, restarts the controller, and asserts that the plan completes. This avoids reliance on a fabricated `ControlNode` with no IP address.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/docs/autopilot-multicommand.md
+++ b/docs/autopilot-multicommand.md
@@ -197,7 +197,7 @@ and treats all other states as an error.
 
 ```mermaid
 flowchart TD
-    Errors --> InconsistentTargets
+    Errors -. no longer in use .-> InconsistentTargets
     Errors --> IncompleteTargets
     Errors --> Restricted
     Errors -. no longer in use .-> MissingSignalNode
@@ -205,7 +205,7 @@ flowchart TD
 
 | Error State | Command | States | Description |
 | ----------- | ------- | ------ | ----------- |
-| **InconsistentTargets** | `k0supdate` | **Schedulable** | Indicates that a **Signal Node** probe has failed for any node that was previously discovered during **NewPlan**. |
+| **~~InconsistentTargets~~** | `k0supdate` | **Schedulable** | Legacy state used by older k0s releases to indicate controller readiness failures; no longer in use. |
 | **IncompleteTargets** | `airgapupdate`, `k0supdate` | **NewPlan**, **Schedulable** | Indicates that a **Signal Node** that existed during the discover phase in **NewPlan** no longer exists (ie. no `ControlNode` or `Node` object) |
 | **Restricted** | `airgapupdate`, `k0supdate` | **NewPlan** | Indicates that a **Plan** has requested an update of a **Signal Node** type that contradicts the startup exclusions (the `--exclude-from-plans` argument) |
 | **~~MissingSignalNode~~** | `airgapupdate`, `k0supdate` | **Schedulable** | Legacy state used by older k0s releases to indicate **IncompleteTargets**; no longer in use. |
--- a/docs/autopilot.md
+++ b/docs/autopilot.md
@@ -75,10 +75,10 @@ additional changes to the plan (other than status) will be recognized.
 ### Controller Quorum Safety
 
 * Prior to scheduling a controller update, **autopilot** queries the API server of **all**
-  controllers to ensure that they report a successful `/ready`
+  controllers to ensure that they report a successful `/ready`.
 * Only once all controllers are `/ready` will the current controller get sent update signaling.
-* In the event that **any** controller reports a non-ready, the `Plan` transitions into an
-  `InconsistentTargets` state, and the `Plan` execution ends.
+* If a previously discovered controller target can no longer be resolved, the `Plan` transitions
+  into an `IncompleteTargets` state and execution ends. Transient readiness failures requeue.
 
 ### Controllers Update Sequentially
 
@@ -329,7 +329,7 @@ update operation. There are a number of statuses available:
 | Status | Description | Ends Plan? |
 | ------ | ----------- | ---------- |
 | `IncompleteTargets` | There are nodes in the resolved `Plan` that do not have associated `Node` (worker) or `ControlNode` (controller) objects. | Yes |
-| `InconsistentTargets` | A controller has reported itself as not-ready during the selection of the next controller to update. | Yes |
+| ~~`InconsistentTargets`~~ | Legacy state used by older k0s releases to indicate controller readiness failures; no longer in use. | Yes |
 | `Schedulable` | Indicates that the `Plan` can be re-evaluated to determine which next node to update. | No |
 | `SchedulableWait` | Scheduling operations are in progress, and no further update scheduling should occur. | No |
 | `Completed` | The `Plan` has run successfully to completion. | Yes |
--- a/inttest/ap-quorumsafety/quorumsafety_test.go
+++ b/inttest/ap-quorumsafety/quorumsafety_test.go
@@ -4,17 +4,29 @@
 package quorumsafety
 
 import (
+	"cmp"
+	"context"
+	"errors"
 	"fmt"
+	"reflect"
+	"runtime"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/k0sproject/k0s/internal/sync/value"
+	apv1beta2 "github.com/k0sproject/k0s/pkg/apis/autopilot/v1beta2"
 	apconst "github.com/k0sproject/k0s/pkg/autopilot/constant"
 	appc "github.com/k0sproject/k0s/pkg/autopilot/controller/plans/core"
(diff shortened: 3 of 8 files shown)


ncopa:
The last commit message says:

> This renders the `IncompleteTargets` plan state unnecessary.

I wonder it this is a typo and that it should be `InconsistentTargets`?

twz123:
> The last commit message says:
> 
> > This renders the `IncompleteTargets` plan state unnecessary.
> 
> I wonder it this is a typo and that it should be `InconsistentTargets`?

It is. I corrected it in the PR description, but didn't bother to update the commit. Will do this now.

pkg/autopilot/controller/readyprober.go
```diff
@@ -1,125 +1,151 @@
...
-// As errors are received, they are collected in a single errors channel for post
-// inspection. This function blocks until *all* spawned goroutines have completed
-// or timed-out.
-func (p readyProber) Probe() error {
-	g := errgroup.Group{}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for _, target := range targets {
+		wg.Go(func() {
+			err := p.probeOne(ctx, target)
```
ncopa, line 74: I think we may need a per-iteration copy of `target` for the closure?

```suggestion
	for _, target := range targets {
	    target := target
		wg.Go(func() {
			err := p.probeOne(ctx, target)
```

twz123: This is no longer necessary in more recent Go versions: https://go.dev/blog/loopvar-preview

pkg/autopilot/controller/readyprober.go
```diff
@@ -1,125 +1,151 @@
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			err = fmt.Errorf("while probing %s: %w", target.Name, err)
+
+			if !errors.Is(err, errReadyProbeTargetResolutionFailed) {
+				result.CompareAndSwap(nil, &err)
```
ncopa: Is it safe to use a `&err` pointer here? As I understand, `err` lives on the stack to the goroutine, so when it exits it is no longer save to use the pointer?

twz123: This would be true in a non-garbage-collected language. Here, as long we retain a reference to something, it "escapes to the heap", as Gophers like to call this. We don't have a way to force something to the stack or heap anyways. It's a decision made by the compiler.

twz123: We can also use a mutex instead. Not important to be lock-free here (although something constantly drags me into this direction :see_no_evil:). I'll sketch up an alternative real quick.

twz123: We could do this instead:

```go
func (p *readyProber) probeTargets(ctx context.Context, targets []apv1beta2.PlanCommandTargetStatus) error {
	var (
		wg     sync.WaitGroup
		mu     sync.Mutex
		result error
	)

	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	for _, target := range targets {
		wg.Go(func() {
			err := p.probeOne(ctx, target)
			if err == nil {
				return
			}

			select {
			case <-ctx.Done():
				return
			default:
			}

			targetResolutionFailed := errors.Is(err, errReadyProbeTargetResolutionFailed)
			if targetResolutionFailed {
				cancel() // Be fail-fast
			}

			mu.Lock()
			defer mu.Unlock()
			if result == nil || targetResolutionFailed {
				result = fmt.Errorf("while probing %s: %w", target.Name, err)
			}
		})
	}

	wg.Wait()
	return result
}
```

ncopa: yeah, I like that.

ncopa (dismissed):
good job!

twz123:
@ncopa PTAL again. I've found a bug concerning context cancellation:

```diff
diff --git a/pkg/autopilot/controller/readyprober.go b/pkg/autopilot/controller/readyprober.go
index 889faa180..8bd77bda1 100644
--- a/pkg/autopilot/controller/readyprober.go
+++ b/pkg/autopilot/controller/readyprober.go
@@ -76,12 +76,6 @@ func (p *readyProber) probeTargets(ctx context.Context, targets []apv1beta2.Plan
 				return
 			}
 
-			select {
-			case <-ctx.Done():
-				return
-			default:
-			}
-
 			targetResolutionFailed := errors.Is(err, errReadyProbeTargetResolutionFailed)
 			if targetResolutionFailed {
 				cancel() // Be fail-fast
@@ -89,7 +83,7 @@ func (p *readyProber) probeTargets(ctx context.Context, targets []apv1beta2.Plan
 
 			mu.Lock()
 			defer mu.Unlock()
-			if result == nil || targetResolutionFailed {
+			if result == nil || (targetResolutionFailed && !errors.Is(result, errReadyProbeTargetResolutionFailed)) {
 				result = fmt.Errorf("while probing %s: %w", target.Name, err)
 			}
 		})
@@ -127,12 +121,12 @@ func (p readyProber) probeOne(ctx context.Context, target apv1beta2.PlanCommandT
 		return fmt.Errorf("unable to create HTTP request for %s:%d: %w", address, p.k8sAPIPort, err)
 	}
 	// The body content is not interesting at the moment.
-	res, _, err := probe.Probe(req, p.timeout)
+	res, msg, err := probe.Probe(req, p.timeout)
 	if err != nil {
 		return fmt.Errorf("failed to probe %s: %w", url, err)
 	}
 	if res != k8sprobe.Success {
-		return fmt.Errorf("%w: %s", errUnsuccessfulReadyProbe, res)
+		return fmt.Errorf("%w: %s: %s", errUnsuccessfulReadyProbe, res, msg)
 	}
 
 	p.log.Infof("Probing %s done: %v", target.Name, res)
diff --git a/pkg/autopilot/controller/readyprober_test.go b/pkg/autopilot/controller/readyprober_test.go
index 83af37aad..91f0b99ef 100644
--- a/pkg/autopilot/controller/readyprober_test.go
+++ b/pkg/autopilot/controller/readyprober_test.go
@@ -6,6 +6,7 @@
 package controller
 
 import (
+	"context"
 	"crypto/tls"
 	"net/http"
 	"net/http/httptest"
@@ -89,6 +90,23 @@ func TestReadyProberSuccess(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestReadyProberContextCanceled(t *testing.T) {
+	hostPort := newTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	cf := testutil.NewFakeClientFactory(newControlNode("controller0", hostPort.Host()))
+	targets := []autopilotv1beta2.PlanCommandTargetStatus{{Name: "controller0"}}
+
+	canceled, cancel := context.WithCancel(context.Background())
+	cancel()
+	prober := newTestProber(t, int(hostPort.Port()), cf)
+	err := prober.probeTargets(canceled, targets)
+
+	assert.NotErrorIs(t, err, errReadyProbeTargetResolutionFailed)
+	assert.ErrorContains(t, err, "while probing controller0")
+	assert.ErrorContains(t, err, "context canceled")
+}
+
 // Probe two controllers, in a way that one of them will fail its readiness
 // probe, and the other won't exist, ensuring that the goroutine that fails the
 // readiness probe will terminate first. This will ensure that a target
```

ncopa:
good catch!

Issue #6771, opened 2025-12-03
IPv6 Dual Stack causes kube-router to not start


simonszu:
### Before creating an issue, make sure you've checked the following:

- [x] You are running the latest released version of k0s
- [x] Make sure you've searched for existing issues, both open and closed
- [x] Make sure you've searched for PRs too, a fix might've been merged already
- [x] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.12.57+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.57-1 (2025-11-05) x86_64 GNU/Linux
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.2
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### Version

v1.34.2+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Total memory: 15.6 GiB (pass)
File system of /var/lib/k0s: ext4 (pass)
Disk space available for /var/lib/k0s: 465.7 GiB (pass)
Relative disk space available for /var/lib/k0s: 92% (pass)
Name resolution: localhost: [::1 127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.12.57+deb13-amd64 (pass)
  Max. file descriptors per process: current: 524287 / max: 524288 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
    cgroup controller "memory": available (is a listed root controller) (pass)
    cgroup controller "devices": available (device filters attachable) (pass)
    cgroup controller "freezer": available (cgroup.freeze exists) (pass)
[...]

PR #6941, opened 2026-01-12
feat: enhance dual-stack node ip detection logic

omar-khallaf:
## Description

Filters candidate IPs so they ensure some of the requirements needed by the `k8s.io/kubernetes/pkg/kubelet.validateNodeIP` function and filters any IPs that are found on the `lo` interface.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist


jnummelin:
@omar-mohamed-khallaf Could you elaborate a bit the description on what exact problem this is trying to overcome? Not really clear, for me at least.

omar-khallaf:
Hi @jnummelin.
Trying to figure out what the problem was with #6771, I stripped the part pointed to by [twz123](https://github.com/k0sproject/k0s/issues/6771#issuecomment-3733365317) and found that after fixing the DNS problem: 
- `lookupNodeName` returned addresses that are prohibited by `k8s.io/kubernetes/pkg/kubelet.validateNodeIP`
- Addresses on the `lo` interface were picked up first, so my `ipv4` and `ipv6` were `lo` addresses.

Those addresses on `lo` are configured for other stuff and not usable for `k0s`, so I filtered those addresses in `lookupNodeName`.

Actually, without this modification, I'm getting correct addresses now, but I'm not sure if it's handled elsewhere or it's by luck.

jnummelin:
Right, so basically when k0s is detecting the `node-ip` in case of dual-stack, we try to lookup via hostname. And in case the hostname maps to some address on `lo` interface, things go sideways. Is that correct understanding?

omar-khallaf:
Interestingly, things didn't go sideways :grin:.
To test, I built a small program with `lookupNodeName` logic, not the whole `k0s`.

After removing `--kubelet-extra-args="--node-ip=<ipv4>,<ipv6>"`, `k0s` seems to pick the correct addresses.

I feel, though, that this is by chance; my program had `lo` interfaces first, but `k0s` didn't.

And what happens if one of those invalid IPs per `k8s.io/kubernetes/pkg/kubelet.validateNodeIP` happens to be first?

simonszu:
> Right, so basically when k0s is detecting the `node-ip` in case of dual-stack, we try to lookup via hostname. And in case the hostname maps to some address on `lo` interface, things go sideways. Is that correct understanding?

Hi, yes, for me the installation, or the startup of several system pods afterwards fails because k0s detects the `lo` IP address as my node's primary IP address.

omar-khallaf:
Thanks, this affirms that this PR is indeed necessary.

docs/dual-stack.md
```diff
@@ -29,6 +29,9 @@ spec:
 This configuration will set up all Kubernetes components and Kube-router
 accordingly for dual-stack networking.
 
+> [!IMPORTANT]
+> A working system DNS resolver returning both IPv4 & IPv6 for the node hostname is necessary. See [issue](https://github.com/k0sproject/k0s/issues/6771).
+
```
twz123, line 53: This is only necessary if you rely on IP auto-detection.

Would be maybe better to explain what's happening, and why: Kubelet doesn't support node IP auto-detection in dual-stack mode, hence k0s replicates it. We could say that

1. Node IPs are being looked up via DNS, based on the node name
2. Node IPs are being looked up via network interfaces.

Similar to what the [comment](https://github.com/kubernetes/kubernetes/blob/v1.34.3/pkg/kubelet/nodestatus/setters.go#L150-L153) in upstream Kubernetes code is mentioning.

Moreover, it'd make a lot of sense IMO  to mention the `--node-ip` kubelet flag here (maybe even before starting to explain the auto-detection mechanism), and that you can simply specify that with the two comma-separated IPs to use. Specifying that flag makes all of the rather involved auto-detection stuff superfluous.

pkg/component/worker/kubelet.go
```diff
@@ -77,12 +78,68 @@ func (k *Kubelet) Init(_ context.Context) (err error) {
...
+				}
+
+				ipaddrs = append(ipaddrs, ip)
+			}
+		}
+	}
+	return ipaddrs, nil
+}
+
 func (k *Kubelet) lookupNodeName(ctx context.Context) (ipv4, ipv6 net.IP, _ error) {
```
twz123, line 82: There's some comments at the call site of `lookupNodeName` which need to be adjusted.

pkg/component/worker/kubelet.go
```diff
@@ -77,12 +78,68 @@ func (k *Kubelet) Init(_ context.Context) (err error) {
...
 	ipaddrs, err := net.DefaultResolver.LookupIPAddr(ctx, string(k.NodeName))
 	if err != nil {
 		return nil, nil, err
 	}
 
+	loopbackIPAddrs, err := getLoopbackIPAddresses()
+	if err != nil {
```
twz123: If we really want to replicate the kubelet behavior, then we should be fail-fast and return this err here.

pkg/component/worker/kubelet.go
```diff
@@ -77,12 +78,68 @@ func (k *Kubelet) Init(_ context.Context) (err error) {
...
+	for _, element := range slice {
+		if !predicate(element) {
+			slice[idx] = element
+			idx++
+		}
+	}
+	return slice[:idx]
+}
+
+func getLoopbackIPAddresses() ([]net.IP, error) {
```
twz123: Not sure what you're after here. I'd say it's enough to check the `IsLoopback()` method on an IP address to figure out if it's a loopback address. No need to check the interface.

Did you maybe try to replicate the `validateNodeIP` checks that ensure that only IPs get selected which are bound to a network interface? Then you could simply iterate over `net.InterfaceAddrs()`.

pkg/component/worker/kubelet.go
```diff
@@ -77,12 +78,68 @@ func (k *Kubelet) Init(_ context.Context) (err error) {
...
+	}
+
+	ipaddrs = removeIf(ipaddrs, func(ipaddr net.IPAddr) bool {
+		ip := ipaddr.IP
+		return (ip.To4() == nil && ip.To16() == nil) ||
+			ip.IsLoopback() ||
+			ip.IsMulticast() ||
+			ip.IsLinkLocalUnicast() ||
+			ip.IsUnspecified() ||
+			slices.IndexFunc(loopbackIPAddrs, func(loopbackIp net.IP) bool { return loopbackIp.Equal(ip) }) != -1
```
twz123: I'd simply stuff this into the for loop below, starting at line 143:

```go
for _, addr := range ipaddrs {
	if ip := addr.IP; ip.IsLoopback() || ip.IsMulticast() || ip.IsLinkLocalUnicast() || ip.IsUnspecified() {
		continue
	}
	if slices.ContainsFunc(loopbackIPAddrs, addr.IP.Equal) {
		continue
	}
	// Or maybe use a switch statement instead ...

	...
```

docs/dual-stack.md
twz123: Just realizing that point 2 is not implemented here. I guess it makes sense to add it, especially given that the node name resolution stuff seems to be more brittle than anticipated.

We could use the upstream Kubernetes [net.ResolveBindAddress](https://pkg.go.dev/k8s.io/apimachinery@v0.34.3/pkg/util/net#ResolveBindAddress) function for this, à la

```go
// import utilnet "k8s.io/apimachinery/pkg/util/net"
v4Addr, err := utilnet.ResolveBindAddress(net.IPv4zero)
v6Addr, err := utilnet.ResolveBindAddress(net.IPv6unspecified)
```

pkg/component/worker/kubelet.go
twz123: IIUC, we could also use [net.IsGlobalUnicast](https://pkg.go.dev/net@go1.25.5#IP.IsGlobalUnicast), so this might boil down to:

```go
for _, addr := range ipaddrs {
	if !addr.IP.IsGlobalUnicast() || slices.ContainsFunc(loopbackIPAddrs, addr.IP.Equal) {
		continue
	}

	...
```

docs/dual-stack.md
omar-khallaf: Should I fall back to this method after failing to get valid IPs from DNS?
It seems superior to the DNS solution. Should I replace the DNS solution altogether?

pkg/component/worker/kubelet.go
omar-khallaf: No, I have non-loopback addresses on the `lo` interface (e.g. `10.45.0.1/32`) for other routing stuff.
If memory serves, those addresses didn't work well with `etcd`.

pkg/component/worker/kubelet.go
omar-khallaf: I swear I searched for such a function `ContainsFunc` but couldn't find it :grin: .

pkg/component/worker/kubelet.go
twz123: Pretty sure this won't fly with kubelet's auto-detection for single-stack. I mean, AFAICT kubelet doesn't check for this, so it would pick that IP. I'd rather not have the auto-detection for dual-stack be different from that for single-stack.

I'd say that you should use `--node-ip` in that rather contrived case. (Or try to change kubelet's auto-detection in upstream Kubernetes to do the interface loopback check, as well.)

WDYT @juanluisvaladas?

docs/dual-stack.md
twz123: I'd do it exactly in the [order kubelet does it](https://github.com/kubernetes/kubernetes/blob/v1.34.3/pkg/kubelet/nodestatus/setters.go#L150-L153), for consistency reasons:

> 			// 3) Lookup the IP from node name by DNS
> 			// 4) Try to get the IP from the network interface used as default gateway

pkg/component/worker/kubelet.go
omar-khallaf: I didn't implement the fallback method in `lookupNodeName`. Did you mean they will need to be adjusted for this?

pkg/component/worker/kubelet.go
omar-khallaf: Should I remove it now or wait for Juan?

docs/dual-stack.md
omar-khallaf: Done

pkg/component/worker/kubelet.go
twz123: You can change this to list all public IPs, and then reverse the ContainsFunc check. That way we're mimicking what `validateNodeIP` is doing.

omar-khallaf: Done

pkg/component/worker/kubelet.go
```diff
@@ -35,6 +36,8 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+import utilnet "k8s.io/apimachinery/pkg/util/net"
```
twz123: Can you move this into the import block above?

pkg/component/worker/kubelet.go
```diff
@@ -77,13 +80,42 @@ func (k *Kubelet) Init(_ context.Context) (err error) {
 	return nil
 }
 
+func Map[T any, R any](input []T, transform func(T) R) []R {
```
twz123: You can just do it inline. No need to have such a generic public helper function here.

pkg/component/worker/kubelet.go
```diff
@@ -94,9 +126,27 @@ func (k *Kubelet) lookupNodeName(ctx context.Context) (ipv4, ipv6 net.IP, _ erro
...
+	if ipv4 == nil || ipv6 == nil {
+		return ipv4, ipv6, fmt.Errorf("node name IP address lookup didn't return addresses for both families: IPv4: %s, IPv6: %s", ipv4, ipv6)
+	}
 	return ipv4, ipv6, nil
 }
 
+func resolveNodeAddress() (ipv4 net.IP, ipv6 net.IP, err error) {
+	ipv4, err = utilnet.ResolveBindAddress(net.IPv4zero)
+	if err != nil {
+		return nil, ipv6, fmt.Errorf("node default gateway interface scanning didn't return addresses for both families: IPv4: %s, IPv6: %s", ipv4, ipv6)
```
twz123:
```suggestion
		return nil, nil, fmt.Errorf("failed to determine preferred IPv4 address: %w, err)
```

pkg/component/worker/kubelet.go
```diff
@@ -94,9 +126,27 @@ func (k *Kubelet) lookupNodeName(ctx context.Context) (ipv4, ipv6 net.IP, _ erro
...
+	}
+
+	ipv6, err = utilnet.ResolveBindAddress(net.IPv6unspecified)
+	if err != nil || ipv6.To4() != nil {
+		return ipv4, nil, fmt.Errorf("node default gateway interface scanning didn't return addresses for both families: IPv4: %s, IPv6: %s", ipv4, ipv6)
+	}
```
twz123:
```suggestion
	if err != nil {
		return nil, nil, fmt.Errorf("failed to determine preferred IPv6 address: %w, err)
	}
```

pkg/component/worker/kubelet.go
```diff
@@ -115,14 +165,28 @@ func (k *Kubelet) Start(ctx context.Context) error {
...
+				logrus.Warnf("%s", err)
+			}
+
+			if ipv4 == nil {
+				ipv4 = fallbackIPv4
+			}
+			if ipv6 == nil {
+				ipv6 = fallbackIPv6
+			}
+		}
```
twz123, line 161: AFAICT, kubelet does it a bit differently. So I'd suggest to change it as follows:

- Remove the last check if IPs for both families are found from `lookupNodeName`.
- Only call `ResolveBindAddress` for IPs that haven't been found by `lookupNodeName`.

simonszu:
@omar-mohamed-khallaf not want to sound rude but care to resolve these conflicts please?

omar-khallaf:
> @omar-mohamed-khallaf not want to sound rude but care to resolve these conflicts please?

No problem, done. I don't know why it has taken that much time.

PR #6947, opened 2026-01-13
Add note about unsupported raw git/github chart URLs in Helm repos

### LINKED ISSUE #6341: Document the (dumb) ways to add non-usual repos.

Oh yes. I am trying to install Rancher's `local-path-provisioner` off of their Helm chart - but not from a chart repo, but rather directly from their Git repo.

It _should_ be possible, and I fully blame Rancher for not just putting this into their common chart repo, but it is also an undocumented edge-case. So - this is why I am making this FR. :)

So I currently use this:

```
  extensions:
    helm:
      concurrencyLevel: 5
      repositories:
        - name: local-path-provisioner
          url: https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/chart/
      charts:
        - name: local-path
          chartname: local-path-provisioner/local-path-provisioner
          #repo: https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/chart/
          #chartname: deploy/chart/local-path-provisioner
          version: "0.0.32"
          namespace: local-path-provisioner
          values: |-
            storageClass:
              defaultClass: true
              pathPattern: "{{ .PVC.Namespace }}-{{ .PVC.Name }}"
```

But I can not find any indication in the logs if this is or is not working.
As you can see from the commented lines, I really tried. x)

The solution would be to document how this is supposed to be done correctly. Bonus if you could also let the `k0s` binary itself show the status of which extensions are or are not applied, and why. Might be helpful.

Alternatively, let me point to a locally installed chart so I can just `git clone` their repo and then point at that chart on my local fs.

### Additional context

I tried to look for this on ArtifactHub - and I was entirely bamboozled that many people had done effective third-party uploads of this considerably miniature chart.

Also, in an earlier iteration, I _was_ able to get an error message by grepping for `k0s-addon`. But this version there is not reporting anything...

DeeBi9:
## Description

Added a note to the Helm Charts docs  that explains, raw git/gitHub chart directories can’t be used as chart sources.
Fixes #6341

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/docs/helm-charts.md
+++ b/docs/helm-charts.md
@@ -41,6 +41,8 @@ See [Chart configuration](#chart-configuration) below for more details on how to
 | `username` | -             | Username for Basic HTTP authentication                                                            |
 | `password` | -             | Password for Basic HTTP authentication                                                            |
 
+**Note:** k0s supports only classic Helm chart repositories that provide a valid `index.yaml`. Direct links to chart folders or files (for example raw GitHub URLs) are not recognized as Helm repositories and will not work unless they follow the full Helm repository structure. For details on how a valid Helm chart repository must be structured, see: https://helm.sh/docs/topics/chart_repository/#create-a-chart-repository
+
 ### Chart configuration
 
 | Field          | Default value | Description                                                                               |
@@ -75,9 +77,10 @@ spec:
       - name: oci-registry-with-private-ca
         # OCI registry URL must not include any path elements
         url: oci://registry-with-private-ca.com:8080
-        # Currently, only caFile is supported for TLS transport
-        # Setting certFile or keyFile will result in an error
+        # certFile and keyFile can be provided to enable mTLS
         caFile: /path/to/ca.crt
+        certFile: /path/to/client.crt
+        keyFile: /path/to/client.key
       charts:
       - name: prometheus-stack
         chartname: prometheus-community/prometheus


docs/helm-charts.md
```diff
@@ -41,6 +41,8 @@ See [Chart configuration](#chart-configuration) below for more details on how to
 | `username` | -             | Username for Basic HTTP authentication                                                            |
 | `password` | -             | Password for Basic HTTP authentication                                                            |
 
+**Note**: Raw Git or GitHub chart directories (for example `https://raw.githubusercontent.com/.../chart/`) can’t be used directly as chart sources. If a chart isn’t  published in a traditional Helm repository, use its OCI reference instead (for example `oci://ghcr.io/...`).
```
twz123: Maybe we can rephrase it in a way that says that we only¹ support "classic Helm chart repositories", and that raw URLs to a random chart won't work (without an `index.yml`)? Maybe also add a reference to https://helm.sh/docs/topics/chart_repository#create-a-chart-repository, so that it's clear what a chart repo looks like.

¹: There's quite some relevant docs about how to configure Helm repos in docs/extensions.md. Maybe it makes sense to move those docs here into this file. There's also some examples on how to use OCI registries over there.

DeeBi9: > Maybe we can rephrase it in a way that says that we only¹ support "classic Helm chart repositories", and that raw URLs to a random chart won't work (without an `index.yml`)? Maybe also add a reference to https://helm.sh/docs/topics/chart_repository#create-a-chart-repository, so that it's clear what a chart repo looks like.

I will update this , thanks !

> ¹: There's quite some relevant docs about how to configure Helm repos in docs/extensions.md. Maybe it makes sense to move those docs here into this file. There's also some examples on how to use OCI registries over there.

Regarding `extensions.md`: `helm-charts.md` already has the complete Prometheus example, so the only difference was the small mTLS change which was made earlier in this commit: https://github.com/k0sproject/k0s/commit/db4bb5bc65862e44f86b2bf745e3c69dd1dce509, which I will update.

PR #6958, opened 2026-01-14
Make NLLB pods tolerate NoExecute taints

### LINKED ISSUE #6955: NLLB on NoExecute-tainted nodes

### Version

v1.34.2+k0s.0

### What happened?

I'm using a NoExecute taint on a subset of my nodes in a cluster to reserve them for specific workloads. K0s starts the nllb pod on these nodes, but k8s immediately evicts the pods due to the untolerated NoExecute taint. Since these are static pods, the eviction of the apiserver pod objects doesn't really have an effect, but it does trigger alerts in my cluster regarding pods staying in terminating state for too long.

### Steps to reproduce

1. Deploy a worker with any custom NoExecute taint and nllb enabled
2. check the kube-system namespace for the nllb pod of that node, it should be Terminating.

### Additional context

Only workaround I see currently would be to use a NoSchedule taint instead, but I haven't tested that yet.

pschichtel:
## Description

I introduced a toleration for all NoExecute taints on the static nllb pod as to prevent kubernetes from trying to evict the pod, which it can't do anyway.

Fixes #6955

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/worker/nllb/envoy.go
+++ b/pkg/component/worker/nllb/envoy.go
@@ -303,6 +303,10 @@ func makePodManifest(params *envoyParams, podParams *envoyPodParams) corev1.Pod 
 					},
 				}},
 			},
+			// without this, kubernetes might try to evict the mirror pod
+			Tolerations: []corev1.Toleration{{
+				Operator: corev1.TolerationOpExists,
+			}},
 		},
 	}
 }


pschichtel:
As mentioned in #6955, this can be easily reproduced by tainting any node in a cluster with any custom NoExecute taint. Without the fix, the nllb pod stays in Pending or Terminating, with the fix it becomes running. I've rolled this change out to a cluster without any issues so far.

pkg/component/worker/nllb/envoy.go
```diff
@@ -303,6 +303,11 @@ func makePodManifest(params *envoyParams, podParams *envoyPodParams) corev1.Pod
 					},
 				}},
 			},
+			// without this kubernetes tries to evict the mirror pod on NoExecute-tainted nodes
+			Tolerations: []corev1.Toleration{{
+				Effect:   corev1.TaintEffectNoExecute,
```
twz123: I wonder if we should bluntly tolerate _all_ taints, instead of _only_ the NoExecute taint... WDYT?

```suggestion
```

pschichtel: Is that an option? I thought you have to at least specify an effect

pschichtel: I haven't tested it, but I don't think NoSchedule or PreferNoSchedule have an impact, since static pods bypass the scheduler anyway

twz123: It is: https://github.com/k0sproject/k0s/blob/v1.34.3+k0s.0/pkg/component/controller/konnectivityagent.go#L261-L262

pschichtel:
> We could add priorityClassName: system-node-critical, as well.

I thought I read in the kubernetes documentation that static pods implicitly have system-node-critical, but I can't find it again, maybe I'm mistaken. It would be a trivial addition, but as you say, don't think it has an impact anyway. Even if kubernetes decides to evict a static pod, it doesn't actually do anything. This entire change is technically just a "cosmetic fix"

pkg/component/worker/nllb/envoy.go
pschichtel: ok then, let's do that.

PR #6968, opened 2026-01-15
Add CNI Health Reporting to k0s status

### LINKED ISSUE #5886: More Detailed CNI Health in `k0s status`

`k0s status` is a good command. It would be even better if it could tell us more about the CNI's health directly. This would help a lot of people figure out network problems faster.

When i set up k0s with Calico using my automation script, i need to know when Calico is really ready. It's not just about the k0s API being up. Calico has its own controllers, like `calico-kube-controllers`, that need to be running. It also creates Custom Resource Definitions (CRDs) in Kubernetes, like `ippools.projectcalico.org`, and these need to be active and working.

Right now, my script has to do a lot of extra steps after `k0s status` says the API is fine. I use `kubectl` to check if the Calico deployments are ready. Then i use more `kubectl` commands to see if the Calico CRDs are established. Sometimes i even have to add `time.sleep()` calls because these Calico parts can take a bit longer to get going, even after the k0s API seems ready. This means my script has many separate `kubectl` calls and custom polling just for Calico.

If `k0s status` could give a clear signal about CNI health, things would be simpler. My script could just check `k0s status`. I wouldn't need so many specific `kubectl` calls just for Calico. It would make the question "is the cluster network actually working?" much easier to answer.

I think this would be useful for other people too.
When pods have network trouble, the CNI is often the first thing to check. A quick CNI status from `k0s status` would be a great help for debugging.
Anyone automating k0s with Calico, or other CNIs k0s might manage by default, would find it easier to confirm the network is truly ready. It helps quickly see if a problem is with k0s itself, the CNI, or something else like an application. Since k0s often deploys and manages the CNI, it makes sense for k0s to report on its status too.

Here's an idea of how it could look. When i run `sudo k0s status`, maybe the output could add something like this:

```
Version: v1.28.5+k0s.0
Process ID: 253118
Role: controller
Workloads: true
SingleNode: false
Kube-api probing successful: true
Kube-api probing last error:
CNI Provider: Calico
CNI Status: Healthy
  Controllers Ready: calico-kube-controllers, calico-node
  Critical CRDs Established: ippools.projectcalico.org, clusterinformations.projectcalico.org
```

And if something was wrong:

```
CNI Provider: Calico
CNI Status: Degraded
  Controllers Ready: calico-node (calico-kube-controllers pending)
  Critical CRDs Established: clusterinformations.projectcalico.org (ippools.projectcalico.org not established)
  CNI Last Error: [some brief error from Calico components if k0s can get it]
```

Another option could be a subcommand like `k0s status cni` that gives this detailed CNI breakdown.

Either way, this would give a clearer picture of the network health right from a core k0s command, making k0s feel more complete for the components it manages.

Thanks.

_No response_

### Additional context

_No response_

DeeBi9:
## Description

This adds CNI health reporting to the k0s status command. It detects CNI provider (kube-router or Calico), overall CNI health, component readiness and error.

Fixes #5886

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test

1. Tested with kube-router

- Degraded State: when pods are in pending state : 

      CNI Status:
       Provider: kuberouter
       Health: Degraded
       Components:
         - kube-router-vwcl7 (Ready: false)
      Error: Component kube-router-vwcl7 is in Pending state

- Running state : 

      CNI Status:
       Provider: kuberouter
       Health: Healthy
       Components:
         - kube-router-vwcl7 (Ready: true)

2. Tested with calico 

- Degraded State: when pods are in pending state : 

      CNI Status:
        Provider: calico
        Health: Degraded
        Components:
          - calico-kube-controllers-74f9fd669d-2dcpd (Ready: false)
          - calico-node-lgrq8 (Ready: false)
        Error: Component calico-node-lgrq8 is in Pending state

- Running state : 

      CNI Status:
        Provider: calico
        Health: Healthy
        Components:
          - calico-kube-controllers-74f9fd669d-2dcpd (Ready: true)
          - calico-node-lgrq8 (Ready: true)

## Checklist

### DIFF
--- a/pkg/component/status/client.go
+++ b/pkg/component/status/client.go
@@ -36,16 +36,15 @@ type ProbeStatus struct {
 	Success bool
 }
 
-type CNI struct {
-	Provider   string
-	Health     string
-	Components []string
-	Error      string
+type Condition struct {
+	Type    string
+	Status  string
+	Reason  string
+	Message string
 }
 
-func (c *CNI) Fail(format string, a ...interface{}) {
-	c.Health = "Degraded"
-	c.Error = fmt.Sprintf(format, a...)
+type CNI struct {
+	Conditions []Condition
 }
 
 // GetStatus returns the status of the k0s process using the status socket
--- a/pkg/component/status/client.go
+++ b/pkg/component/status/client.go
@@ -13,6 +13,7 @@ import (
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 	"github.com/k0sproject/k0s/pkg/component/prober"
 	"github.com/k0sproject/k0s/pkg/config"
+	corev1 "k8s.io/api/core/v1"
 )
 
 type K0sStatus struct {
@@ -29,7 +30,7 @@ type K0sStatus struct {
 	WorkerToAPIConnectionStatus ProbeStatus
 	ClusterConfig               *v1beta1.ClusterConfig
 	K0sVars                     *config.CfgVars
-	CNI                         *CNI
+	Conditions                  []Condition
 }
 type ProbeStatus struct {
 	Message string
@@ -38,13 +39,9 @@ type ProbeStatus struct {
 
 type Condition struct {
 	Type    string
-	Status  string
+	Status  corev1.ConditionStatus
 	Reason  string
 	Message string
-}
-
-type CNI struct {
-	Conditions []Condition
 }
 
 // GetStatus returns the status of the k0s process using the status socket
--- a/pkg/component/status/client.go
+++ b/pkg/component/status/client.go
@@ -29,10 +29,23 @@ type K0sStatus struct {
 	WorkerToAPIConnectionStatus ProbeStatus
 	ClusterConfig               *v1beta1.ClusterConfig
 	K0sVars                     *config.CfgVars
+	CNI                         *CNI
 }
 type ProbeStatus struct {
 	Message string
 	Success bool
+}
+
+type CNI struct {
+	Provider   string
+	Health     string
+	Components []string
+	Error      string
+}
+
+func (c *CNI) Fail(format string, a ...interface{}) {
+	c.Health = "Degraded"
+	c.Error = fmt.Sprintf(format, a...)
 }
 
 // GetStatus returns the status of the k0s process using the status socket
--- a/pkg/component/status/status.go
+++ b/pkg/component/status/status.go
@@ -10,13 +10,14 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/component/prober"
 	kubeutil "github.com/k0sproject/k0s/pkg/kubernetes"
 	"github.com/sirupsen/logrus"
+	appsv1 "k8s.io/api/apps/v1"
+	authv1 "k8s.io/api/authentication/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
@@ -34,6 +35,10 @@ type Status struct {
 	L                 *logrus.Entry
 	httpserver        http.Server
 	CertManager       certManager
+
+	WorkerClient       kubernetes.Interface
+	AdminClient        kubernetes.Interface
+	AdminClientFactory *kubeutil.ClientFactory
 }
 
 type certManager interface {
@@ -95,7 +100,6 @@ func (s *Status) Stop() error {
 
 type statusHandler struct {
 	Status *Status
-	client kubernetes.Interface
 }
 
 // ServerHTTP implementation of handler interface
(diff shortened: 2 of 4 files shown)


DeeBi9:
A couple of clarifications about this implementation:

1. I haven’t added/updated documentation yet. Before doing so, I’d like to confirm the expected level of detail and whether this new CNI status block should also be documented under k0s status.

2. The CNI health evaluation is currently implemented directly inside getCurrentStatus().

3. The current logic does not include support for custom CNIs

4. Should CRD checks be implemented?. I'd like to confirm this

pkg/component/status/client.go
```diff
@@ -29,12 +29,25 @@ type K0sStatus struct {
...
+	Provider   string
+	Health     string
+	Components []string
+	Error      string
+}
+
+func (c *CNI) Fail(format string, a ...interface{}) {
+	c.Health = "Degraded"
+	c.Error = fmt.Sprintf(format, a...)
+}
```
twz123: I'd rather move this into status.go.

pkg/component/status/status.go
```diff
@@ -125,12 +127,72 @@ func (sh *statusHandler) getCurrentStatus(ctx context.Context) K0sStatus {
 		}
 		sh.client = kubeClient
 	}
-	_, err := sh.client.CoreV1().Nodes().List(context.Background(), v1.ListOptions{})
+	nodes, err := sh.client.CoreV1().Nodes().List(ctx, v1.ListOptions{})
```
twz123: This is a bit tricky, from the semantics point of view. All the things happening at this point are executed on worker nodes only. That's probably not what we want for the CNI checks, so we need to separate this.

That node check uses a special client, and is solely used to check the API server reachability from pure worker nodes, mainly to check that kubelet certificate rotation didn't break the node, but you're now using it for something else.

I fear that the status component needs a bit of refactoring first. Thing is that we have three different scenarios:

* Pure worker nodes: All they have is the kubelet kubeconfig anyways, so in this case, it doesn't matter.
* Pure controller nodes: They don't have a kubelet kubeconfig at all. None of this code is executed on them. However, having the CNI checks would probably make sense there, too. So we'd need to get the right client factory (the admin one from the controller command) into the status component, and use that instead.
* Controller nodes that run worker components: They probably want to use the same admin client factory that pure controllers use, except for this connectivity check, in which they should exercise the kubelet kubeconfig. Granted, listing all the nodes is probably not the best connectivity check in the first place. Feel free to find something more lightweight. Maybe a call to whoami or sth. along those lines.

pkg/component/status/status.go
```diff
@@ -125,12 +127,72 @@ func (sh *statusHandler) getCurrentStatus(ctx context.Context) K0sStatus {
...
+
+	for _, node := range nodes.Items {
+		for _, cond := range node.Status.Conditions {
+			if cond.Type == "NetworkUnavailable" && cond.Status == "True" {
+				status.CNI.Fail("Node %s reports NetworkUnavailable: %s", node.Name, cond.Message)
+			}
+		}
+	}
+
+	if status.ClusterConfig != nil && status.ClusterConfig.Spec.Network.Provider != "" {
```
twz123: I'd separate this into three cases:

* kube-router: Just check kube-router components
* calico: Just check calico components.
* custom CNI: Do nothing.

pkg/component/status/status.go
```diff
@@ -125,12 +127,72 @@ func (sh *statusHandler) getCurrentStatus(ctx context.Context) K0sStatus {
...
+		status.CNI.Provider = status.ClusterConfig.Spec.Network.Provider
+	}
+
+	pods, err := sh.client.CoreV1().Pods("kube-system").List(ctx, v1.ListOptions{
```
twz123: Might be better to check the state of Deployments and DaemonSets instead of doing this for individual pods. That would scale better with bigger clusters. No need to have hundreds of pods listed here.

pkg/component/status/client.go
```diff
@@ -29,12 +29,25 @@ type K0sStatus struct {
...
 	Message string
 	Success bool
 }
 
+type CNI struct {
+	Provider   string
+	Health     string
+	Components []string
+	Error      string
+}
```
twz123: I'm mainly okay with this. I'm just thinking that we might consider a more general approach by using conditions, which is quite common for status fields. Just as an example:

```yaml
status:
  conditions:
  - type: CNIReady
    status: "True"
    reason: AllComponentsReady
    message: kube-router DaemonSet reports 7/7 ready pods
    # The typical lastTransitionTime would only make sense if we'd check for this proactively.
```

WDYT?

We could then add other stuff in the future, as well, without changing the structs.

twz123 (commented):
Nice! I've had a first pass on this. There's some general things that need to be discussed, please see the individual review comments for that.

> A couple of clarifications about this implementation:
> 
> 1. I haven’t added/updated documentation yet. Before doing so, I’d like to confirm the expected level of detail and whether this new CNI status block should also be documented under k0s status.

Let's postpone the docs until the implementation has settled.

> 3. The current logic does not include support for custom CNIs

The node status checks that you added is probably all we can do in that case. No way to try to guess around.

> 4. Should CRD checks be implemented?. I'd like to confirm this

Not sure what you mean by this?

pkg/component/status/client.go
DeeBi9: I’ll move the Fail() into `status.go`.

pkg/component/status/status.go
DeeBi9: Thanks for the explanation, I understand now

pkg/component/status/status.go
DeeBi9: I will restructure this logic .

pkg/component/status/status.go
DeeBi9: Agreed !, this is much cleaner approach . I'll update this

pkg/component/status/client.go
DeeBi9: I like this idea actually, I will try to update the fields and test the changes.

DeeBi9:
> Not sure what you mean by this?

My initial implementation included scanning CRDs to ensure the CNI installation was fully established 😅. 

Hey thanks a lot for the detailed review really appreciate . Just to mention I am fairly new to k8s k0s , so I am learning patterns and architecture as I go. 
Thanks for the guidance

DeeBi9:
PTAL @twz123 

I tested the changes locally :
1. With kube-router 
        
        // Running state
        Version: v1.34.1+k0s.0-533-g370ff77b3
        Process ID: 753629
        Role: controller
        Workloads: true
        SingleNode: true
        Kube-api probing successful: true
        Kube-api probing last error:  
        
        Status:
         conditions:
          - type:   CNIReady
            status: True
            reason: AllReady
            message: kube-router DaemonSet reports 1/1 ready
            
        // Degraded state   
        Version: v1.34.1+k0s.0-533-g370ff77b3
        Process ID: 753629
        Role: controller
        Workloads: true
        SingleNode: true
        Kube-api probing successful: true
        Kube-api probing last error:  
        
        Status:
         conditions:
          - type:   CNIReady
            status: False
            reason: NotReady
            message: kube-router DaemonSet reports 0/1 ready

2. With Calico 

       // Degraded State
        Version: v1.34.1+k0s.0-533-g370ff77b3
        Process ID: 762972
        Role: controller
        Workloads: true
        SingleNode: true
        Kube-api probing successful: true
        Kube-api probing last error:  
        
        Status:
         conditions:
          - type:   CNIReady
            status: False
            reason: NotReady
            message: calico-node DaemonSet reports 0/1 ready
          - type:   CNIReady
            status: False
            reason: ComponentsNotReady
            message: Calico kube-controllers reports 0/1 ready
            
        // Running state    
        Version: v1.34.1+k0s.0-533-g370ff77b3
        Process ID: 762972
        Role: controller
        Workloads: true
        SingleNode: true
        Kube-api probing successful: true
        Kube-api probing last error:  
        
        Status:
         conditions:
          - type:   CNIReady
            status: True
            reason: AllReady
            message: calico-node DaemonSet reports 1/1 ready
          - type:   CNIReady
            status: True
            reason: AllComponetsReady
            message: Calico kube-controllers reports 1/1 ready

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
+		Message: message,
+	})
+}
+
+func (sh *statusHandler) checkCNIDaemonSet(ds *appsv1.DaemonSet, cni *CNI) {
+	desired := ds.Status.DesiredNumberScheduled
+	ready := ds.Status.NumberReady
+
+	if desired > 0 && ready == desired {
+		addCNICondition(cni, "True", "AllReady", fmt.Sprintf("%s DaemonSet reports %d/%d ready", ds.Name, ready, desired))
```
twz123: Might make sense to have some constants for the "reasons", so we have them consistent between Kube-Router and Calico.

pkg/component/status/client.go
```diff
@@ -29,12 +29,24 @@ type K0sStatus struct {
...
+	CNI                         *CNI
 }
 type ProbeStatus struct {
 	Message string
 	Success bool
 }
 
+type Condition struct {
+	Type    string
+	Status  string
```
twz123: I wonder if we could hijack `corev1.ConditionStatus` here:


```suggestion
	Status  corev1.ConditionStatus // import corev1 "k8s.io/api/core/v1"
```

Or at least create our own newtype, say `K0sStatusCondition`.

pkg/component/status/client.go
```diff
@@ -29,12 +29,24 @@ type K0sStatus struct {
...
+	Reason  string
+	Message string
+}
+
+type CNI struct {
+	Conditions []Condition
```
twz123: I thought about adding this to the `K0sStatus` struct, and not having a `CNI` struct at all. Otherwise, "CNIReady" seems a bit redundant, and could only be "Ready".

Having it on the K0sStatus level would allow us to add other non-CNI related conditions, say etcd status or similar.

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
+	} else {
+		addCNICondition(cni, "False", "NotReady", fmt.Sprintf("%s DaemonSet reports %d/%d ready", ds.Name, ready, desired))
+	}
+}
+
+func (sh *statusHandler) checkCalicoDeployment(deploy *appsv1.Deployment, cni *CNI) {
+	ready := deploy.Status.ReadyReplicas
+	total := deploy.Status.Replicas
+
+	if ready == total {
```
twz123: Deployments have their own "available" condition, which we can inspect, like we do e.g. in the [integration tests](https://github.com/k0sproject/k0s/blob/v1.34.3+k0s.0/inttest/common/util.go#L190-L200).

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
+		addCNICondition(cni, "True", "AllComponetsReady", fmt.Sprintf("Calico kube-controllers reports %d/%d ready", ready, total))
+	} else {
+		addCNICondition(cni, "False", "ComponentsNotReady", fmt.Sprintf("Calico kube-controllers reports %d/%d ready", ready, total))
+	}
+}
+
 func (sh *statusHandler) getCurrentStatus(ctx context.Context) K0sStatus {
+	if sh.Status.AdminClient == nil && sh.Status.AdminClientFactory != nil {
```
twz123: I wouldn't cache that, but always consult the client factory, especially as this is a handler and can be called concurrently.

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
+	}
+
+	provider := ""
+	if status.ClusterConfig != nil {
+		provider = status.ClusterConfig.Spec.Network.Provider
+	}
+
+	switch provider {
+	case "kuberouter":
+		ds, err := apiClient.AppsV1().DaemonSets("kube-system").
```
twz123: Let's use constants here and in the other places.

```suggestion
		ds, err := apiClient.AppsV1().DaemonSets(metav1.NamespaceSystem).
```

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
 	}
 	status.WorkerToAPIConnectionStatus.Success = true
+
+	status.CNI = &CNI{
+		Conditions: []Condition{},
+	}
+
+	apiClient := sh.Status.AdminClient
+	if apiClient == nil {
+		addCNICondition(status.CNI, "Unknown", "AdminClientNotReady", "Admin API client not initialized yet")
```
twz123: I wonder if it makes sense at all to even include a CNI condition in this case. If the admin client factory reports an error, something is fundamentally wrong, and I feel that the status endpoint should respond with a 500 error. Not sure if that's a good idea, but I'm not entirely happy with this artificial condition either 🤔

pkg/component/status/status.go
```diff
@@ -111,26 +117,108 @@ const (
...
+	switch provider {
+	case "kuberouter":
+		ds, err := apiClient.AppsV1().DaemonSets("kube-system").
+			Get(ctx, "kube-router", v1.GetOptions{})
+		if err == nil {
+			sh.checkCNIDaemonSet(ds, status.CNI)
+		} else {
+			addCNICondition(status.CNI, "False", "NotFound",
+				"kube-router DaemonSet not found")
+		}
```
twz123, line 204: It could be anything here, so we need to check the error... (same for Calico further down)

```suggestion
		} else if apierrors.IsNotFound(err) {
			addCNICondition(status.CNI, "False", "NotFound", "kube-router DaemonSet not found")
		} else {
			addCNICondition(status.CNI, "Unknown", "Error", err.Error())
		}
```

twz123 (commented):
This is looking very promising!

pkg/component/status/status.go
DeeBi9: Okay I will update this n add some constants

pkg/component/status/client.go
DeeBi9: Agree it's more extensible , I will update this too

pkg/component/status/status.go
DeeBi9: Returning a 500 here would be make debugging harder or what is actually broken, for now maybe I will just leave it empty and return status, without a working client we cant report CNI status anyway

PR #7053, opened 2026-02-03
Replace kardianos/service

### LINKED ISSUE #4962: Drop kardianos/service dependency
We anyways already provide our own templates for system units and thus do not depend _that much_ on the library functionality.

kardianos/service is not really maintained, there's e.g. a bug fix PR opened by myself on June 2023 which has not been merged.

I propose that we drop the whole dependency, use our own templates fully and provide needed wrappers around. We handle much of the same/similar logic in k0sctl/rig side already and thus known what needs to be done.

ncopa:
## Description

Fixes https://github.com/k0sproject/k0s/issues/4962

Note: This PR drops implicit sysvinit and upstart support that was previously provided by kardianos/service. These init systems were never documented as supported by k0s (docs/install.md only lists systemd and OpenRC), so this is not considered a user-facing breaking change.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist


pkg/install/service.go
```diff
@@ -17,128 +18,121 @@ var (
...
+		return err
+	}
+	logrus.Infof("Enabling %s service", spec.Name)
+	return svc.Enable(ctx)
 }
 
 func UninstallService(role string) error {
-	prg := &Program{}
-
+	ctx := context.Background()
```
twz123:
```suggestion
	ctx := context.TODO()
```

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,140 @@
...
+	return err
+}
+
+func (s *openrcService) Restart(ctx context.Context) error {
+	_, _, _, err := s.runner.Run(ctx, "rc-service", s.spec.Name, "restart")
+	return err
+}
+
+func (s *openrcService) Status(ctx context.Context) (Status, error) {
+	if !file.Exists(s.initPath()) {
```
twz123: I'm a bit on the fence with the implementation of `file.Exists`. It will just swallow errors. What about just using something like this:

```suggestion
	if stat, err := os.Stat(s.initPath()); err != nil {
		if errors.Is(err, os.ErrNotExist) {
			return StatusNotInstalled, nil
		}
		return StatusUnknown, err
	} else if stat.IsDir() {
		return StatusNotInstalled, nil
	}
```

Moreover, this feels like a bit of a TOCTOU thing, as I'd rather let the runner return an error that indicates that the service doesn't exist...

pkg/sysservice/backend_openrc_test.go
```diff
@@ -0,0 +1,172 @@
...
+	ctx := context.Background()
+	root := t.TempDir()
+	r := newFakeRunner()
+
+	spec := Spec{
+		Name:        "k0scontroller",
+		Description: "k0s controller",
+		Exec:        "/usr/local/bin/k0s",
+		Args:        []string{"controller", "--config", "/etc/k0s/k0s.yaml"},
+		Env:         []string{"K0S_LOG_LEVEL=info", "K0S_FOO=bar"},
```
twz123: I guess this is only an example, but AFAICT we don't support the `K0S_LOG_LEVEL` variable. I'd rather choose an example from which it's clear that this is made up (like `K0S_FOO`).

pkg/sysservice/detect.go
```diff
@@ -0,0 +1,52 @@
...
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+)
+
+// Detect returns the best guess service backend kind: "systemd", "openrc", or "windows".
+// If root != "", it will check marker paths under that root first (useful for tests).
+func Detect(root string) (string, error) {
```
twz123: I'm not sure if we gain much by having the `root` arg here, which is exclusively used for testing, especially because the tests are so simple, and the extra noise is significant. I'd just drop detect_unix_test.go and detect_windows_test.go.

If you really think it's worth having this, I'd rather use an `io/fs.FS` implementation, and call this with `os.DirFS("/")` from production code, and with `os.DirFS(t.TempDir())` from tests. But I'm really not sure if the noise/signal ratio is worth it.

twz123: This doesn't need to be exported, btw.

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,33 @@
...
+
+package sysservice
+
+import (
+	"context"
+	"errors"
+	"os/exec"
+)
+
+type Runner interface {
```
twz123: Doesn't need to be exported. It's never used outside of this package.

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,33 @@
...
+	Run(ctx context.Context, name string, args ...string) (exitCode int, stdout string, stderr string, err error)
+}
+
+type ExecRunner struct{}
+
+func (ExecRunner) Run(ctx context.Context, name string, args ...string) (int, string, string, error) {
+	cmd := exec.CommandContext(ctx, name, args...)
+	out, err := cmd.CombinedOutput()
```
twz123: Is mixing together stdout/stderr done on purpose here? The interface's Run method explicitly distinguishes stdout and stderr.

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		var ee *exec.ExitError
+		if errors.As(err, &ee) {
+			exit = ee.ExitCode()
+			err = nil
+		} else {
+			exit = 1
+		}
+	}
+	return exit, string(out), "", err
```
twz123: I'd rather let this be done on the call sites. Especially the err=nil in case of a non-zero exit code is a bit of a red flag for me: There's a lot of invocations like these: `_, _, _, err := s.runner.Run(ctx, ...); return err`, which will just silently drop non zero exit codes.

pkg/sysservice/backend_windows.go
twz123, line 1: Did you consider to use `golang.org/x/sys/windows/svc/mgr` to implement this? I'm always a bit reluctant to use scripting.

pkg/sysservice/runner_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+	"testing"
+)
+
+func TestExecRunnerRun_ReturnsExitCodeAndCombinedOutputWithoutError(t *testing.T) {
+	runner := ExecRunner{}
+	t.Setenv("GO_WANT_HELPER_PROCESS", "1")
+
+	exit, stdout, stderr, err := runner.Run(
+		context.Background(),
```
twz123:
```suggestion
		t.Context(),
```

pkg/sysservice/runner_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	if stderr != "" {
+		t.Fatalf("expected empty stderr, got %q", stderr)
+	}
+	if !strings.Contains(stdout, "stdout-line") {
+		t.Fatalf("expected stdout to contain stdout output, got %q", stdout)
+	}
+	if !strings.Contains(stdout, "stderr-line") {
+		t.Fatalf("expected stdout to contain stderr output from CombinedOutput, got %q", stdout)
+	}
```
twz123: Usually, we use testify for assertions.

pkg/sysservice/runner_test.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func TestExecRunnerRun_ReturnsExitCodeOneWhenCommandCannotStart(t *testing.T) {
+	runner := ExecRunner{}
+
+	exit, stdout, stderr, err := runner.Run(context.Background(), "definitely-not-a-real-command-k0s-test")
```
twz123:
```suggestion
	exit, stdout, stderr, err := runner.Run(t.Context(), "definitely-not-a-real-command-k0s-test")
```

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,79 @@
...
+	case "openrc":
+		return newOpenRC(spec, cfg), nil
+	case "systemd":
+		return newSystemd(spec, cfg), nil
+	case "windows":
+		return newWindows(spec, cfg), nil
+	default:
+		return nil, errors.New("unsupported kind: " + kind)
+	}
+}
```
twz123: I'd rather remove the whole Options struct, as this is solely used for tests, which in turn just shortcuts this whole function (Detect is never called, for example).

We could go with something like this:

```go
func New(spec Spec) (Service, error) {
	if spec.Name == "" || spec.Exec == "" {
		return nil, ErrInvalidSpec
	}

	kind, err := detectInitSystem()
	if err != nil {
		return nil, err
	}

	switch kind {
	case "openrc":
		newOpenRC(spec), nil
	case "systemd":
		newSystemd(spec), nil
	case "windows":
		newWindows(spec), nil
	default:
		return nil, errors.New("unsupported kind: " + kind)
	}
}
```

The individual tests can then call `newOpenRC(...)` and `newSystemd(...)` directly, override some fields in the returned structs (e.g. the root folder, or the runner implementation), and then proceed with their tests.

We could even have a separate version of this for Windows, which just looks like this:

```go
func New(spec Spec) (Service, error) {
	if spec.Name == "" || spec.Exec == "" {
		return nil, ErrInvalidSpec
	}

	return newWindows(spec), nil
}
```

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,79 @@
...
+	Disable(ctx context.Context) error
+
+	Start(ctx context.Context) error
+	Stop(ctx context.Context) error
+	Restart(ctx context.Context) error
+
+	Status(ctx context.Context) (Status, error)
+}
+
+var ErrInvalidSpec = errors.New("invalid spec")
```
twz123:
```suggestion
var errInvalidSpec = errors.New("invalid service spec")
```

pkg/install/service.go
```diff
@@ -17,128 +18,121 @@ var (
...
+	if err != nil {
+		return sysservice.Spec{}, err
+	}
+	return sysservice.Spec{
 		Name:        k0sServiceName,
 		DisplayName: k0sDisplayName,
 		Description: k0sDescription,
-	}
+		Exec:        k0sExec,
+	}, nil
```
twz123: If I'm not mistaken, the display name, the description and the executable path are only required when installing a service. Wouldn't it make sense to not include them in the general service spec, but just return the service name? The install method could get them as arguments. Since the display name is always the service name, the description is just a constant, and the executable path is always the current executable's one, maybe we don't even need them as arguments, but can just hard-code them in the init system implementations?

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,79 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package sysservice
+
+import (
+	"context"
+	"errors"
+)
+
+type Spec struct {
```
twz123: This is more like an InstallSpec. None of this is required for any of the `Service` methods, except the service name.

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,79 @@
...
+	Name        string
+	DisplayName string
+	Description string
+
+	Exec string
+	Args []string
+	Env  []string // ["K=V", ...]
+
+	WorkingDir string
+	User       string
```
twz123: I don't think those are used anywhere in the codebase, except in the tests?

pkg/install/service.go
```diff
@@ -17,128 +18,121 @@ var (
...
+		}
 	}
 
-	var s service.Service
-	return s, errors.New("k0s has not been installed as a service")
+	return nil, errors.New("k0s has not been installed as a service")
 }
 
 // InstallService installs the k0s service, per the given arguments, and the detected platform
-func InstallService(args []string, envVars []string, force bool) error {
```
twz123, line 57: The environment variables aren't used at all, currently.

pkg/install/service.go
```diff
@@ -17,128 +18,121 @@ var (
...
-		svcConfig.Option["Environment"] = envVars
-	}
-
-	if runtime.GOOS == "windows" {
-		args = append([]string{"service=" + svcConfig.Name}, args...)
-	}
-
-	svcConfig.Arguments = args
-
 	if force {
```
twz123, line 59: I wonder if it makes sense to implement the force flag differently, since we now have control over how things are done. Maybe just pass it on to the service Install method. The backing implementation can then figure out what to do. Pretty sure that most of them will allow to just override the service definitions. No need to uninstall things first.

pkg/sysservice/backend_openrc.go
ncopa: Yeah, I'm not too happy with the TOCTOU thingy, but IIRC it was not really possible to get the proper error return code without parsing the output or something. I'll investigate.

pkg/sysservice/backend_windows.go
ncopa: I wasnt aware of `golang.org/x/sys/windows/svc/mgr`. Sounds like a better approach.

pkg/install/service.go
ncopa: Lets do that in separate PR if this change of behavior is needed

pkg/sysservice/backend_openrc.go
ncopa: solved. good catch.

pkg/sysservice/detect.go
ncopa: fixed

pkg/sysservice/sysservice.go
ncopa: I removed Spec type, we only need name.

pkg/sysservice/sysservice.go
ncopa: I have removed them

pkg/sysservice/backend_openrc_test.go
ncopa: solved.

pkg/sysservice/detect.go
ncopa: fixed

pkg/install/service.go
ncopa: I removed `DisplayName`, `Description`, and `Exec` and the entire `Spec` type, and only pass the name as arg to `New`.

pkg/install/service.go
ncopa: fixed

pkg/sysservice/backend_systemd_test.go
```diff
@@ -0,0 +1,187 @@
...
+	st, err := svc.Status(ctx)
+	if err != nil {
+		t.Fatalf("Status() error: %v", err)
+	}
+	if st != StatusNotInstalled {
+		t.Fatalf("expected StatusNotInstalled, got %v", st)
+	}
+}
+
+func TestSystemdStatus_RunningWhenSystemctlExit0(t *testing.T) {
```
juanluisvaladas: Probably we should add a test case where the status is stopped too

pkg/sysservice/backend_systemd.go
```diff
@@ -0,0 +1,206 @@
...
+		return StatusUnknown, err
+	}
+
+	_, _, _, err := s.runner.Run(ctx, "systemctl", "is-active", "--quiet", s.unitName())
+	if err == nil {
+		return StatusRunning, nil
+	}
+	var ee *exec.ExitError
+	if errors.As(err, &ee) {
+		return StatusStopped, nil
```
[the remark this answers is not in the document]
ncopa: I also removed the Stat while at it.

pkg/sysservice/backend_systemd_test.go
ncopa: good point. done.

pkg/sysservice/detect.go
```diff
@@ -0,0 +1,38 @@
...
+		return "systemd", nil
+	}
+	if _, err := exec.LookPath("openrc-init"); err == nil {
+		return "openrc", nil
+	}
+	if _, err := exec.LookPath("rc-service"); err == nil {
+		return "openrc", nil
+	}
+
+	return "", errors.New("unable to detect init/service manager (systemd/openrc)")
```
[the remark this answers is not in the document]
ncopa: we already documented that only openrc and systemd are supported in `docs/install.md`.

pkg/sysservice/detect.go
```diff
@@ -0,0 +1,38 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sysservice
```
twz123: There's the internal/supervised package which has some abstractions for dealing with process supervisors. Maybe we can merge these things?

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"os/exec"
+)
+
+type runner interface {
+	Run(ctx context.Context, name string, args ...string) (exitCode int, stdout string, stderr string, err error)
```
twz123: All these return values are mostly unused. What about just having an err return value and figure out the rest from there?

If we need the exit code, then we can use `errors.As(err, ...)` and probe for it.

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+type execRunner struct{}
+
+func (execRunner) Run(ctx context.Context, name string, args ...string) (int, string, string, error) {
+	cmd := exec.CommandContext(ctx, name, args...)
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	err := cmd.Run()
```
twz123: Since stdout/stderr aren't really used anywere, I'd use `Output` instead. It will capture stdout into a byte slice and do some smarts about stderr, e.g. keeping some of the first bytes, some of the last bytes and embed it into the error message, if the exit code is non-zero.

```suggestion
	stdout, err := cmd.Output()
```

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,55 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sysservice
```
twz123, line 4: There's the internal/supervised package which has some abstractions for dealing with process supervisors. Maybe we can merge these things?

pkg/sysservice/detect.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"runtime"
+)
+
+func detectInitSystem() (string, error) {
```
twz123: I think we could merge this directly with `New()`?

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,55 @@
...
+type Status int
+
+const (
+	StatusUnknown Status = iota
+	StatusNotInstalled
+	StatusStopped
+	StatusRunning
+)
+
+type Service interface {
```
twz123, line 23: Some Methods aren't used at all (Kind, Disable). Maybe we can just remove them from the interface and the implementations?

pkg/cleanup/services.go
```diff
@@ -23,13 +23,13 @@ func (s *services) Name() string {
...
-func (s *services) Run() error {
+func (s *services) Run(ctx context.Context) error {
 	var errs []error
 
 	for _, role := range []string{"controller", "worker"} {
 		logrus.Debugf("attempting to uninstall k0s%s service", role)
-		if err := install.UninstallService(role); err != nil {
-			if !errors.Is(err, service.ErrNotInstalled) && !errors.Is(err, fs.ErrNotExist) && !isExitCode(err, 1) {
+		if err := install.UninstallService(ctx, role); err != nil {
+			if !errors.Is(err, fs.ErrNotExist) && !isExitCode(err, 1) {
```
twz123: That's a bit wonky? Why should this be an fs error? We shouldn't assume the internal details of the service manager. On Windows, this is definitely not backed by traditional files. Moreover, I'd question the isExitCode check. What's this good for?

I get that _only_ the ErrNotInstalled error has been removed, but I'm not confident that the rest of the remaining if statement makes much sense?

pkg/install/service.go
twz123: I just think it'd be saner, and that this is just an implementation detail of kardianos that we can get rid of.

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,174 @@
...
+command="{{.Exec}}"
+command_args="{{range .Args}}'{{.}}' {{end}}"
+name=$(basename $(readlink -f $command))
+supervise_daemon_args="--stdout /var/log/${name}.log --stderr /var/log/${name}.err"
+
+: "${rc_ulimit=-n 1048576 -u unlimited}"
+
+{{- if .Env}}{{range .Env}}
+export {{.}}{{end}}{{- end}}
+
```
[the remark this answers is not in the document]
twz123: I think this is a valid point, isn't it?

pkg/sysservice/backend_openrc_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if err := svc.Install(ctx, []string{"controller", "--config", "/etc/k0s/k0s.yaml"}, []string{"K0S_FOO=bar", "K0S_BAR=baz"}); err != nil {
+		t.Fatalf("Install() error: %v", err)
+	}
+
+	initPath := filepath.Join(root, "etc/init.d/k0scontroller")
+
+	initBytes, err := os.ReadFile(initPath)
+	if err != nil {
+		t.Fatalf("expected init script at %s: %v", initPath, err)
```
twz123: Usually, we're using testify assertions.

pkg/sysservice/backend_winservice.go
```diff
@@ -0,0 +1,220 @@
+//go:build windows
```
twz123: We'd call this file backend_windows.go, so we don't need to specify the build flags explicitly.

pkg/sysservice/detect.go
```diff
@@ -0,0 +1,38 @@
...
+
+	// Prefer filesystem markers (most reliable), then fall back to PATH checks.
+	if _, err := os.Stat("/run/systemd/system"); err == nil {
+		return "systemd", nil
+	}
+	if _, err := os.Stat("/run/openrc"); err == nil {
+		return "openrc", nil
+	}
+
+	// PATH fallbacks (for real installs where markers might not be visible)
```
twz123: Did you consider to check what's PID1?

ncopa: No. I have bad experience with making assumptions on pid 1. You could for example run k0s install in a chroot.

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,55 @@
...
+	Stop(ctx context.Context) error
+	Restart(ctx context.Context) error
+
+	Status(ctx context.Context) (Status, error)
+}
+
+func New(name string) (Service, error) {
+	if name == "" {
+		return nil, errors.New("service name must not be empty")
+	}
```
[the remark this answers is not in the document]
ncopa: I think this is irrelevant. We have good control of the input name.

pkg/install/service.go
ncopa: yeah, i think the `--force` is moot actually. Current implementation will overwrite the current service file and not exit with error. enabling an already enabled service will also be ignored with exit 0 on openrc at least. I think that is expected behavior.

We should probably mark the `--force` flag as deprecated, but keep it for backwards compat.

ncopa: The behavior for `k0s install  controller --start` will change though. Currently it will fail if the service is already running. With `--force` it will restart it. So we should probably keep `--force`.

On Windows though, an install will fail if service already exists. That is different from Linux. What behavior do we want on Windows?

internal/pkg/file/file.go
```diff
@@ -57,6 +57,18 @@ func Copy(src, dst string) (err error) {
...
+// data to it. It returns an error wrapping [fs.ErrExist] if the file already
+// exists, so callers can use errors.Is(err, fs.ErrExist) to detect that case.
+func WriteNew(name string, data []byte, perm os.FileMode) error {
+	f, err := os.OpenFile(name, os.O_WRONLY|os.O_CREATE|os.O_EXCL, perm)
+	if err != nil {
+		return err
+	}
+	_, writeErr := f.Write(data)
+	return errors.Join(writeErr, f.Close())
+}
```
twz123, line 70: I don't see a major difference to `os.WriteFile()`, which just uses `O_TRUNC` instead of `O_EXCL`? Is this really worth its own helper function, especially given that we have the atomic helpers, too?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+		return fmt.Errorf("get executable: %w", err)
+	}
+	if err := dir.Init(s.initDir(), 0o755); err != nil {
+		return err
+	}
+	initScript, err := renderOpenRCInit(exec, args, env)
+	if err != nil {
+		return err
+	}
+	if err := file.WriteNew(s.initPath(), initScript, 0o755); err != nil {
```
twz123, line 50: I'd rather use the atomic file helper here. Same for the systemd implementation.

```suggestion
	if err := file.AtomicWithTarget(s.initPath()).WithPermissions(0755).Write(initScript); err != nil {
```

pkg/sysservice/sysservice.go
```diff
@@ -0,0 +1,62 @@
...
+type Status int
+
+const (
+	StatusUnknown Status = iota
+	StatusNotInstalled
+	StatusStopped
+	StatusRunning
+)
+
+type Service interface {
```
twz123, line 23: We should document the expected behavior of those methods. Might sound trivial, but I find it noteworthy, that e.g. `Uninstall` will silently succeed, even if the service didn't exist in the first place. I'd have expected it to return some "no such service" error. Also, but maybe more intuitive: Restart won't fail if the service isn't running.

That said, maybe we don't need `Restart` to begin with? The only function which is using it is `install.StartinstalledService`, which is _solely_ called in places where it's expected for the service _not_ to be running, and hence failing if it does is fine.

Similar for `Enable`. We can just say that `Install` installs k0s in a way that it launches automatically. No need for an extra method.

pkg/sysservice/runner.go
```diff
@@ -0,0 +1,54 @@
...
+
+type runner interface {
+	Run(ctx context.Context, name string, args ...string) error
+}
+
+type execRunner struct{}
+
+func (execRunner) Run(ctx context.Context, name string, args ...string) error {
+	_, err := exec.CommandContext(ctx, name, args...).Output()
+	return err
```
twz123, line 29: I'd rater directly inline `wrapStderr` here. No need to do this at the call sites.

```suggestion
	if err != nil {
		var ee *exec.ExitError
		if errors.As(err, &ee) && len(ee.Stderr) > 0 {
			return fmt.Errorf("%w: %s", err, string(bytes.TrimSpace(ee.Stderr)))
		}
		return err
	}
	return nil
```

pkg/sysservice/backend_systemd.go
```diff
@@ -0,0 +1,215 @@
...
+func (s *systemdService) Status(ctx context.Context) (Status, error) {
+	err := s.runner.Run(ctx, "systemctl", "is-active", "--quiet", s.unitName())
+	code := exitCode(err)
+	if code < 0 {
+		return StatusUnknown, err
+	}
+	switch code {
+	case 0:
+		return StatusRunning, nil
+	case 1, 2, 3: // inactive, reloading, failed
```
twz123: IIUC "reloading" means that the service _is_ running, but just reloading its config? Could you maybe include a link to the upstream docs explaining the exit codes?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (s *openrcService) run(ctx context.Context, cmd string, args ...string) error {
+	return wrapStderr(s.runner.Run(ctx, cmd, args...))
+}
+
+func (s *openrcService) initDir() string {
+	if s.root != "" {
+		return filepath.Join(s.root, "etc/init.d")
+	}
+	return filepath.Join(string(filepath.Separator), "etc/init.d")
```
twz123: Maybe like so?

```suggestion
	root := cmp.Or(s.root, string(filepath.Separator))
	return filepath.Join(root, "etc", "init.d")
```

Same for the systemd variant of this.

pkg/sysservice/backend_systemd.go
```diff
@@ -0,0 +1,215 @@
...
+		runner: execRunner{},
+		exec:   os.Executable,
+	}
+}
+
+func (s *systemdService) runSystemctl(ctx context.Context, args ...string) error {
+	return wrapStderr(s.runner.Run(ctx, "systemctl", args...))
+}
+
+func (s *systemdService) unitDir() string {
```
twz123, line 34: Ugh, this _really_ looks like a typo, but isn't :see_no_evil:

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (s *openrcService) initPath() string { return filepath.Join(s.initDir(), s.name) }
+
+func (s *openrcService) Install(ctx context.Context, args []string, env []string) error {
+	exec, err := s.exec()
+	if err != nil {
+		return fmt.Errorf("get executable: %w", err)
+	}
+	if err := dir.Init(s.initDir(), 0o755); err != nil {
```
twz123: `dir.Init` is meant to be used for paths owned by k0s. Not sure if we should use it in this context? E.g. it will _always_ change directory permissions, even if the directory already exists (and the distro has applied different permissions to it, for whichever reason). Moreover, the more I think about it: Do we even want to create those directories? Shouldn't those already exist?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+}
+
+func renderOpenRCInit(exec string, args []string, env []string) ([]byte, error) {
+	const initTmpl = `#!/sbin/openrc-run
+supervisor=supervise-daemon
+description="k0s - Zero Friction Kubernetes"
+command="{{.Exec}}"
+command_args="{{range .Args}}'{{.}}' {{end}}"
+name=$(basename $(readlink -f $command))
```
twz123: I think the evaluation result will always be the same. Could we just hardcode this?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if err := file.WriteNew(s.initPath(), initScript, 0o755); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *openrcService) Uninstall(ctx context.Context) error {
+	// best-effort disable
+	_ = s.Disable(ctx)
```
twz123: What's the rationale behind this, and why don't we check the error here?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+func (s *openrcService) Stop(ctx context.Context) error {
+	return s.run(ctx, "rc-service", s.name, "stop")
+}
+
+func (s *openrcService) Restart(ctx context.Context) error {
+	return s.run(ctx, "rc-service", s.name, "restart")
+}
+
+func (s *openrcService) Status(ctx context.Context) (Status, error) {
+	// rc-service exit codes:
```
twz123, line 82: Would be a bit nicer to read if we'd place the comment lines directly above the case arms to which they belong, similar to how it's done for systemd.

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+	code := exitCode(err)
+	if code < 0 {
+		return StatusUnknown, err
+	}
+	switch code {
+	case 0, 8:
+		return StatusRunning, nil
+	case 1:
+		return StatusNotInstalled, nil
+	case 3, 4, 16, 32:
```
twz123: Hmm, stopping != stopped?

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"github.com/k0sproject/k0s/internal/pkg/dir"
+	"github.com/k0sproject/k0s/internal/pkg/file"
+)
+
+type openrcService struct {
+	name   string
+	root   string
+	runner runner
+	exec   func() (string, error)
```
twz123, line 24: We could rip out this field, and instead add it as an arg for Service.Install à la:

`Install(ctx context.Context, exe string, args []string, env []string) error`, or maybe even struct this as `Install(ctx context.Context, opts InstallOpts) error` and

```go
type InstallOpts struct {
	ExecutablePath string
	Args           []string
	Environment    []string
}
```

pkg/install/service.go
```diff
@@ -4,141 +4,97 @@
...
 
-	"github.com/kardianos/service"
+	"github.com/k0sproject/k0s/pkg/sysservice"
 	"github.com/sirupsen/logrus"
 )
 
-var (
-	k0sServiceName = "k0s"
-	k0sDescription = "k0s - Zero Friction Kubernetes"
-)
```
twz123, line 18: Might make sense to keep them around (as consts), so we can use them in the appropriate places in the different service implementations?

pkg/sysservice/backend_openrc_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}{
+		{`KEY=value`, `KEY='value'`},
+		{`KEY=has space`, `KEY='has space'`},
+		{`KEY=$dollar`, `KEY='$dollar'`},
+		{`KEY=$(subshell)`, `KEY='$(subshell)'`},
+		{`KEY=it's`, `KEY='it'\''s'`},
+		{`NOEQUAL`, `NOEQUAL`},
+	}
+	for _, tt := range tests {
+		assert.Equal(t, tt.want, quoteOpenRCEnv(tt.in), "quoteOpenRCEnv(%q)", tt.in)
```
twz123, line 73: I'd rather test that the resulting service script contains the escaped values, or let's phrase it differently: I'd at least verify _once_ that quoting is actually applied in the script. Testing all the edge cases like this is fine, though.

pkg/sysservice/backend_openrc.go
```diff
@@ -0,0 +1,190 @@
...
+	switch code {
+	case 0, 8:
+		return StatusRunning, nil
+	case 1:
+		return StatusNotInstalled, nil
+	case 3, 4, 16, 32:
+		return StatusStopped, nil
+	default:
+		return StatusUnknown, nil
+	}
```
twz123: I'm a bit on the fence with the `exitCode()` function. I somehow think this might be more idiomatic: 

```suggestion
	if err := s.runner.Run(ctx, "rc-service", s.name, "status"); err != nil {
		var exitCoder interface{ ExitCode() int }
		if errors.As(err, &exitCoder) {
			switch exitCoder.ExitCode() {
			case 1:
				return StatusNotInstalled, nil
			case 3, 16, 32:
				return StatusStopped, nil
			case 4, 8:
				return StatusRunning, nil
			}
		}
		return StatusUnknown, err
	}

	return StatusRunning, nil
```

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func (s *windowsService) Install(ctx context.Context, args []string, env []string) error {
+	m, err := mgr.Connect()
+	if err != nil {
+		return fmt.Errorf("connect to SCM: %w", err)
+	}
+	defer func() { _ = m.Disconnect() }()
+
+	displayName := strings.Replace(s.name, "k0s", "k0s ", 1)
```
twz123: That looks kinda brittle :see_no_evil:

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		if isNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("open service: %w", err)
+	}
+	defer sv.Close()
+
+	// Best-effort stop before deleting.
+	_ = stopAndWait(ctx, sv)
```
twz123: Intuitively, I'd say we should check and forward the error here?

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (s *windowsService) Enable(ctx context.Context) error {
+	return s.withService(ctx, func(sv *mgr.Service) error {
+		return setStartType(sv, mgr.StartAutomatic)
+	})
+}
+
+func (s *windowsService) Disable(ctx context.Context) error {
+	return s.withService(ctx, func(sv *mgr.Service) error {
+		return setStartType(sv, mgr.StartDisabled)
```
twz123: Not `StartManual`?

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+			`SYSTEM\CurrentControlSet\Services\`+s.name,
+			registry.SET_VALUE)
+		if err != nil {
+			return fmt.Errorf("open service registry key: %w", err)
+		}
+		defer key.Close()
+		if err := key.SetStringsValue("Environment", env); err != nil {
+			return fmt.Errorf("set service environment: %w", err)
+		}
+	}
```
twz123, line 76: Should we try to clean up after ourselves if those things fail, i.e. try to delete the half-installed service?

pkg/sysservice/backend_windows.go
twz123: btw the whole method looks unused

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func stopAndWait(ctx context.Context, sv *mgr.Service) error {
+	if _, err := sv.Control(svc.Stop); err != nil {
+		if errors.Is(err, windows.ERROR_SERVICE_NOT_ACTIVE) {
+			return nil
+		}
+		return err
+	}
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
twz123: Why are we adding this artificial timeout? We don't do it for the other service implementations, which simply block for as long as the service manager deems necessary. Windows is not different here, AFAICT. It will kill the service process eventually, if it doesn't react to a service shutdown request in time, and I think we can just block for as long as that takes. Users still have the possibility to interrupt this via Ctrl+C or the like.

pkg/sysservice/backend_windows.go
```diff
@@ -0,0 +1,216 @@
...
+			return err
+		}
+		if status.State == svc.Stopped {
+			return nil
+		}
+		time.Sleep(200 * time.Millisecond)
+	}
+}
+
+func isNotExist(err error) bool {
```
twz123: Inlining this function would probably improve readability.

internal/pkg/file/file.go
ncopa: Not sure how the atomic helpers would help here. The behavior we want is "try write file and fail if it exists", and the simplest way to do that is with `O_EXCL`. Using `O_TRUNC` will make it overwrite an existing file, which is why we cannot use `os.WriteFile`.

We could go back to test if file exists like first implementations did and like kardianos does, but that gives us a TOCTOU race window.

pkg/sysservice/backend_openrc.go
ncopa: Will that not just overwrite the file if it already exists? we want the `O_EXCL` behavior so it fails (unless `--force` flag is used).

pkg/sysservice/sysservice.go
ncopa: I'd like to keep them separate for now.

pkg/sysservice/sysservice.go
ncopa: > I'd have expected it to return some "no such service" error.

I think that is how kardianos works, so maybe we should do that as well.

pkg/install/service.go
ncopa: Not sure what value that brings? Should we also have a const for "controller" and "worker"?

pkg/sysservice/backend_openrc.go
ncopa: we could but then we need to refactor the tests.

ncopa:
I think we should move this to next milestone. It was way more work than expected to get this into a good enough shape. I won't make it for 1.36.

pkg/sysservice/backend_systemd.go
ncopa: thatIsThePriceWePayForCamelCase. Personally_i_find_snake_case_much_easier_to_read_but_who_am_I_to_have_opinion_on_that.

pkg/sysservice/runner.go
ncopa: done

pkg/sysservice/backend_windows.go
ncopa: done

pkg/sysservice/runner_test.go
ncopa: done

pkg/sysservice/sysservice.go
ncopa: `Uninstall` should now return error.
`Restart` removed.
I kept `Enable` as I think that makes it cleaner.

pkg/sysservice/backend_systemd.go
ncopa: link to sources was added. I think the return codes was wrong, but they have been fixed now.

pkg/sysservice/backend_openrc.go
ncopa: done

pkg/sysservice/backend_openrc.go
ncopa: good point. fixed

pkg/sysservice/backend_openrc.go
ncopa: service may been manually disabled, and that is fine and can be ignored.
I have added comment.

pkg/sysservice/backend_openrc.go
ncopa: done.

pkg/sysservice/backend_openrc.go
ncopa: done

pkg/sysservice/backend_openrc_test.go
ncopa: done

pkg/install/service.go
ncopa: done

pkg/sysservice/backend_openrc.go
ncopa: done

pkg/sysservice/backend_openrc.go
ncopa: fixed.

twz123:
@ncopa wanna pick up on this, once again?

Issue #6915, opened 2025-12-19
Possibility to reference Secret values for helm registry authentication


a13x5:
### Is your feature request related to a problem? Please describe.

Currently when it's required to use registry with authentication the credentials must be passed directly to the object in plain text in the following fields:

- `spec.k0sConfigSpec.k0s.extensins.helm.repositories[0].username`
- `spec.k0sConfigSpec.k0s.extensins.helm.repositories[0].password`
- `spec.k0sConfig.extensins.helm.repositories[0].username`
- `spec.k0sConfig.extensins.helm.repositories[0].password`

This is insecure in a sense that may expose actual credentials to unwanted audience and k8s objects generally not meant for holding sensitive data.

I understand that k0s is section currently just a simple k0s configuration and it's pretty simple in a sense that there is no additional processing of it (or at least not much).

### Describe the solution you would like

The registry authentication currently may be done from k0smotron by passing containerd configuration to k0s, which will be applied pretty much for all interactions with the registry. Additionally helm must be configured which has the problem described above. 
There could be several possible solutions to the problem:
- Helm controller in k0s will read registry configuration from containerd and uses credentials passed there.
- Helm controller in k0s use special credential secret in `kube-system` namespace as a source for helm repository. This secret can be the passed as a static file using k0smotron `files.[].contentsFrom.secretRef` mechanism
- k0smotron adds additional field to provide registry authentication parameters in for of a Secret and creates the needed containerd configuration + modifies k0s configuration as needed to incorporate the auth details
- k0smtron adds additional field to pass Secret reference only for helm registry auth details and modifies the k0s configuration fields as needed.

### Describe alternatives you've considered

The plain text values were considered, but because of use of helm as a templating system the plain text authentication values are exposed in multiple objects throughout the system.

jnummelin:
Transferred from k0smotron since this is more of an k0s related challenge

jnummelin:
> Helm controller in k0s will read registry configuration from containerd and uses credentials passed there.

I don't htink this is actually possible. The Helm controller is running on the CP node(s) which might not even have containerd running.

> Helm controller in k0s use special credential secret in kube-system namespace as a source for helm repository

I don't think a single secret suffices, user might configure multiple repos each requiring it's own creds. We'd need at least some "well-known" naming scheme, something like `helm-<repo-name>`.

> k0smtron adds additional field to pass Secret reference only for helm registry auth details and modifies the k0s configuration fields as needed.

[...]

PR #7060, opened 2026-02-04
Refactor Helm reconciler to use ephemeral environments

jnummelin:
## Description

Replace persistent Helm repository initialization with ephemeral, per-chart environments. This eliminates timing dependencies between repository setup and chart installation while simplifying the architecture.

Allows much easier implementation for things like #6915

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Existing cases cover this

## Checklist

### DIFF
--- a/pkg/component/controller/extensions_controller.go
+++ b/pkg/component/controller/extensions_controller.go
@@ -9,10 +9,13 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
 	"slices"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
@@ -36,6 +39,7 @@ import (
 	"github.com/avast/retry-go"
 	"github.com/bombsimon/logrusr/v4"
 	"github.com/sirupsen/logrus"
+	"helm.sh/helm/v3/pkg/registry"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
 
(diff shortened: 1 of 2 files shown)


pkg/component/controller/extensions_controller.go
```diff
@@ -51,14 +53,53 @@ import (
...
+	return &repo
+}
+
+// update replaces the entire cache contents with the provided repositories
+func (rc *repositoryCache) update(repositories []k0sv1beta1.Repository) {
+	// Clear existing cache
+	rc.cache.Range(func(key, value interface{}) bool {
+		rc.cache.Delete(key)
+		return true
+	})
```
twz123: Sine we replace the whole map, maybe it's easier to simply use a `map[string]k0sv1beta1.Repository` plus a `sync.Mutex` instead of a `sync.Map`?

We could simply build the new map before acquiring the mutex. Then the critical section would become a simple assignment.

pkg/component/controller/extensions_controller.go
```diff
@@ -312,6 +372,46 @@ func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv
...
+		// Remove the chart path, keep only the hostname
+		parts := strings.Split(repoURL, "/")
+		if len(parts) > 0 {
+			hostname := parts[0]
+			// Look up repository by hostname (OCI repos are keyed by hostname)
+			// For OCI charts, repository configuration is optional (allows anonymous access)
+			return cr.repositoryCache.get(hostname), nil
+		}
+		return nil, fmt.Errorf("invalid OCI chart URL format: %s", chartName)
+	}
```
twz123: I'm not too happy with this ad-hoc parsing of URLs here 🙈 

We have already helpers for this in pkg/helm/oci.go (ociRegistryManager.AddRegistry). They're currently not public, but it might make sense to refactor them a bit to make them reusable here, as well, so we don't have the same logic in multiple places.

pkg/component/controller/extensions_controller.go
```diff
@@ -51,14 +55,60 @@ import (
...
+				logrus.WithError(err).Warnf("Invalid repository URL %q, skipping", repo.URL)
+				continue
+			}
+			repoName = url.Host
+		} else {
+			// Traditional repo: use name as key
+			repoName = repo.Name
+		}
+		rc.cache[repoName] = repo
+	}
```
twz123: I was thinking about building the new cache without holding the lock:

```suggestion
	cache := make(map[string]k0sv1beta1.Repository)

	// Populate with new repositories
	// OCI repos are keyed by hostname, traditional repos by name
	var repoName string
	for _, repo := range repositories {
		if registry.IsOCI(repo.URL) {
			// Extract hostname from OCI URL and use as key
			url, err := url.Parse(repo.URL)
			// This shouldn't really happen at this point but just in case let's log it
			if err != nil {
				logrus.WithError(err).Warnf("Invalid repository URL %q, skipping", repo.URL)
				continue
			}
			repoName = url.Host
		} else {
			// Traditional repo: use name as key
			repoName = repo.Name
		}
		cache[repoName] = repo
	}

	// Replace old cache with new one
	rc.mutex.Lock()
	rc.cache = cache
	rc.mutex.Unlock()
```

pkg/component/controller/extensions_controller.go
```diff
@@ -312,6 +381,46 @@ func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv
...
+	parts := strings.SplitN(chartName, "/", 2)
+	if len(parts) != 2 {
+		return nil, fmt.Errorf("invalid chart name format (expected 'reponame/chartname'): %s", chartName)
+	}
+
+	repoName := parts[0]
+	repo := cr.repositoryCache.get(repoName)
+	if repo == nil {
+		return nil, fmt.Errorf("repository '%s' not found in cluster configuration", repoName)
+	}
```
twz123, line 422: I'm still not too happy with the URL parsing. Also, `strings.Cut` is a better fit than `strings.SplitN`.

```suggestion
	var repoName string

	// Check if it's an OCI chart
	if chartURL, err := url.Parse(chartName); err == nil && chartURL.Scheme == "oci" {
		if chartURL.Host == "" {
			return nil, fmt.Errorf("invalid OCI chart URL format: %s", chartName)
		}

		repoName = chartURL.Host
	} else if repo, _, ok := strings.Cut(chartName, "/"); !ok || repo == "" { // Traditional format: "reponame/chartname"
		repoName = repo
	} else {
		return nil, fmt.Errorf("invalid chart name format (expected 'reponame/chartname' or a valid OCI registry URL): %s", chartName)
	}

	repo := cr.repositoryCache.get(repoName)
	if repo == nil {
		return nil, fmt.Errorf("repository '%s' not found in cluster configuration", repoName)
	}
```

pkg/component/controller/extensions_controller.go
```diff
@@ -312,6 +381,46 @@ func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv
...
+	if strings.HasPrefix(chartName, "oci://") {
+		// Extract hostname from OCI URL
+		repoURL := strings.TrimPrefix(chartName, "oci://")
+		// Remove the chart path, keep only the hostname
+		parts := strings.Split(repoURL, "/")
+		if len(parts) > 0 {
+			hostname := parts[0]
+			// Look up repository by hostname (OCI repos are keyed by hostname)
+			// For OCI charts, repository configuration is optional (allows anonymous access)
+			return cr.repositoryCache.get(hostname), nil
```
twz123: I think this is missing a check for the empty string.

pkg/component/controller/extensions_controller.go
jnummelin: > I'm still not too happy with the URL parsing

Not happy about it either, missed this part of fixing things while fixing the other 🤦

PR #7063, opened 2026-02-04
Fix NLLB SELinux labeling on enforcing systems

### LINKED ISSUE #7057: NLLB doesn't work with SELinux

### Platform

```shell

```

### Version

v1.32.6+k0s.0

### Sysinfo

<pre>

</pre>
</details>

### What happened?

When NLLB is enabled, k0s fails to install on SELinux nodes

Logs in k0s service (controller and worker) : 
```
[critical][main] [source/server/server.cc:412] error initializing config '/etc/envoy/envoy.yaml': unable to read file: /etc/envoy/envoy.yaml
[info][main] [source/server/server.cc:1038] exiting
unable to read file: /etc/envoy/envoy.yaml
```

My root cause analysis is that NLLB features works by deploying Envoy Proxy static Pods which are mounting config files from the host.

```
   - mountPath: /etc/envoy
      name: envoy-config
      readOnly: true
(...)
  volumes:
  - hostPath:
      path: /run/k0s/nllb/envoy
      type: Directory
    name: envoy-config
```

When containerd is configured with SELinux, Pods are started with the `container_t` context and can only access files with the `container_file_t` label

```
# on a node
ps -eZ | grep envoy
system_u:system_r:container_t:s0:c198,c388 119165 ? 00:00:01 envoy
```
The problem is that k0s creates files in `/run/k0s/nllb/envoy` which are tagged with the `var_run_t` label

```
[root@pac-manager-0 rocky]# ls -lZ /run/k0s/nllb/envoy/
total 8
-r--r--r--. 1 root root system_u:object_r:var_run_t:s0 1122 Jan 19 22:26 cds.yaml
-r--r--r--. 1 root root system_u:object_r:var_run_t:s0  850 Jan 19 22:26 envoy.yaml
```
And SELinux blocks the Envoy Proxy process from accessing the file, which makes the Pod fails.

Message in SELinux audit logs confirms that :  

```
type=AVC msg=audit(...): avc:  denied  { read } for  pid=30201 comm="envoy" name="envoy.yaml" dev="tmpfs" ino=2276 scontext=system_u:system_r:container_t:s0:c263,c846 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=0
 ```

### Steps to reproduce

1. Enable NLLB in the k0s config
2. Deploy k0s on an SELinux node

### Expected behavior

NLLB works

### Actual behavior

NLLB doesn't work

### Screenshots and logs

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by jnummelin**
## Description

Set SELinux labels on NLLB Envoy configuration files to allow containers to read them. Files created in /run/k0s/nllb/envoy/ inherit var_run_t label by default, which containers with container_t context cannot access. The fix programmatically sets container_file_t label on the directory and config files when SELinux is enabled.

Scoped the file labeling only to NLLB since currently it's the only component using static pods and such mounts.

Fixes #7057

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

Manually tested on Rocky9 in enforcing mode:
```sh
[root@lima-rocky9-test ~]# getenforce 
Enforcing
[root@lima-rocky9-test ~]# ls -laZ /run/k0s/nllb/envoy/
total 8
drwxr-xr-x. 2 root root system_u:object_r:container_file_t:s0   80 Feb  4 15:04 .
drwx------. 3 root root system_u:object_r:var_run_t:s0          80 Feb  4 15:03 ..
-r--r--r--. 1 root root system_u:object_r:container_file_t:s0 1128 Feb  4 15:04 cds.yaml
-r--r--r--. 1 root root system_u:object_r:container_file_t:s0  850 Feb  4 15:04 envoy.yaml
[root@lima-rocky9-test ~]# k0s kc get pod -A
NAMESPACE     NAME                             READY   STATUS    RESTARTS   AGE
kube-system   coredns-55c758887c-dfcpz         1/1     Running   0          13m
kube-system   konnectivity-agent-2hmgs         1/1     Running   0          12m
kube-system   kube-proxy-t8x4s                 1/1     Running   0          13m
kube-system   kube-router-zn8m6                1/1     Running   0          13m
kube-system   metrics-server-654657c67-rljqr   1/1     Running   0          13m
kube-system   nllb-lima-rocky9-test            1/1     Running   0          13m
```

## Checklist

### DIFF
--- a/docs/selinux.md
+++ b/docs/selinux.md
@@ -57,11 +57,15 @@ sudo restorecon -R -v ${DATA_DIR}/bin
 sudo semanage fcontext -a -t container_var_lib_t "${DATA_DIR}/containerd(/.*)?"
 sudo semanage fcontext -a -t container_ro_file_t "${DATA_DIR}/containerd/io.containerd.snapshotter.*/snapshots(/.*)?"
 sudo restorecon -R -v ${DATA_DIR}/containerd
+sudo semanage fcontext -a -t container_file_t "/etc/cni/.*"
+sudo restorecon -R -v /etc/cni
+sudo semanage fcontext -a -t container_file_t "/opt/cni/.*"
+sudo restorecon -R -v /opt/cni
 ```
 
 ### Enable SELinux in containerd of k0s
 
-Add below lines to `/etc/k0s/containerd.toml` of the worker nodes. You need to restart k0s service on the node to make the change take effect.
+Add below lines to `/etc/k0s/containerd.d/selinux.toml` of the worker nodes. k0s will automatically restart containerd with the new configuration if needed.
 
 ```toml
 [plugins."io.containerd.grpc.v1.cri"]


pkg/component/worker/nllb/envoy.go
```diff
@@ -87,16 +88,41 @@ type envoyConfig struct {
 const (
 	envoyBootstrapFile = "envoy.yaml"
 	envoyCDSFile       = "cds.yaml"
+
+	// containerFileLabel is the SELinux label applied to files that need to be
+	// readable by containers. Containers typically run with the container_t type,
+	// which can read files labeled with container_file_t.
+	containerFileLabel = "system_u:object_r:container_file_t:s0"
```
twz123: No idea about SELinux, but I hope this is not something that changes from distro to distro? Is there some spec that we can reference here, to be sure to have the right label?

pkg/component/worker/nllb/envoy.go
```diff
@@ -87,16 +88,41 @@ type envoyConfig struct {
...
+
 	return nil
 }
 
+// setContainerFileLabel sets the SELinux label on a file or directory to allow
+// containers to access it. This is necessary on SELinux-enabled systems where
+// files created in /run get the var_run_t label by default, which containers
+// with container_t context cannot read.
+func setContainerFileLabel(path string) {
+	if !selinux.GetEnabled() {
```
twz123: Don't we have the sysinfo stuff that checks for SELinux? Might be only for AppArmor.

pkg/component/worker/nllb/envoy.go
```diff
@@ -87,16 +88,41 @@ type envoyConfig struct {
...
+		return
+	}
+
+	// Set the label to container_file_t which is readable by containers
+	if err := selinux.SetFileLabel(path, containerFileLabel); err != nil {
+		// Log as warning rather than error - SELinux labeling is best-effort
+		// The system might have SELinux in permissive mode or different policies
+		logrus.WithError(err).WithField("path", path).Warn("Failed to set SELinux label on file")
```
twz123: Any reason not to return the error here? I mean we already know that things won't work when the label is not in place.

pkg/component/worker/nllb/envoy.go
```diff
@@ -229,6 +256,9 @@ func writeEnvoyConfigFiles(params *envoyParams, filesParams *envoyFilesParams) e
 		})
 		if err != nil {
 			errs = append(errs, fmt.Errorf("failed to write %s: %w", fileName, err))
+		} else {
+			// Set SELinux label after successful write
+			setContainerFileLabel(filePath)
```
twz123: This should happen before we're replacing the file. WDYT about adding this properly to atomic.go?

jnummelin: > This should happen before we're replacing the file.

oh, right. Envoy could pick up the file after writing before the label is applied. 

> WDYT about adding this properly to atomic.go?

I was thinking about that. But it'd have to be some "after file is created" kind of hook since we do not do these for other files that these, currently at least. And since we do this for 2 files in NLLB only, seemed bit overkill. But since there is a timing issue in this way, need to think how to hook it up properly for atomic.go

pkg/component/worker/nllb/envoy.go
jnummelin: In the docs we instruct folks to install this:

```
dnf install -y container-selinux
```

This is where those come from. See https://github.com/containers/container-selinux

I could add some commentary on these in the code

pkg/component/worker/nllb/envoy.go
jnummelin: Nope, at least I can't find any for SELinux. We do have some apparmor checks though

pkg/component/worker/nllb/envoy.go
jnummelin: Added this functionality to atomix and dir packages now. PTAL

twz123 (approved):

PR #7083, opened 2026-02-09
Embed repo info to Chart CRD and enabled repo config via secrets

jnummelin:
## Description

This way the Chart becomes a self-sufficient entity and does not have any dependency to k0s configuration.

Enables Charts to reference Kubernetes Secrets for repository credentials and TLS certificates, supporting both traditional Helm repos and OCI registries.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


pkg/apis/helm/v1beta1/chart_types.go
```diff
@@ -10,8 +10,47 @@ import (
...
+
+	k0sv1beta1 "github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 )
 
+// RepositorySpec describes a Helm repository configuration for a Chart.
+// Fields map to the CLI flags for the "helm repo add" command.
+type RepositorySpec struct {
+	// The repository URL.
+	// +kubebuilder:validation:MinLength=1
+	URL string `json:"url,omitempty"`
```
juanluisvaladas, line 39: I don't think this should have omitempty, also there might be a better validation.

```suggestion
	// The repository URL.
	// +kubebuilder:validation:Pattern=`^((file|https?|oci)://)|/\S+$`
	URL string `json:"url"`
```

I haven't actually checked the validation pattern. Can be incorrect.

juanluisvaladas (changes requested):

pkg/apis/helm/v1beta1/chart_types.go
jnummelin: hmm, technically it can be also http, I _think_. Wonder if in theory it could be also some local file system path 🤔

juanluisvaladas (dismissed):

juanluisvaladas (approved):

pkg/helm/helm.go
```diff
@@ -27,10 +27,32 @@ import (
...
+	// These take precedence over file paths when both are set
+	CAData   []byte
+	CertData []byte
+	KeyData  []byte
+	Insecure *bool
+}
+
+// IsInsecure returns true if TLS verification should be skipped.
+func (r *Repository) IsInsecure() bool {
+	return r.Insecure != nil && *r.Insecure
```
twz123, line 53: Oops, this switched the default behavior from insecure-by-default to secure-by-default. Not saying that this isn't a good change per se, but it would have been good to call out in the release notes. See #4660.

Issue #7109, opened 2026-02-12
k0s controller restart may permanently break extensions


byDimasik:
### Before creating an issue, make sure you've checked the following:

- [x] You are running the latest released version of k0s
- [x] Make sure you've searched for existing issues, both open and closed
- [x] Make sure you've searched for PRs too, a fix might've been merged already
- [x] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.8.0-1044-aws #46~22.04.1-Ubuntu SMP Tue Dec  2 12:52:18 UTC 2025 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Version

v1.32.11+k0s.0

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>

Total memory: 15.4 GiB (pass)
File system of /var/lib/k0s: ext4 (pass)
Disk space available for /var/lib/k0s: 85.2 GiB (pass)
Relative disk space available for /var/lib/k0s: 88% (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.8.0-1044-aws (pass)
  Max. file descriptors per process: current: 1048575 / max: 1048576 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
[...]

PR #7130, opened 2026-02-16
Fix Helm extensions stuck in pending states after controller restart

jnummelin:
## Description

When k0s controller stops (systemd restart, process kill), controller-runtime cancels the manager's context as part of its shutdown sequence. Since leader election runs the extensions controller via RunLeaderTasks, losing the leader lease triggers immediate context cancellation (pkg/leaderelection/tasks.go:32).

This context cancellation propagates to in-flight Helm operations. When Helm's atomic install/upgrade operations see a cancelled context, they abort mid-execution, leaving releases stuck in pending states (pending-install, pending-upgrade, pending-rollback, or uninstalling). GracefulShutdownTimeout cannot prevent this because the context is already cancelled before the timeout mechanism activates.

This implements automatic recovery by detecting and cleaning up broken releases before each install/upgrade operation:

- pending-install: Complete the interrupted uninstall (with hooks disabled to avoid "already exists" errors from partially-created hook resources), then proceed with fresh install
- pending-upgrade/pending-rollback: Roll back to last successful deployment, then retry the desired upgrade
- uninstalling: Complete the interrupted uninstall (with hooks disabled)

The cleanup mechanism uses structured logging for operational visibility and is idempotent (safe to call multiple times). Integration test validates recovery with a chart containing a 60-second post-install hook that guarantees catching controller restart during Helm execution.

Fixes: https://github.com/k0sproject/k0s/issues/7109
Related: https://github.com/helm/helm/issues/7476, https://github.com/operator-framework/helm-operator-plugins/issues/94

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


jnummelin:
Flux uses similar logic in https://github.com/fluxcd/helm-controller/blob/main/internal/reconcile/atomic_release.go#L519-L547

inttest/addons/testdata/slow-hook-chart/templates/_helpers.tpl
twz123, line 1: While those are typical for a real chart, I'd rather toss away these helpers. We can get along without them for testing (just using .Chart.Name and .Release.Name directly). This means less lines/files to review and maintain.

inttest/addons/testdata/slow-hook-0.1.0.tgz
twz123, line 1: Not too fond of binary files in repositories 🙈 What's in here, and can't we get along without it?

inttest/addons/testdata/slow-hook-chart/templates/job-postinstall.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "slow-hook.fullname" . }}-postinstall
+  annotations:
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": hook-succeeded
```
twz123, line 8: The double quotes around the keys aren't needed here.

pkg/helm/helm.go
```diff
@@ -417,17 +418,39 @@ func (hc *Commands) ListReleases(namespace string) ([]*release.Release, error) {
...
 	helmAction := action.NewList(cfg)
+	helmAction.StateMask = action.ListAll // Include all release states: deployed, failed, pending-install, pending-upgrade, pending-rollback, uninstalling, superseded
 	return helmAction.Run()
 }
 
+// GetReleaseStatus returns the status of a specific Helm release by searching
+// through all releases in the namespace. We use ListReleases instead of Helm's
+// native status check to ensure we can find releases in any state (not just deployed).
+// This is primarily used by the cleanup logic to detect stuck releases.
+func (hc *Commands) GetReleaseStatus(releaseName string, namespace string) (release.Status, error) {
```
twz123, line 429:
```suggestion
func (hc *Commands) GetReleaseStatus(releaseName, namespace string) (release.Status, error) {
```

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+//
+// When k0s stops, the leader lease is lost immediately, causing leaderelection.RunLeaderTasks
+// to cancel the manager's context (see pkg/leaderelection/tasks.go:32). This bypasses
+// GracefulShutdownTimeout, interrupting in-flight Helm operations mid-execution.
+//
+// If Helm's atomic rollback/uninstall is interrupted, releases are left in unusable states:
+// - pending-install: Install was interrupted before completion
+// - pending-upgrade: Upgrade was interrupted before completion
+// - pending-rollback: Atomic rollback was interrupted
+// - uninstalling: Uninstall operation was interrupted
```
twz123, line 615: Indent this (and the other bullet points) by two additional spaces. This way, it's properly rendered as a list by godoc: https://go.dev/doc/comment#lists

```suggestion
//   - pending-install: Install was interrupted before completion
//   - pending-upgrade: Upgrade was interrupted before completion
//   - pending-rollback: Atomic rollback was interrupted
//   - uninstalling: Uninstall operation was interrupted
```

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
 		chart.Status.ValuesHash != chart.Spec.HashValues()
 }
 
+// cleanupBrokenRelease detects and recovers from Helm releases stuck in pending states.
+// This situation occurs when k0s controller restarts during a Helm operation:
+//
+// When k0s stops, the leader lease is lost immediately, causing leaderelection.RunLeaderTasks
+// to cancel the manager's context (see pkg/leaderelection/tasks.go:32). This bypasses
```
twz123, line 608: I don't think this kind of inter source file links is sustainable.

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+// GracefulShutdownTimeout, interrupting in-flight Helm operations mid-execution.
```
twz123, line 609: Well, that's not true, is it? This is actually the problem: There's no way to really interrupt any operations at all. We can just stop waiting for them. I've seen this wording occurs in the commit message, as well.

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+//
+// If Helm's atomic rollback/uninstall is interrupted, releases are left in unusable states:
+// - pending-install: Install was interrupted before completion
+// - pending-upgrade: Upgrade was interrupted before completion
+// - pending-rollback: Atomic rollback was interrupted
+// - uninstalling: Uninstall operation was interrupted
+//
+// Recovery strategy:
```
twz123, line 617: Maybe, a bit more text would be better:

```go
// As per [Helm issue 7476], the following cleanup operations will be performed:
//   - ...
//
// [Helm issue 7476]: https://github.com/helm/helm/issues/7476
```

Or something along these lines.

pkg/component/controller/extensions_controller.go
twz123, line 609: We're mixing different things here without much context. First, there's k0s's leader election. Then, there's the controller-runtime's graceful shutdown timeout. Finally, there's Helm itself. None of these are important here except Helm. It doesn't matter what cancels the context because there isn't a way to interrupt a Helm action, which is the root cause IMO.

pkg/helm/helm.go
```diff
@@ -417,17 +418,39 @@ func (hc *Commands) ListReleases(namespace string) ([]*release.Release, error) {
...
+	releases, err := hc.ListReleases(namespace)
+	if err != nil {
+		return "", fmt.Errorf("can't list releases: %w", err)
+	}
+	for _, rel := range releases {
+		if rel.Name == releaseName {
+			return rel.Info.Status, nil
+		}
+	}
+	return "", fmt.Errorf("release %q not found in namespace %q", releaseName, namespace)
```
twz123, line 439: Would be more intuitive to use a Helm action, wouldn't it?


```suggestion
	cfg, err := hc.getActionCfg(namespace)
	if err != nil {
		return "", fmt.Errorf("can't create helmAction configuration: %w", err)
	}
	helmAction := action.NewGet(cfg)
	rel, err := helmAction.Run(releaseName)
	if err != nil {
		return "", err
	}

	return rel.Info.Status, nil
```

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+// - pending-install: Uninstall (no previous version to preserve)
+// - pending-upgrade/pending-rollback: Rollback to last deployed revision (preserves workloads)
+// - uninstalling: Complete the uninstall
+//
+// See: https://github.com/helm/helm/issues/7476
+func (cr *ChartReconciler) cleanupBrokenRelease(ctx context.Context, helmCmd *helm.Commands, releaseName, namespace string) error {
+	status, err := helmCmd.GetReleaseStatus(releaseName, namespace)
+	if err != nil {
+		// Release not found is not an error - nothing to clean up
+		if errors.Is(err, driver.ErrReleaseNotFound) || strings.Contains(err.Error(), "not found") {
```
twz123, line 627: We should rely on a proper error for the second part, as well. I guess switching to a proper Helm action in `GetReleaseStatus` will do the trick already, making the `strings.Contains` superfluous.

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+// - pending-rollback: Atomic rollback was interrupted
+// - uninstalling: Uninstall operation was interrupted
+//
+// Recovery strategy:
+// - pending-install: Uninstall (no previous version to preserve)
+// - pending-upgrade/pending-rollback: Rollback to last deployed revision (preserves workloads)
+// - uninstalling: Complete the uninstall
+//
+// See: https://github.com/helm/helm/issues/7476
+func (cr *ChartReconciler) cleanupBrokenRelease(ctx context.Context, helmCmd *helm.Commands, releaseName, namespace string) error {
```
twz123, line 623: Maybe a better method name would be something along the lines of `detectAndRecoverBrokenRelease`? This is what the first godoc sentence says and I think it better describes what this is doing. "Cleanup" sounds like "delete", but that's not necessarily what happens.

jnummelin:
I tested the upgrade case manually, this is the resulting history for the related chart:
```
[root@lima-rocky9-test ~]# helm history -n sleeper sleeper
REVISION	UPDATED                 	STATUS          	CHART        	APP VERSION	DESCRIPTION                               
1       	Mon Feb 16 22:59:48 2026	superseded      	sleeper-0.1.0	1.0        	Install complete                          
2       	Mon Feb 16 23:10:24 2026	failed          	sleeper-0.2.0	2.0        	Upgrade "sleeper" failed: context canceled
3       	Mon Feb 16 23:10:37 2026	pending-rollback	sleeper-0.1.0	1.0        	Rollback to 1                             
4       	Mon Feb 16 23:12:04 2026	superseded      	sleeper-0.2.0	2.0        	Rollback to 2                             
5       	Mon Feb 16 23:12:04 2026	deployed        	sleeper-0.2.0	2.0        	Upgrade complete  
```

So first the upgrade failed (2), since I restarted the controller mid-fly which results Helm trying to "undo" automatically (3) because we use `Atomic=true`. After k0s restart it detects the broken state and rollbacks (4) and after that successfully upgrades (5).

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+	case release.StatusPendingInstall:
+		// Install was interrupted - uninstall to allow retry
+		// DisableHooks=true because if the install was interrupted during atomic rollback,
+		// hook resources may already exist from the failed install attempt
+		action = "uninstall"
+		cr.L.WithFields(logrus.Fields{
+			"release":   releaseName,
+			"namespace": namespace,
+			"status":    status,
+			"action":    action,
```
twz123, line 647: I'd rather put the log setup in front of the switch, and do all the error handling inside the individual cases:

```go
	log := cr.L.WithFields(logrus.Fields{
		"release":   releaseName,
		"namespace": namespace,
		"status":    status,
	})

	switch status {
	case release.StatusPendingInstall:
		// ...
		log.Warn("Detected stuck installation, uninstalling to allow retry")
		if err := helmCmd.UninstallRelease(ctx, releaseName, namespace, true); err != nil {
			return err
		}

	// ...
	}
```

pkg/component/controller/extensions_controller.go
```diff
@@ -590,6 +601,105 @@ func (cr *ChartReconciler) chartNeedsUpgrade(chart helmv1beta1.Chart) bool {
...
+		}).Warn("Detected interrupted uninstall, completing operation (skipping hooks)")
+		recoveryErr = helmCmd.UninstallRelease(ctx, releaseName, namespace, true)
+
+	default:
+		// Release is in a stable state - no cleanup needed
```
twz123, line 676:
```suggestion
		log.Debug("Release is in a stable state, no recovery needed")
```

pkg/component/controller/extensions_controller.go
```diff
@@ -723,13 +833,22 @@ func (ec *ExtensionsController) instantiateManager(ctx context.Context) (crman.M
...
 		Logger:     logrusr.New(ec.L),
 		Controller: ctrlconfig.Controller{},
+		// Note: GracefulShutdownTimeout is set to 30 minutes, but in practice this timeout
+		// is largely ineffective for the extensions controller. When k0s stops, the leader
+		// lease is immediately lost, which causes leaderelection.RunLeaderTasks to cancel
+		// the manager's context immediately (see pkg/leaderelection/tasks.go:32).
+		// This means controller-runtime never gets time to complete in-flight Helm operations.
+		// The primary defense against interrupted Helm operations is the cleanupBrokenRelease
+		// mechanism in the ChartReconciler, which detects and recovers from stuck releases.
+		GracefulShutdownTimeout: &shutDownDuration,
```
twz123, line 851: I think we can just drop this. I'm also not following the reasoning given in the comment (see my comment about the godoc of cleanupBrokenRelease).

```suggestion
```

pkg/helm/helm.go
```diff
@@ -358,6 +358,7 @@ func (hc *Commands) InstallChart(ctx context.Context, chartName string, version
 	if err != nil {
 		return nil, fmt.Errorf("can't install loadedChart `%s`: %w", loadedChart.Name(), err)
 	}
+
```
twz123, line 361: Stray whitespace


```suggestion
```

pkg/helm/helm.go
twz123, line 415: We can drop this whole method here if we switch `GetReleaseStatus` to use Helm's Get action.

pkg/helm/helm.go
```diff
@@ -438,3 +461,27 @@ func (hc *Commands) UninstallRelease(ctx context.Context, releaseName string, na
 	}
 	return nil
 }
+
+// RollbackRelease rolls back a release to the previous deployed revision.
+// This is used to recover from interrupted upgrades where the release is stuck in
+// pending-upgrade or pending-rollback state. By rolling back, we preserve the
+// previously deployed workloads instead of uninstalling everything.
+// InstallChart, UpgradeChart, UninstallRelease and RollbackRelease are *NOT* thread-safe
+func (hc *Commands) RollbackRelease(ctx context.Context, releaseName string, namespace string) error {
```
twz123, line 470: I wonder what happens if this method itself gets interrupted? What state would the release be in then? Would we be able to recover? Would we try rolling back to the previous version of the previous version if we ended up here again?

pkg/component/controller/extensions_controller.go
```diff
@@ -456,6 +457,11 @@ func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv
 	if chart.Status.ReleaseName == "" {
 		// new chartRelease
 		cr.L.Tracef("Start update or install %s", chartName)
+		// Check for and clean up any broken release from previous interrupted install
+		if err = cr.cleanupBrokenRelease(ctx, helmCmd, chart.Spec.ReleaseName, chart.Spec.Namespace); err != nil {
+			err = fmt.Errorf("failed to cleanup broken release before install: %w", err)
+			return
+		}
```
twz123, line 464: Let's make this more idiomatic (including the other similar occurrences):
 
```suggestion
		if err := cr.cleanupBrokenRelease(ctx, helmCmd, chart.Spec.ReleaseName, chart.Spec.Namespace); err != nil {
			return fmt.Errorf("failed to cleanup broken release before install: %w", err)
		}
```

inttest/addons/addons_test.go
```diff
@@ -626,6 +628,89 @@ func (as *AddonsSuite) doTestAddonUpdate(addonName string, values map[string]any
...
+
+	as.Require().NoError(wait.PollUntilContextTimeout(ctx, 2*time.Second, 1*time.Minute, true, func(ctx context.Context) (bool, error) {
+		secretList, err := kc.CoreV1().Secrets(metav1.NamespaceDefault).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			as.T().Logf("Error listing secrets: %v", err)
+			return false, err
+		}
+
+		for _, secret := range secretList.Items {
+			if secret.Labels["name"] == restartAddonName && secret.Labels["owner"] == "helm" {
```
twz123, line 688: Can we add a check for the desired release status here (pendihg install)?

inttest/addons/addons_test.go
```diff
@@ -626,6 +628,89 @@ func (as *AddonsSuite) doTestAddonUpdate(addonName string, values map[string]any
 	as.PutFile(as.ControllerNode(0), path, buf.String())
 }
 
+// testControllerRestartRecovery tests that charts stuck in pending-install state
+// after controller restart are automatically recovered via the cleanup mechanism.
+// This reproduces issue https://github.com/k0sproject/k0s/issues/7109
+func (as *AddonsSuite) testControllerRestartRecovery(ctx context.Context, kc *k8s.Clientset) {
```
twz123, line 634: Might be appropriate to add more test cases, e.g. for recovering from a broken release upgrade? The RollbackRelease method feels like the most dangerous one to me.

inttest/addons/testdata/slow-hook-0.1.0.tgz
jnummelin: that's the packaged chart. I'll look into if we could package it on-the-fly without adding any new dependencies for the build

banjoh:
In **Helm 4**, some work was done to allow SDK users to pass to install/upgrade actions, fine grained go context objects that can be used to cancel a waiting operation. The effect of the cancellation would be the same as a timeout i.e release is would be left in `failed` status. https://github.com/helm/helm/pull/31435 is the PR which was contributed to support this

These follow up PRs are related to the same work 

https://github.com/helm/helm/pull/31735 - fine grained contexts to allow choosing what phase to cancel
https://github.com/helm/helm/pull/31730

For **Helm 3** which k0s is using today, adding this capability will be harder since [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md), which these cancellations are built on, is not supported.

jnummelin:
Closing as this does NOT really solve the problem properly. @twz123 is working on a better approach to stop Helm doing what ever it's doing when we stop k0s.

PR #7131, opened 2026-02-17
[renovate] Disable python datasource

juanluisvaladas:
## Description

We're hitting rate limits on python API and the python dependencies are handled by dependabot so we don't really need this.

This is happening in most runs: https://github.com/k0sproject/k0s/actions/runs/22090661444/job/63834839388#step:3:22077

This causes that some PRs marked to retry/rebase aren't done, such as https://github.com/k0sproject/k0s/pull/7117

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -182,6 +182,23 @@
       }
     },
     {
+      "description": "Disable Python managers",
+      "enabled": false,
+      "matchManagers": [
+        "pip_requirements",
+        "pip_setup",
+        "pipenv",
+        "poetry"
+      ]
+    },
+    {
+      "description": "Disable python datasource",
+      "enabled": false,
+      "matchDatasources": [
+        "python-version"
+      ]
+    },
+    {
       "description": "Only bump the main branch by default",
       "enabled": false,
       "prTitleStrict": false,


twz123:
I think there's a way to ignore the 429 responses for the data source instead via the host rules, so we don't need to disable it completely.

juanluisvaladas:
Is there a point on having this datasource though? We're using dependabot for this and it works well.

twz123:
> Is there a point on having this datasource though? We're using dependabot for this and it works well.

We are using this to update Python itself in `docs/Makefile.variables` (e.g. #7072)

juanluisvaladas:
That's a problem indeed. Closing it.

twz123:
I do wonder why we see the rate limits, tho. I'm seeing those even from my private IP (where I ran Renovate for the first time since quite a while). I wonder if the Python folks changed something and introduced a general ban based on user agents? :shrug:

twz123:
I think this is the only practical temporary fix. I've inspected the Renovate sources, and it seems that you can't bypass this via host rules. So we need to disable this until those 429 errors go away (unlikely), or upstream Renovate does something clever with the Python data source.

Other projects seem to experience comparable problems (see e.g. apache/airflow#61904). I can repro this even from my personal IP (although I'm pretty sure I didn't do anything with python.org in the last days).

twz123:
Oh, found the upstream issue: https://github.com/renovatebot/renovate/issues/27802#issuecomment-3902083358

twz123:
Alternatively, we could go for #7135

PR #7138, opened 2026-02-17
chore: adopt go fix suggestions

shinebayar-g:
## Description

Used `go fix ./...` command to adopt go modernizer fixes. All seems reasonable fixes.
- interface{} to any
- wg.Add(1) to wg.Go()
- ~~reflect.TypeOf() to reflect.TypeFor()~~
- and some more minor changes

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/internal/pkg/log/k0s.go
+++ b/internal/pkg/log/k0s.go
@@ -11,7 +11,7 @@ import (
 	crlog "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-type Backend interface{}
+type Backend any
 
 type ShutdownLoggingFunc func()
 
--- a/internal/sync/value/latest_test.go
+++ b/internal/sync/value/latest_test.go
@@ -21,12 +21,10 @@ func TestLatest(t *testing.T) {
 
 	var got int
 	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		<-expired
 		got, _ = underTest.Peek()
-	}()
+	})
 
 	time.Sleep(10 * time.Millisecond) // Simulate some delay
 	underTest.Set(42)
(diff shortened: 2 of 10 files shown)


internal/testutil/fakeclient/tracker.go
```diff
@@ -43,7 +43,7 @@ func NewClientset[T any](discovery *discoveryfake.FakeDiscovery, tracker testing
 	})
 
 	// Set the fake clientset's discovery and tracker.
-	ty := reflect.TypeOf(p).Elem()
+	ty := reflect.TypeFor[T]()
```
twz123: I think the previous version was more expressive. Yes, `p` is of type `*T`, but we're not operating on the type itself, but on the concrete pointer `p`. See the usages of `unsafe.Pointer(p)` in the loop below.

pkg/autopilot/signaling/v2/signaling_v2.go
```diff
@@ -87,7 +87,7 @@ func (s *SignalData) Unmarshal(m map[string]string) error {
 	Unmarshal(
 		m,
 		func() reflect.Type {
-			return reflect.TypeOf(*e)
+			return reflect.TypeFor[Signal]()
```
twz123: Same here, we're operating on a concrete variable, not on a type, so I think it's more expressive to use this:

```suggestion
			return reflect.TypeOf(e).Elem()
```

That said, I bet that probably the whole `Unmarshal` function would benefit from being generic, instead of using these clunky reflection callback functions 😅

shinebayar-g: Thanks reverted this change.

shinebayar-g: Created https://github.com/k0sproject/k0s/pull/7142 for follow up. Not sure if I'm in the right direction lol.

PR #7145, opened 2026-02-18
[Release 1.33] Bump alpine to 3.21.6

jnummelin:


juanluisvaladas:
The windows error is probably due to the caches, this happens in 1.32. I'll clear the windows cache.

PR #7173, opened 2026-02-24
docs: add version skew policy page

### LINKED ISSUE #6305: Document version skew policy
Different kube components have different supported version skews. It [seems](https://kubernetes.io/releases/version-skew-policy/) the control plane components mostly only allow one minor version difference (1.22.x => 1.23.x) both for running and when upgrading. Major upgrades are not supported at all (but there isn't a v2 anyway), but other upgrades should be done 1 minor at a time and preferably from latest patch to latest patch.

Kubelet seems to allow it to be 3 minor versions behind (1.22.x should still work with a 1.25.x control plane) when running but doesn't seem to directly mention if it's ok to upgrade in steps of 3 minors. Kubelet can never be newer than kube-api.

I think k0s wants to just follow these and not offer any "extended version skew support". Perhaps even only officially support 1 minor for everything.

It's only shortly mentioned in [autopilot docs](https://docs.k0sproject.io/head/autopilot/#q-i-upgraded-my-workers-and-now-kubelets-are-no-longer-reporting).

Related: https://github.com/k0sproject/k0sctl/pull/923

turanalmammadov:
Fixes #6305

## What this adds

Creates `docs/version-skew-policy.md` with a dedicated page covering:

1. **Kubernetes version skew rules** inherited by k0s — the table of what is allowed between kube-apiserver, kube-controller-manager, kubelet, etc.
2. **k0s official policy** — k0s supports 1 minor version of skew for all components (stricter than Kubernetes upstream which allows 3 for kubelet).
3. **Upgrade path** — one minor version at a time, using the latest patch release.
4. **Autopilot behavior** — how Autopilot enforces these rules.
5. **Commands** to check current versions of all components.
6. **Summary table** mapping common scenarios to supported/not-supported.

## Why

This information was only briefly mentioned in `docs/autopilot.md`. Operators frequently ask about version compatibility (see related k0sctl PR referenced in the issue) and a dedicated page makes it easy to find.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/version-skew-policy.md
+++ b/docs/version-skew-policy.md
@@ -1,0 +1,110 @@
+# Version Skew Policy
+
+This document describes the supported version skew between k0s components
+and between k0s and the embedded Kubernetes components.
+
+k0s follows the [Kubernetes version skew policy][k8s-version-skew] and adds
+a few k0s-specific notes on top.
+
+## Kubernetes Version Skew
+
+k0s embeds specific Kubernetes versions and inherits their skew rules.
+
+### Control Plane Components
+
+The core Kubernetes control plane components (`kube-apiserver`,
+`kube-controller-manager`, `kube-scheduler`) must be at the **same minor
+version** or at most **one minor version** apart:
+
+| Component | Supported skew relative to kube-apiserver |
+|-----------|------------------------------------------|
+| kube-controller-manager | ≤ 1 minor version older |
+| kube-scheduler | ≤ 1 minor version older |
+| kube-proxy | ≤ 1 minor version older |
+| kubelet | ≤ 3 minor versions older (never newer) |
+
+> **Important:** `kubelet` can be up to **3 minor versions behind**
+> `kube-apiserver` at runtime, but must **never be newer** than `kube-apiserver`.
+
+### Worker Node Kubelet
+
+While the Kubernetes project supports kubelet being up to 3 minor versions
+behind the control plane, k0s officially supports **1 minor version** of skew
+for all components to keep upgrades predictable.
+
+This means a k0s worker running v1.29.x is supported with a control plane
+running v1.30.x, but **not** with v1.31.x or newer.
+
+## k0s Component Version Skew
+
+When running a mixed-version cluster during a rolling upgrade:
+
+- The k0s **controller** version must be **≥** the k0s **worker** version.
+- You should upgrade controllers **before** workers.
+- Major version upgrades (e.g., `v1.x` → `v2.x`) are not supported. k0s
+  currently only has a `v1` API.
+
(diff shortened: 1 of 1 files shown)


docs/version-skew-policy.md
```diff
@@ -0,0 +1,110 @@
+# Version Skew Policy
+
+This document describes the supported version skew between k0s components
+and between k0s and the embedded Kubernetes components.
+
+k0s follows the [Kubernetes version skew policy][k8s-version-skew] and adds
```
juanluisvaladas, line 6: When we use k0s at the beginning of a sentence we write "K0s" (with a capital k)

docs/version-skew-policy.md
```diff
@@ -0,0 +1,110 @@
...
+a few k0s-specific notes on top.
+
+## Kubernetes Version Skew
+
+k0s embeds specific Kubernetes versions and inherits their skew rules.
+
+### Control Plane Components
```
juanluisvaladas, line 13: Same

docs/version-skew-policy.md
```diff
@@ -0,0 +1,110 @@
...
+
+The core Kubernetes control plane components (`kube-apiserver`,
```
juanluisvaladas, line 15: Since these are all part of k0s controller, perhaps this could be phrased in a way that makes it clear that the user only needs to worry about the k0s version on a per node basis.

docs/version-skew-policy.md
```diff
@@ -0,0 +1,110 @@
...
+`kube-controller-manager`, `kube-scheduler`) must be at the **same minor
+version** or at most **one minor version** apart:
```
juanluisvaladas, line 17: Perhaps clarify that this is only for upgrades.

docs/version-skew-policy.md
```diff
@@ -0,0 +1,110 @@
...
+
+| Component | Supported skew relative to kube-apiserver |
```
juanluisvaladas, line 19: Pretty sure the linter will fail with both tables, they should be formatted like this: https://github.com/k0sproject/k0s/blob/main/docs/cplb.md?plain=1#L637-L649

Anyway I'll run the CI so that we get the linter output.

juanluisvaladas:
One more thing, you need to sign off your commits, this is just a matter of running  `git commit --amend -s`

docs/version-skew-policy.md
juanluisvaladas: You can probably just feed the linter output to an LLM, they usually do a pretty good job at fixing those.

twz123:
Closing due to inactivity. If anyone wants to pick this up, feel free!

PR #7177, opened 2026-02-27
Make Helm chart reconciliation interruption-aware and recover interrupted installs

twz123:
## Description

K0s couldn't recover from interrupted Helm operations, e.g. due to controller restarts. Helm releases that were left in pending or failed states would not converge, and Helm action interruption behavior did not match k0s's requirements.

Refactor the helm package to support external interruption handling: Helm commands now use a Kubernetes client with interruption-aware wait logic and an HTTP transport that cancels in-flight requests and closes open connections and streams on interruption.

Chart installation behavior is now driven by the observed Helm release state:

* `pending-install`, `failed`, `uninstalling` -> Uninstall, then reinstall.
* `uninstalled` -> Reinstall with replace.
*  all other statuses -> Persist observed status on the Chart and return a reconciliation error.

Add an integration-test scenario in the addons inttest that verifies recovery of a chart stuck in pending-install.

Fixes:

* #7109

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/inttest/addons/addons_test.go
+++ b/inttest/addons/addons_test.go
@@ -4,17 +4,22 @@
 package addons
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"crypto"
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
 	"slices"
 	"strings"
+	"sync"
 	"testing"
 	"text/template"
 	"time"
@@ -36,9 +41,11 @@ import (
 	"github.com/stretchr/testify/require"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/wait"
 	k8s "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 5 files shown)


inttest/addons/addons_test.go
```diff
@@ -521,59 +533,43 @@ func (as *AddonsSuite) waitForTestRelease(addonName, appVersion string, namespac
 	ctx := as.Context()
 	as.T().Logf("waiting to see %s release ready in kube API, generation %d", addonName, rev)
 
-	cfg, err := as.GetKubeConfig(as.ControllerNode(0))
+	kc, err := as.AutopilotClient(as.ControllerNode(0))
```
jnummelin, line 536: `as.AutopilotClient(...` gives us a client that can do `kc.HelmV1beta1().Charts....`? Feels bit fishy

pkg/helm/client.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+
+		timer.Reset(2 * time.Second)
+		select {
+		case <-timer.C:
+		case <-ctx.Done():
+			return context.Cause(ctx)
+		}
+	}
+}
```
jnummelin, line 106: Not sure I read this right. So if the goroutine does NOT get all deleted, and ctx is NOT yet canceled, we wait 2 secs and spin the loop again, try to wait again for the deletion?

twz123: Yep. We're doing it in a goroutine so we can cancel early. The `v.Get()` is actually a call to the API server that should have a context parameter.

twz123: This is the cancellabe version of https://github.com/helm/helm/blob/v3.20.0/pkg/kube/wait.go#L112-L128

inttest/addons/addons_test.go
twz123: It's actually the k0s clientset which has all the clients, autopilot, Helm, etcd members. The method name comes from the time where the clientset only had the autopilot client.

pkg/helm/client.go
jnummelin: I'm failing to see how this is really different from the `wait.PollUntilContextCancel` given that now the `v.Get()` would get "properly" canceled since it's wired with our custom client now, right? Or am I looking this just wrong?

twz123: The original method will never return until it times out, because the context is derived from `ctx.Background()`. Not using wait.PollUntilContextCancel here because it won't return `context.Cause(ctx)` but `ctx.Err()`, hence swallowing error information.

twz123:
This is a bit more intertwined with #7083 than I originally anticipated. So I'm no longer planning to backport this beyond 1.35.

PR #7194, opened 2026-03-02
Fix calico-node restarts on BIRD mode

### LINKED ISSUE #7191: Calico pod restarts due to livenessProbe failures when using BIRD mode.
Created to track this issue from https://github.com/k0sproject/k0s/issues/6988

As referenced in: https://github.com/k0sproject/k0s/issues/6988#issuecomment-3974145203 when calico is using BIRD mode livenessProbes fail causing the pod to restart

juanluisvaladas:
## Description

Bumps calico to v3.31.4-1 which fixes #7191 and also adds tests that would have caught this issue.

Fixes #7191

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/inttest/common/util.go
+++ b/inttest/common/util.go
@@ -419,3 +419,49 @@ func (s *LineWriter) Flush() {
 		s.buf = s.buf[:0]
 	}
 }
+
+// VerifyNoRestartedPods returns a slice of errors with the restarted pods in kube-system.
+// In order to verify this, the test makes sure that the node has been ready for at least 80 seconds,
+// which is enough to validate that the calico (slowest component to restart due to a liveness probe error)
+// would have restarted.
+func VerifyNoRestartedPods(ctx context.Context, client *kubernetes.Clientset) []error {
+	var errs []error
+
+	// Check that every node has been ready for at least 80 seconds before we check if
+	// there are restarts. Otherwise we won't be able to catch failed livenessProbes.
+	nodes, err := client.CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return []error{err}
+	}
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady {
+				if condition.Status != corev1.ConditionTrue {
+					errs = append(errs, fmt.Errorf("node %q condition %s is %q", node.Name, corev1.NodeReady, condition.Status))
+					continue
+				}
+				// Wait 80 seconds after the node becomes ready so that we ensure that
+				// livenessProbes had enough time to fail and that the pods have restarted.
+				select {
+				case <-time.After(time.Until(condition.LastTransitionTime.Add(80 * time.Second))):
+				case <-ctx.Done():
+					errs = append(errs, ctx.Err())
+					return errs
+				}
+			}
+		}
+	}
+
+	pods, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(ctx, metav1.ListOptions{})
+	if err != nil {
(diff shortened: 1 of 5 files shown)


inttest/common/util.go
```diff
@@ -419,3 +419,40 @@ func (s *LineWriter) Flush() {
...
+			}
+		}
+	}
+	pods, err := client.CoreV1().Pods(metav1.NamespaceSystem).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		errs = append(errs, err)
+	}
+	for _, pod := range pods.Items {
+		for _, cs := range pod.Status.ContainerStatuses {
+			if cs.RestartCount != int32(0) {
```
twz123, line 461: I'm a bit reluctant to add a test for zero pod restarts. Isn't it enough if a Pod is ready?

inttest/common/util.go
```diff
@@ -419,3 +419,40 @@ func (s *LineWriter) Flush() {
 		s.buf = s.buf[:0]
 	}
 }
+
+// VerifyNoRestartedPods returns a slice of errors with the restarted on kube-system.VerifyNoRestartedPods
+// In order to. verify this, the test makes sure that the node has been ready for at least 80 seconds,
+// which is enough to validate that the calico (slowest component to restart due to a liveness probe error)
+// would have restarted.
```
twz123: Some full stops at places where they shouldn't be.

inttest/common/util.go
```diff
@@ -419,3 +419,40 @@ func (s *LineWriter) Flush() {
...
+
+	for _, node := range nodes.Items {
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == corev1.NodeReady {
+				if condition.Status == corev1.ConditionFalse {
+					errs = append(errs, fmt.Errorf("node %s is not ready", node.Name))
+					continue
+				}
+				// Wait at least 80 seconds so that the container mau have rebooted
+				<-time.After(time.Until(condition.LastTransitionTime.Add(80 * time.Second)))
```
twz123: Not sure what we're after, here?

inttest/kuberouter/kuberouter_hairpin_test.go
```diff
@@ -130,6 +135,7 @@ func TestKubeRouterHairpinSuite(t *testing.T) {
 		s.K0sExtraImageBundleMountPoints = []string{"/var/lib/k0s/images/ipv6.tar"}
 	}
 	suite.Run(t, &s)
+
```
twz123: Stray whitespace

inttest/common/util.go
juanluisvaladas: I don't think we can because calico-node readynessProbe works but livenessProbe fails.

We could limit the pods we search for using labels so that we only validate calico/kube-router pods but if we don't check restarts I can't think of a way to actually validate that this doesn't happen

inttest/common/util.go
juanluisvaladas: Fixed it thanks

inttest/common/util.go
juanluisvaladas: We want the node to be ready for at least 80 seconds before checking the pods for restarts, otherwise it's possible that the pods haven't had enough time to restart.

Calico needs roughly 70 seconds to fail, but if we test it with `v3.31.4-0` with only 70 seconds the test doesn't fail consistently. At 75 seconds the test fails consistently so I rounded it up to 80 seconds for a bit more reliability.

I added a couple comments in the code hoping that the intent and reason for it is clear now.

Also this isn't respecting the context timeout, I modified it to fix that problem.

PR #7302, opened 2026-03-17
Pass CAP_NET_BIND_SERVICE to kube-apiserver

### LINKED ISSUE #7221: Pass CAP_NET_BIND_SERVICE to kube-apiserver

`kube-apiserver` component is unable to bind to ports <1024 (eg 443) on it's own without root permissions.

Detect when `spec.api.port` is <1024 and pass ambient CAP_NET_BIND_SERVICE capability to `kube-apiserver` allowing it to bind to the lower port number.

1. Run the API server as root allowing it to bind however this decreases security due to root permissions.
2. Add nftables/iptables redirect rule but if all traffic is going to talk on port 443 anyway, might as well let `kube-apiserver` bind to the port cutting out unnecessary NAPT.
3. During k0sctl install/upgrades before hooks, start a service that monitors the `kube-apiserver` file and applies the CAP_NET_BIND_SERVICE permission.

We're using option 3 via systemd transient units which looks something like this:

```yaml
hooks:
      install:
        before: 
          - >
            sudo systemd-run
            --unit=k0s-apiserver-setcap
            --path-property=PathModified=/var/lib/k0s/bin/kube-apiserver
            -- /usr/sbin/setcap 'cap_net_bind_service=+ep' /var/lib/k0s/bin/kube-apiserver
```

 It works but going to have to test the setup long term to determine how reliable it is.

### Additional context

Spoke a bit to @twz123 in the Slack channel on this, but basically we're trying to bind to port 443 as an "ease of access". This is to avoid needing to adjust the upstream firewall and prevent users from having to specify the port, mainly the former tho. The cluster's also running CPLB so there's no external LB to handle ports redirects.

vladiskuz:
## Description

When the Kubernetes API server is configured to listen on a privileged port (< 1024), k0s now automatically grants the CAP_NET_BIND_SERVICE Linux capabilibty to the kube-apiserver process. This allows the non-root process to bind to ports like 443 without requiring full root privileges.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

Closes #7221


pkg/supervisor/detachattr_unix.go
```diff
@@ -22,8 +24,9 @@ func DetachAttr(uid, gid int) *syscall.SysProcAttr {
 	}
 
 	return &syscall.SysProcAttr{
-		Setpgid:    true,
-		Pgid:       0,
-		Credential: creds,
+		Setpgid:     true,
+		Pgid:        0,
+		Credential:  creds,
+		AmbientCaps: ambientCaps,
```
twz123: This is a Linux-only API, hence I think we need to split this file into `detachattr_linux.go` and `detachattr_unix.go` with a go:build directive of `unix && !linux`.

pkg/constant/constant.go
```diff
@@ -62,6 +62,12 @@ const (
 	// KeepalivedUser defines the user to use for running keepalived
 	KeepalivedUser = "keepalived"
 
+	/* Linux Capabilities */
+
+	// CapNetBindService is the Linux capability to bind to privileged ports (< 1024)
+	// See: https://man7.org/linux/man-pages/man7/capabilities.7.html
+	CapNetBindService = 10
```
twz123: There's https://pkg.go.dev/golang.org/x/sys/unix#CAP_NET_BIND_SERVICE, too.

pkg/supervisor/detachattr_unix.go
```diff
@@ -10,8 +10,10 @@ import (
 	"syscall"
 )
 
-// DetachAttr creates the proper syscall attributes to run the managed processes
-func DetachAttr(uid, gid int) *syscall.SysProcAttr {
+// DetachAttr creates the proper syscall attributes to run the managed processes.
+// If ambientCaps is provided and not empty, the process will be granted those
+// ambient capabilities.
+func DetachAttr(uid, gid int, ambientCaps ...uintptr) *syscall.SysProcAttr {
```
twz123: Passing around uintptrs feels wrong and insecure (even if in this case it's just some flags) 🙈

pkg/supervisor/supervisor.go
```diff
@@ -42,6 +42,8 @@ type Supervisor struct {
 	KeepEnvPrefix bool
 	// A function to clean some leftovers before starting or restarting the supervised process
 	CleanBeforeFn func() error
+	// Ambient capabilities to pass to the process
+	AmbientCaps []uintptr
```
twz123: Given that we need to be multi-platform here, and ambient capabilities are Linux only, what about abstracting this into sth. like a `RequiredPrivileges` struct à la 

```go
type RequiredPrivileges struct {
	BindsPrivilegedPorts bool
}
```

...that we can pass around? (As always, naming is up for debate 😅 ) That encodes the intent, which can then be translated into platform specifics (i.e. adding ambient capabilities on Linux).

pkg/supervisor/detachattr_unix.go
vladiskuz: Yes, good point. I fixed it

pkg/constant/constant.go
vladiskuz: Good catch, changed to https://pkg.go.dev/golang.org/x/sys/unix#CAP_NET_BIND_SERVICE

pkg/supervisor/supervisor.go
vladiskuz: Added logic with RequiredPrivileges to use the capability in only with Linux

twz123:
Looks good so far! The linter has some nits, tho ...

vladiskuz:
@twz123t looks like all the issues are fixed. But I don't see the new integration test running. What did I do wrong?
upd: looks like I need to add it to Makefile.variables

inttest/privileged-port/privileged_port_test.go
```diff
@@ -0,0 +1,132 @@
...
+	config := s.getControllerConfig()
+	s.PutFile(s.ControllerNode(0), "/tmp/k0s.yaml", config)
+
+	s.Require().NoError(s.InitController(0, "--config=/tmp/k0s.yaml"))
+
+	kc, err := s.KubeClient(s.ControllerNode(0))
+	s.Require().NoError(err)
+
+	s.AssertSomeKubeSystemPods(kc)
+
```
twz123, line 74: As you're not running any worker nodes, this will always fail.

```suggestion
```

twz123:
@vladiskuz Are you still interested in bringing this over the finish line? I think the test failure fix is the last missing bit.

twz123:
Merged as #8045. Thanks!

PR #7011, opened 2026-01-22
Bump containerd to v2.2.1

morlay:
## Description

Fixes https://github.com/k0sproject/k0s/issues/4581

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

- [x] My code follows the style [guidelines](https://docs.k0sproject.io/head/contributors/) of this project
- [x] My commit messages are [signed-off](https://docs.k0sproject.io/head/contributors/github_workflow/)
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #7376, opened 2026-03-31
Require v3 configuration on containerd imports

### LINKED ISSUE #7377: Handle containerd config with version = 2
Containerd won't crash with a config file that has `version = 2`.

This is dangerous because it's possible that a user has a configuration that is invalid and simply is ignored. I think the safest option for k0s is to exit exit during bootstrap instead of attempting to handle the situation.

juanluisvaladas:
## Description

Enforces containerd v3 configuration expectations when processing imported containerd config fragments, and updates tests/integration coverage to validate failures on v2/invalid configs. Also simplifies detection of k0s managed containerd configuration because there is no need for us to track older versions than 1.35.

Fixes #7377

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/component/worker/containerd/component.go
+++ b/pkg/component/worker/containerd/component.go
@@ -5,13 +5,9 @@ package containerd
 
 import (
 	"bufio"
-	"bytes"
 	"context"
-	"crypto/md5"
-	"encoding/hex"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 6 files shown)


pkg/component/worker/containerd/component.go
```diff
@@ -307,68 +303,24 @@ func (c *Component) Stop() error {
...
-		return false, err
-	}
-
-	if bytes.Equal(pre1_27ConfigSumBytes, sum) {
+	if scanner.Scan() && scanner.Text() == "# k0s_managed=true" {
 		return true, nil
 	}
-
-	return false, nil
+	return false, scanner.Err()
```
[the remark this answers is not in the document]
juanluisvaladas: This isn't an issue as 1.35 rewrites the file during the bootstrap if it's k0s managed, so we can safely ignore it.

pkg/component/worker/containerd/configurer.go
```diff
@@ -116,11 +133,30 @@ func generateDefaultCRIConfig(sandboxContainerImage string) ([]byte, error) {
...
+// hasV2CRIConfig checks if the given containerd configuration tree contains any plugins."io.containerd.grpc.v1.cri"
+// configuration section. This is the main difference between containerd v2 and v3 configuration formats, so if this is present we
+// want the user to fix it manually instead of trying to merge it and potentially causing unexpected breakage.
+func hasV2CRIConfig(tree *toml.Tree) bool {
+	return tree.HasPath([]string{"plugins", "io.containerd.grpc.v1.cri"})
+}
+
+func assertV3Config(tree *toml.Tree) error {
+	version := tree.Get("version")
+	if version == nil {
```
juanluisvaladas: I'm not too sure if it actually makes sense to fail if version is undefined. According to our docs, users should always set the version field. However containerd accepts it.

Maybe it's just saner just return nil if version is not defined.

juanluisvaladas: I decided to remove this because even k0s add a snippet without the version field explicitly defined: https://github.com/k0sproject/k0s/pull/7011/changes/a2ab5916b36a9ca9a3bcff1ed1006f460a27d024#diff-5ce55b33fdea68d04798d1490596b0293a68fc70375cbf254de6f3880dcbf95eR163-R168

ncopa (approved):

Issue #1919, opened 2022-07-14
RISC-V support


jekader:
### Is your feature request related to a problem? Please describe.

The RISC-V ISA is getting more traction as real hardware is starting to appear. This includes smaller SBC's which don't have that many resources (i.e. 1 CPU core and 1 GB of RAM) so k8s would be a great fit to try out kubernetes on this hardware.

### Describe the solution you would like

k0s available not only for amd64 and arm64 but for riscv64 as well

### Describe alternatives you've considered

Currently, I do not see any kubernetes distros supporting RISC-V. Podman/containerd are available on both Ubuntu and Debian and work just fine - I was able to pull containers from Docker Hub and run them.

### Additional context

Software supporting this ISA:
Debian: http://ftp.ports.debian.org/debian-ports/pool-riscv64/main/
Ubuntu: http://ports.ubuntu.com/ubuntu-ports/pool/
Alpine: https://dl-cdn.alpinelinux.org/alpine/edge/main/riscv64/
Images on Docker Hub: https://hub.docker.com/u/riscv64

trawler:
@jekader thank you for this issue. I've added it to our backlog candidates.

Additional interesting resources:
* https://github.com/carlosedp/riscv-bringup/
* https://carlosedp.medium.com/docker-containers-on-risc-v-architecture-5bc45725624b
* https://riscv.org/wp-content/uploads/2019/12/12.10-14.50c-The-RISC-V-Journey-Through-Containers-to-the-Cloud.pdf

ncopa:
Does the other upstream components support riscv? Like [etcd](https://etcd.io/), [kine](https://github.com/k3s-io/kine), [kube-router](https://github.com/cloudnativelabs/kube-router), [calico](https://github.com/projectcalico/calico)?

Does github actions support riscv64 or do we need to cross compile? If we need to cross compile, how do we run unit and integration tests?

twz123:
> Is there precedent for having k0s support an arch for a worker, but not a controller?

This is the case for Windows. For Windows, k0s only ships kubelet and kube-proxy, relying on an external CRI and some manual shenanigans for the Calico setup. Although there's currently some ongoing work to bundle containerd also for Windows and to streamline the Calico support.

What's the status of containerd for RISC-V? I think that'd be very important for a RISC-V based k0s worker.

iggy:
Updated the table above. Both containerd and runc have supported riscv64 for multiple releases.

jnummelin:
[...]

PR #7405, opened 2026-04-08
Add Traefik as an alternative NLLB backend

twz123:
## Description

Since envoy is not supported on all platforms supported by k0s, a more portable alternative is needed. Traefik supports all these platforms and can replace Envoy for NLLB functionality.

Duplicate all of the envoy specific code and modify it to manage Traefik. Add additional variants for Traefik to unit tests, whenever there was a mention of Envoy. Add another parameter to the NLLB integration test and run it once for envoy and once for Traefik. Enable NLLB using Traefik in the Windows WSL integration test, just to prove that it works on Windows.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

## Checklist

- [x] My code follows the style [guidelines](https://docs.k0sproject.io/head/contributors/) of this project
- [x] My commit messages are [signed-off](https://docs.k0sproject.io/head/contributors/github_workflow/)
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #7414, opened 2026-04-09
ci: Add support for linux-riscv64

luhenry:
## Description

_This is currently an experiment. Happy to get reviews as always._

Add GitHub Actions CI on `linux-riscv64` using [RISE RISC-V Runners](https://riseproject-dev.github.io/riscv-runner/).

Relates to https://github.com/k0sproject/k0s/issues/1919
Depends on https://github.com/k0sproject/image-builder/pull/253 https://github.com/k0sproject/k0s/pull/7459

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/.github/workflows/build-image-bundle.yml
+++ b/.github/workflows/build-image-bundle.yml
@@ -14,6 +14,10 @@ on:
         type: string
         required: true
         description: The platform suffix of the image bundle to be built.
+      runs-on:
+        type: string
+        required: false
+        description: The platform to build the image bundle on.
     outputs:
       cache-key:
         description: The image bundle's cache key.
@@ -25,7 +29,7 @@ env:
 jobs:
   build:
     name: "${{ inputs.image-bundle-name }}-${{ inputs.image-bundle-platform }}"
-    runs-on: ubuntu-24.04
+    runs-on: ${{ inputs.runs-on || 'ubuntu-24.04' }}
 
     env:
       IMAGE_BUNDLE_NAME: ${{ inputs.image-bundle-name }}
@@ -40,7 +44,7 @@ jobs:
       - name: "Download :: k0s"
         uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
         with:
-          name: k0s-linux-amd64
+          name: ${{ inputs.runs-on == 'ubuntu-24.04-riscv' && 'k0s-linux-riscv64' || 'k0s-linux-amd64' }}
 
       - name: "Download :: Image list"
         uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
--- a/.github/workflows/build-k0s.yml
+++ b/.github/workflows/build-k0s.yml
@@ -13,7 +13,7 @@ on:
         description: The architecture to build k0s for.
 
 env:
-  MAKEFLAGS: -j
+  MAKEFLAGS: -j --output-sync=line
 
 jobs:
   build:
@@ -22,6 +22,7 @@ jobs:
       ${{
            inputs.target-arch == 'arm'   && fromJSON('["self-hosted", "linux", "arm"]')
         || inputs.target-arch == 'arm64' && 'ubuntu-24.04-arm'
+        || inputs.target-arch == 'riscv64' && 'ubuntu-24.04-riscv'
         || 'ubuntu-24.04'
       }}
 
@@ -103,9 +104,11 @@ jobs:
             k0s${{ steps.build-prepare.outputs.executable-suffix }}
 
       - name: "Build :: SBOM"
+        if: inputs.target-arch != 'riscv64' # FIXME: https://github.com/anchore/syft/pull/4757, it doesn't support linux-riscv64 yet
         run: make spdx.json
 
       - name: "Upload :: SBOM"
+        if: inputs.target-arch != 'riscv64' # FIXME: https://github.com/anchore/syft/pull/4757, it doesn't support linux-riscv64 yet
         uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
         with:
           name: sbom-${{ inputs.target-os }}-${{ inputs.target-arch }}
--- a/.github/workflows/build-k0s.yml
+++ b/.github/workflows/build-k0s.yml
@@ -23,6 +23,7 @@ jobs:
         case(
           inputs.target-arch == 'arm', fromJSON('["self-hosted", "linux", "arm"]'),
           inputs.target-arch == 'arm64', 'ubuntu-24.04-arm',
+          inputs.target-arch == 'riscv64', 'ubuntu-24.04-riscv',
           'ubuntu-24.04'
         )
       }}
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -100,6 +100,8 @@ jobs:
             target-arch: amd64
           - target-os: linux
             target-arch: arm64
+          - target-os: linux
+            target-arch: riscv64
           - target-os: windows
             target-arch: amd64
 
@@ -113,7 +115,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target-arch: [amd64, arm64]
+        target-arch: [amd64, arm64, riscv64]
 
     name: "Build :: Airgap image bundle"
     needs: [build-k0s]
@@ -161,6 +163,8 @@ jobs:
         include:
           - name: linux-amd64
             runs-on: ubuntu-24.04
+          - name: linux-riscv64
+            runs-on: ubuntu-24.04-riscv
           - name: windows-amd64
             runs-on: windows-2022
             target-os: windows
@@ -267,6 +271,24 @@ jobs:
     uses: ./.github/workflows/smoketest.yaml
     with:
       arch: arm64
+      name: ${{ matrix.smoke-suite }}
+
+  smoketests-linux-riscv64:
+    strategy:
+      fail-fast: false
+      matrix:
+        smoke-suite:
+          - airgap
+          - basic
+          - network-conformance-calico
+          - network-conformance-kuberouter
+
+    name: "check-${{ matrix.smoke-suite }} :: riscv64"
+    needs: [build-k0s, build-airgap-image-bundle]
+
+    uses: ./.github/workflows/smoketest.yaml
+    with:
+      arch: riscv64
       name: ${{ matrix.smoke-suite }}
 
   autopilot-tests:
--- a/.github/workflows/riscv64.yml
+++ b/.github/workflows/riscv64.yml
@@ -1,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2026 k0s authors
+
+name: "RISC-V build"
+
+on:
+  workflow_dispatch:
+    inputs:
+      smoketests:
+        description: Which smoketests to run.
+        type: string
+  schedule:
+    - cron: 45 2 * * *
+
+jobs:
+  build-k0s:
+    name: "Build :: k0s :: linux-riscv64"
+    uses: ./.github/workflows/build-k0s.yml
+    with:
+      target-os: linux
+      target-arch: riscv64
+
+  build-airgap-image-bundle:
+    name: "Build :: Airgap image bundle"
+    needs: [build-k0s]
+    uses: ./.github/workflows/build-image-bundle.yml
+    with:
+      image-bundle-name: airgap
+      image-bundle-platform: linux-riscv64
+      runs-on: ubuntu-24.04-riscv
+
+  unittests-k0s:
+    name: "check-unit :: k0s :: linux-riscv64"
+    uses: ./.github/workflows/unittests-k0s.yml
+    with:
+      runs-on: ubuntu-24.04-riscv
+
+  smoketests:
+    strategy:
+      fail-fast: false
+      matrix:
+        smoke-suite: >-
+          ${{ fromJSON(inputs.smoketests || '
+              [
+                "airgap",
+                "basic"
+              ]
+            ')
+          }}
+    name: "check-${{ matrix.smoke-suite }} :: riscv64"
+    needs: [build-k0s, build-airgap-image-bundle]
+
+    uses: ./.github/workflows/smoketest.yaml
+    with:
+      arch: riscv64
+      name: ${{ matrix.smoke-suite }}
--- a/.github/workflows/smoketest.yaml
+++ b/.github/workflows/smoketest.yaml
@@ -26,7 +26,7 @@ permissions:
 jobs:
   smoketest:
     name: Test
-    runs-on: ${{ inputs.arch == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    runs-on: ${{ (inputs.arch == 'arm64' && 'ubuntu-24.04-arm') || (inputs.arch == 'riscv64' && 'ubuntu-24.04-riscv') || 'ubuntu-24.04' }}
     steps:
       - name: Check out code into the Go module directory
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
--- a/.github/workflows/smoketest.yaml
+++ b/.github/workflows/smoketest.yaml
@@ -30,6 +30,7 @@ jobs:
       ${{
         case(
           inputs.arch == 'arm64', 'ubuntu-24.04-arm',
+          inputs.arch == 'riscv64', 'ubuntu-24.04-riscv',
           'ubuntu-24.04'
         )
       }}


luhenry:
Currently blocked on missing https://github.com/anchore/syft support for `linux-riscv64`. I'll submit a PR upstream and update here accordingly. EDIT: work in progress at https://github.com/anchore/syft/pull/4757, I disabled it here for now for `riscv64` specifically.

twz123:
Nice! IIUC we need to add the RISE app to our GitHub org, right? There's also https://www.riscvrunners.com/. Do you know the differences?

luhenry:
> Nice! IIUC we need to add the RISE app to our GitHub org, right? There's also https://www.riscvrunners.com/. Do you know the differences?

Yes, you’d need to add https://github.com/apps/rise-risc-v-runners (only require permissions on “self-hosted runners read+write”).

On https://www.riscvrunners.com/, the main difference is the RISE ones are free (and will stay free, [RISE](https://riseproject.dev/) is not a business) and you have access to the whole machine. I’m not intimately familiar with the base image they are using (which packages, what OS, etc) but I’ll look more into it.

luhenry:
Trying to add the proper image dependencies at https://github.com/k0sproject/image-builder/compare/main...luhenry:k0s-image-builder:main

twz123:
> Trying to add the proper image dependencies at [k0sproject/image-builder@main...luhenry:k0s-image-builder:main](https://github.com/k0sproject/image-builder/compare/main...luhenry:k0s-image-builder:main)

I need to follow up on k0sproject/image-builder#256. I don't think we can generally enable RISC-V builds for everything.

luhenry:
> > Trying to add the proper image dependencies at [k0sproject/image-builder@main...luhenry:k0s-image-builder:main](https://github.com/k0sproject/image-builder/compare/main...luhenry:k0s-image-builder:main)
> 
> I need to follow up on [k0sproject/image-builder#256](https://github.com/k0sproject/image-builder/pull/256). I don't think we can generally enable RISC-V builds for everything.

Sounds good. Do you know if there will be other images that will be a required dependency?

luhenry:
I'm brute-forcing my way through figuring out which images/versions are working: https://github.com/luhenry/k0s-image-builder/actions/runs/24236376271

Good news is quite a few are already working, so adding them is a simple "adding a new version" away. Once that workflow finished, I'll look through all the failures and categorize them, that should give us a good todo list, and list of things to exclude/ignore as well

twz123:
> Sounds good. Do you know if there will be other images that will be a required dependency?

No. In fact, Calico is not a requirement. By default, k0s will use kube-router, which is already [available for RISC-V](https://quay.io/repository/k0sproject/kube-router?tab=tags). You won't be able to use NLLB (another optional feature), as this is would require envoy, and we didn't have the time to try to make it compile for RISC-V. But I'm currently looking into adding support for Traefik (#7405), mainly for Windows, but Traefik supports RISC-V and ARMv7 out of the box, so this might be a viable alternative to make NLLB usable on RISC-V, as well.

luhenry:
Recent development: BIRD has just added (like 3 days ago) builds for riscv64: https://gitlab.nic.cz/labs/bird/-/commit/0aca5b325fb6606d620ef349d70892105e1242c3. I ran into that issue in the build for `calico-node` at least.

> In fact, Calico is not a requirement

Good, then at least it's not a blocker. It should get resolved soon though. I never like diverging from how it's done on `linux-amd64` and `linux-arm64` because I like things to "just work". But gotta be pragmatic here 😅

luhenry:
Also, another finding: `etcd` is broken on all platforms, since [bitnami/etcd](https://hub.docker.com/r/bitnami/etcd) seems to have no more tags (?). That seems broken not just for `linux-riscv64` but for all other platforms as well.

twz123:
> Also, another finding: `etcd` is broken on all platforms, since [bitnami/etcd](https://hub.docker.com/r/bitnami/etcd) seems to have no more tags (?). That seems broken not just for `linux-riscv64` but for all other platforms as well.

Not all the images in the image-builder repo are used by k0s itself. There's also some utility things, like the GH Actions runners which we use to build k0s for ARMv7. Not sure if the etcd stuff is used anywhere, still.

Pretty sure that envoy is the last image that's directly used by k0s which is not available for RISC-V.

twz123:
> Recent development: BIRD has just added (like 3 days ago) builds for riscv64: https://gitlab.nic.cz/labs/bird/-/commit/0aca5b325fb6606d620ef349d70892105e1242c3. I ran into that issue in the build for calico-node at least.

We started to build BIRD from source in the Calico Dockerfiles recently, so this should be fine.

luhenry:
Triggering with `v3.31.4-2` from https://github.com/k0sproject/image-builder/pull/256, running at https://github.com/luhenry/k0s/actions/runs/24244825789

luhenry:
And for the missing `quay.io/k0sproject/pushgateway-ttl:1.4.0-k0s.0` image: https://github.com/k0sproject/pushgateway-ttl-builder/pull/5

pkg/constant/constant.go
```diff
@@ -71,8 +71,8 @@ const (
 
 	KonnectivityImage                     = "quay.io/k0sproject/apiserver-network-proxy-agent"
 	KonnectivityImageVersion              = "v0.34.0-1"
-	PushGatewayImage                      = "quay.io/k0sproject/pushgateway-ttl"
-	PushGatewayImageVersion               = "1.4.0-k0s.0"
+	PushGatewayImage                      = "ttl.sh/luhenry/pushgateway-ttl" //FIXME
+	PushGatewayImageVersion               = "1.4.0-k0s.1"
```
twz123: I'm a bit reluctant to approve CI runs that use ttl.sh images without a content hash 🙈

I can't find the CI run that built this image either.

luhenry: Absolutely! That's WIP 😅 . As soon as it's a bit cleaner and I've figured out all the things we need to get to green, I'll summarize it here, submit relevant PRs, link them back here, and we can run the whole CI only then.

In the meantime, the CI is running at https://github.com/luhenry/k0s/actions/workflows/go.yml?query=branch%3Amain and it's working well enough.

luhenry:
Following https://github.com/k0sproject/pushgateway-ttl-builder/pull/5#issuecomment-4225052426, I've cherry-picked [your commit](https://github.com/k0sproject/k0s/pull/7414/changes/4ea78ce3853f234f170438c9e26958de6ed58099) to mark Envoy as not-supported.

pkg/constant/constant.go
luhenry: Cherry-picking your commit has fixed the `linux-riscv64` airgap step!

luhenry:
https://github.com/replicatedhq/troubleshoot is missing on riscv64 now for:
- [check-basic :: riscv64 / Test](https://github.com/luhenry/k0s/actions/runs/24253466129/job/70822232168#logs)
- [check-network-conformance-calico :: riscv64 / Test](https://github.com/luhenry/k0s/actions/runs/24253466129/job/70822232081#logs)
- [check-network-conformance-kuberouter :: riscv64 / Test](https://github.com/luhenry/k0s/actions/runs/24253466129/job/70822232111#logs)
- [check-airgap :: riscv64 / Test](https://github.com/luhenry/k0s/actions/runs/24253466129/job/70822232105#logs)

The simplest solution is to simply _not_ run the `smoketests-linux-riscv64` for now. I'll submit a PR upstream to try to get them to enable and release `linux-riscv64` as well.

I've also verified that https://github.com/k0sproject/image-builder/pull/253 is working: I [rebased it](https://github.com/k0sproject/image-builder/compare/main...luhenry:k0s-image-builder:main) and used the [resulting image](ttl.sh/luhenry-pushgateway-ttl-24254219886-1:1d) at https://github.com/luhenry/k0s/actions/runs/24254357499/job/70824468191. The [Build :: Airgap image bundle (riscv64) / linux-riscv64](https://github.com/luhenry/k0s/actions/runs/24254357499/job/70824468191#logs) is passing.

luhenry:
Opened a PR on https://github.com/replicatedhq/troubleshoot/pull/2010 for `linux-riscv64`

twz123:
I've added RISC-V support to troubleshoot back in the day, exactly for that reason. But it got removed again afterwards, it seems. Too bad.

twz123:
Note that we can't merge this until we get the permit from CNCF to add the RISE RISC-V runners to the k0sproject org. 🤞 that we can sort this out quickly. In the meantime, maybe you want to go ahead and add RISC-V builds to k0sctl, as well? 🙃

luhenry:
> In the meantime, maybe you want to go ahead and add RISC-V builds to k0sctl, as well? 🙃

https://github.com/k0sproject/k0sctl/pull/1067. Not sure how to test it though.

twz123:
NB: you can drop dba5c64a49ba5b48a3d1fca416bd7984fc1ec48e in favor of #7459.

luhenry:
A full CI workflow has passed for the latest commit: https://github.com/luhenry/k0s/actions/runs/24552328790

> NB: you can drop [dba5c64](https://github.com/k0sproject/k0s/commit/dba5c64a49ba5b48a3d1fca416bd7984fc1ec48e) in favor of #7459.

I'll rebase once that PR is merged

luhenry:
troubleshoot has just released [v0.127.0](https://github.com/replicatedhq/troubleshoot/releases/tag/v0.127.0) with riscv64 support re-enabled :tada:

twz123:
> A full CI workflow has passed for the latest commit: https://github.com/luhenry/k0s/actions/runs/24552328790
>

... hot stuff 🔥

luhenry:
@twz123 from talking with @caniszczyk yesterday, we should be good to enable the [RISE RISC-V Runners](https://github.com/apps/rise-risc-v-runners) on `k0sproject/k0s`. For next steps, we can either 1. enable it right away if you want and we iterate on the different needs that you have, or 2. figure out your requirements ahead of time and I can implement them in next days/weeks. Let me know what you would prefer.

luhenry:
[anchore/syft](https://github.com/anchore/syft) has [integrated support for linux/riscv64](https://github.com/anchore/syft/commit/05cc8ee5f43e4dff3f01cf24bc38c7af3d0fdfa5), so it should show up on their next release. I'll update this PR when available.

@twz123 anything else that's blocking this PR from being merged?

luhenry:
I'll explore enabling the tests at https://github.com/luhenry/k0s/pull/19

jnummelin:
> anything else that's blocking this PR from being merged?

I think the main blocker currently is the official agreement from CNCF side which they are looking into

luhenry:
From debugging the smoketests on the machines directly (added a ` || sleep 30d` after `make -C inttest`), the issue has to do with the current setup:
- each workflow runs inside a k8s pod, so inside a container with its own overlayfs
- the dockerd is launched as a background process _inside_ that pod (it's not shared with the host as that would lead to leaking information and data across pods and workflows)
- the docker container for worker0, worker1, and controller0 are then docker-in-container, and so `overlayfs` on top of `overlayfs`.
- (that's where it gets a bit confusing to me as I'm not entirely aware of how k0s and overlayfs works)
- something somewhere is trying to allocate an overlayfs filesystem on top of overlayfs, using the "host" (aka overlayfs) filesystem as an "upper" <- that's not possible

The error I get in `dmesg` is `[Tue Apr 28 23:22:42 2026] overlayfs: filesystem on '/var/lib/k0s/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/13/fs' not supported as upperdir`.

Some workarounds that Claude proposed me are:
- Maybe a newer version of the Kernel supports something that would allow us to circumvent that issue. Currently the underlying machines are stuck on kernel 5.9, RISE is working with a kernel contributor to upstream the necessary patches for the TH1520 so that we can go to mainline kernel. The timeline is multiple weeks at minimum
- Allocating a block device backed by a file and formatted as ext4 to mount /var/lib/k0s` onto it, and do that as part of the pod startup process. It's possible to do in RISE infra or to do as part of the k0s workflow.
- Modify something somewhere in k0s CI infra to not mount it as "upperdir", but I don't understand what that means.

The longer term fix will be to switch from pods to VMs, that's going to happen in a couple of months when we have RVA23 hardware available and hooked in a datacenter that we can use. I would make sure to prioritize k0s to use these machines.

luhenry:
I got some progress in enabling the test suite but hitting the next blocker.

On the overlay-on-overlay, I fixed it by using an empty volume for the pod at `/var/lib/k0s` (see [[1]](https://github.com/riseproject-dev/riscv-runner-app/blob/71ac66b8ac484c354ee302b1f0c16347c5af4ed1/container/k8s.py#L84-L87)). Because a k8s pod's volumne is a direct mount from the host's filesystem into the container, and not an overlayfs, the `/var/lib/k0s` folder inside the pod is a `ext4` mount [2].

The next issue I ran into was the DNS from kube-proxy inside the k0s cluster was getting confused because the CIDR was the same between the k8s's pod and the k0s running inside the pod. The solution was simply to use the hostNetwork for the pods [[3]](https://github.com/riseproject-dev/riscv-runner-app/blob/71ac66b8ac484c354ee302b1f0c16347c5af4ed1/container/k8s.py#L66). This actually makes more sense since the pod should _not_ have access to anything internal to the k8s cluster.

Finally, the issue I'm not running into has to do with `kube-router` trying to setup a firewall which depends on the `ip_set_hash_net` kernel module. However, that module is not available on the Scaleway EM-RV1 out-of-the-box (even as a module). So the solution would be to build the kernel module so that it can be loaded. I'm currently working on that.

As a side note, we are also working with a Kernel contributor to upstream any patches necessary to mainline Linux to support the TH1520 (the SoC used by the EM-RV1) so we can go to mainline kernel instead of an older kernel version.

Overall, it's making good progress and I'm hoping to get things working in next weeks. Good news is nearly nothing has to change on k0s itself, and all in the infrastructure.

[2]:
```
$> kubectl exec -it rise-riscv-runner-9572x75pv -- findmnt -R
TARGET SOURCE FSTYPE  OPTIONS
/      overlay
              overlay rw,relatime,lowerdir=/var/lib/containerd/io.containerd.sna
|-/var/lib/k0s
|      /dev/mmcblk0p3[/var/lib/kubelet/pods/a8bffb66-a287-4d9f-a4bf-7cb8c30db806/volumes/kubernetes.io~empty-dir/k0s]
|             ext4    rw,relatime,errors=remount-ro
```

twz123:
Good news! CNCF approved to add the RISE RISC-V runners to the k0sproject org :rocket:

@luhenry could you rework this PR in a way that it doesn't trigger for every PR? Instead, I think it's best to start with a nightly job that also has a workflow_dispatch, similar to what we have for the ostests.

If the integration tests aren't working yet, we could add this as an optional parameter for workflow_dispatch (again, in a way similar to what we're doing for the ostests, this time in ostests-matrix.yaml). This way, we can execute and iterate on the integration tests without needing to change the workflows.

luhenry:
I split the `unittests` in a separate workflow_call file, similar to `build-k0s` and `smoketests`. I added a `go-nightly.yml` which only runs the `build-k0s`, `unittests`, `build-airgap-image-bundle` and `smoketests-linux-riscv64` (disabled for now [[1]](https://github.com/k0sproject/k0s/pull/7414#issuecomment-4343769538)[[2]](https://github.com/k0sproject/k0s/pull/7414#issuecomment-4374112491)) for `linux-riscv64`.

I'm testing it at:
- `go.yml`: https://github.com/luhenry/k0s/actions/runs/25448582937
- `go-nightly.yml`:
  - w/o smoketests: https://github.com/luhenry/k0s/actions/runs/25454194134
  - w/ smoketests: https://github.com/luhenry/k0s/actions/runs/25454212706

luhenry:
For the latest on the issue with `kube-router`, I met with my contact at Scaleway today, and they are going to share the sources for the Linux Kernel they have on the Scaleway EM-RV1. I'll build the necessary modules from it and add them to all the machines of the pool. I'll keep you posted as soon as I have more things to show.

> @luhenry could you rework this PR in a way that it doesn't trigger for every PR? Instead, I think it's best to start with a nightly job that also has a workflow_dispatch, similar to what we have for the ostests.
>
> If the integration tests aren't working yet, we could add this as an optional parameter for workflow_dispatch (again, in a way similar to what we're doing for the ostests, this time in ostests-matrix.yaml). This way, we can execute and iterate on the integration tests without needing to change the workflows.

@twz123 that should be all done.

.github/workflows/build-k0s.yml
```diff
@@ -20,9 +20,12 @@ jobs:
     name: ${{ inputs.target-os }}-${{ inputs.target-arch }}
     runs-on: >-
       ${{
-           inputs.target-arch == 'arm'   && fromJSON('["self-hosted", "linux", "arm"]')
-        || inputs.target-arch == 'arm64' && 'ubuntu-24.04-arm'
-        || 'ubuntu-24.04'
+        case(
```
twz123, line 23: Cool! Is this case function something new?

.github/workflows/go-nightly.yml
```diff
@@ -0,0 +1,86 @@
...
+
+name: "Go build :: Nightly"
+
+on:
+  workflow_dispatch:
+    inputs:
+      run-riscv64-smoketests:
+        description: Whether to run the riscv64 smoketests; disabled by default
+        type: boolean
+        default: false
```
twz123: I was more thinking about a list of names, like this:

```suggestion
      smoketests:
        description: Which smoketests to run.
        type: string
        default: "[]"
```

This could be fed directly into the matrix build using `${{ fromJSON(inputs.smoketests) }}`. Whenever we want to run specific tests, we can set this e.g. to `[ "check-basic", "check-calico", "check-airgap" ]`.

.github/workflows/go-nightly.yml
```diff
@@ -0,0 +1,86 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2026 k0s authors
+
+# For Go build that you only want to run nightly
+
+name: "Go build :: Nightly"
```
twz123: Since the platform isn't parameterized, I'd simply call it "RISC-V build" or something like that.

.github/workflows/unittests-k0s.yml
```diff
@@ -0,0 +1,93 @@
...
+            touch -t "$(TZ=UTC+24 date +%Y%m%d%H%M.%S)" build/cache/_cache_sentinel
+            find build/cache/go/build -type f \( -name '*-a' -o -name '*-d' \) -exec touch -r build/cache/_cache_sentinel {} +
+          fi
+
+          make check-unit $UNITTEST_EXTRA_ARGS
+
+          if [ -n "$CACHE_HIT_GOCACHE" ]; then
+            echo Trimming Go cache
+            find build/cache/go/build -type f \( -name '*-a' -o -name '*-d' \) -not -newer build/cache/_cache_sentinel -delete
+          fi
```
twz123: Missing newline at the end of the file.

.github/workflows/unittests-k0s.yml
```diff
@@ -0,0 +1,94 @@
...
+    runs-on: >-
+      ${{
+        case(
+          inputs.target-os == 'windows' && inputs.target-arch == 'amd64', 'windows-2022',
+          inputs.target-os == 'darwin' && inputs.target-arch == 'arm64', 'macos-15',
+          inputs.target-os == 'linux' && inputs.target-arch == 'arm64', 'ubuntu-24.04-arm',
+          inputs.target-os == 'linux' && inputs.target-arch == 'riscv64', 'ubuntu-24.04-riscv',
+          'ubuntu-24.04'
+        )
+      }}
```
twz123: Did you pass in runs-on directly via an input parameter?

```suggestion
    runs-on: ${{ inputs.runs-on }}
```

This is probably more straight forward than the case expression. Not entirely sure why I did this in `build-k0s.yaml` back in the day :thinking:

Ah, probably because of `["self-hosted", "linux", "arm"]` being an array, and not a string. Since we don't need to support this here, I'd rather go without the case expression.

.github/workflows/go-nightly.yml
```diff
@@ -0,0 +1,86 @@
...
+        description: Whether to run the riscv64 smoketests; disabled by default
+        type: boolean
+        default: false
+  schedule:
+    - cron: 45 2 * * *
+
+env:
+  MAKEFLAGS: -j
+  CURL_OPTS: --proto =https --tlsv1.2 --retry 5 --retry-all-errors --silent --show-error --location --fail
+
```
twz123: Since this workflow doesn't run commands on its own, we don't need those here.

```suggestion
```

.github/workflows/go-nightly.yml
```diff
@@ -0,0 +1,86 @@
...
+    name: "Build :: Airgap image bundle"
+    needs: [build-k0s]
+    uses: ./.github/workflows/build-image-bundle.yml
+    with:
+      image-bundle-name: airgap
+      image-bundle-platform: linux-${{ matrix.target-arch }}
+
+  unittests-k0s:
+    strategy:
```
twz123: Do we need a matrix build here? I'd have guessed we could just call the unit tests once for RISC-V?

Same for the build-airgap-image-bundle job.

.github/workflows/go-nightly.yml
```diff
@@ -0,0 +1,86 @@
...
+
+jobs:
+  build-k0s:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # linux-amd64 is required for .github/workflows/build-image-bundle.yml
+          - target-os: linux
+            target-arch: amd64
```
twz123: Hmm. That's unfortunate. I wonder if it'd make sense to add some parameters to the build-image job, so it can work with the RISC-V binary, too :thinking:

The easiest would probably be to add some optional inputs, say `runs-on` (defaulting to ubuntu-slim) and `k0s-artifact-name` (defaulting to k0s-linux-amd64). We could then call the workflow from here using `ubuntu-24.04-riscv` and `k0s-linux-riscv64`?

twz123:
If you want, you can split out 6fd5939534796d49f576f6ab1fb204ded2b3c53e into a separate PR, so we can merge that already.

.github/workflows/build-k0s.yml
luhenry: I don't know when it was added but it's documented at https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case

.github/workflows/go-nightly.yml
luhenry: Happy to. Are there any other platforms you'd like to add to this nightly? For example linux-arm64?

.github/workflows/unittests-k0s.yml
luhenry: We will need the `target-os` parameter nonetheless as it's used elsewhere in that file.

.github/workflows/go-nightly.yml
luhenry: We don't, it's mostly for easily extending it in the future. Happy to go either way.

luhenry:
> If you want, you can split out [6fd5939](https://github.com/k0sproject/k0s/commit/6fd5939534796d49f576f6ab1fb204ded2b3c53e) into a separate PR, so we can merge that already.

https://github.com/k0sproject/k0s/pull/7591

I'll work on the rest of the feedback tomorrow

twz123:
> > If you want, you can split out [6fd5939](https://github.com/k0sproject/k0s/commit/6fd5939534796d49f576f6ab1fb204ded2b3c53e) into a separate PR, so we can merge that already.
> 
> https://github.com/k0sproject/k0s/pull/7591
> 
> I'll work on the rest of the feedback tomorrow
> 

I meant the split-off of the unit tests into their own callable workflow 😅

luhenry:
> > > If you want, you can split out [6fd5939](https://github.com/k0sproject/k0s/commit/6fd5939534796d49f576f6ab1fb204ded2b3c53e) into a separate PR, so we can merge that already.
> > 
> > 
> > #7591
> > I'll work on the rest of the feedback tomorrow
> 
> I meant the split-off of the unit tests into their own callable workflow 😅

Ah sorry, my bad, will do that first thing tomorrow!

luhenry:
@twz123 I split it out as you requested.

Also, I [got `check-basic` to work](https://github.com/luhenry/k0s/actions/runs/25637625662/job/75252983716
) 🎉  I updated the underlying workers with the right kernel modules and it works. I'll check other test suites now.

UPDATE: `check-basic` [1] and `check-airgap` [2] are working. The `check-network-conformance-calico` [3] and `check-network-conformance-kuberouter` [4] are not, but only because https://hub.docker.com/r/sonobuoy/sonobuoy is not available on riscv64. I'll try submitting a PR this week

UPDATE 2: The PR for sonobuoy: https://github.com/vmware-tanzu/sonobuoy/pull/2049

[1]:
https://github.com/luhenry/k0s/actions/runs/25638493227/job/75258697918
[2]:
https://github.com/luhenry/k0s/actions/runs/25638493227/job/75255285406
[3]:
https://github.com/luhenry/k0s/actions/runs/25638493227/job/75255285407
[4]:
https://github.com/luhenry/k0s/actions/runs/25638493227/job/75255285295

luhenry:
@twz123 anything else I should do to get this merged? Thanks 🙏

luhenry:
I verified a build works: https://github.com/luhenry/k0s/actions/runs/26634638919

luhenry:
Enabling the `basic` smoketests as I've verified it works at https://github.com/luhenry/k0s/actions/runs/26634649027

luhenry:
Also enabling `airgap` as it works: https://github.com/luhenry/k0s/actions/runs/26709975774/job/78719293615

.github/workflows/build-image-bundle.yml
twz123, line 1: The changes here are a bit hacky, but I don't have a better proposal right now that wouldn't require lots of work. So just let's live with it for now...

twz123 (approved):
Thanks for pushing this forward!

luhenry:
It's been succeeding for past few days:
- https://github.com/k0sproject/k0s/actions/runs/28001505151
- https://github.com/k0sproject/k0s/actions/runs/27928956848

PR #7438, opened 2026-04-13
feat: delegate konnectivity server counting to lease controller

### LINKED ISSUE #5736: Enable lease counting on konnectivity
https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/643

This means we can get rid of the lease counter thingy on k0s side and "just" set those flags on konnectivity.

So we can get rid of some code 🎉  Removing code makes me happy 😄

fooxlj07:
## Description

Delegates konnectivity server counting to the built-in lease controller instead of k0s tracking it manually.
  - konnectivity-server: add `--enable-lease-controller=true`, drop `--server-count`
  - konnectivity-agent: add `--count-server-leases`, drop the `K0S_CONTROLLER_COUNT` env var

### Changes:                                                                                          
  - The server now publishes its own coordination Lease and the agent counts them directly, so k0s no longer 
  needs to watch the controller count, restart the server, or re-render the agent DaemonSet on every HA
  event. New RBAC is added so both sides can manage leases in kube-system.                                   
 - Adds an integration test asserting the new flags/RBAC and that the agent DaemonSet's generation is
  unchanged when a controller leaves.
 
### Notes:                                                                                             
We do not pass `--lease-label` to either konnectivity-server or konnectivity-agent. Both sides rely on the upstream default `k8s-app=konnectivity-server` — see [server default](https://github.com/kubernetes-sigs/apiservernetwork-proxy/blob/ea0a0d314b4d7ae02383adb6f8b74cee5ae330d4/cmd/server/app/options/options.go#L381) and [agent default](https://github.com/kubernetes-sigs/apiserver-network-proxy/blob/ea0a0d314b4d7ae02383adb6f8b74cee5ae330d4/cmd/agent/app/options/options.go#L283) (added in[#670](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/670)). The server publishes leases with that label, the agent selects them by the same label, so they match by design. If either default ever changes upstream, `--lease-label` will need to be set explicitly on both sides.

(Note: [#643](https://github.com/kubernetes-sigs/apiserver-network-proxy/pull/643)'s description still refers to a `--server-lease-selector` flag, but that name only existed during PR development — the merged code uses `--count-server-leases` to enable the system and `--lease-label` to configure the selector.)

Co-authored with [Claude](https://claude.com/claude-code). 

Closes #5736 

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/konnectivityagent.go
+++ b/pkg/component/controller/konnectivityagent.go
@@ -23,7 +23,6 @@ import (
 type KonnectivityAgent struct {
 	ManifestsDir           string
 	KonnectivityServerHost string
-	ServerCount            func() (uint, <-chan struct{})
 
 	configChangeChan chan *v1beta1.ClusterConfig
 	log              *logrus.Entry
@@ -45,22 +44,12 @@ func (k *KonnectivityAgent) Init(_ context.Context) error {
 func (k *KonnectivityAgent) Start(ctx context.Context) error {
 
 	go func() {
-		serverCount, serverCountChanged := k.ServerCount()
-
 		var clusterConfig *v1beta1.ClusterConfig
 		var retry <-chan time.Time
 		for {
 			select {
 			case config := <-k.configChangeChan:
 				clusterConfig = config
-
-			case <-serverCountChanged:
-				prevServerCount := serverCount
-				serverCount, serverCountChanged = k.ServerCount()
-				// write only if the server count actually changed
-				if serverCount == prevServerCount {
-					continue
-				}
 
 			case <-retry:
 				k.log.Info("Retrying to write konnectivity agent manifest")
@@ -76,7 +65,7 @@ func (k *KonnectivityAgent) Start(ctx context.Context) error {
 				continue
 			}
 
-			if err := k.writeKonnectivityAgent(clusterConfig, serverCount); err != nil {
+			if err := k.writeKonnectivityAgent(clusterConfig); err != nil {
 				k.log.Errorf("failed to write konnectivity agent manifest: %v", err)
 				retry = time.After(10 * time.Second)
 				continue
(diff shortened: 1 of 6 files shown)


pkg/component/controller/konnectivityagent.go
```diff
@@ -205,6 +193,34 @@ subjects:
...
+    kubernetes.io/cluster-service: "true"
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: system:konnectivity-server
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: system:konnectivity-server
+---
```
jnummelin: Do we really need this role and binding? In case of k0s the server is running as plain process on the CP node and thus uses (IIRC) the main admin credentials anyways.

pkg/component/controller/konnectivityagent.go
```diff
@@ -213,6 +229,32 @@ metadata:
...
+  - apiGroups: ["coordination.k8s.io"]
+    resources: ["leases"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:konnectivity-agent
+  labels:
+    kubernetes.io/cluster-service: "true"
```
jnummelin: Why do we need this label?

pkg/component/controller/konnectivityagent.go
fooxlj07: Hi Jussi, thank you for taking the time to review the PR.

True, it is not needed, I've removed the Role and the RoleBinding, and also the ClusterRoleBinding for the server
I noticed this [certificate](https://github.com/k0sproject/k0s/blob/main/cmd/controller/certificates.go#L111)— as I understand it, this cert grants the konnectivity-server full permissions
  to make requests to the API server?

pkg/component/controller/konnectivityagent.go
fooxlj07: True, No need, removed them, 

I had copied the example config from this [PR](https://github.com/carreter/apiserver-network-proxy/blob/714d09261101800120373cb1a3bb10278f32220d/examples/kind-multinode/templates/k8s/konnectivity-agent-ds.yaml#L33) without verifying it.

PR #7561, opened 2026-04-30
Improve network conformance tests flakyness

juanluisvaladas:
## Description

We were disabling some tests because they were flaky. While fixing the kube-router IPv6 tests in b2de9fb we found out these were flaky because kube-router could become ready before konnectivity. The conformance test rely heavily on the api-server connecting to the kubelet and these simply did not work. Turns out as a side effect the tests we were skipping because they were flaky, aren't flaky anymore.

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/inttest/network-conformance/network_test.go
+++ b/inttest/network-conformance/network_test.go
@@ -4,6 +4,7 @@
 package network
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -12,7 +13,11 @@ import (
 	"time"
 
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+	"github.com/k0sproject/k0s/pkg/kubernetes/watch"
+	"k8s.io/client-go/kubernetes"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	sc "github.com/vmware-tanzu/sonobuoy/pkg/client"
@@ -86,6 +91,9 @@ func (s *networkSuite) TestK0sGetsUp() {
 	s.T().Log("waiting to see CNI pods ready for", daemonSetName)
 	s.NoErrorf(common.WaitForDaemonSet(s.Context(), kc, daemonSetName, metav1.NamespaceSystem), "%s did not start", daemonSetName)
 
+	s.T().Log("waiting for CoreDNS to be ready")
+	s.Require().NoError(waitForCoreDNS(s.Context(), kc), "While waiting for CoreDNS to become ready")
+
 	s.T().Log("waiting for konnectivity-agent to be ready")
 	s.NoErrorf(common.WaitForDaemonSet(s.Context(), kc, "konnectivity-agent", metav1.NamespaceSystem), "konnectivity-agent did not start")
 
@@ -96,6 +104,13 @@ func (s *networkSuite) TestK0sGetsUp() {
 	s.Require().NoError(err)
 	client, err := sc.NewSonobuoyClient(restConfig, skc)
 	s.Require().NoError(err)
+
+	e2eExtraArgs := ""
+
+	// kube-router's ipv6 tests are flaky, so we skip some
+	if s.cni == "kuberouter" && s.isIPv6Only {
+		e2eExtraArgs = "--ginkgo.flake-attempts=2"
+	}
 
 	deadline, _ := s.Context().Deadline()
 	err = client.Run(&sc.RunConfig{
@@ -114,6 +129,7 @@ func (s *networkSuite) TestK0sGetsUp() {
 					"E2E_SKIP":          "\\[Serial\\]|(Services\\ should.*session\\ affinity\\ .*for\\ service\\ with\\ type\\ clusterIP)",
 					"E2E_PARALLEL":      "y",
 					"E2E_USE_GO_RUNNER": "true",
+					"E2E_EXTRA_ARGS":    e2eExtraArgs,
 				},
 			},
 			KubeVersion: semver.Canonical(k8sVersion.String()),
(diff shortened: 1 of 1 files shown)


inttest/network-conformance/network_test.go
```diff
@@ -97,23 +100,26 @@ func (s *networkSuite) TestK0sGetsUp() {
 	client, err := sc.NewSonobuoyClient(restConfig, skc)
 	s.Require().NoError(err)
 
+	var e2eExtraArgs, e2eSkip string
+
+	// kube-router's ipv6 tests are flaky, so we skip some
+	if s.cni == "kuberouter" && s.isIPv6Only {
+		e2eSkip = "\\[Flaky\\]|\\[Serial\\]|\\[Disruptive\\]|\\[Feature:.+\\]"
```
twz123: Which tests are flaky that we want to skip? I think this is a bit too broad, isn't it?

inttest/common/util.go
```diff
@@ -189,7 +189,7 @@ func WaitForDeployment(ctx context.Context, kc kubernetes.Interface, name, names
 		Until(ctx, func(deployment *appsv1.Deployment) (bool, error) {
 			for _, c := range deployment.Status.Conditions {
 				if c.Type == appsv1.DeploymentAvailable {
-					if c.Status == corev1.ConditionTrue {
+					if c.Status == corev1.ConditionTrue && deployment.Status.ReadyReplicas == *deployment.Spec.Replicas {
```
twz123: I don't think we should do this here. This is explicitly testing the available status, which is based on the Deployment spec. If the cluster has some hickups because of this, we should probably fix the Deployment settings themselves?

Do you have a good theory which Deployment introduces the trouble and why?

inttest/network-conformance/network_test.go
```diff
@@ -104,14 +104,7 @@ func (s *networkSuite) TestK0sGetsUp() {
...
-					"E2E_FOCUS": "\\[sig-network\\].*\\[Conformance\\]",
-					//
-					// Skipping flaky tests:
-					// - [It] [sig-network] Services should be able to switch session affinity for service with type clusterIP [LinuxOnly] [Conformance]
-					// - [It] [sig-network] Services should have session affinity work for service with type clusterIP [LinuxOnly] [Conformance]
-					// - [It] [sig-network] Services should have session affinity timeout work for service with type clusterIP [LinuxOnly] [Conformance]
-					//
-					"E2E_SKIP":          "\\[Serial\\]|(Services\\ should.*session\\ affinity\\ .*for\\ service\\ with\\ type\\ clusterIP)",
+					"E2E_FOCUS":         "\\[sig-network\\].*\\[Conformance\\]",
 					"E2E_PARALLEL":      "y",
```
twz123, line 115: We can't remove the E2E_SKIP for `[Serial]` here if we still set E2E_PARALLEL to `y` I guess?

inttest/network-conformance/network_test.go
juanluisvaladas: I'm just preserving the old list. The list of most commonly failing tests I have are:

- [sig-network] DNS should resolve DNS of partial qualified names for services [LinuxOnly] [Conformance]
- [sig-network] DNS should provide DNS for pods for Subdomain [Conformance]
- [sig-network] DNS should provide DNS for services [Conformance]
- [sig-network] DNS should provide DNS for ExternalName services [Conformance]
- [sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]
- [sig-network] Services should serve endpoints on same port and different protocols [Conformance]

These aren't the ones we disable explicitly on this list, which is initially odd but these fail either because coredns fails or because konnectivity fails. These seem to fail because either coreDNS or konnectivity time out at some point and the timeouts seem to be triggered by the disruptive tests.

I'm timing how long it takes to run these tests serially (all of them instead of in parallel) and maybe it's better to just disable parallelism for these tests.

inttest/common/util.go
juanluisvaladas: Without this the deployment is ready as soon as there is one replica, the conformance tests make a lot of DNS queries and if there is only one replica there might be a few timeouts. I'm still trying to figure out why there are a few timeouts only on kube-router and only on IPv6.

We can make this a new function just for the conformance tests, but without it the tests become problematic.

inttest/network-conformance/network_test.go
juanluisvaladas: I don't know how it works internally but it doesn't seem to be a problem, I haven't seen it fail with calico or kuberouter ipv4 once.

inttest/common/util.go
juanluisvaladas: I moved it to a new function just for the conformance test

inttest/network-conformance/network_test.go
twz123: Maybe there aren't any sig-network conformance tests that need to be run in serial yet? Anyhow, I'd rather keep this comination in lockstep: ignore serial, enable parallel, or drop both. In fact, we could also try to run this with parallel tests disabled. Maybe that helps with the flakiness?

inttest/common/util.go
twz123: Okidoki. It would be just much more reassuring if we'd have a good understanding why one replica isn't enough. If this is simply a load problem, that's one thing, but maybe there's something else that we're overlooking.

inttest/network-conformance/network_test.go
twz123: WDYT @juanluisvaladas?

inttest/common/util.go
juanluisvaladas: I don't think it's a coredns problem but a kube-router problem... I just don't see why this happens...

juanluisvaladas:
I reduced the scope of the PR, basically added a check for coreDNS and maintained the kuberouter-ipv6 specific changes.

inttest/network-conformance/network_test.go
juanluisvaladas: I removed all these changes and left it as it was.

twz123 (approved):

Issue #3719, opened 2023-11-16
Autopilot update inttests succeed even if cluster is not operational


twz123:
### Before creating an issue, make sure you've checked the following:

- [X] You are running the latest released version of k0s
- [X] Make sure you've searched for existing issues, both open and closed
- [X] Make sure you've searched for PRs too, a fix might've been merged already
- [X] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
GitHub Actions
```

### Version

main

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Machine ID: "9ca147e6762965990e6ecf9753f1382c0f5d91df2d488a81ba51f938a4b64b25" (from machine) (pass)
Total memory: 15.6 GiB (pass)
Disk space available for /var/lib/k0s: 42.7 GiB (pass)
Name resolution: localhost: [::1 127.0.0.1] (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.2.0-1015-azure (pass)
  Max. file descriptors per process: current: 65536 / max: 65536 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
    cgroup controller "memory": available (is a listed root controller) (pass)
    cgroup controller "devices": unknown (warning: insufficient permissions, try with elevated permissions)
    cgroup controller "freezer": available (cgroup.freeze exists) (pass)
    cgroup controller "pids": available (is a listed root controller) (pass)
    cgroup controller "hugetlb": available (is a listed root controller) (pass)
    cgroup controller "blkio": available (via io in version 2) (pass)
  CONFIG_CGROUPS: Control Group support: built-in (pass)
    CONFIG_CGROUP_FREEZER: Freezer cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_PIDS: PIDs cgroup subsystem: built-in (pass)
    CONFIG_CGROUP_DEVICE: Device controller for cgroups: built-in (pass)
[...]

PR #7605, opened 2026-05-09
fix: thread StatusSocketPath through autopilot controllers

### LINKED ISSUE #6750: Non-default --status-socket or --run-dir break autopilot

### Platform

```shell

```

### Version

1.34.2

### Sysinfo

_No response_

### What happened?

Setting `--status-socket` to anything but the default value will break autopilot, which is hardcoded to connect to the default path.  Same happens if you set a custom run-dir as the default socket path is based on run dir, but autopilot has hardcoded it as `/run/k0s/status.sock`.

### Steps to reproduce

1. Start with `--status-socket=/tmp/foo`
2. Enable autopilot

### Expected behavior

Autopilot works.

### Actual behavior

Autopilot does not work.

### Screenshots and logs

_No response_

### Additional context

https://github.com/k0sproject/k0s/blob/c0998466cdf0752bdf3d601a973265790974097b/pkg/autopilot/controller/setup.go#L204

kimjune01:
## Description

Autopilot hardcoded `status.DefaultSocketPath` in multiple controllers, breaking `--status-socket` and custom `--run-dir` configurations. The socket path configured via `CfgVars.StatusSocketPath` was never threaded through to autopilot's signal controllers or the cron updater.

This adds `StatusSocketPath` to `RootConfig` and passes it through:
- `signal.RegisterControllers` → `k0s.RegisterControllers`
- `restart` / `restarted` reconcilers (version check and PID lookup)
- `updates.RegisterControllers` → `newCronUpdater` (status info fetch)

Falls back to `status.DefaultSocketPath` when `StatusSocketPath` is empty, preserving existing behavior for default configurations.

Fixes #6750

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

Verified the `StatusSocketPath` propagation manually by tracing the call chain from `component/controller/autopilot.go` and `component/worker/autopilot.go` through to each consumer. All previously hardcoded `status.DefaultSocketPath` references are replaced with the threaded value.

## Checklist:


twz123:
Thanks! Were you able to test this?

kimjune01:
No runtime testing yet. The verification was through manual code inspection of the call chain. I can set up a test configuration with --status-socket if that would be helpful for validating the fix.

kimjune01:
Yes, I've now run the full autopilot unit test suite locally against this branch:

```
$ go test ./pkg/autopilot/... -count=1
ok   github.com/k0sproject/k0s/pkg/autopilot/channels         0.518s
ok   github.com/k0sproject/k0s/pkg/autopilot/checks            0.889s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller         1.390s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/delegate      2.761s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans         2.462s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans/cmdprovider/airgapupdate  6.282s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans/cmdprovider/k0supdate     3.652s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans/cmdprovider/k0supdate/discovery  7.935s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans/cmdprovider/k0supdate/utils      1.954s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/plans/core     4.132s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/signal/airgap  5.481s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/signal/common  4.849s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/signal/common/predicate  4.412s
ok   github.com/k0sproject/k0s/pkg/autopilot/controller/signal/k0s     7.166s
ok   github.com/k0sproject/k0s/pkg/autopilot/signaling/v2      7.189s
ok   github.com/k0sproject/k0s/pkg/autopilot/updater            6.175s
```

`go build` and `go vet` also pass clean on all changed packages (`pkg/autopilot/...`, `pkg/component/controller/`, `pkg/component/worker/`).

What I haven't tested: running an actual k0s cluster with `--status-socket` or a custom `--run-dir` to confirm the socket path reaches the autopilot controllers at runtime. That would require a multi-node cluster with autopilot enabled, which I don't have set up locally. If there's a preferred integration test setup or a smoke test I should run, happy to do that.

twz123:
> What I haven't tested: running an actual k0s cluster with --status-socket or a custom --run-dir to confirm the socket path reaches the autopilot controllers at runtime. That would require a multi-node cluster with autopilot enabled, which I don't have set up locally. If there's a preferred integration test setup or a smoke test I should run, happy to do that.

You could parameterize one of the existing integration tests to use a custom run dir and a custom status socket. Not sure if it even has to be a multi-node cluster that gets updated? At least for the status socket, the code _looks_ like it should fail for a single node cluster, as well, AFAICT.

I'd try to add a variant for, say, the `ap-single` integration test à la `ap-single-custom-socket` or sth. along the lines (see how `os.Getenv("K0S_INTTEST_TARGET")` is used for parameterization in other inttests), and then see if that's enough to repro the problem. But beware, the AP tests have some ... limitations :see_no_evil:: #3719. You might need to double check manually, or fix that issue first :sweat_smile:

kimjune01:
Pushed 6a6574c parameterizing `ap-single` per your suggestion. New target `K0S_INTTEST_TARGET=ap-single-custom-socket` reuses the existing `ap-single` package; when set, `SetupTest` launches the controller with `--status-socket=/run/k0s/custom/status.sock --data-dir=/var/lib/k0s-custom`, then `TestApply` runs the same `k0supdate` Plan it always runs. After the Plan reaches `PlanCompleted`, the test asserts the custom socket exists and `/run/k0s/status.sock` does not — that second check is the regression guard against the bug, since prior to this PR the autopilot signal controllers' post-restart probe would have hit the hardcoded default path. (I added it explicitly because of #3719: PlanCompleted alone isn't sufficient evidence the cluster path actually worked.) Wired into `smoketests` so CI runs it.

Verified locally: `go vet ./inttest/ap-single/...`, `go vet ./pkg/autopilot/...`, and a compile-only `go test -run NONE ./inttest/ap-single/...` all pass; `make -n check-ap-single-custom-socket` resolves to the expected `K0S_INTTEST_TARGET=ap-single-custom-socket go test ... github.com/k0sproject/k0s/inttest/ap-single`. The actual bootloose run needs CI — I don't have a Docker-on-Linux setup that can spin the alpine bootloose image. If the post-Plan socket assertions turn out to be too strict on bootloose (e.g. some other component touches `/run/k0s/status.sock` before the autopilot controllers run), happy to relax to "custom socket exists" only.

PR #7620, opened 2026-05-13
Bump kube-router to v2.9.0

juanluisvaladas:
Looking into why dependabot hasn't done it, but in the meantime I'm doing it manually.


juanluisvaladas:
Figured out the issue with renovate, testing on my local repo.

PR #7633, opened 2026-05-15
Fix autopilot periodic update window evaluation

immanuwell:
## Description

`IsWithinPeriod(t)` looked like it checked the given time, but the window start was built from `time.Now()`. That made fixed/non-current evals miss the window; lowkey nasty for tests and edge timing.

No related issue found.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

Repro, on the old code with the new test:

```sh
go test ./pkg/apis/autopilot/v1beta2 -run 'TestPeriodicUpgradeStrategy_IsWithinPeriod/Fixed_future_time_within_window' -count=1
```

It fails with `got false, want true`. After this fix, it passes.

Also ran:

```sh
go test ./pkg/apis/autopilot/v1beta2 ./pkg/autopilot/controller/updates -count=1
git diff --check -- pkg/apis/autopilot/v1beta2/updateconfig.go pkg/apis/autopilot/v1beta2/updateconfig_test.go
```

## Checklist

### DIFF
--- a/pkg/apis/autopilot/v1beta2/updateconfig.go
+++ b/pkg/apis/autopilot/v1beta2/updateconfig.go
@@ -118,7 +118,7 @@ func (p *PeriodicUpgradeStrategy) IsWithinPeriod(t time.Time) bool {
 		return false
 	}
 
-	startTime := startTimeForCurrentDay(st)
+	startTime := startTimeForDay(st, t)
 
 	windowDuration, err := time.ParseDuration(p.Length)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


twz123:
Thanks! Just to better understand the issue: Did you experience any problems with the old version, or did you come across this randomly while inspecting the sources? I'm asking because the only usage of `IsWithinPeriod` is calling it with `time.Now()` anyway, so the only observable glitch could happen if the periodic updater is called exactly at midnight.

immanuwell:
@twz123 yes, I came across this randomly while inspecting the sources
since the bug is rare, maybe it's better just to close this PR

twz123:
No. I'm fine with it. Just wondering about the impact. Thanks for confirming.

twz123 (approved):

PR #7644, opened 2026-05-15
Detect and reconcile stale EtcdMember objects

twz123:
## Description

If a controller is permanently removed without going through the normal leave procedure (e.g. after a backup restore), its `EtcdMember` object remains marked as joined even though the controller is no longer present in the etcd cluster. This is confusing and may cause operators to draw false conclusions.

Make the etcd member reconciler detect such stale members and mark them as `Joined=false` (if not already marked as such), without requiring `Spec.Leave` to be set. Members are now matched against the etcd cluster by member ID rather than by name, which is more reliable after renames or re-adds. To keep the reconcile loop consistent, a single etcd client and member list is created once per resync cycle and shared across all member evaluations.

Extend the backup integration test to verify that stale members are detected after restore and that all members are correctly marked as joined once the cluster has been fully restored. Fan out the three backup test variants into separate CI runs.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/inttest/Makefile.variables
+++ b/inttest/Makefile.variables
@@ -19,6 +19,8 @@ smoketests := \
 	check-ap-updater \
 	check-ap-updater-periodic \
 	check-backup \
+	check-backup-stream \
+	check-backup-kine \
 	check-basic \
 	check-bind-address \
 	check-byocri \
--- a/cmd/etcd/list.go
+++ b/cmd/etcd/list.go
@@ -4,14 +4,21 @@
 package etcd
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 
 	"github.com/k0sproject/k0s/pkg/config"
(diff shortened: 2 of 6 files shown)


inttest/Makefile.variables
```diff
@@ -19,6 +19,8 @@ smoketests := \
 	check-ap-updater \
 	check-ap-updater-periodic \
 	check-backup \
+	check-backup-stream \
+	check-backup-kine \
```
jnummelin, line 23: I fail to see how these map into `inttest/backup/backup_test.go`?

twz123: Via the `K0S_INTTEST_TARGET` environment variable introduced in #6174.

jnummelin (dismissed):

jnummelin (approved):

PR #7649, opened 2026-05-18
Self-remove from etcd cluster before stopping

twz123:
## Description

The EtcdMember reconciler removed a leaving node from the etcd cluster only after that node had already stopped (i.e. its controller lease was no longer held). This could break the cluster if stopping of the leaving node dropped the cluster below quorum, effectively preventing downscaling below quorum, e.g. from two nodes to one.

Fix this by having the leaving node remove itself from etcd before stopping its process, while it can still contribute to quorum. When the k0s leader requests a node shutdown and the node's EtcdMember resource is flagged for leaving, the etcd component now removes itself from the cluster before handing off to the supervisor.

Extended the etcd member integration test to cover a full scale-down/scale-up cycle rather than a single leave-and-rejoin.

Fixes:

* #6901

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -16,6 +16,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"sync/atomic"
 	"time"
 
 	"github.com/k0sproject/k0s/cmd/internal"
@@ -249,6 +250,7 @@ func (c *command) start(ctx context.Context, rtc *config.RuntimeConfig, nodeConf
 	logrus.Infof("DNS address: %s", dnsAddress)
 
 	var storageBackend manager.Component
+	var leaveEtcdClusterOnStop *atomic.Bool
 	storageType := nodeConfig.Spec.Storage.Type
 
 	switch storageType {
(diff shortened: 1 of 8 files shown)


makhov (dismissed):

makhov:
@twz123 shall we proceed with this?

twz123:
> @twz123 shall we proceed with this?

Yes! The only thing I'm not too sure of is the [one minute delay](https://github.com/k0sproject/k0s/pull/7649/files#diff-8079d80fc69eee5885b2ab1e45866b577101202b3fdc311eaf9eeb94234fbd99R613) before deleting stale members. This might be too short for longer shutdown cycles, especially if the API server on the node that's shutting down has been started with a long request timeout. Do you have any opinion on that?

We can't keep the controller lease alive while the API server is in its graceful shutdown sequence, hence we can't send a signal to the current leader that we're still in orderly shutdown. OTOH, the leader needs to kick out stale members eventually, if they failed to do that by themselves.

twz123:
For that matter, as another security measure to prevent the cluster from breaking, I was also thinking about limiting the number of controllers scheduled for shutdown to one. So we never have more than one controller trying to shutdown and leave the cluster concurrently. Although we can do this in a separate PR. Would that make sense?

jnummelin:
> OTOH, the leader needs to kick out stale members eventually, if they failed to do that by themselves.

Yeah, I don't see any other way around either. 1 minute feels maybe bit too fast for that though, like you mentioned too

> ... never have more than one controller trying to shutdown and leave the cluster concurrently.

SGTM. But definitely worth as a separate followup PR IMO.

twz123:
> Yeah, I don't see any other way around either. 1 minute feels maybe bit too fast for that though, like you mentioned too

Changed this to 5 minutes.

jnummelin (approved):

Issue #7186, opened 2026-03-01
Introducing IPv6 on existing Cluster


pschichtel:
Hi there! This is more of a discussion/question, since I'm pretty sure this is not officially supported, neither here nor upstream. I still want to try though. I have an existing 3 node (controller+worker) bare-metal cluster on 1.35.1. This is a home lab, so I'm willing to toy around and potentially break things, even though I'd prefer not to break it.

The cluster is currently on IPv4-only since back when I originally set the cluster up, kube-router didn't support IPv6/Dual-Stack.

Just simply enabling dual stack and settings the cidrs is clearly not enough.

Doing that makes the kube-controller-manage fail with:

```
failed to mark cidr[10.0.0.0/24] at idx [0] as occupied for node: server: cidr 10.0.0.0/24 is out the range of cluster cidr 1111:1111:1111:1111::1:0/112
```

From my research that might be because the node objects don't have an IP in the new IPv6 cluster cidr. Unfortunately the IP can't just be added to an existing node.

Can I just drain and delete a node and have k0s register it again?

Any input on this? Thanks in advance!

juanluisvaladas:
Hi,

This error is very suspicoius and makes us think you changed the [primary address family](https://docs.k0sproject.io/stable/dual-stack/?h=primary+family#specifying-the-default-ip-family) (if this happened this was most certainly an accidental change). Could you please provide:

1- The full k0s configuration of the control plane node (one node should be enough. Feel free to redact stuff but if it's a hostname we need to know if it resolves to IPv4 and/or IPv6 and we need to know if IP addresses are either IPv4 or IPv6( 
2- `ps auxww | grep k0s` in a control plane node.
3- `ps auxww | grep -e k0s -e kube-router` in a worker node.

Regarding if this is supported or not, kubernetes itself supports it and I don't see why it wouldn't be possible in kube-router, but it's not something that we actively test.

Regarding your actual question:

> Can I just drain and delete a node and have k0s register it again?

This should be fine if the problem is that, but we haven't actually tested it. But since the error is coming from kube-controller manager I don't think that's the issue.

pschichtel:
I added the primary address family explicitly as IPv4 to the config. From what was documented, the default value should have been IPv4 anyway, that's why I thought adding IPv4 explicitly shouldn't change a thing.

[...]

PR #7684, opened 2026-05-24
More consistently build pod and service CIDRs

pschichtel:
## Description

This PR makes BuildPodCIDR consistent with BuildServiceCIDR in that it now also puts the primary address first. I don't think the order of the pod cidrs actually matters. More importantly this PR also make use of BuildServiceCIDR for building the `--service-cluster-ip-range` argument for kuberouter, which so far only had the primary address.

Doesn't fix, but is related to: #7186

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -282,16 +282,27 @@ func (n *Network) BuildServiceCIDR(primaryAddressFamily PrimaryAddressFamilyType
 	case PrimaryFamilyIPv6:
 		return n.DualStack.IPv6ServiceCIDR + "," + n.ServiceCIDR
 	default:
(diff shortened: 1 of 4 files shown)


twz123:
> More importantly this PR also make use of BuildServiceCIDR for building the `--service-cluster-ip-range` argument for kuberouter, which so far only had the primary address.

Aha, this sounds like it might fix #6771?

pschichtel:
Not sure, I don't think my kube-router had trouble starting without it.

twz123 (approved):
Thanks for this!

pschichtel:
Also another note: I haven't looked at calico _at all_, so not sure if there is something similar that needs a dualstack adjustment. Maybe worth having a look.

PR #7685, opened 2026-05-24
Allow workers to use cached config

### LINKED ISSUE #7379: Support offline worker startup with cached kubelet configuration

When a k0s worker node restarts and the control plane is unavailable, the worker blocks indefinitely waiting to fetch the kubelet configuration ConfigMap from the API server. This prevents the kubelet from starting, which means static pods (in our case, critical infrastructure) cannot run.

Workers that have previously joined the cluster should be able to start with their cached configuration when the control plane is temporarily unreachable.

Add a `--allow-cached-config` flag to k0s worker that allows using the cached worker profile when it exists and the API server is unreachable. When the API server is reachable, the behavior would remain unchanged. This logic to handle caching the kubelet config is already implemented, but is only used to [load API server addresses](https://github.com/k0sproject/k0s/blob/70099d08e1b7739384d6e0c26a8118dac0883696/pkg/component/worker/config/loader.go#L109)

### Proposed behavior:
- Attempt to fetch configuration from the API server (current behavior)
- If the API server is unreachable and `--allow-cached-config` is set:
  - Use the cached profile if it exists and matches the requested profile name
  - Fail with a clear error if no valid cache exists
  - Keep the same backoff retry behavior for attempting to retrieve config via ConfigMap

_No response_

### Additional context

_No response_

azmifarih:
Fixes #7379

Adds `--allow-cached-config` for workers. When enabled, k0s still tries to load the worker profile from the Kubernetes API first. If the API server is unavailable, it falls back to the cached worker profile when the cached profile exists and matches the requested profile.

Default behavior is unchanged.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

Unit tests:

```sh
go test ./pkg/component/worker/config ./pkg/config ./cmd/worker ./cmd/controller ./cmd/install
go test ./cmd/... ./pkg/component/worker/... ./pkg/config
```

Manual bootloose verification:

1. Started a `check-basic` bootloose cluster.
2. Confirmed `/var/lib/k0s/worker-profile.yaml` exists on `worker0`.
3. Stopped the controller.
4. Restarted `worker0` with `--allow-cached-config`.
5. Confirmed the worker used the cached profile:

```text
Using cached worker profile "default" because the Kubernetes API is unavailable
```

6. Confirmed `k0s`, `containerd`, and `kubelet` were running on the worker.
7. Started the controller again.
8. Confirmed both workers returned to `Ready` and all kube-system pods were `Running`.

## Checklist

### DIFF
--- a/cmd/controller/controller_test.go
+++ b/cmd/controller/controller_test.go
@@ -50,6 +50,7 @@ Examples:
 	Note: Token can be passed either as a CLI argument, a flag, or an environment variable
 
 Flags:
+      --allow-cached-config                            allow worker startup with cached configuration when the Kubernetes API is unavailable
       --api-server-stop-timeout duration               time to wait for the API server to stop
   -c, --config string                                  config file, use '-' to read the config from stdin (default `+defaultConfigPath+`)
       --cri-socket string                              container runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]
--- a/cmd/install/controller_test.go
+++ b/cmd/install/controller_test.go
@@ -43,6 +43,7 @@ With the controller subcommand you can setup a single node cluster by running:
 	
 
 Flags:
+      --allow-cached-config                            allow worker startup with cached configuration when the Kubernetes API is unavailable
       --api-server-stop-timeout duration               time to wait for the API server to stop
   -c, --config string                                  config file, use '-' to read the config from stdin (default `+defaultConfigPath+`)
       --cri-socket string                              container runtime socket to use, default to internal containerd. Format: [remote|docker]:[path-to-socket]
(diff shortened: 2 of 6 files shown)


jnummelin:
@azmifarih What's the use case where you need this functionality?

azmifarih:
> @azmifarih What's the use case where you need this functionality?

The main use case is edge/offline recovery.

If a worker that has already joined the cluster reboots while the control plane is temporarily unreachable, it currently blocks on fetching the worker profile ConfigMap. That means kubelet does not start, so local/static pod workloads on the worker cannot recover.

Dwilborn:
Bump on this?

PR #7686, opened 2026-05-24
Configure CoreDNS and kubelet for dual stack DNS

pschichtel:
## Description

Previously the kube-dns service only received the primary IP address and the service was single stack. It is now configured with 2 address and address families and PreferDualStack (maybe that should be require?). To utilize the IP addresses, kubelet's config now also includes both IP addresses when running in dual stack mode.

Doesn't fix, but is related to: https://github.com/k0sproject/k0s/issues/7186

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [X] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -187,15 +187,6 @@ func (n *Network) Validate() []error {
 	return errors
 }
 
-// DNSAddress calculates the 10th address of configured service CIDR block.
-func (n *Network) DNSAddress(primaryAddressFamily PrimaryAddressFamilyType) (string, error) {
-	addresses, err := n.DNSAddresses(primaryAddressFamily)
-	if err != nil {
-		return "", err
-	}
-	return addresses[0].String(), nil
-}
-
 func (n *Network) DNSAddresses(primaryAddressFamily PrimaryAddressFamilyType) ([]net.IP, error) {
 	var cidrs []string
 	if n.DualStack.Enabled {
--- a/pkg/apis/k0s/v1beta1/network.go
+++ b/pkg/apis/k0s/v1beta1/network.go
@@ -189,30 +189,51 @@ func (n *Network) Validate() []error {
 
 // DNSAddress calculates the 10th address of configured service CIDR block.
 func (n *Network) DNSAddress(primaryAddressFamily PrimaryAddressFamilyType) (string, error) {
-	serviceCIDR := n.ServiceCIDR
-	if n.DualStack.Enabled && primaryAddressFamily == PrimaryFamilyIPv6 {
-		serviceCIDR = n.DualStack.IPv6ServiceCIDR
-	}
-
-	_, ipnet, err := net.ParseCIDR(serviceCIDR)
+	addresses, err := n.DNSAddresses(primaryAddressFamily)
 	if err != nil {
-		return "", fmt.Errorf("failed to parse service CIDR %q: %w", serviceCIDR, err)
+		return "", err
 	}
+	return addresses[0].String(), nil
+}
 
-	addr := slices.Clone(ipnet.IP)
-
-	maskLen, netLen := ipnet.Mask.Size()
-	if netLen-maskLen > 3 {
-		addr[len(addr)-1] += 10
+func (n *Network) DNSAddresses(primaryAddressFamily PrimaryAddressFamilyType) ([]net.IP, error) {
+	var cidrs []string
+	if n.DualStack.Enabled {
+		if primaryAddressFamily == PrimaryFamilyIPv6 {
+			cidrs = []string{n.DualStack.IPv6ServiceCIDR, n.ServiceCIDR}
+		} else {
+			cidrs = []string{n.ServiceCIDR, n.DualStack.IPv6ServiceCIDR}
+		}
 	} else {
-		addr[len(addr)-1] += 2
+		cidrs = []string{n.ServiceCIDR}
 	}
 
-	if !ipnet.Contains(addr) {
-		return "", fmt.Errorf("failed to calculate DNS address: CIDR too narrow: %s", n.ServiceCIDR)
+	var addresses []net.IP
+
+	for _, cidr := range cidrs {
+
+		_, ipnet, err := net.ParseCIDR(cidr)
+		if err != nil {
+			return []net.IP{}, fmt.Errorf("failed to parse service CIDR %q: %w", cidr, err)
+		}
+
+		addr := slices.Clone(ipnet.IP)
+
+		maskLen, netLen := ipnet.Mask.Size()
+		if netLen-maskLen > 3 {
+			addr[len(addr)-1] += 10
+		} else {
+			addr[len(addr)-1] += 2
+		}
+
+		if !ipnet.Contains(addr) {
+			return []net.IP{}, fmt.Errorf("failed to calculate DNS address: CIDR too narrow: %s", cidr)
+		}
+
+		addresses = append(addresses, addr)
 	}
 
-	return addr.String(), nil
+	return addresses, nil
 }
 
 // InternalAPIAddresses calculates the internal API address of configured service CIDR block.
--- a/pkg/component/controller/calico.go
+++ b/pkg/component/controller/calico.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"net"
 	"path"
 	"path/filepath"
 	"strings"
@@ -73,7 +74,7 @@ type calicoConfig struct {
 type calicoNodeConfig struct {
 	APIServer       *k0snet.HostPort
 	ServiceCIDRIPv4 string
-	ClusterDNSIP    string
+	ClusterDNSIPs   []net.IP
 }
 
 type calicoClusterConfig struct {
@@ -102,7 +103,7 @@ type calicoClusterConfig struct {
 
 // NewCalico creates new Calico reconciler component
 func NewCalico(nodeConfig *v1beta1.ClusterConfig, manifestsDir string, hasWindowsNodes func() (*bool, <-chan struct{})) (*Calico, error) {
-	dnsAddress, err := nodeConfig.Spec.Network.DNSAddress(nodeConfig.Spec.PrimaryAddressFamily())
+	dnsAddresses, err := nodeConfig.Spec.Network.DNSAddresses(nodeConfig.Spec.PrimaryAddressFamily())
 	if err != nil {
 		return nil, err
 	}
@@ -117,7 +118,7 @@ func NewCalico(nodeConfig *v1beta1.ClusterConfig, manifestsDir string, hasWindow
 		nodeConfig: calicoNodeConfig{
 			APIServer:       apiServer,
 			ServiceCIDRIPv4: nodeConfig.Spec.Network.ServiceCIDR,
-			ClusterDNSIP:    dnsAddress,
+			ClusterDNSIPs:   dnsAddresses,
 		},
 		primaryAddressFamily: nodeConfig.Spec.PrimaryAddressFamily(),
 		manifestsDir:         manifestsDir,
--- a/pkg/component/controller/coredns.go
+++ b/pkg/component/controller/coredns.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"net"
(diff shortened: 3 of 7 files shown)


jnummelin:
> PreferDualStack (maybe that should be require?)

hmm, in this case I don't think there's much of a difference. In our case we know when things are dual-stack. This is what kube docs say, so if I read that correctly both Prefer.. and Require... behave exact same in our case.

```
PreferDualStack: Allocates both IPv4 and IPv6 cluster IPs for the Service when dual-stack is enabled. If dual-stack is not enabled or supported, it falls back to single-stack behavior.
```

pschichtel:
yeah that's how I interpreted the docs as well. I went with "prefer" only for the remote chance of the service somehow being misaligned with the cluster dualstack configuration and then being blocked because it is RequireDualStack but doesn't have/receive an IPv6. not sure if that is ever a realistic scenario.

pkg/component/controller/coredns.go
```diff
@@ -251,6 +252,17 @@ spec:
...
   clusterIP: {{ .ClusterDNSIP }}
+  {{- if .DualStack }}
+  clusterIPs:
+  {{- range .ClusterDNSIPs }}
+  - {{ . | quote }}
+  {{- end }}
+  ipFamilies:
+  {{- range .ClusterDNSIPFamilies }}
+  - {{ . | quote }}
+  {{- end }}
```
pschichtel, line 262: I'm not entirely sure if the order of the families matters. I currently implemented it in a way that ensures that if IPv4 is primary, than IPv4 will be the first IP family here. if it doesn't matter, then I can get rid of quite a bit of code and simplify the template here.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -186,30 +186,51 @@ func (n *Network) Validate() []error {
 
 // DNSAddress calculates the 10th address of configured service CIDR block.
 func (n *Network) DNSAddress(primaryAddressFamily PrimaryAddressFamilyType) (string, error) {
```
pschichtel: I retained the original function mostly to not have to touch any other code. I guess ideally it would be completely replaced.

pkg/component/controller/coredns.go
pschichtel: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/563-dual-stack

the KEP suggests (I haven't fully read it) that clusterIPs and ipFamilies are desigend as parallel arrays, so clusterIPs[n] must match ipFamilies[n]. In that case the implementation like this is necessary.

pkg/apis/k0s/v1beta1/network.go
twz123: I think it actually makes sense to remove it, so all call sites are aware that the DNS may actually have multiple addresses. Would you be up for that?

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -186,30 +186,51 @@ func (n *Network) Validate() []error {
...
 	}
+	return addresses[0], nil
+}
 
-	addr := slices.Clone(ipnet.IP)
-
-	maskLen, netLen := ipnet.Mask.Size()
-	if netLen-maskLen > 3 {
-		addr[len(addr)-1] += 10
+func (n *Network) DNSAddresses(primaryAddressFamily PrimaryAddressFamilyType) ([]string, error) {
```
twz123: I think it's fair to return a `net.IP` slice here, instead of a string slice.

pkg/component/controller/coredns.go
```diff
@@ -360,12 +375,28 @@ func (c *CoreDNS) getConfig(ctx context.Context, clusterConfig *v1beta1.ClusterC
...
+
 	config := coreDNSConfig{
-		Replicas:      replicaCount(nodeCount),
-		ClusterDomain: c.clusterDomain,
-		ClusterDNSIP:  c.dnsAddress,
-		Image:         clusterConfig.Spec.Images.CoreDNS.URI(),
-		PullPolicy:    clusterConfig.Spec.Images.DefaultPullPolicy,
+		Replicas:             replicaCount(nodeCount),
+		ClusterDomain:        c.clusterDomain,
+		ClusterDNSIP:         c.dnsAddresses[0],
```
twz123: That could be done in the template, as well, so we'd have one field less in the template struct.

pschichtel: I wasn't sure what functions are available in template. how can I find out which functions are available?

pkg/apis/k0s/v1beta1/network.go
pschichtel: I can have a look in the coming days.

pkg/component/controller/coredns.go
twz123: I guess `{{ index .ClusterDNSIPs 0 }}` should do the trick. The upstream go template docs explain this: <https://pkg.go.dev/text/template>

pschichtel: pretty much 100% of my go template usage is through helm, but I know they have a ton of extra functionality, so I'm never sure what's available by default. I'll check that.

pkg/component/controller/calico.go
```diff
@@ -64,7 +65,7 @@ type calicoConfig struct {
 type calicoNodeConfig struct {
 	APIServer       *k0snet.HostPort
 	ServiceCIDRIPv4 string
-	ClusterDNSIP    string
+	ClusterDNSIPs   []net.IP
```
pschichtel, line 77: @twz123 Removing DNSAddress() revealed that this field is unused, which is kinda concerning.

pkg/component/controller/coredns.go
pschichtel: works as expected

pschichtel:
I'll rebase in the evening.

Any input regarding the unused calico fields?

pschichtel:
Is there still a chance this could land for 1.36 or 1.36.1?

pschichtel:
Fixed the failing test and rebased.

twz123:
> Is there still a chance this could land for 1.36 or 1.36.1?

I guess we can backport this.

pschichtel:
Ok, i fixed the golangci-lint violation now. I also rebased onto latest main.

pschichtel:
The CI failures don't seem to be related to my change, or do they?

twz123:
> The CI failures don't seem to be related to my change, or do they?

Will retrigger this. We need to work on the integration test stability. I'd love to have some automated LLM based triage on failed tests, so contributors get quick feedback, and maintainers won't just hit retry without looking at the actual failure cause (No, I'm definitely not talking about me, here 🙈)

pschichtel:
I just rebased this onto main. Any updates on the flaky CI?

pschichtel:
\*friendly reminder*

pkg/component/controller/coredns.go
twz123: Can we add a test for this? Something along the lines of providing a dual stack config, and then checking if the generated manifest is a valid service that has the correct IPs and families in the right order?

pkg/component/controller/calico.go
twz123: This is because it's used in the Windows manifests. The WSL integration test uncovered this, but I figured a unit test would've been helpful as well: #8320

IIRC Windows only supports IPv4, so you somehow need to extract the v4 CIDR in the template. Perhaps via a special function that the template can call? Dunno...

pkg/component/controller/coredns.go
```diff
@@ -360,12 +374,28 @@ func (c *CoreDNS) getConfig(ctx context.Context, clusterConfig *v1beta1.ClusterC
...
+		addresses = append(addresses, address.String())
+		if address == nil {
+			return coreDNSConfig{}, fmt.Errorf("invalid IP address %q", address)
+		}
+		if address.To4() == nil {
+			addressFamilies = append(addressFamilies, "IPv6")
+		} else {
+			addressFamilies = append(addressFamilies, "IPv4")
+		}
+	}
```
twz123, line 389: This whole block can be moved into `NewCoreDNS` and you can store the prefabricated addresses and families in the struct directly.

pkg/component/controller/coredns.go
```diff
@@ -360,12 +374,28 @@ func (c *CoreDNS) getConfig(ctx context.Context, clusterConfig *v1beta1.ClusterC
 
 	nodeCount := len(nodes.Items)
 
+	var addresses []string
+	var addressFamilies []string
+	for _, address := range c.dnsAddresses {
+		addresses = append(addresses, address.String())
+		if address == nil {
+			return coreDNSConfig{}, fmt.Errorf("invalid IP address %q", address)
+		}
```
twz123, line 383: I wouldn't check for a `nil` `net.IP`. You think this is a scenario that's worth it? We don't do it anywhere else, AFAICT. To me, `net.IP` is more like a value type, and if it can be `nil`, I'd rather expect `*net.IP`, in a way. (That's one of the reasons why I prefer `netip.Addr` for new code these days, which doesn't have that problem.)

```suggestion
```

pkg/component/controller/coredns.go
```diff
@@ -360,12 +374,28 @@ func (c *CoreDNS) getConfig(ctx context.Context, clusterConfig *v1beta1.ClusterC
...
-		ClusterDNSIP:  c.dnsAddress,
-		Image:         clusterConfig.Spec.Images.CoreDNS.URI(),
-		PullPolicy:    clusterConfig.Spec.Images.DefaultPullPolicy,
+		Replicas:             replicaCount(nodeCount),
+		ClusterDomain:        c.clusterDomain,
+		ClusterDNSIPs:        addresses,
+		ClusterDNSIPFamilies: addressFamilies,
+		Image:                clusterConfig.Spec.Images.CoreDNS.URI(),
+		PullPolicy:           clusterConfig.Spec.Images.DefaultPullPolicy,
+		DualStack:            clusterConfig.Spec.Network.DualStack.Enabled,
```
twz123, line 398: That's one of the nice "wrinkles" of the k0s code base. The distinction between node config and cluster config (even the names are suboptimal 🙈). I guess it's a bug that the DualStack config even ends up in the cluster config. We can't dynamically reconcile a switch from single stack to dual stack or vice versa. You'd need to do this by hand (probably by re-installing the whole cluster).

So you have to pick that one up from the node config only, which is coming from the k0s config file only and fixed for the whole process lifetime.

That said, I think you can drop this field altogether. DualStack is already implied by the len of DNS address and family slices.

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -187,32 +187,44 @@ func (n *Network) Validate() []error {
 	return errors
 }
 
-// DNSAddress calculates the 10th address of configured service CIDR block.
-func (n *Network) DNSAddress(primaryAddressFamily PrimaryAddressFamilyType) (string, error) {
-	serviceCIDR := n.ServiceCIDR
-	if n.DualStack.Enabled && primaryAddressFamily == PrimaryFamilyIPv6 {
-		serviceCIDR = n.DualStack.IPv6ServiceCIDR
+func (n *Network) DNSAddresses(primaryAddressFamily PrimaryAddressFamilyType) ([]net.IP, error) {
```
twz123, line 190: You've swallowed the godoc here. Might make sense to re-add it and expand it, saying that the function now returns two addresses for dual stack, the primary family first.

pkg/component/controller/coredns.go
```diff
@@ -249,7 +250,18 @@ metadata:
 spec:
   selector:
     k8s-app: kube-dns
-  clusterIP: {{ .ClusterDNSIP }}
+  clusterIP: {{ index .ClusterDNSIPs 0 }}
+  {{- if .DualStack }}
+  clusterIPs:
```
twz123, line 255: Could we just unconditionally write clusterIPs here, even if we only have one? Same for ipFamilies? Would allow us to get rid of DualStack completely here.

Moreover, I think we could do the mapping from IP addresses to address families right here in the template. That way, we'd get rid of another struct/template field, and clusterIPs and ipFamilies  can't diverge anymore, although not sure if it'd be very nice to express this in go template primitives. Anyhow, just mentioning...

twz123 (commented):
Thanks for your patience!

pschichtel:
Nice to see this rolling again. I'm currently traveling and will get back to this by mid of October.

PR #7697, opened 2026-05-28
Makefile: prune build/cache and inttest from GO_SRCS find

ncopa:
-not -path is a filter, not a prune: find still traversed the entire build/cache tree before discarding the results, costing up to ~20 seconds on every make invocation when build cached is big (17 GB, 437K+ files).

Switch to -prune so find skips those directories entirely, bringing make startup time from ~20s to ~0.3s.

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,12 @@ DOCKER ?= docker
 
 K0S_GO_BUILD_CACHE ?= build/cache
 
-GO_SRCS := $(shell $(FIND) . -type f -name "*.go" -not -path "./$(K0S_GO_BUILD_CACHE)/*" -not -path "./inttest/*" -not -name "*_test.go" -not -name "zz_generated*")
+# find evaluates left-to-right: match the dirs first, -prune them (skip their
+# entire subtree without traversing), then -o match the actual .go files. The
+# explicit -print is needed because without it find's implicit -print applies
+# to both sides of -o and would also print the pruned directory names.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,7 +32,7 @@ DOCKER ?= docker
 
 K0S_GO_BUILD_CACHE ?= build/cache
 
-GO_SRCS := $(shell $(FIND) . -type f -name "*.go" -not -path "./$(K0S_GO_BUILD_CACHE)/*" -not -path "./inttest/*" -not -name "*_test.go" -not -name "zz_generated*")
+GO_SRCS := $(shell $(FIND) . \( -path "./$(K0S_GO_BUILD_CACHE)" -o -path "./inttest" \) -prune -o -type f -name "*.go" -not -name "*_test.go" -not -name "zz_generated*" -print)
```
jnummelin, line 39: The flag ordering _feels_ bit weird. I trust that it works but for the next reader of the code, it might help with some comment what the prune actually drops out

ncopa: comment was added

twz123: TIL!

Might be clearer if we set the braces in a way that makes clear that we have two ORed expressions, one ending in -prune, the other ending in -print:

```suggestion
GO_SRCS := $(shell $(FIND) . \( \( -path "./$(K0S_GO_BUILD_CACHE)" -o -path "./inttest" \) -prune \) -o \( -type f -name "*.go" -not -name "*_test.go" -not -name "zz_generated*" -print \))
```

At least this helps me in understanding how this works, as I'm otherwise unsure about the binding precedence of the different tokens.

twz123 (approved):

PR #7706, opened 2026-05-30
Make applyingUpdate reconciler idempotent

### LINKED ISSUE #7703: Autopilot `applyingUpdate` reconciler is non-idempotent — workers wedge forever after `client.Update` conflict

### Platform

```shell
Linux (observed on workers running k0s in HA)
```

### Version

v1.35.1+k0s.1

### Sysinfo

<pre>
➡️ Please replace this text with the output of `k0s sysinfo`. ⬅️
</pre>
</details>

### What happened?

The `applyingUpdate` reconciler in `pkg/autopilot/controller/signal/k0s/apply.go` is **not idempotent**. If the worker successfully renames `/usr/local/bin/k0s.tmp` → `/usr/local/bin/k0s` but the subsequent `client.Update(ctx, node)` fails (e.g. with a `resourceVersion` conflict — "the object has been modified"), controller-runtime retries the reconciler. On retry, `os.Stat("/usr/local/bin/k0s.tmp")` fails because the file was already moved, and the reconciler enters an infinite error loop:

```
Applying update
Reconciler error  error="unable to find update file 'k0s.tmp': stat /usr/local/bin/k0s.tmp: no such file or directory"
```

The Node's autopilot signal annotation stays stuck at `ApplyingUpdate` forever, even though the target binary is already correctly in place at `/usr/local/bin/k0s`. Restarting the worker (or the k0s service) is **not** sufficient — on restart, the reconciler reads the same stuck annotation and re-enters the same failing loop. The only recovery is operator intervention: delete the autopilot `Plan` and remove the `k0sproject.io/autopilot-signal-data` annotation from each affected worker Node before a new Plan can make progress.

Reference source on `main`:
https://github.com/k0sproject/k0s/blob/main/pkg/autopilot/controller/signal/k0s/apply.go

The reconciler does:

1. `client.Get(ctx, node)` — read Node
2. `os.Stat("/usr/local/bin/k0s.tmp")` — check downloaded binary
3. `os.Rename("/usr/local/bin/k0s.tmp", "/usr/local/bin/k0s")` — replace binary
4. `client.Update(ctx, node)` — write `Restart` status to Node annotation

If step 4 fails after step 3 has run, the controller-runtime retry hits step 2 and fails forever.

Note: PR #6994 ("Requeue Autopilot signal node updates on conflict", merged 2026-01-27) fixes an analogous conflict-on-update bug in  `pkg/autopilot/controller/plans/cmdprovider/{k0supdate,airgapupdate}/schedulable.go`, but does **not** touch `signal/k0s/apply.go`. The worker-side `applyingUpdate` reconciler remains non-idempotent.

### Steps to reproduce

1. Create a 2-node cluster (1 controller + 1 worker) running k0s **v1.35.1+k0s.1** with autopilot enabled.
   **On the controller node:**

   ```bash
   curl --proto '=https' --tlsv1.2 -sSf https://get.k0s.sh | sudo K0S_VERSION=v1.35.1+k0s.1 sh
   sudo k0s install controller --enable-worker
   sudo k0s start
   ```

   **On the controller, generate a worker join token:**

   ```bash
   sudo k0s token create --role=worker > /tmp/worker-token
   ```

   Copy `/tmp/worker-token` over to the worker node.

   **On the worker node:**

   ```bash
   curl --proto '=https' --tlsv1.2 -sSf https://get.k0s.sh | sudo K0S_VERSION=v1.35.1+k0s.1 sh
   sudo k0s install worker --token-file /tmp/worker-token
   sudo k0s start
   ```

   **Wait for the cluster to be ready** (run on the controller):

   ```bash
   sudo k0s kubectl wait --for=condition=Ready node --all --timeout=120s
   ```

2. On the worker, start a tight loop that moves `k0s.tmp` into place as soon as it appears:

   ```bash
   while true; do
     { [ -f /usr/local/bin/k0s.tmp ] && mv /usr/local/bin/k0s.tmp /usr/local/bin/k0s; } 2>/dev/null || true
     sleep 0.05 2>/dev/null || true
   done
   ```

3. From the controller, create an autopilot `Plan` that upgrades to a newer version (e.g. v1.35.4+k0s.0) with `selector: {}` discovery so all nodes are targeted.
4. Observe the worker's autopilot logs:

   ```
   journalctl -u k0sworker -f | grep k0s.tmp
   ```

   The `applyingUpdate` reconciler loops forever with:

   ```
   unable to find update file 'k0s.tmp': stat /usr/local/bin/k0s.tmp: no such file or directory
   ```

5. Verify the binary was moved into place and is the correct target version:

   ```bash
   /usr/local/bin/k0s version
   # v1.35.4+k0s.0
   ```

   The Node annotation `k0sproject.io/autopilot-signal-data` stays stuck at `status.status: ApplyingUpdate` indefinitely.

**Note:** This reproducer does **not** simulate the original `client.Update` API-conflict trigger. It instead deterministically produces the same on-disk state the race leaves behind — `k0s.tmp` already moved into place before the reconciler's `os.Stat` runs. The reconciler then enters the same infinite error loop, which proves the underlying bug is the reconciler's non-idempotency, independent of what triggers the second pass.

### Expected behavior

The `applyingUpdate` reconciler should be idempotent: if `k0s.tmp` is missing, it should check whether `/usr/local/bin/k0s` is already the requested target version and, if so, proceed to write the `Restart` status to the Node annotation instead of erroring out. This would make the reconciler resilient to:

1. The `client.Update` conflict / retry path.
2. Any other scenario in which the binary is already in place from a prior partial run.

### Actual behavior

The reconciler enters an infinite error loop. The Node annotation `k0sproject.io/autopilot-signal-data` is permanently stuck at `status.status: ApplyingUpdate`, even though the upgrade binary is correctly installed and `k0s version` reports the new version. The autopilot Plan never progresses to `Restart` on the affected worker.

### Screenshots and logs

```
# journalctl -u k0sworker -f | grep autopilot
...
applying-update  Applying update
applying-update  Reconciler error  error="unable to find update file 'k0s.tmp': stat /usr/local/bin/k0s.tmp: no such file or directory"
applying-update  Applying update
applying-update  Reconciler error  error="unable to find update file 'k0s.tmp': stat /usr/local/bin/k0s.tmp: no such file or directory"
... (repeats every few seconds indefinitely)
```

When the trigger is the `client.Update` conflict, the preceding log line is:

```
failed to update signal node to status 'Restart':
Operation cannot be fulfilled on nodes "<node>": the object has been modified;
please apply your changes to the latest version and try again
```

### Additional context

_No response_

bainaichang:
## Description

When the client.Update call fails after successfully renaming k0s.tmp to k0s
(e.g. due to a resourceVersion conflict), the reconciler retries but fails
because k0s.tmp no longer exists. This results in an infinite error loop that
leaves the node stuck in ApplyingUpdate status.

Fix this by checking if the target k0s binary is already in place before
attempting the file rename. If it is, skip the file operations and proceed
directly to updating the signaling status to Restart.

Fixes: https://github.com/k0sproject/k0s/issues/7703

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

- `go build` passes
- Existing unit tests pass (`make check-unit`)

The logic follows the same pattern as the fix in PR #6994 for `schedulable.go`,
which addressed a similar non-idempotent reconciler issue.

## Checklist

- [x] My code follows the style of this project
- [x] My commit messages are in the proper format

### DIFF
--- a/pkg/autopilot/controller/signal/k0s/apply.go
+++ b/pkg/autopilot/controller/signal/k0s/apply.go
@@ -108,21 +108,41 @@ func (r *applyingUpdate) Reconcile(ctx context.Context, req cr.Request) (cr.Resu
 		return cr.Result{}, nil
 	}
 
+	k0sBinaryFilenamePath := filepath.Join(r.k0sBinaryDir, "k0s")
 	updateFilenamePath := filepath.Join(r.k0sBinaryDir, apconst.K0sTempFilename)
+	updateLinkFilenamePath := filepath.Join(r.k0sBinaryDir, apconst.K0sTempLinkFilename)
 
-	// Ensure that the expected file exists
-	if _, err := os.Stat(updateFilenamePath); errors.Is(err, os.ErrNotExist) {
-		return cr.Result{}, fmt.Errorf("unable to find update file '%s': %w", apconst.K0sTempFilename, err)
-	}
+	// Check if the update file still exists. If not, the rename was already
+	// performed in a previous reconciler run whose client.Update failed.
+	// In that case the file operations can be skipped and we can proceed
+	// directly to updating the signaling status to Restart.
+	if _, err := os.Stat(updateFilenamePath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return cr.Result{}, fmt.Errorf("unable to stat update file '%s': %w", updateFilenamePath, err)
+		}
+		logger.Info("Update file already applied, skipping file operations")
+	} else {
+		// Ensure the downloaded temporary file is executable
+		if err := os.Chmod(updateFilenamePath, 0755); err != nil {
+			return cr.Result{}, fmt.Errorf("unable to chmod update file '%s': %w", updateFilenamePath, err)
+		}
 
-	// Ensure that the new file is executable
-	if err := os.Chmod(updateFilenamePath, 0755); err != nil {
-		return cr.Result{}, fmt.Errorf("unable to chmod update file '%s': %w", apconst.K0sTempFilename, err)
-	}
+		// Clean up any stale link file from a previous failed rename attempt
+		os.Remove(updateLinkFilenamePath)
 
-	// Perform the update atomically
-	if err := os.Rename(updateFilenamePath, filepath.Join(r.k0sBinaryDir, "k0s")); err != nil {
-		return cr.Result{}, fmt.Errorf("unable to update (rename) to the new file: %w", err)
+		// Create k0s.new as a hard link to k0s.tmp, sharing the same
+		// inode. This way k0s.tmp survives the subsequent rename,
+		// providing idempotency: if client.Update fails and the
+		// reconciler is re-triggered, k0s.tmp will still exist and
+		// the whole sequence can be replayed.
+		if err := os.Link(updateFilenamePath, updateLinkFilenamePath); err != nil {
+			return cr.Result{}, fmt.Errorf("unable to create hard link '%s' -> '%s': %w", updateLinkFilenamePath, updateFilenamePath, err)
+		}
+
+		// Atomically replace the running k0s binary with the new version
+		if err := os.Rename(updateLinkFilenamePath, k0sBinaryFilenamePath); err != nil {
+			return cr.Result{}, fmt.Errorf("unable to rename '%s' -> '%s': %w", updateLinkFilenamePath, k0sBinaryFilenamePath, err)
+		}
 	}
 
 	// When the k0s process has been terminated, move to 'Restart'
@@ -136,6 +156,14 @@ func (r *applyingUpdate) Reconcile(ctx context.Context, req cr.Request) (cr.Resu
 	logger.Infof("Updating signaling response to '%s'", signalData.Status.Status)
 	if err := r.client.Update(ctx, signalNodeCopy, &crcli.UpdateOptions{}); err != nil {
 		return cr.Result{Requeue: true}, fmt.Errorf("failed to update signal node to status '%s': %w", signalData.Status.Status, err)
+	}
+
+	// Clean up k0s.tmp after a successful apply. If the file does not exist
+	// (e.g. this is a retry where the file was already removed), ignore the error.
+	if err := os.Remove(updateFilenamePath); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			logger.WithError(err).Warn("Failed to remove update file")
+		}
 	}
 
 	return cr.Result{}, nil
--- a/pkg/autopilot/controller/signal/k0s/apply.go
+++ b/pkg/autopilot/controller/signal/k0s/apply.go
@@ -111,21 +111,43 @@ func (r *applyingUpdate) Reconcile(ctx context.Context, req cr.Request) (cr.Resu
 		return cr.Result{}, nil
 	}
 
+	k0sBinaryFilenamePath := filepath.Join(r.k0sBinaryDir, "k0s")
 	updateFilenamePath := filepath.Join(r.k0sBinaryDir, apconst.K0sTempFilename)
+	updateLinkFilenamePath := filepath.Join(r.k0sBinaryDir, apconst.K0sTempLinkFilename)
 
-	// Ensure that the expected file exists
-	if _, err := os.Stat(updateFilenamePath); errors.Is(err, os.ErrNotExist) {
-		return cr.Result{}, fmt.Errorf("unable to find update file '%s': %w", apconst.K0sTempFilename, err)
-	}
+	// Check if the update file still exists. If it doesn't, the binary was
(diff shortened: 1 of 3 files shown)


bainaichang:
Thanks for taking a look! The 6 failing checks appear to be pre-existing
flaky tests unrelated to this change (calico-ipv6, kuberouter-ipv6, etc.).
Let me know if any changes are needed.

twz123:
Thanks for having a look! I guess there's a fundamental logic error in your current version (and I am a bit appalled that none of the Autopilot integration tests failed). The target k0s executable _always_ exists, so renaming will never happen and the update is never applied ...

What about the following idea: Using hard links to keep the executable around, even after replacing the target k0s path. In the apply phase, we could:

1. Create a hard link from `k0s.tmp` to `k0s.new`, overwriting `k0s.new` if that already exists (or removing it first, I think there's no way for link to overwrite the target path).
2. Do the rename from `k0s.new` to `k0s`.
3. Update the resource status.
4. Remove `k0s.tmp`.


A note on the integration tests not failing: The autopilot upgrade tests from v1.35.4 to this PR's version start the cluster with the old executable so this PR's update logic is never executed, and the other self-update integration tests can't detect this, since the old and the new executable are the same anyways.

pkg/autopilot/controller/signal/k0s/apply.go
```diff
@@ -108,21 +108,41 @@ func (r *applyingUpdate) Reconcile(ctx context.Context, req cr.Request) (cr.Resu
...
-	// Perform the update atomically
-	if err := os.Rename(updateFilenamePath, filepath.Join(r.k0sBinaryDir, "k0s")); err != nil {
-		return cr.Result{}, fmt.Errorf("unable to update (rename) to the new file: %w", err)
+	// Check if the target binary has already been applied. This can happen if
+	// a previous reconciler run completed the file rename but then failed during
+	// the subsequent client.Update call (e.g. due to a resourceVersion conflict).
+	// By checking the target first, we make the reconciler idempotent.
+	k0sBinaryFilename := filepath.Join(r.k0sBinaryDir, "k0s")
+
+	if _, err := os.Stat(k0sBinaryFilename); err != nil && !errors.Is(err, os.ErrNotExist) {
```
makhov: But won't it detect the old k0s binary as well as existing and do nothing? 

Also I think this form is easier to read:
```
if _, err := os.Stat(k0sBinaryFilename); err != nil {
     if !errors.Is(err, os.ErrNotExist) {
          return cr.Result{}, fmt.Errorf("unable to stat k0s binary '%s': %w", k0sBinaryFilename, err)
     }
     // Target binary not in place — proceed with the normal update flow
     ...
} else {
}
```

banjoh: I had pre-empted earlier in https://github.com/k0sproject/k0s/issues/7703#issuecomment-4582937823.

Won't it be better to compare the binary with the expected one using the checksum? The info is available in the signal annotation already

```
...v1.34.7+k0s.0-amd64","version":"v1.34.7+k0s.0","sha256":"f9e1335e2c4cc6e1cea3970d38bd5282d6382f4aea5050924ff50c520194619f"}...
```

twz123: I'd rather avoid trying to figure out if the replace already happened. I'd rather simply run the replacement again, unconditionally. My favorite is currently the hard link approach. No I/O, just a single syscall.

bainaichang:
Thank you for catching this bug. I was indeed mistaken. I think it's definitely a good idea to go with the solution you proposed. I'll update the code accordingly. Thank you!

bainaichang:
Oh! I pushed a bit too many times; I didn’t expect the PR to update automatically!

banjoh:
@twz123 is there anything pending here, or are you able to give it another review?

banjoh:
@bainaichang it looks like the PR has some conflicts. Are you able to address them?

blorenz-ra:
@bainaichang we are blocked by this issue, any estimation when you can resolve the conflicts and merge the fix?

banjoh:
@bainaichang when you get a chance, could you please address conflicts so maintainers can review?

bainaichang:
Got it. I’ll resolve the conflicts and push the updates shortly.

jnummelin:
@bainaichang Linter has some nags, could you resolve those too 🙏 

Onco those are green, I can push through CI to get it green and then approve

bainaichang:
I will fix it. Thank you for the feedback. I’ll commit and push the change so CI can rerun.

jnummelin:
Now the DCO is not happy, you need to sign-off the commit(s)

bainaichang:
Thanks! The DCO check has been fixed. Please let me know if anything else is needed.

jnummelin (approved):

PR #7713, opened 2026-06-02
containerd 2.x: use native config merging instead of custom CRI merge logic

jnummelin:
## Description

Containerd 2.x natively deep-merges plugin sections from imported config files, making k0s's manual RFC 7396 merge patch redundant. Replace the custom merge pipeline with a single glob import in the main config:

  imports = ["/etc/k0s/containerd.d/*.toml"]

K0s CRI defaults (sandbox image, Windows CNI paths) are now written inline in the main config so containerd merges user drop-ins on top natively. The fsnotify watcher is retained to trigger a SIGHUP whenever the drop-in directory changes, preserving the dynamic reconfiguration behaviour.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added
- [x] Existing cases cover the functional changes, the existing smoke for the containerd drop-ins preserved as-is and passes.

## Checklist


pkg/component/worker/containerd/configurer.go
```diff
@@ -121,18 +32,15 @@ func generateDefaultCRIConfig(sandboxContainerImage string) ([]byte, error) {
...
 		},
 	})
 }
 
-func hasCRIPluginConfig(tree *toml.Tree) bool {
-	return tree.HasPath([]string{"plugins", "io.containerd.cri.v1.runtime"}) || tree.HasPath([]string{"plugins", "io.containerd.cri.v1.images"})
-}
-
-// Checks if the given config data is valid and compatible with the k0s-managed
+// ValidateConfigFile checks if the given config data is valid and compatible with the k0s-managed
```
twz123, line 43: Why are LLMs _always_ doing this? Changing random godoc without having any business to do so just because there _used to be_ this old convention 🤷

twz123:
We also need to update the documentation. It still contains quite some infos about the old "special treatment" that we're able to remove now.

pkg/component/worker/containerd/configurer.go
jnummelin: True. But TBH I should've also spotted that change and not committed it 🤦

jnummelin:
> We also need to update the documentation.

Another 🤦 . I had it done locally but forgot to `git add`...

twz123 (approved):
We don't have an integration test for this, do we? I wonder how that might look like 🤔

jnummelin:
> We don't have an integration test for this, do we?

We do, in `inttest/containerdimports/containerd_imports_test.go`. That drops in a faked new CRI runtime in `/etc/k0s/containerd.d/foo.toml` and verifies we can actually use it properly, thus making sure containerd has been properly reconfigured.

And as you can see, I did NOT modify that test at all, it passes as-is.

PR #7727, opened 2026-06-03
Resource patches

### LINKED ISSUE #1349: Allow setting resources for kubernetes components via config
We don't set resources for kube-proxy, kube-router, metrics-server and also don't have a way to customize them with k0s config.

Add resources section to the k0s config and allow users to customize them if needed. Probably worth adding reasonable default values.

**Additional context**
As a side-effect YAML linter returns some errors because of that

```
❌  Ensure each container has a configured memory request  [1 occurrence]
    — metadata.name: kube-proxy (kind: DaemonSet)
💡  Missing property object `requests.memory` - value should be within the accepted boundaries recommended by the organization

❌  Ensure each container has a configured CPU request  [1 occurrence]
    — metadata.name: kube-proxy (kind: DaemonSet)
💡  Missing property object `requests.cpu` - value should be within the accepted boundaries recommended by the organization

❌  Ensure each container has a configured memory limit  [1 occurrence]
    — metadata.name: kube-proxy (kind: DaemonSet)
💡  Missing property object `limits.memory` - value should be within the accepted boundaries recommended by the organization

❌  Ensure each container has a configured CPU limit  [1 occurrence]
    — metadata.name: kube-proxy (kind: DaemonSet)
💡  Missing property object `limits.cpu` - value should be within the accepted boundaries recommended by the organization
```

https://github.com/k0sproject/k0s/issues/1230

### LINKED ISSUE #3246: Enable custom nodeSelector for built-in components

nodeSelector is hardcoded for most components (think metrics-server for ex).
It does not enable to customize the node (or pool) you want it to be deployed to.

Have a variable in k0s.yml/k0sctl.yaml to do it.

Manual edit but it is not very robust.

### Additional context

_No response_

### LINKED ISSUE #7651: Make preferred scheduling of k0s system components (Deployments) on control plane nodes

The intention is to make worker nodes free and predictable for planning and scheduling customer workloads when the cluster has control plane nodes bootstrapped as workers

Suggested solutions is to add the `preferredDuringSchedulingIgnoredDuringExecution` affinity to all the default k0s components that are deployed as Deployments and don't have the strict requirement to be deployed on all nodes in the cluster. This should schedule pods to control plane nodes by default with the safe fallback if CP nodes don't have kubelet or required resources.

```
spec:
      tolerations:
        - key: "node-role.kubernetes.io/control-plane"
          operator: "Exists"
          effect: "NoSchedule"

      affinity:
        nodeAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 100
              preference:
                matchExpressions:
                  - key: node-role.kubernetes.io/control-plane
                    operator: Exists
```

_No response_

### Additional context

We plan to pin all the system component in MKE product to CP nodes and this configuration should cover our requirement

makhov:
## Description

Introduce patches for k0s generated resources to k0s config. We [use](https://docs.k0smotron.io/head/advanced/components/) similar approach in k0smotron.

Some things worth mentioning:
- Patches are applied in TemplateWriter to keep the resulting yamls on disk in manifest folders. This allows user to see the result yaml and debug patches easily.
- The new config field is just a single array with selector and patches, not split by component or anything else. It makes support easier and it will work for any new component we may introduce in the future.

Fixes #1349
Fixes #3246
Fixes #7651

Example: 

```yaml
spec:
  network:
    coreDNS:
      patches:
        - target:
            kind: Deployment
            name: coredns
            namespace: kube-system
          patch:
            type: strategic
            content: |
              spec:
                replicas: 3
                template:
                  spec:
                    containers:
                      - name: coredns
                        resources:
                          limits:
                            memory: 256Mi
```

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


jnummelin:
```
spec:
  patches:
  ...
```

I think this is maybe not 100% obvious for _what_ these patches are. An alternative could be something like `spec.generatedResourcePatches` to make it slightly more obvious this is patching all the generated resources.

Also thinking if the top-level placement is optimal. We could hang it under `spec.extensions` too. TBH, not sure if that's any better though 😆

jnummelin:
@byDimasik As a potential user for this stuff, your view on the UX of the API would be highly appreciated

makhov:
I'm not sure `spec.extensions` is the right place for it, for me "extensions" are about adding something new, not tuning existing.

Regarding the naming, I agree but there is no option that I like more yet, also keeping it broad allows us to extend it if we need to in the future

byDimasik:
> @byDimasik As a potential user for this stuff, your view on the UX of the API would be highly appreciated

Since we're most likely staying on 1.35 for the next year or so, and you don't backport API changes, high chance I won't consume these changes for a good while. I wonder, though, if while working on this, you saw a chance of making this work in 1.35 with some kind of workaround or clever trick? Otherwise, we may have BYO metrics-server.

Besides that, from the code, it seems that any patch change will trigger template rewrite for all components. 

```go
    patches := m.clusterConfig.Spec.Patches
    if previousConfig == newConfig && reflect.DeepEqual(previousPatches, patches) {
	    continue
    }
```

Thus, if I change a patch intended for coredns, it will make metrics-server, kube-proxy, etc, rewrite their template files. Does this mean all of these components will also be reconciled? Or if the template is rewritten identically to the previous template, nothing will happen?

I do not have a strong preference for the placement of these patches in the `.spec`. In my scenarios, the `.spec` generation is automated, so I can easily add it wherever. One comment kinda related to my previous one is that it may be easier to maintain patches for different components independently. E.g. I may wanna have some logic like `if patch was not applied or outdated then apply it now`, and if the patches for all components are one big list as it is now, it may not be easy to find the exact patch in that list. Contrastly, if the patches were separated in the API into the sections where they belong, it could be easier to understand what applies to what, e.g.

```yaml
spec:
  network:
    kubeProxy:
      patches:
  coredns:
    patches:
```

but I totally understand how much of the dev overhead it brings, so maybe the current implementation makes it more robust.

makhov:
```
spec:
  network:
    kubeProxy:
      patches:
  coredns:
    patches:
```

This exactly what I want to avoid. Not every internal resource has it's own config section, and in general, we can add some other resources at some point, so with this approach we will have to change config structure. I personally prefer to make it intentionally broad to give users more freedom to tune whatever they want.

apedriza:
what about

```yaml
spec:
  components:
    patches:
      . . .
```
Future resources don't change the API. Nesting it also gives more context + allows more actions over components in the future if needed. So a bit more flexible

twz123:
> ```
> spec:
>   network:
>     kubeProxy:
>       patches:
>   coredns:
>     patches:
> ```
>
> This exactly what I want to avoid. Not every internal resource has it's own config section, and in general, we can add some other resources at some point, so with this approach we will have to change config structure. I personally prefer to make it intentionally broad to give users more freedom to tune whatever they want.

I understand that keeping it broad looks promising at first, especially to avoid adding a config section for each component, and being able to sneak in support for other components in patch releases. Personally I still prefer the "explicit section per component" approach. In my experience, grouping things functionally instead of technically pays off in the long run.

In k0s, we have the component abstraction, and all the k0s-managed resources belong exactly to one of them, so grouping them with the rest of that component's config makes the API easier to reason about. For example, if I want to copy, remove, or review all kube-router related config settings, it is very helpful to have everything under `spec.network.kuberouter` instead of having to remember that there might still be patches lying around in an otherwise unrelated configuration section.

A global `spec.patches` also feels like a fairly broad promise. Users will probably simply assume that every single k0s-managed resource is a legitimate patch target. Currently, the docs need to explain which components are actually patch-aware, and users have to know this and look it up. We know how this will pan out :sweat_smile:

This has implications for the implementation, too. With a single global set of patches, every component needs to consider the entire list, even if it might very well not be concerned with any of these in the first place. We'd need to reconcile all components whenever there's a change to an arbitrary patch. Moreover, it becomes much more difficult to issue warnings about patches for resources that k0s doesn't know about, since a single component only knows about its own resources, and has no idea what other components are reconciling and whether an unknown resource legitimately belongs to another component.

Finally, the lack of a separate section for each k0s component shouldn't prevent us from adding them. The fact that some components don't have config knobs doesn't mean that they have to stay that way. At a minimum, I could at least think of having an enable/disable field for each of them, as an alternative (or potential successor) to the `--disable-components` flag. That would actually feel similar to the idiomatic way Helm chart values are being structured.

makhov:
> grouping things functionally instead of technically pays off in the long run.

Everything under `spec.network.kuberouter` is semantic configuration, knobs k0s understands and translates into behavior. A patch is the opposite - it's a mechanical override of a rendered Kubernetes yaml, not component concepts. Grouping it with kuberouter's real knobs actually mixes two abstraction levels.

Also, manual overrides are exactly what you usually want to see all in one place.

I'm not sure that which one easier to use:

<details>
<summary>Per component</summary>

```
spec:
  network:
    kubeProxy:
      patches:
      - target:
          kind: Daemonset
          name: kubeproxy
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                   - name: kubeproxy
                    resources:
                      limits:
                        memory: 256Mi
    nodeLocalLoadBalancing:
      envoyProxy:
        patches: 
  coredns:
    patches:
      - target:
          kind: Deployment
          name: coredns
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                   - name: coredns
                    resources:
                      limits:
                        memory: 256Mi
  metricsserver:
    patches:
     - target:
          kind: Deployment
          name: metricsserver
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                    - name: metricsserver
                      resources:
                        limits:
                          memory: 256Mi
```
</details>

<details>
<summary>Plain array</summary>

```
spec:
   patches:
      - target:
          kind: Daemonset
          name: kubeproxy
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                   - name: kubeproxy
                    resources:
                      limits:
                        memory: 256Mi
      - target:
          kind: Deployment
          name: coredns
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                   - name: coredns
                    resources:
                      limits:
                        memory: 256Mi
     - target:
          kind: Deployment
          name: metricsserver
        patch:
          type: strategic
          content: |
            spec:
              replicas: 3
              template:
                spec:
                  containers:
                    - name: metricsserver
                      resources:
                        limits:
                          memory: 256Mi
```
</details>

twz123:
> Everything under `spec.network.kuberouter` is semantic configuration, knobs k0s understands and translates into behavior. A patch is the opposite - it's a mechanical override of a rendered Kubernetes yaml, not component concepts. Grouping it with kuberouter's real knobs actually mixes two abstraction levels.

I kinda disagree that patches are not a component concept. The components are dealing with their own set of resources, and they are responsible to actually apply the patches. Patches wouldn't make sense without components that own and generate resources. Yes, patches are on a different _technical_ abstraction level compared to other knobs, but the same could very well also be said of `rawArgs`. One could argue that this is a "mechanical override of a rendered command line", in a way. Yet, we still declare them on a per-component level.

makhov:
I rewrote it to per-component configuration

internal/pkg/patches/patches.go
```diff
@@ -0,0 +1,121 @@
...
+			current, err = applyJSONPatch(current, patchJSON)
+		case v1beta1.MergePatchType:
+			current, err = jsonpatch.MergePatch(current, patchJSON)
+		case v1beta1.StrategicMergePatchType:
+			current, err = applyStrategicMerge(obj.GroupVersionKind(), current, patchJSON)
+		default:
+			err = fmt.Errorf("unknown patch type %q", p.Patch.Type)
+		}
+		if err != nil {
+			return nil, fmt.Errorf("patches[%d] (%s/%s): %w", i, kind, name, err)
```
jnummelin: The error message could be bit clearer

pkg/component/controller/calico.go
```diff
@@ -113,6 +119,7 @@ func NewCalico(nodeConfig *v1beta1.ClusterConfig, manifestsDir string, hasWindow
 		primaryAddressFamily: nodeConfig.Spec.PrimaryAddressFamily(),
 		manifestsDir:         manifestsDir,
 		hasWindowsNodes:      hasWindowsNodes,
+		patches:              patches,
```
jnummelin: Why do we add patches only at construction time?  We don't _seem_ to get them while reconciling dynamically (IF I'm not missing anything), is this intentional?

pkg/apis/k0s/v1beta1/calico.go
```diff
@@ -47,6 +47,11 @@ type Calico struct {
 	// The virtual network ID for VXLAN (default: 4096)
 	// +kubebuilder:default=4096
 	VxlanVNI int `json:"vxlanVNI,omitempty"`
+
+	// Patches holds customizations applied to the Calico resources generated by
+	// k0s before they are written and applied.
```
twz123: I think "written" is an implementation detail. Same for the other `Patches` fields.

```suggestion
	// k0s before they are applied.
```

pkg/apis/k0s/v1beta1/network.go
```diff
@@ -181,6 +184,19 @@ func (n *Network) Validate() []error {
 		errors = append(errors, err)
 	}
 
+	if n.KubeProxy != nil {
+		errors = append(errors, n.KubeProxy.Patches.validate(field.NewPath("kubeProxy", "patches"))...)
```
twz123: Doesn't this belong to `KubeProxy.Validate()`? Same for the other patch validations that follow. I'd rather expect the Validate methods of the individual structs to cover this.

jnummelin:
Many of the tests might benefit from using `unstructured.NestedXYZ` helpers to dig out the expected patched values

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	Name string `json:"name"`
+	// Namespace optionally narrows the match to a namespace.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// PatchSpec is the patch type and body.
+type PatchSpec struct {
+	// Type is the patch type to apply.
+	// +kubebuilder:validation:Enum=json;strategic;merge
```
twz123: Wouldn't "JSON", "StrategicMergePatch" and "MergePatch" be more typical wordings for enum constants in a Kubernetes resource?

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	// Content is the patch body (JSON or YAML; YAML is converted to JSON).
+	// +kubebuilder:validation:Required
+	Content string `json:"content"`
+}
+
+// PatchType is the supported patch encoding.
+type PatchType string
+
+const (
+	// JSONPatchType is RFC 6902 JSON Patch (array of operations).
```
twz123:
```suggestion
	// JSONPatchType is an RFC 6902 JSON Patch (array of operations).
```

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	JSONPatchType PatchType = "json"
+	// StrategicMergePatchType is Kubernetes strategic merge patch.
```
twz123:
```suggestion
	// StrategicMergePatchType is a Kubernetes Strategic Merge Patch.
```

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	StrategicMergePatchType PatchType = "strategic"
+	// MergePatchType is RFC 7386 JSON Merge Patch.
```
twz123:
```suggestion
	// MergePatchType is an RFC 7386 JSON Merge Patch.
```

pkg/apis/k0s/v1beta1/calico.go
```diff
@@ -47,6 +47,11 @@ type Calico struct {
...
+	// +optional
+	Patches Patches `json:"patches,omitempty"`
```
twz123, line 53: I wonder if a better name would be `resourcePatches` 🤷

pkg/apis/k0s/v1beta1/clusterconfig_types.go
```diff
@@ -468,6 +439,10 @@ func (s *ClusterSpec) Validate() (errs []error) {
 		}
 	}
 
+	if s.MetricsServer != nil {
+		errs = append(errs, s.MetricsServer.Patches.validate(field.NewPath("metricsServer", "patches"))...)
+	}
```
twz123: I'd rather have expected something like that:

```suggestion
	errs = append(errs, s.MetricsServer.Validate(field.NewPath("metricsServer"))...)
```

(The Validate method could be aware of nil receivers.)

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	MergePatchType PatchType = "merge"
+)
+
+// Validate checks every patch for a known type and a non-empty target.
+func (p Patches) Validate() []error {
+	return p.validate(field.NewPath("patches"))
```
twz123, line 66: Hmm, why should the field name always be "patches"? This is something that the caller would know. I'd rather make the `validate` method below public and remove this one.

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// validate checks every patch for a known type and a non-empty target,
+// reporting errors relative to the given field path.
+func (p Patches) validate(path *field.Path) (errs []error) {
```
twz123: I'd rather implement it like this:

```go
	var errors []error
	for i, patch := range p {
		if errs := patch.Validate(path.Index(i)); err != nil {
			errors = append(errors, errs...)
		}
	}
	return errors
```

and move the loop body into a `Patch.Validate(path *field.Path) (errs []error)` method.

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+	for i, patch := range p {
+		item := path.Index(i)
+		switch patch.Patch.Type {
+		case JSONPatchType, StrategicMergePatchType, MergePatchType:
+			// valid
+		default:
+			errs = append(errs, fmt.Errorf("%s: invalid type %q: must be one of json, strategic, merge", item.Child("patch", "type"), patch.Patch.Type))
```
twz123: For enums, there's usually [`field.NotSupported(...)`](https://pkg.go.dev/k8s.io/apimachinery@v0.36.0/pkg/util/validation/field#NotSupported).

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		if patch.Target.Kind == "" {
+			errs = append(errs, fmt.Errorf("%s is required", item.Child("target", "kind")))
```
twz123: This one and the next error could be [`field.Required(...)`](https://pkg.go.dev/k8s.io/apimachinery@v0.36.0/pkg/util/validation/field#Required).

pkg/component/controller/calico.go
```diff
@@ -44,8 +44,9 @@ type Calico struct {
 	manifestsDir         string
 	hasWindowsNodes      func() (*bool, <-chan struct{})
 
-	config value.Latest[*calicoClusterConfig]
-	stop   func()
+	config  value.Latest[*calicoClusterConfig]
+	patches v1beta1.Patches
```
twz123, line 57: This should probably go into the `calicoClusterConfig` struct.

pkg/component/controller/coredns.go
```diff
@@ -276,6 +276,7 @@ type CoreDNS struct {
 	log                    *logrus.Entry
 	manifestDir            string
 	previousConfig         coreDNSConfig
+	previousPatches        v1beta1.Patches
```
twz123, line 279: I'd put this into the `coreDNSConfig` struct.

pkg/apis/k0s/v1beta1/patches.go
```diff
@@ -0,0 +1,88 @@
...
+
+// PatchSpec is the patch type and body.
+type PatchSpec struct {
+	// Type is the patch type to apply.
+	// +kubebuilder:validation:Enum=json;strategic;merge
+	// +kubebuilder:validation:Required
+	Type PatchType `json:"type"`
+	// Content is the patch body (JSON or YAML; YAML is converted to JSON).
+	// +kubebuilder:validation:Required
+	Content string `json:"content"`
```
twz123, line 49: I'm wondering: If we only support JSON/YAML style content anyways, why don't we declare this as `map[string]any` directly, so we don't have to deal with the validation and conversion ourselves?

docs/configuration.md
```diff
@@ -606,6 +606,65 @@ spec:
     enabled: true
 ```
 
+### Component patches
```
twz123, line 609: Would this be a better fit for one level further up?

```suggestion
## Component patches
```

docs/configuration.md
```diff
@@ -606,6 +606,65 @@ spec:
...
+
+!!! warning "Experimental feature"
+     Patching is an experimental feature, which means the behavior of this feature might be unstable and may change without a major version bump.
+
+Several k0s-managed components let you customize the Kubernetes resources they
```
twz123, line 614: What about mentioning the patches field in each component's section, and then add a hyperlink to this section? That way, we wouldn't need to list the components here.

We could then rephrase this paragraph maybe like this:

> Several k0s components allow you to customize the Kubernetes resources they generate before those resources are applied. This is an escape hatch for making adjustments that k0s does not offer as dedicated configuration options. Components that support resource patches typically have a field called `patches`.
> 
> Each patch selects a target resource by `kind` and `name` (optionally narrowed by `namespace`) and provides the patch `type` and its `content`:
>

twz123 (approved):

PR #7738, opened 2026-06-03
Strip default images from the config with non-default repo

### LINKED ISSUE #7708: Dynamic config materializes and pins bundled image versions when a custom `spec.images.repository` is set, blocking updates across k0s binary upgrades
# Dynamic config materializes and pins bundled image versions when a custom `spec.images.repository` is set, blocking updates across k0s binary upgrades

## Summary

When dynamic config is enabled and only `spec.images.repository` is set in `k0s.yaml`, k0s materializes the full set of bundled component image definitions (CoreDNS, Calico, etc.) - including explicit `version` tags - into the persisted `clusterconfig/k0s`. After a k0s binary upgrade, these pinned versions remain unchanged and no longer follow the new k0s defaults.

This creates mixed-version upgrade hazards and, in at least one case, a hard incompatibility. It also forces operators to manually strip `spec.images` during upgrades or manually bump versions under the `images` section, with no documented guidance on when or how to do so.

## Environment

- Cluster initially bootstrapped with k0s **v1.33.9**
- Dynamic config **enabled**
- Custom (non-default) image repository

## Steps to reproduce

1. Bootstrap a cluster with dynamic config enabled and a `k0s.yaml` that sets only the custom repository:

   ```yaml
   images:
     repository: registry.example.com/lcm
   ```

2. Inspect the persisted ClusterConfig:

   ```sh
   kubectl get clusterconfig -n kube-system k0s -o yaml
   ```

3. Observe that explicit image versions have been materialized into the config:

   ```yaml
   images:
     calico:
       cni:
         image: registry.example.com/lcm/k0sproject/calico-cni
         version: v3.29.7-0
       kubecontrollers:
         image: registry.example.com/lcm/k0sproject/calico-kube-controllers
         version: v3.29.7-0
       node:
         image: registry.example.com/lcm/k0sproject/calico-node
         version: v3.29.7-0
     coredns:
       image: registry.example.com/lcm/k0sproject/coredns
       version: 1.12.4
     # ...
   ```

4. Upgrade the k0s binary on all controllers.
5. Observe that the old image tags persist in `clusterconfig/k0s`.

**Note on when this happens:** The materialization appears specifically during the **initial cluster bootstrap**, when the dynamic ClusterConfig is first written to the API. It does **not** reproduce by editing an already-existing ClusterConfig. I suspect it may also re-occur after a k0s service restart / reconciliation cycle, but I have not yet fully confirmed that part.

## Current behavior

Old image tags persist indefinitely. Bundled component images stay pinned to the versions that were current at bootstrap time, even after the k0s binary is upgraded. Materialization happens only for non-default repositories - with the default `quay.io/k0sproject` the defaults are stripped away.

## Upgrade procedure (for reference)

Our rolling control-plane upgrade is:

1. cordon/drain the first control-plane node;
2. stop the k0s service, replace the binary, start the k0s service;
3. repeat for the remaining control-plane nodes.

During steps 2–3 the control plane runs mixed k0s binary versions and leadership can move between nodes, so it is not obvious which binary's defaults are "in effect" at any given point - which is the core of the "during the upgrade" ambiguity below.

## Why this is still a problem

After a binary upgrade there are effectively **two separate reconciliation domains** that can drift apart - the image section in ClusterConfig, and the manifests applied by the current leader - and there is no defined, safe moment to reconcile the image tags. We hit a hard incompatibility where the CoreDNS version bundled with k0s 1.33 is incompatible with the 1.34 manifests / k8s version and crash-loops continuously until the version is manually bumped.

In practice, every available moment to manually bump or strip the images has its own problem:

- **Before the control plane upgrade.** Issue: it is unclear whether the new images are compatible with the still-running older kubernetes / k0s version.
- **During the control plane upgrade.** Issue: the exact moment is unclear. I don't know the k0s internals well, but I suspect only the leader applies the manifests for the new bundled components, and there is no obvious way to synchronize the image update with k0s's own manifest application.
- **After the control plane upgrade.** Issue: the new manifests run against the old images until they are bumped, producing exactly the incompatibility we already observed (the CoreDNS crash-loop above).

## Desired behavior

Either of:

- only `repository` persists in the dynamic config (defaults are not materialized), **or**
- defaults re-materialize for the new binary version on upgrade.

## Requests

**At minimum (documentation):**

A documented procedure for operators using a custom registry: how, and at what point during a rolling k0s upgrade, the bundled component image tags should be updated to match the new binary's defaults - so that mixed-version states like the CoreDNS crash-loop above can be avoided reliably rather than discovered in production.

**Highly desirable (feature):**

An opt-in "unpinned" / "follow-defaults" mode for custom registries, where setting `spec.images.repository` overrides **only** the repository while component image names and versions continue to track the running k0s binary's defaults. This would let operators mirror images into a private registry - a very common air-gapped / enterprise requirement - without inheriting indefinitely pinned versions, and without having to hand-edit the `images` section on every upgrade.

makhov:
## Description

Fixes #7708

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/apis/k0s/v1beta1/images_test.go
+++ b/pkg/apis/k0s/v1beta1/images_test.go
@@ -86,6 +86,32 @@ spec:
 	}
 }
 
+func TestStripDefaultsForRepositoryOnlyImages(t *testing.T) {
+	yamlData := []byte(`
+apiVersion: k0s.k0sproject.io/v1beta1
+kind: ClusterConfig
+spec:
+  images:
+    repository: registry.acme.corp
+`)
+
+	cfg, err := ConfigFromBytes(yamlData)
+	require.NoError(t, err)
+
+	persisted := cfg.GetClusterWideConfig().StripDefaults().CRValidator()
+
+	require.NotNil(t, persisted.Spec.Images)
+	require.Equal(t, "registry.acme.corp", persisted.Spec.Images.Repository)
+	require.Nil(t, persisted.Spec.Images.Konnectivity)
+	require.Nil(t, persisted.Spec.Images.PushGateway)
+	require.Nil(t, persisted.Spec.Images.MetricsServer)
+	require.Nil(t, persisted.Spec.Images.KubeProxy)
+	require.Nil(t, persisted.Spec.Images.CoreDNS)
+	require.Nil(t, persisted.Spec.Images.Pause)
+	require.Nil(t, persisted.Spec.Images.Calico)
+	require.Nil(t, persisted.Spec.Images.KubeRouter)
+}
+
 func TestImagesRepoOverrideInConfiguration(t *testing.T) {
 	t.Run("if_has_repository_not_empty_add_prefix_to_all_images", func(t *testing.T) {
 		t.Run("default_config", func(t *testing.T) {
--- a/pkg/apis/k0s/v1beta1/images_test.go
+++ b/pkg/apis/k0s/v1beta1/images_test.go
@@ -150,6 +150,48 @@ func TestImagesRepoOverrideInConfiguration(t *testing.T) {
 			require.Equal(t, "my.repo/k0sproject/pushgateway-ttl:"+constant.PushGatewayImageVersion, testingConfig.Spec.Images.PushGateway.URI())
 		})
 	})
+	// NLLB Envoy/Traefik images are rewritten with spec.images.repository, so
+	// they must be stripped when they only differ from the defaults by that
+	// repository.
+	t.Run("strips_default_nllb_images_with_repository", func(t *testing.T) {
+		for _, tt := range []struct {
+			nllbType NllbType
+			verify   func(t *testing.T, nllb *NodeLocalLoadBalancing)
+		}{
+			{NllbTypeEnvoyProxy, func(t *testing.T, nllb *NodeLocalLoadBalancing) {
+				require.NotNil(t, nllb.EnvoyProxy)
+				require.Nil(t, nllb.EnvoyProxy.Image, "default Envoy image rewritten with a custom repository should be stripped")
+				require.Nil(t, nllb.Traefik)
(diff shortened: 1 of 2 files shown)


jnummelin:
Envoy image gets still written as full image:
```
    nodeLocalLoadBalancing:
      enabled: false
      envoyProxy:
        apiServerBindPort: 7443
        image:
          image: foo.bar/k0sproject/k0sproject/envoy-distroless
          version: v1.37.2
```

pkg/apis/k0s/v1beta1/images_test.go
```diff
@@ -86,6 +86,32 @@ spec:
...
+	require.NotNil(t, persisted.Spec.Images)
+	require.Equal(t, "registry.acme.corp", persisted.Spec.Images.Repository)
+	require.Nil(t, persisted.Spec.Images.Konnectivity)
+	require.Nil(t, persisted.Spec.Images.PushGateway)
+	require.Nil(t, persisted.Spec.Images.MetricsServer)
+	require.Nil(t, persisted.Spec.Images.KubeProxy)
+	require.Nil(t, persisted.Spec.Images.CoreDNS)
+	require.Nil(t, persisted.Spec.Images.Pause)
+	require.Nil(t, persisted.Spec.Images.Calico)
+	require.Nil(t, persisted.Spec.Images.KubeRouter)
```
jnummelin, line 112: IMO we should add a check that the NLLB image is also nullified correctly.

jnummelin (approved):
We need to rework the whole config and images handling quite a bit anyways in near-ish (hopefully) future. 

This solution works as an intermediate fix for the pressing issue, so LGTM!

jnummelin:
oh dear, manual backporting it is then...

PR #7783, opened 2026-06-11
feat: Allow configuring priorityClassName and terminationGracePeriodS…

### LINKED ISSUE #7782: Configure NLLB Envoy Pod Settings - Proposal for Enhancing Envoy Pod Configurability

The node-local load balancing (NLLB) Envoy Pod is a static pod generated entirely in k0s code (`makePodManifest()` in `pkg/component/worker/nllb/envoy.go`) and served to the kubelet over the internal static-pod HTTP server (staticPodURL). Its `EnvoyProxy` config (`pkg/apis/k0s/v1beta1/nllb.go`) currently only exposes `image`, `imagePullPolicy`, `apiServerBindPort`, and `konnectivityServerBindPort`.

Because of this limitation, two operationally critical Pod settings cannot be configured:

- No `priorityClassName`: The Envoy Pod runs at priority 0. Since NLLB is the node's load-balanced path to the control plane (`kube-apiserver` and `konnectivity`), its failure causes the node to lose its API path and become unstable. Under node-pressure, the kubelet's eviction manager prioritizes killing low-priority pods first, making this critical infrastructure pod a primary candidate for termination exactly when it is most needed.
- No `terminationGracePeriodSeconds`: Envoy is locked to the default 30-second grace period, preventing clean draining of in-flight connections during shutdowns or restarts.

These fields are immutable once a Pod is created, and the manifest is regenerated by k0s on every restart, rendering manual edits or external workarounds like admission webhooks ineffective.

## Concrete failure: graceful node shutdown breaks on NLLB nodes

This configuration gap has direct negative impacts on reliability. When [graceful node shutdown](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) is enabled via a worker profile, the kubelet's shutdown manager separates pods into phases based on priority.

```yaml
spec:
  workerProfiles:
  - name: default
    values:
      shutdownGracePeriod: 30s
```

Pods at or above the system-cluster-critical threshold are terminated last, while all others are terminated first. Because the NLLB Envoy pod has a priority of 0, it is killed immediately in the first phase:

`nodeshutdown_manager.go:153 "Shutdown manager killing pod with gracePeriod" pod="kube-system/nllb-<node>" gracePeriod=15`. Since this pod serves as the worker's path to the API server, killing it early severs the connection for all other pods still attempting to drain. These remaining pods are left in inconsistent states (Running, Pending, or Unknown) as they can no longer report status to the API server. Consequently, graceful node shutdown is effectively broken on any node using NLLB.

Assigning the Envoy pod system-node-critical ensures it is shut down last, maintaining the API path while other workloads finish draining.

Expose two new optional fields on `spec.network.nodeLocalLoadBalancing.envoyProxy`:

- `priorityClassName`: Assign a PriorityClass (e.g., system-node-critical) to protect the Envoy Pod from node-pressure eviction and ensure it is shut down last during graceful node shutdown events.
- `terminationGracePeriodSeconds`: Allow overriding the default termination grace period for cleaner connection draining.

These fields should be plumbed through `envoyPodParams` into `makePodManifest()` and set on the Pod spec. To maintain backward compatibility, both will default to unset.

Recommended default: Set priorityClassName to `system-node-critical` by default. This change is considered a reliability fix rather than just an optimization, as it addresses the malfunction of the graceful node shutdown feature on NLLB nodes.

- Mutating HTTP proxy for `staticPodURL`: Highly fragile due to random upstream ports, tmpfs-based kubelet configs, and startup races.
- `Restart=alway`s systemd unit: Does not address the kill-ordering problem during shutdown and cannot durably set priority on a regenerated manifest.
- Mutating admission webhooks: Ineffective as they only see the mirror pod, not the actual pod run by the kubelet.
- Self-hosting the Envoy static pod: Requires manual reimplementation of k0s's automated Envoy configuration generation.

### Additional context

We do not always have the luxury of draining nodes or performing maintenance under ideal conditions, as our operating environments can be subject to unexpected power loss and abrupt node failures. This change improves resiliency by ensuring critical components are available wherever workloads are scheduled, allowing the system to recover more reliably from real-world disruptions.

@Josh-Tracy

cmwylie19:
## Description

The node-local load balancing (NLLB) Envoy Pod is a static pod generated by k0s in `makePodManifest()`. Its `EnvoyProxy` config only exposes `image`, `imagePullPolicy`, `apiServerBindPort` and `konnectivityServerBindPort`, so there is no way to set a `priorityClassName` or a `terminationGracePeriodSeconds` on the Pod.

This matters in practice. The Envoy Pod runs at priority `0`, yet it *is* the worker's load-balanced path to the control plane. With graceful node shutdown enabled (`shutdownGracePeriod` / `shutdownGracePeriodCriticalPods` via a worker profile), the kubelet shutdown manager kills non-critical pods first and critical pods last. Because the Envoy Pod is priority `0`, it is killed in the first phase, severing the worker's path to the API server (`[::1]:7443`) before the remaining pods can drain or report status:

```plaintext
Failed to update status for pod ...: Patch "https://[::1]:7443/...": unexpected EOF
... dial tcp [::1]:7443: connect: connection refused
```

This change exposes two new optional fields on `spec.network.nodeLocalLoadBalancing.envoyProxy`:

- `priorityClassName` (string) — e.g. `system-node-critical`, so the Pod is protected from node-pressure eviction and shut down last during graceful node shutdown.
- `terminationGracePeriodSeconds` (int64, `>= 0`) — override the default 30s grace period to let Envoy drain in-flight connections.

Both fields are plumbed through `envoyPodParams` into `makePodManifest()` and set on the Pod spec. Both default to unset, so existing behavior is unchanged.

Fixes #7782

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

Manual: configured `envoyProxy.priorityClassName: system-node-critical` and `terminationGracePeriodSeconds` in `k0s.yaml`, restarted the worker, and confirmed the resulting static Pod carries both fields. With `system-node-critical` set, the Envoy Pod is shut down in the critical-pods phase during graceful node shutdown instead of immediately.

Auto: `TestMakePodManifest` asserts the fields propagate to the Pod spec (and are absent by default); `TestEnvoyProxy_PriorityClassAndGracePeriod` covers config parsing and rejection of a negative grace period.

## Checklist


jnummelin:
I wonder if we should rather actually "hardcode" the priority to `system-node-critical` 🤔 

Like you said, having priority at 0 makes no sense as this really is node critical component. I can't think of a reason why anyone would actually have it on any other prio level.

The other thing I'm wondering is if we'd be able to use the recently merged [resource patches](https://github.com/k0sproject/k0s/pull/7727) feature for this. That'd allow users to basically override _anything_ on the generated manifest without us having to maintain bazillion different opts in the config.

cmwylie19:
> I wonder if we should rather actually "hardcode" the priority to `system-node-critical` 🤔
> 
> Like you said, having priority at 0 makes no sense as this really is node critical component. I can't think of a reason why anyone would actually have it on any other prio level.
> 
> The other thing I'm wondering is if we'd be able to use the recently merged [resource patches](https://github.com/k0sproject/k0s/pull/7727) feature for this. That'd allow users to basically override _anything_ on the generated manifest without us having to maintain bazillion different opts in the config.

I couldn't agree more, it definitely needs `system-node-critical`, yes that looks like it may give us what we need in terms of the`terminationGracePeriodSeconds` too.

cmwylie19:
> > I wonder if we should rather actually "hardcode" the priority to `system-node-critical` 🤔
> > Like you said, having priority at 0 makes no sense as this really is node critical component. I can't think of a reason why anyone would actually have it on any other prio level.
> > The other thing I'm wondering is if we'd be able to use the recently merged [resource patches](https://github.com/k0sproject/k0s/pull/7727) feature for this. That'd allow users to basically override _anything_ on the generated manifest without us having to maintain bazillion different opts in the config.
> 
> I couldn't agree more, it definitely needs `system-node-critical`, yes that looks like it may give us what we need in terms of the`terminationGracePeriodSeconds` too.

Just let me know how you would like me to proceed forward, if you want a default priorityClass or anything or if we need to close this. Happy to help!

cmwylie19:
> I wonder if we should rather actually "hardcode" the priority to `system-node-critical` 🤔
> 
> Like you said, having priority at 0 makes no sense as this really is node critical component. I can't think of a reason why anyone would actually have it on any other prio level.
> 
> The other thing I'm wondering is if we'd be able to use the recently merged [resource patches](https://github.com/k0sproject/k0s/pull/7727) feature for this. That'd allow users to basically override _anything_ on the generated manifest without us having to maintain bazillion different opts in the config.

@jnummelin I am not sure if we can use [resource patches](https://github.com/k0sproject/k0s/pull/7727) if I am following correctly based on the fact that the name of the nllb pod is based on the node name and the node-name is going to be random.

```bash
nllb-canes-b29b                   1/1     Running   0               7d19h   192.168.4.67   canes-b29b 
```

```yaml
kind: ClusterConfig
metadata:
  name: k0s
  namespace: kube-system
spec:
  network:
    nodeLocalLoadBalancing:
      envoyProxy:
        patches:
        - target:
            kind: Pod
            name: ???
            namespace: kube-system
          patch:
            type: StrategicMergePatch
            content: |
              spec:
                terminationGracePeriodSeconds: 60
```

cmwylie19:
i think we can close this now. looks like there is a workaround

PR #7841, opened 2026-06-20
fix: avoid recursive API error writes

immanuwell:
## Description

`sendError` called itself again if `ResponseWriter.Write` failed.
A client disconnect can hit that path for real. I reproduced it with a tiny HTTP server: send a request, read the status line, close the socket early, then the server write fails with `connection reset by peer`.

This patch keeps the same status and body on the normal path, but on write failure it just logs and stops. no more looping on the same broken writer.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

`go test ./cmd/api`
`go test $(go list ./... | grep -v '/inttest/')`

Repro:
1. Start a handler that sleeps, then writes a response body.
2. Send a request and close the client connection before the write completes.
3. The server write fails.
4. Before this patch, `sendError` called itself again on that same writer.
5. With this patch, it logs the write failure and bails out. pretty much it.

## Checklist

### DIFF
--- a/cmd/api/util.go
+++ b/cmd/api/util.go
@@ -18,7 +18,7 @@ func sendError(err error, resp http.ResponseWriter, status ...int) {
 	resp.Header().Set("Content-Type", "text/plain")
 	resp.WriteHeader(code)
 	if _, err := resp.Write([]byte(err.Error())); err != nil {
-		sendError(err, resp)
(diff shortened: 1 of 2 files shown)


twz123 (approved):
Thank you!

PR #7862, opened 2026-06-23
Use correct make variable to pass build tags for containerd

jnummelin:
## Description

https://github.com/containerd/containerd/blob/v2.3.2/Makefile#L91-L103

Containerd expect to see GO_BUILDTAGS variable

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist


embedded-bins/containerd/Dockerfile
```diff
@@ -42,7 +42,7 @@ ARG TARGET_OS \
 RUN make \
 	CGO_ENABLED=${BUILD_GO_CGO_ENABLED} \
 	SHIM_CGO_ENABLED=${BUILD_SHIM_GO_CGO_ENABLED} \
-	GO_TAGS="-tags=${BUILD_GO_TAGS}" \
+	GO_BUILDTAGS="${BUILD_GO_TAGS}" \
```
twz123, line 45: I don't see how this would help? AFAICT the upstream Makefile is using GO_BUILDTAGS to populate GO_TAGS, so there shouldn't be a difference?

jnummelin: But the `SHIM_GO_TAGS` is only populated via the `GO_BUILDTAGS`

jnummelin: The other alternative would be to pass in `SHIM_GO_TAGS` in addition but that would then omit some of the "defaulting" which containerd does itself

ncopa: and we also lose the `GO_BUILDTAGS += urfave_cli_no_docs ` that is [added by containerd](https://github.com/containerd/containerd/blob/57489ef654fbba7f2a70eb104d2dcbca8427b3f7/Makefile#L94) when overriding `GO_TAGS`. I think we want keep the flags added by upstream containerd and only append ours?

ncopa (approved):

embedded-bins/containerd/Dockerfile
jnummelin: > I think we want keep the flags added by upstream containerd and only append ours?

That was my thinking here, upstream adds those by "default" for a reason, I think 😄

twz123: This is a bit unintuitive, as we usually want to replace all the build tags, don't we? Otherwise, we should probably try to make use of `STATIC`, as well? Anyhow, I think the shim go tags behaving differently seems to be unintentional. Need to have a look at the containerd git history.

jnummelin: I see two approaches:
- the way I proposed, plus need to add the `STATIC` stuff in. Originally just overlooked that part
- keep as-is in main, but add option to pass in the needed `SHIM_GO_BUILDTAGS`

PR #7894, opened 2026-06-26
Allow set network for all bootloose machines

apedriza:
## Description

We want to switch from footloose to bootloose in k0smtoron but in some tests when running containers as remotemachines, we need to put that remote machine in the same network as the management cluster, which seems to be a configuration lost in bootloose.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/inttest/common/bootloosesuite.go
+++ b/inttest/common/bootloosesuite.go
@@ -1226,6 +1226,7 @@ func (s *BootlooseSuite) initializeBootlooseClusterInDir(dir string) error {
 					Privileged:   true,
 					Volumes:      volumes,
 					PortMappings: portMaps,
+					Networks:     s.Networks,
 				},
 			},
 		},
@@ -1239,6 +1240,7 @@ func (s *BootlooseSuite) initializeBootlooseClusterInDir(dir string) error {
 				Privileged:   true,
 				Volumes:      volumes,
 				PortMappings: portMaps,
+				Networks:     s.Networks,
 			},
 			Count: 1,
 		})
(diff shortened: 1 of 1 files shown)


inttest/common/bootloosesuite.go
```diff
@@ -1226,6 +1226,7 @@ func (s *BootlooseSuite) initializeBootlooseClusterInDir(dir string) error {
 					Privileged:   true,
 					Volumes:      volumes,
 					PortMappings: portMaps,
+					Networks:     s.Networks,
```
twz123, line 1229: Wouldn't it make sense to add this line to _all_ the machines that are spawned? I've seen quite a few where the networks are missing. I think this has been an oversight?

apedriza: > all the machines that are spawned

You mean controllers, worker and k0smotron workers?

Controllers and workers have already set the network from `BootlooseSuite` config, the only one missing is the k0smotron worker machine

twz123: I mean basically every machine that's provisoned by the suite. I don't think it makes sense to keep any of those outside the custom networks. There are quite a few machines that don't have the networks set yet, if I'm not mistaken, e.g. the LB, the Docker registry, and so on...

apedriza: Alright, forgot about those. Changed to allow setting all of them

twz123 (approved):

PR #7895, opened 2026-06-26
Add patching for NLLB

### LINKED ISSUE #7782: Configure NLLB Envoy Pod Settings - Proposal for Enhancing Envoy Pod Configurability

jnummelin:
## Description

Add patching for NLLB via the generic patching mechanism.

Fixes #7782

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


pkg/apis/k0s/v1beta1/nllb.go
```diff
@@ -40,6 +40,10 @@ type NodeLocalLoadBalancing struct {
 	// traefik contains configuration options related to the "Traefik"
 	// type of load balancing.
 	Traefik *Traefik `json:"traefik,omitempty"`
+
+	// patches holds customizations applied to the NLLB resources generated by k0s before they are applied.
+	// +optional
+	Patches Patches `json:"patches,omitempty"`
```
twz123: I'd rather place this into the Envoy and Traefik structs. Maybe there's one day an NLLB implementation that doesn't require any Kubernetes resources at all. If I think about the userspace CPLB proxy, it feels like it might be another valid NLLB mode, as well.

pkg/apis/k0s/v1beta1/nllb.go
```diff
@@ -108,6 +112,11 @@ func (n *NodeLocalLoadBalancing) Validate(path *field.Path) (errs field.ErrorLis
 
 	errs = append(errs, n.EnvoyProxy.Validate(path.Child("envoyProxy"))...)
 	errs = append(errs, n.Traefik.Validate(path.Child("traefik"))...)
+	if patchErrs := n.Patches.Validate(); len(patchErrs) > 0 {
+		for _, err := range patchErrs {
+			errs = append(errs, field.Invalid(path.Child("patches"), n.Patches, err.Error()))
```
twz123: https://github.com/k0sproject/k0s/pull/7727/changes#r3387093531

```suggestion
			errs = append(errs, err)
```

pkg/component/worker/nllb/reconciler.go
```diff
@@ -439,3 +442,23 @@ func getLoopbackIP(ctx context.Context) (net.IP, error) {
...
 }
+
+func patchPod(pod *corev1.Pod, patches v1beta1.Patches) (*corev1.Pod, error) {
+	// patches work via marshaled data, so convert to pod to yaml and back to apply the patches
+	podBytes, err := yaml.Marshal(pod)
+	if err != nil {
+		return pod, fmt.Errorf("failed to marshal pod manifest for patching: %w", err)
+	}
+
+	patchedBytes, err := k0spatches.Apply(podBytes, patches)
```
twz123, line 453: We might want to add some more functions to the patches package in the future, i.e. I could imagine some `func ApplyAll([]runtime.Object)` that's using a proper scheme, so we don't rely on multiple serialization roundtrips.

pkg/apis/k0s/v1beta1/nllb.go
jnummelin: `err` is of type `Error`, cannot append to type `[]*field.Error`

twz123: Then we should fix this on Patches.validate(). The current approach produces weird error messages:

> envoyProxy.patches[0]: Invalid value: {"target":{"kind":"Foo","name":"foo","namespace":"none"},"patch":{"type":"","content":""}}: patches[0].patch.type: invalid type "": must be one of JSON, StrategicMergePatch, MergePatch


Moreover, this seems to panic for empty patches:

```go
func TestNLLB_Validate(t *testing.T) {
	minimalValid := func() *NodeLocalLoadBalancing {
		return &NodeLocalLoadBalancing{
			EnvoyProxy: &EnvoyProxy{
				APIServerBindPort: 7443,
				Image:             &ImageSpec{Image: "envoy", Version: "latest"},
			},
			Traefik: &Traefik{
				APIServerBindPort: 7443,
				Image:             &ImageSpec{Image: "traefik", Version: "latest"},
			},
		}
	}

	require.NoError(t, minimalValid().Validate(nil).ToAggregate())

	t.Run("empty patch", func(t *testing.T) {
		patches := Patches{{}}
		underTest := minimalValid()
		underTest.EnvoyProxy.Patches = patches
		underTest.Traefik.Patches = patches

		// This panics ...
		assert.NoError(t, underTest.Validate(nil).ToAggregate())
	})

	t.Run("bogus patch", func(t *testing.T) {
		patches := Patches{{
			Target: PatchTarget{
				Kind:      "Foo",
				Name:      "foo",
				Namespace: "none",
			},
			Patch: PatchSpec{},
		}}
		underTest := minimalValid()
		underTest.EnvoyProxy.Patches = patches
		underTest.Traefik.Patches = patches

		// This is just for demonstration purposes
		assert.NoError(t, underTest.Validate(nil).ToAggregate())
	})
}
```

jnummelin: Fixed in #[0e63c76](https://github.com/k0sproject/k0s/pull/7895/commits/0e63c7632aedb2ef7d29c2384feac037f8300591)

Scoped this to patches only. We need to decide what the `Validateable` interface should return in the long run, plain `[]error` or `field.ErrorList`. IMO that's completely a separate PR, if we decide to harmonize that.

jnummelin:
@twz123 resolved the comments, PTAL

pkg/component/worker/nllb/reconciler.go
```diff
@@ -439,3 +442,23 @@ func getLoopbackIP(ctx context.Context) (net.IP, error) {
 
 	return net.IP{127, 0, 0, 1}, err
 }
+
+func patchPod(pod *corev1.Pod, patches v1beta1.Patches) (*corev1.Pod, error) {
+	// patches work via marshaled data, so convert to pod to yaml and back to apply the patches
+	podBytes, err := yaml.Marshal(pod)
+	if err != nil {
+		return pod, fmt.Errorf("failed to marshal pod manifest for patching: %w", err)
```
twz123, line 450: I don't think it makes sense to wrap the error here. In fact, I guess this call can't fail at all in practice? If that ever occurs, then the error itself will carry enough information, given that callers of this function will wrap the returned error themselves.

Moreover, I think we should return a nil pod on error.

Same for the unmarshal error wrapping at the end of the function.

```suggestion
		return nil, err
```

pkg/component/worker/nllb/reconciler.go
```diff
@@ -439,3 +442,23 @@ func getLoopbackIP(ctx context.Context) (net.IP, error) {
...
+	}
+
+	patchedBytes, err := k0spatches.Apply(podBytes, patches)
+	if err != nil {
+		return pod, fmt.Errorf("failed to apply patches to pod manifest: %w", err)
```
twz123, line 455: This is error wrapping is duplicated at the call site. I wonder if we should return just the plain error here, as well?


```suggestion
		return nil, err
```

pkg/component/worker/nllb/reconciler.go
```diff
@@ -20,14 +20,17 @@ import (
...
 	k0snet "github.com/k0sproject/k0s/internal/pkg/net"
+	k0spatches "github.com/k0sproject/k0s/internal/pkg/patches"
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
 	"github.com/k0sproject/k0s/pkg/component/manager"
 	"github.com/k0sproject/k0s/pkg/component/worker"
 	workerconfig "github.com/k0sproject/k0s/pkg/component/worker/config"
 	"github.com/k0sproject/k0s/pkg/config"
 	"github.com/k0sproject/k0s/pkg/constant"
 	kubeutil "github.com/k0sproject/k0s/pkg/kubernetes"
+	"sigs.k8s.io/yaml"
```
twz123, line 31: Can you place this further down, together with the logrus import?

twz123 (commented):
We probably want to document the patches feature in both the configuration and in the NLLB doc pages?

There's quite a few places in which non-nil pointers are returned on the error path, which is rather unusual.

We need to consider one thing carefully, since we're patching on the worker side for the first time. A bad patch can prevent workers from starting up. If a patch fails to apply, or tries to do nasty things like changing the NLLB pod name, the whole NLLB component will fail to start, and with it, the whole worker. That's a different story than the one for controllers, where only the actually affected parts fail to be reconciled.

Also nice combined with dynamic config. A bad patch passes the config validation and gets written to the worker config. Since we're currently _still_ not reconciling the worker config, just nothing happens. But then, if workers get restarted, things go south. Especially nasty when you're trying to update a cluster, and you're running controller nodes with enabled workers. That way, you might be able to bring down your control plane, but will have a hard time to fix the broken config, because the controllers are crash-looping and you need to patch the bad `ClusterConfig`.

jnummelin:
> If a patch fails to apply, or tries to do nasty things like changing the NLLB pod name, the whole NLLB component will fail to start, and with it, the whole worker. That's a different story than the one for controllers, where only the actually affected parts fail to be reconciled.

You're right, I didn't consider this. I think the question is: If a patch fails to apply, what _should_ we do? I see couple options, each with their own pros and cons:

- Run with the default; User would still get NLLB running but just without any of the patches.
- Fail to run the entire NLLB; The worker would get up and running, just without NLLB

PR #7902, opened 2026-06-28
Propagate supervisor stop failures

### LINKED ISSUE #7901: konnectivity respawns duplicate server and fails to bind :8132

### Platform

```shell
Linux 6.19.14-300.fc44.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Apr 23 15:17:50 UTC 2026 x86_64 GNU/Linux

NAME="Fedora Linux"
VERSION="44 (Server Edition)"
ID=fedora
VERSION_ID=44
PRETTY_NAME="Fedora Linux 44 (Server Edition)"
VARIANT="Server Edition"
VARIANT_ID=server
```

### Version

```shell
v1.36.1+k0s.0
```

### What happened?

`k0scontroller.service` is running, but the `konnectivity` component is continuously respawning every 5 seconds. Each new `konnectivity-server` process fails because `:8132` is already in use.

The port is already held by an existing `konnectivity-server` child process of the same `k0s controller` process, so this looks like k0s is supervising/spawning a duplicate konnectivity process while the old one is still alive.

Current k0s status:

```shell
$ k0s status
Version: v1.36.1+k0s.0
Process ID: 1342
Role: controller
Workloads: false
SingleNode: false
```

Port owner:

```shell
$ ss -ltnp 'sport = :8132'
State  Recv-Q Send-Q  Local Address:Port Peer Address:PortProcess
LISTEN 0      1048576             *:8132            *:*    users:(("konnectivity-se",pid=3129,fd=7))
```

Process tree details:

```shell
$ ps -fp 3129,1342
UID        PID  PPID  C STIME TTY          TIME CMD
konnect+  3129  1342  0 13:37 ?        00:00:05 /var/lib/k0s/bin/konnectivity-server --admin-port=8133 --agent-port=8132 ...
root      1342     1  0 13:37 ?        00:01:26 /usr/local/bin/k0s controller --config=/etc/k0s/k0s.yaml --data-dir=/var/lib/k0s
```

Relevant config:

```yaml
spec:
  konnectivity:
    adminPort: 8133
    agentPort: 8132
```

### Steps to reproduce

I do not yet have a clean reproducer. This was observed on an existing controller running `v1.36.1+k0s.0`.

Observed state:

1. Start/run `k0scontroller.service`.
2. `konnectivity-server` is already listening on `:8132`.
3. k0s keeps spawning another `konnectivity` process.
4. The new process exits with `bind: address already in use`.
5. k0s logs `respawning in 5s` and repeats.

### Expected behavior

k0s should not repeatedly start a second `konnectivity-server` instance when one of its own child processes is already running and listening on the configured `agentPort`.

### Actual behavior

`konnectivity` is restarted every 5 seconds and exits with `address already in use`.

### Screenshots and logs

```text
2026-06-28T19:14:23+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:23" level=info msg="Restarted (4025)" component=konnectivity
2026-06-28T19:14:23+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:23" level=info msg="Error: failed to run the agent server: failed to listen on :8132: listen tcp :8132: bind: address already in use" component=konnectivity stream=stderr
2026-06-28T19:14:23+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:23" level=info msg="E0628 19:14:23.617582   31891 main.go:48] error: failed to run the agent server: failed to listen on :8132: listen tcp :8132: bind: address already in use" component=konnectivity stream=stderr
2026-06-28T19:14:23+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:23" level=error msg="Process terminated unexpectedly: exit status 1" component=konnectivity
2026-06-28T19:14:23+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:23" level=info msg="respawning in 5s" component=konnectivity
2026-06-28T19:14:28+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:28" level=info msg="Restarted (4026)" component=konnectivity
2026-06-28T19:14:28+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:28" level=info msg="Error: failed to run the agent server: failed to listen on :8132: listen tcp :8132: bind: address already in use" component=konnectivity stream=stderr
2026-06-28T19:14:28+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:28" level=error msg="Process terminated unexpectedly: exit status 1" component=konnectivity
2026-06-28T19:14:28+08:00 k0s-share2 k0s[1342]: time="2026-06-28 19:14:28" level=info msg="respawning in 5s" component=konnectivity
```

### Additional context

I searched the existing issues for the exact error strings:

- `failed to listen on :8132`
- `failed to run the agent server`
- `bind: address already in use 8132`

I found related konnectivity issues such as `No agent available`, but did not find an exact match for this duplicate `konnectivity-server` / `:8132` bind conflict.

flyzstu:
Fixes #7901.

## Summary

- propagate supervisor stop/termination errors back to `Stop()` callers
- keep the supervisor marked as running when stop times out, so callers cannot accidentally start a replacement while the old process still owns resources
- make konnectivity server reconciliation abort when stopping the previous server fails
- treat a process exiting due to SIGTERM as a successful stop to preserve existing normal shutdown semantics

## Test plan

- `GOCACHE=/tmp/k0s-go-cache GOMODCACHE=/tmp/k0s-go-mod go test ./pkg/supervisor ./pkg/component/controller`

### DIFF
--- a/pkg/component/controller/konnectivity.go
+++ b/pkg/component/controller/konnectivity.go
@@ -186,7 +186,7 @@ func (k *Konnectivity) runServer(ctx context.Context, count uint) error {
 	err := k.supervisor.Supervise(ctx)
 	if err != nil {
 		k.supervisor = nil // not to make the next loop to try to stop it first
-		return err
+		return fmt.Errorf("failed to supervise konnectivity server: %w", err)
 	}
 	k.EmitWithPayload("started konnectivity server", map[string]any{"serverCount": count})
 
--- a/pkg/supervisor/supervisor.go
+++ b/pkg/supervisor/supervisor.go
@@ -47,14 +47,14 @@ type Supervisor struct {
 	log            logrus.FieldLogger
 	mutex          sync.Mutex
 	startStopMutex sync.Mutex
-	stop           func(opts StopOpts)
+	stop           func(opts StopOpts) error
 }
 
 const k0sManaged = "_K0S_MANAGED=yes"
 
-// processWaitQuit waits for a process to exit or a shut down signal
-// returns true if shutdown is requested
-func (s *Supervisor) processWaitQuit(ctx context.Context, cmd *exec.Cmd) bool {
+// processWaitQuit waits for a process to exit or a shut down signal.
+// It returns true if shutdown is requested and an error if the shutdown failed.
+func (s *Supervisor) processWaitQuit(ctx context.Context, cmd *exec.Cmd) (bool, error) {
 	waitresult := make(chan error, 1)
 	go func() {
 		defer close(waitresult)
@@ -73,12 +73,15 @@ func (s *Supervisor) processWaitQuit(ctx context.Context, cmd *exec.Cmd) bool {
 		if stoppingErr, ok := errors.AsType[*stoppingErr](cause); ok {
 			stopOpts = stoppingErr.opts
 		}
-		if err := s.terminateSupervisedProcess(cmd, waitresult, stopOpts); err != nil {
+		if terminated, err := s.terminateSupervisedProcess(cmd, waitresult, stopOpts); err != nil && !terminated {
 			s.log.WithError(err).Error("Error while terminating process")
+			return true, err
+		} else if err != nil {
+			s.log.WithError(err).Warn("Process terminated with errors")
 		} else {
 			s.log.Info("Process terminated successfully")
 		}
-		return true
+		return true, nil
 
 	case err, ok := <-waitresult:
 		var exitErr *exec.ExitError
@@ -94,25 +97,27 @@ func (s *Supervisor) processWaitQuit(ctx context.Context, cmd *exec.Cmd) bool {
 		default:
 			s.log.WithError(err).Error("Failed to wait for process: ", state)
 		}
-		return false
+		return false, nil
 	}
 }
 
-func (s *Supervisor) terminateSupervisedProcess(cmd *exec.Cmd, waitresult <-chan error, stopOpts StopOpts) error {
+func (s *Supervisor) terminateSupervisedProcess(cmd *exec.Cmd, waitresult <-chan error, stopOpts StopOpts) (bool, error) {
 	if timeout := stopOpts.DeferGracefulTerminationUntil; timeout != nil {
 		// Termination request deferred, wait for process to finish on its own.
 		s.log.Debug("Awaiting process termination")
 
 		select {
-		case err := <-waitresult:
+		case err, ok := <-waitresult:
 			var exitErr *exec.ExitError
 			switch {
+			case !ok:
+				return true, errors.New("process wait result unavailable")
 			case err == nil:
-				return nil
+				return true, nil
 			case errors.As(err, &exitErr):
-				return exitErr
+				return true, fmt.Errorf("process terminated while awaiting deferred stop: %w", exitErr)
 			default:
-				return fmt.Errorf("failed to wait for process: %w", err)
+				return true, fmt.Errorf("failed to wait for process: %w", err)
 			}
 		case <-timeout:
 			s.log.Debug("Timed out while waiting for process to terminate, requesting graceful termination")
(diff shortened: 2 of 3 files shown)


pkg/supervisor/supervisor_test.go
```diff
@@ -216,6 +216,37 @@ func TestStopWhileRespawn(t *testing.T) {
...
+		TimeoutRespawn: 1 * time.Hour,
+	}
+
+	require.NoError(t, s.Supervise(t.Context()))
+	require.NoError(t, pingPong.AwaitPing())
+
+	process := s.GetProcess()
+	t.Cleanup(func() {
+		_ = process.Kill()
+	})
```
twz123: We don't need the kill here if we move up the `SendPong` from line 247 and defer it like this:

```suggestion
	t.Cleanup(func() {
		assert.NoError(t, pingPong.SendPong())
	})
```

I don't know why LLMs do this all the time in this test case :sweat_smile:

pkg/supervisor/supervisor_test.go
```diff
@@ -216,6 +216,37 @@ func TestStopWhileRespawn(t *testing.T) {
...
+	s := Supervisor{
+		Name:           t.Name(),
+		BinPath:        pingPong.BinPath(),
+		Args:           pingPong.BinArgs(),
+		RunDir:         t.TempDir(),
+		TimeoutStop:    10 * time.Millisecond,
+		TimeoutRespawn: 1 * time.Hour,
+	}
+
+	require.NoError(t, s.Supervise(t.Context()))
```
twz123: After a call to supervise, it's best to always defer the Stop call.

```suggestion
	require.NoError(t, s.Supervise(t.Context()))
	t.Cleanup(func() { _ = s.Stop() })
```

pkg/component/controller/konnectivity.go
```diff
@@ -172,6 +169,7 @@ func (k *Konnectivity) runServer(ctx context.Context, count uint) error {
 			map[string]any{"serverCount": count})
 		if err := k.supervisor.Stop(); err != nil {
 			k.log.WithError(err).Error("Failed to stop executable")
+			return fmt.Errorf("failed to stop existing konnectivity server: %w", err)
```
twz123: The general idea is not to err out here, but around lines 185-192. So whenever stopping fails, we log it. Attempting to start a process when the old process is still around is the point at which k0s should give up.

pkg/supervisor/supervisor.go
```diff
@@ -133,7 +129,7 @@ func (s *Supervisor) terminateSupervisedProcess(cmd *exec.Cmd, waitresult <-chan
 				return nil
 			case errors.As(err, &exitErr):
 				if status, ok := exitErr.Sys().(syscall.WaitStatus); ok && status.Signal() == syscall.SIGTERM {
-					return errors.New("process terminated without handling SIGTERM")
+					return nil
```
twz123: Let's keep this as a warning for now.

```suggestion
					s.log.Warn("Process terminated without handling SIGTERM")
					return nil
```

pkg/supervisor/supervisor.go
```diff
@@ -216,10 +212,14 @@ func (s *Supervisor) Supervise(ctx context.Context) error {
...
-	started, done := make(chan error, 1), make(chan bool)
+	started, done := make(chan error, 1), make(chan error, 1)
 
 	go func() {
-		defer close(done)
+		var stopErr error
+		defer func() {
+			done <- stopErr
+			close(done)
+		}()
```
twz123: I think this can be much simpler. We can just declare `stopErr` outside the goroutine, and leave the done channel as `chan struct{}`. (I guess the `chan bool` is a leftover from a previous refactoring.)


```go
	var stopErr error
	go func() {
		defer close(done)

		// ...
```

pkg/supervisor/supervisor.go
```diff
@@ -293,9 +294,14 @@ func (s *Supervisor) Supervise(ctx context.Context) error {
...
-		<-done
+	var stopOnce sync.Once
+	var stopResult error
+	s.stop = func(opts StopOpts) error {
+		stopOnce.Do(func() {
+			cancel(&stoppingErr{opts})
+			stopResult = <-done
+		})
+		return stopResult
 	}
```
twz123: Since we only ever read `stopErr` after `done` has been closed, we achieve a happens-before relationship between the assignment in the goroutine and the access in this stop func. So we could simplify this to:

```go
	s.stop = func(opts StopOpts) error {
		cancel(&stoppingErr{opts})
		<-done
		return stopErr // Assuming stopErr has been declared outside the goroutine. 
	}
```

pkg/supervisor/supervisor.go
```diff
@@ -313,9 +319,11 @@ func (s *Supervisor) StopWith(opts StopOpts) error {
...
 
-	s.stop(opts)
-	s.stop = nil
-	return nil
+	err := s.stop(opts)
+	if err == nil {
+		s.stop = nil
+	}
+	return err
```
twz123, line 331: The general idea is that the Supervisor does all it can to stop the process, and then gives up and leaves the supervised process alone. The process is now expected to terminate on its own, eventually. If that doesn't happen, k0s requires manual intervention from cluster operators. It deliberately no longer kills processes on its own, as it never knows if that will induce even more harm. We have PID files in place which are used to guard against double-starting an orphaned process. That's also helpful when e.g. k0s itself gets terminated abnormally. So the expectation is that the supervisor will fail when `Supervise` is called on an orphaned process, so it's safe to reset the internal state to "stopped" here.

If we change that contract, we'd need to change the callers, too. I.e. they'd need to call Stop before Supervise, in order to retry a previously attempted stop request.

pkg/supervisor/supervisor.go
```diff
@@ -47,45 +47,41 @@ type Supervisor struct {
...
 		cause := context.Cause(ctx)
 		s.log.Debugf("Attempting to terminate supervised process (%v)", cause)
 		var stopOpts StopOpts
-		if stoppingErr, ok := errors.AsType[*stoppingErr](cause); ok {
+		if stoppingErr := (*stoppingErr)(nil); errors.As(cause, &stoppingErr) {
 			stopOpts = stoppingErr.opts
 		}
 		if err := s.terminateSupervisedProcess(cmd, waitresult, stopOpts); err != nil {
 			s.log.WithError(err).Error("Error while terminating process")
+			return true, err
```
twz123, line 78: It's not that easy here unfortunately.

What's the actual goal of this PR? Right now, it will just forward all the errors from `terminateSupervisedProcess` back to callers of the `Stop` method. But that function may return a multitude of errors, and not all of them are IMO interesting/relevant to the callers. In fact, I think the only class of errors which makes sense to be forwarded is whenever the supervised process is still running when `Stop` returns. Would you agree?

But that's not how `terminateSupervisedProcess` works right now. It might return errors even if the process _did_ terminate, e.g. because it didn't exit cleanly.

twz123 (commented):
Thanks for this. A few things that need to be sorted out:

> * propagate supervisor stop/termination errors back to `Stop()` callers

What kind of errors would make sense to be forwarded to callers of `Stop`? As mentioned in one of the comments, I think we need to agree on this first.

> * keep the supervisor marked as running when stop times out, so callers cannot accidentally start a replacement while the old process still owns resources

This is not how `Supervisor` is intended to work. The double start prevention is supposed to be done by checking PID files inside the `Supervise` method. This covers e.g. also cases in which k0s was terminated abnormally and wasn't able to reap its supervised processes properly.

> * make konnectivity server reconciliation abort when stopping the previous server fails

The idea is to err out if the new server fails to start because another one is still running.

twz123:
Btw, this needs a rebase on top of the current main branch. There's been some relevant changes for this PR.

flyzstu:
Thanks a lot for the detailed review and guidance. I really appreciate the time you spent explaining the intended supervisor semantics and the related main branch changes.

I'm not a professional software developer, so I'm relying on tools like Codex to help me understand and implement the changes correctly. That said, I do want to contribute something useful to the project, and your feedback is very helpful for learning the right direction here.

I'll rebase this PR on top of the current main branch, adjust the implementation according to your comments, and rerun the relevant tests before pushing an update.

PR #7929, opened 2026-07-03
Support K0S_JOIN_TOKEN and deprecate CLI token args

### LINKED ISSUE #5253: Support getting join token from environment variable

didiberman:
### Description
This PR resolves #5253 by:
1. Adding support for the `K0S_JOIN_TOKEN` environment variable as the preferred source for join tokens.
2. Falling back to the legacy `K0S_TOKEN` environment variable for backwards compatibility.
3. Adding deprecation warnings to both `k0s controller` and `k0s worker` subcommands when a join token is passed as a positional command-line argument.

This ensures credentials are passed securely and gives users a clear deprecation warning timeline to transition off positional arguments.

### Testing
Ran unit tests for validation and retrieval:
- `go test -v ./cmd/internal/...`
- `go test -v ./cmd/worker/...`
All tests passed successfully.

### AI Assistance Note
This PR was written with AI assistance.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -93,6 +93,7 @@ func NewControllerCmd() *cobra.Command {
 
 			if len(args) > 0 {
 				c.TokenArg = args[0]
+				logrus.Warn("Passing the join token as a positional CLI argument is deprecated and will be removed in a future release. Please use the --token-file flag or the K0S_JOIN_TOKEN environment variable instead.")
 			}
 			if err := internal.CheckSingleTokenSource(c.TokenArg, c.TokenFile); err != nil {
 				return err
--- a/cmd/internal/tokendata.go
+++ b/cmd/internal/tokendata.go
@@ -9,8 +9,11 @@ import (
 	"os"
 )
 
-// EnvVarToken is the environment variable name for the join token
+// EnvVarToken is the environment variable name for the join token (legacy/fallback)
 const EnvVarToken = "K0S_TOKEN"
+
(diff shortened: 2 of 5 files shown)


jnummelin:
I don't get this 🤔 

We already support this functionality via `K0S_TOKEN`, why do we need another env var for this?

s3rj1k:
We might want to add ref of https://github.com/k0sproject/k0s/pull/6727#issuecomment-3897249695 into parent issue and close it

twz123:
We still need to add the depreciation note to the CLI arg, I guess.

PR #8023, opened 2026-07-28
Distinguish pending from performed restarts in Autopilot

twz123:
## Description

The k0s-restart reconciler determined whether to terminate k0s solely by comparing the version reported via the status socket to the version requested in the update plan. It couldn't tell whether a signaled restart was still pending or had already been performed. Once the restart brings up the new k0s process, node object updates (such as kubelet status heartbeats) retrigger the reconciler. For forced updates whose plan version doesn't match the version reported by the binary, the comparison never succeeds, and the reconciler terminated the freshly started process over and over again. The k0s-restarted reconciler, which honors forceupdate, then had to win a race against the next kill to move the signal to UnCordoning.

Break this cycle by tracking restart initiations in the process itself. The applying-update reconciler records the signal data right before making the Restart status visible to the API, and the restart reconciler only terminates k0s for restarts recorded by the running process. If the restart status was not written by the process itself, it has to predate the process, meaning a restart must have already occurred. For forced updates, move the signal to UnCordoning in that case instead. Conversely, the restarted reconciler no longer advances the signal while a restart initiated by the running process is pending. Previously, a restart could be skipped entirely if the controller managers happened to be rebuilt in the short window between the status update and the process's termination.

Create the tracker in the Autopilot roots and hand it down to the controllers. Its lifetime needs to be tied to the process, not to any particular controller manager, as the controller root rebuilds its managers whenever the lease status changes, and the worker root does so on every retry attempt.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/autopilot/controller/root_controller.go
+++ b/pkg/autopilot/controller/root_controller.go
@@ -20,6 +20,7 @@ import (
 	"github.com/k0sproject/k0s/pkg/autopilot/controller/plans"
 	aproot "github.com/k0sproject/k0s/pkg/autopilot/controller/root"
 	"github.com/k0sproject/k0s/pkg/autopilot/controller/signal"
+	apsigk0s "github.com/k0sproject/k0s/pkg/autopilot/controller/signal/k0s"
 	"github.com/k0sproject/k0s/pkg/autopilot/controller/updates"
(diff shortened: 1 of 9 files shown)


jnummelin (approved):
LGTM

PR #8033, opened 2026-07-29
Stop and wait for svc deletion on Windows before re-install

### LINKED ISSUE #8029: `k0s install worker --force` fails on Windows when the service is already running (leaves service in DELETE_PENDING)
### Before creating an issue, make sure you've browsed the existing issues
Checked — no existing issue for `install --force` failing against a running Windows service.

## Version
```
v1.35.4+k0s.0
```

## Platform
```
Microsoft Windows Server 2022 Datacenter, Version 10.0.20348 (x86_64)
```

## Sysinfo
Standard worker node; `k0s sysinfo` passes. The bug is in the service (re)install path, not node prerequisites.

## What happened?

On a Windows node where the `k0sworker` service is **already installed and running**, re-running `k0s install worker --force` with any changed argument (in our case a changed `--kubelet-extra-args`) fails immediately (exit code 1, in well under a second).

The failed run leaves the node in a broken half-reinstalled state:

- The old `k0sworker` process keeps running with the **old** arguments (the new args are never applied).
- The Windows service is left with `StartType: Disabled` while `Status: Running` — i.e. the service was marked for deletion but the deletion never completed because the process was never stopped.
- No new "A service was installed in the system" event is ever emitted, confirming the re-create step failed.

Observed service state immediately after the failure (`Get-Service k0sworker`):
```
Name       : k0sworker
DisplayName: k0s worker
Status     : Running        <-- old process still alive
StartType  : Disabled       <-- delete pending, never finalized
```

`k0s status` still reports the pre-reinstall arguments (no new flag present), confirming the running process was never replaced.

The most recent SCM "service installed" event on the node is the *original* install; there is no second install event after the `--force` attempt.

## Root cause

The `--force` branch of `InstallService` uninstalls and immediately reinstalls the service **without stopping the running service first**:

`pkg/install/service.go` (v1.35.4+k0s.0), lines 85–94:
```go
if force {
    logrus.Infof("Uninstalling %s service", svcConfig.Name)
    err = s.Uninstall()                 // <-- DeleteService, but the process is still running
    if err != nil && !errors.Is(err, service.ErrNotInstalled) {
        logrus.Warnf("failed to uninstall service: %v", err)
    }
}

logrus.Infof("Installing %s service", svcConfig.Name)
return s.Install()                      // <-- fails: service still "exists" (DELETE_PENDING)
```

On Windows, `kardianos/service`'s `Uninstall()` calls `DeleteService` **without stopping the service**. The Windows Service Control Manager does not delete a service whose process is still running — it transitions it to `DELETE_PENDING` (which is why `StartType` flips to `Disabled`). The immediately-following `Install()` then calls `OpenService`/`CreateService` for the same name and fails, because from SCM's point of view the service still exists / is marked for deletion (`ERROR_SERVICE_MARKED_FOR_DELETE`, 1072). `InstallService` returns that error, so `k0s install --force` exits non-zero.

Why this is Windows-specific: on Linux the "uninstall + install" pair just rewrites a systemd unit file, which is harmless while the old process keeps running (it's restarted separately afterwards). There is no SCM-style "can't recreate a running service" constraint, so the same code path works on Linux.

Note that k0s already knows how to do this correctly elsewhere: `StartInstalledService(force)` in the same file calls `svc.Restart()` when the service is already running. Only the `--force` reinstall path in `InstallService` skips the stop.

## Steps to reproduce

On a Windows host:

1. Install and start a worker:
   ```
   k0s install worker --token-file C:\etc\k0s\k0stoken --kubelet-extra-args="--node-ip=10.0.0.11"
   k0s start
   ```
   Confirm `Get-Service k0sworker` shows `Running`, `StartType: Automatic`.

2. Re-run install with `--force` and a changed argument while the worker is still running:
   ```
   k0s install worker --force --token-file C:\etc\k0s\k0stoken --kubelet-extra-args="--node-ip=10.0.0.11 --register-with-taints=example.com/notready=true:NoExecute"
   ```

**Actual:** the command exits 1 almost instantly. `Get-Service k0sworker` now shows `Status: Running`, `StartType: Disabled`; the running process still has the old args; no new service-installed event is logged.

(Any changed arg reproduces it — the specific flag is irrelevant. A fresh `install worker` with the same new args on a clean node works fine, which isolates the failure to the reinstall-over-running-service case.)

## Expected behavior

`k0s install worker --force` on an already-running Windows node should either:
- stop the running service before uninstalling, wait for the deletion to finalize, then create the service with the new arguments; or
- update the existing service's arguments in place (e.g. `ChangeServiceConfig`).

Either way the command should succeed and the worker should come back up with the new arguments, matching the Linux behavior.

## Suggested fix

In `InstallService` (`pkg/install/service.go`), when `force` is set, stop the service before uninstalling and wait for the service to actually be gone before reinstalling. Sketch:

```go
if force {
    if status, _ := s.Status(); status == service.StatusRunning {
        logrus.Infof("Stopping %s service before reinstall", svcConfig.Name)
        if err := s.Stop(); err != nil {
            logrus.Warnf("failed to stop service before reinstall: %v", err)
        }
    }
    logrus.Infof("Uninstalling %s service", svcConfig.Name)
    if err := s.Uninstall(); err != nil && !errors.Is(err, service.ErrNotInstalled) {
        logrus.Warnf("failed to uninstall service: %v", err)
    }
    // On Windows, DeleteService is asynchronous; wait until the service is
    // actually gone (ErrNotInstalled) before CreateService, otherwise Install()
    // races DELETE_PENDING and fails with ERROR_SERVICE_MARKED_FOR_DELETE.
}
```
(with a bounded wait/poll on `s.Status()` returning `ErrNotInstalled` on Windows before calling `s.Install()`).

## Impact

On Windows, k0s cannot change **any** service argument (`--kubelet-extra-args`, labels, profile, etc.) on an already-running worker via `install --force` — the reinstall is guaranteed to fail and leaves the service in `DELETE_PENDING`/`Disabled` while orphaning the old process on the old config. This breaks tools like k0sctl that reconfigure workers by re-running `k0s install --force` when install flags change (the node is left running the old config, with the service disabled). A node reboot is currently required to clear the pending deletion.

jnummelin:
## Description

On Windows we cannot re-install a svc unless the svc is stopped. So stop the svc first on Windows and then proceed with the uninstall and reinstall steps. Uninstall is also async so we need to poll until the svc is actually gone.

Fixes #8029

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/install/service.go
+++ b/pkg/install/service.go
@@ -4,12 +4,15 @@
 package install
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"runtime"
+	"time"
 
 	"github.com/kardianos/service"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 var (
@@ -54,7 +57,7 @@ func InstalledService() (service.Service, error) {
 }
 
 // InstallService installs the k0s service, per the given arguments, and the detected platform
-func InstallService(args []string, envVars []string, force bool) error {
+func InstallService(ctx context.Context, args []string, envVars []string, force bool) error {
 	var svcConfig *service.Config
 
 	prg := &Program{}
@@ -83,10 +86,31 @@ func InstallService(args []string, envVars []string, force bool) error {
 	svcConfig.Arguments = args
 
 	if force {
+		if runtime.GOOS == "windows" {
+			// On Windows, the service must be stopped first
+			if err = s.Stop(); err != nil && !errors.Is(err, service.ErrNotInstalled) {
+				logrus.Warnf("failed to stop service before re-install: %v", err)
+			}
+		}
+
 		logrus.Infof("Uninstalling %s service", svcConfig.Name)
 		err = s.Uninstall()
 		if err != nil && !errors.Is(err, service.ErrNotInstalled) {
 			logrus.Warnf("failed to uninstall service: %v", err)
+		}
+		// On windows the service delete/uninstall is async, wait till it is done
+		if runtime.GOOS == "windows" {
+			if err = wait.PollUntilContextCancel(ctx, 1*time.Second, true, func(ctx context.Context) (done bool, err error) {
+				_, err = s.Status()
+				if errors.Is(err, service.ErrNotInstalled) {
+					return true, nil
+				}
+				logrus.Info("waiting for the service to be actually deleted")
+				return false, nil
+			}); err != nil {
+				logrus.Errorf("failed to wait service to be deleted: %v", err)
+				return err
+			}
 		}
 	}
 
--- a/.github/workflows/win-wsl.yaml
+++ b/.github/workflows/win-wsl.yaml
@@ -260,6 +260,10 @@ jobs:
           & kubectl get nodes,pods -A
           go test -v -count=1 ./inttest/windows
 
+      - name: Reinstall with force
+        run: |
+          & .\k0s.exe install worker --force --token-file=k0s-worker-token --debug --labels test.k0sproject.io/foo=bar --kubelet-extra-args="--hostname-override=windowsworker"
+          & .\k0s.exe start
       - name: Stop & reset k0s worker
         run: |
           & .\k0s.exe stop --debug
--- a/cmd/install/controller.go
+++ b/cmd/install/controller.go
@@ -66,7 +66,7 @@ With the controller subcommand you can setup a single node cluster by running:
 			}
 
 			args := append([]string{"controller"}, flagsAndVals...)
-			if err := install.InstallService(args, envVars, installFlags.force); err != nil {
+			if err := install.InstallService(cmd.Context(), args, envVars, installFlags.force); err != nil {
 				return fmt.Errorf("failed to install controller service: %w", err)
 			}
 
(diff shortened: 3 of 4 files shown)


pkg/install/service.go
```diff
@@ -83,11 +86,33 @@ func InstallService(args []string, envVars []string, force bool) error {
...
+		}
+
 		logrus.Infof("Uninstalling %s service", svcConfig.Name)
 		err = s.Uninstall()
 		if err != nil && !errors.Is(err, service.ErrNotInstalled) {
 			logrus.Warnf("failed to uninstall service: %v", err)
 		}
+		// On windows the service delete/uninstall is async, wait till it is done
+		if runtime.GOOS == "windows" {
+			ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
twz123: Any chance we get a real context in here? Not a fan of hard coded timeouts, nor that we can't exit early using Ctrl+C.

That whole thing makes me wonder if we wanna bypass kardianos completely here and try to do it directly via `golang.org/x/sys/windows/svc`.

jnummelin: None of the install commands currently wire up context, I'll see how that could be done

jnummelin: > Not a fan of hard coded timeouts

You mean we should wait indefinitely, or for the ctrl-c event/sigterm?

twz123 (approved):

PR #8034, opened 2026-07-29
Unzip k0s embedded bins and set capabilities during docker build

makhov:
## Description

Currently we do hacks in k0smotron entrypoint but since now k0s is just an archive, we can do it during build time

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,29 @@
 ARG ARCH
 ARG ALPINE_VERSION
-FROM docker.io/library/${ARCH}alpine:$ALPINE_VERSION
+
+# Unpack the zip payload that's appended to the k0s executable while building
+# the image, so that k0s doesn't have to do it on every container start.
+FROM --platform=$BUILDPLATFORM docker.io/library/alpine:$ALPINE_VERSION AS bins
 ARG TARGETARCH
+RUN apk add --no-cache libcap unzip
+ADD ./k0s-${TARGETARCH}/k0s /k0s
+RUN mkdir -p /out/bin \
+  # unzip exits 1 because of the "extra bytes at beginning" warning that's
+  # caused by the k0s executable preceding the appended zip payload.
+  && { unzip -oqq /k0s -d /out/bin || [ $? -le 1 ]; } \
+  && { [ -n "$(ls -A /out/bin)" ] || { echo 'no zip payload appended to k0s' >&2; exit 1; }; } \
+  # The zip entries carry no external attributes, so unzip creates the files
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,7 +1,29 @@
...
+  # unzip exits 1 because of the "extra bytes at beginning" warning that's
+  # caused by the k0s executable preceding the appended zip payload.
+  && { unzip -oqq /k0s -d /out/bin || [ $? -le 1 ]; } \
+  && { [ -n "$(ls -A /out/bin)" ] || { echo 'no zip payload appended to k0s' >&2; exit 1; }; } \
+  # The zip entries carry no external attributes, so unzip creates the files
+  # with 0600. Use the same permissions that the runtime staging would use.
+  && chmod 0750 /out/bin/* && chmod 0755 /k0s \
+  # Allow kube-apiserver to bind to privileged ports without being root. This
+  # only survives because k0s won't rewrite the pre-unpacked files at runtime.
+  && setcap cap_net_bind_service=+ep /out/bin/kube-apiserver \
```
twz123, line 20: I guess this is no longer necessary as of #8045.

Dockerfile
```diff
@@ -1,7 +1,29 @@
...
+  # k0s reuses an already staged executable only if its modification time is
+  # equal to the one of the k0s executable, compared with nanosecond precision.
+  # Stamp both with a whole-second timestamp, so that they compare equal.
+  && touch -d "@$(stat -c %Y /k0s)" /k0s /out/bin/*
```
twz123, line 24: I think this should work, as well:

```suggestion
  && touch -r /k0s /k0s /out/bin/*
```

Would be interesting to know if BusyBox stat truncates the timestamp precision even when it's read from a reference file internally.

Dockerfile
```diff
@@ -11,7 +33,10 @@ RUN apk add --no-cache iptables tini \
 ENV KUBECONFIG=/var/lib/k0s/pki/admin.conf
 
 ADD docker-entrypoint.sh /entrypoint.sh
-ADD ./k0s-${TARGETARCH}/k0s /usr/local/bin/k0s
+# Both of these come from the bins stage, so that the modification times and
+# file capabilities established there are preserved.
+COPY --from=bins /k0s /usr/local/bin/k0s
```
twz123, line 38: Can we somehow strip the now no longer required zip archive from the k0s executable? We'd inflate the docker image quite a bit otherwise, since we'd ship the embedded executables twice.

PR #8066, opened 2026-08-07
Wire up context for all reset steps

### LINKED ISSUE #8049: k0s reset can hang indefinitely in the containers cleanup step: StopPodSandbox is called with context.TODO()
**Before creating an issue, make sure you've checked the following**
- [x] You're looking at docs for the released version

### Platform
```
Ubuntu 22.04.5 LTS, kernel 6.8.0-1061-aws, x86_64
AWS EC2 t3.medium (2 vCPU, 3.7 GiB), single node
```

### Version
```
v1.36.3+k0s.0
```

`pkg/cleanup/containers.go` and `pkg/container/runtime/cri.go` are identical between `v1.36.3+k0s.0` and `v1.35.4+k0s.0`, so both are affected.

### Sysinfo

<details><summary><code>k0s sysinfo</code></summary>

```
Total memory: 3.7 GiB (pass)
File system of /var/lib: ext4 (pass)
Disk space available for /var/lib/k0s: 26.8 GiB (pass)
Relative disk space available for /var/lib/k0s: 92% (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Containerd config snippets in /etc/k0s/containerd.d/: no config snippets found (pass)
  Linux kernel release: 6.8.0-1061-aws (pass)
  AppArmor: active (pass)
      CONFIG_NF_NAT: module (pass)
      CONFIG_LLC: module (pass)
      CONFIG_STP: module (pass)
```
</details>

### What happened?

`k0s reset` can block forever in its `containers` cleanup step. If the container runtime does not answer a `StopPodSandbox` call, reset waits indefinitely, because the call is made with a context that has neither a deadline nor cancellation.

In `pkg/cleanup/containers.go`, `stopAllContainers()`:

```go
func (c *containers) stopAllContainers() error {
	var errs []error

	var pods []string
	ctx := context.TODO()
	err := retry.Do(func() error {
		logrus.Debugf("trying to list all pods")
		var err error
		pods, err = c.containerRuntime.ListContainers(ctx)
		...
	}, retry.Context(ctx), retry.LastErrorOnly(true))
	...
	for _, pod := range pods {
		logrus.Debugf("stopping container: %v", pod)
		err := c.containerRuntime.StopContainer(ctx, pod)
```

Note the asymmetry: `ListContainers` is wrapped in `retry.Do`, but the `StopContainer` loop is not bounded in any way. That `ctx` reaches `pkg/container/runtime/cri.go`, where it is handed straight to the gRPC call:

```go
func (cri *CRIRuntime) StopContainer(ctx context.Context, id string) error {
	...
	request := &pb.StopPodSandboxRequest{PodSandboxId: id}
	logrus.Debugf("StopPodSandboxRequest: %v", request)
	r, err := client.StopPodSandbox(ctx, request)
```

With `context.TODO()` there is nothing to bound the wait, so a runtime that is slow or wedged stops reset permanently rather than producing an error.

I reproduced it by making one container genuinely unstoppable: pinning its rootfs mount so it cannot be released. `StopPodSandbox` then never returns.

### Steps to reproduce

```bash
k0s install controller --single && k0s start        # wait for the node to be Ready
kubectl run holder --image=busybox:1.36 -- sleep 100000
k0s stop

# pin one container's rootfs so it cannot be unmounted: a process whose cwd is the mountpoint
RF=$(mount | awk '/io.containerd.runtime.v2.task/ {print $3}' | head -1)
setsid bash -c "cd '$RF' && exec tail -f /dev/null" &
sleep 3
umount "$RF"    # confirms the pin: "target is busy"

k0s reset --data-dir /var/lib/k0s --kubelet-root-dir /var/lib/kubelet   # never returns
```

An open file descriptor inside the mountpoint works as well as a cwd.

### Expected behavior

The CRI calls in the cleanup path are bounded, so an unresponsive container runtime makes `k0s reset` fail with a clear error instead of hanging. Reset should terminate one way or the other.

### Actual behavior

Reset never returns. Observed still running at 30 s, 60 s and 90 s in the instrumented run, and for 25 minutes in an earlier unattended one.

Process state while hung:

```
State:   S (sleeping)
Threads: 9

per-thread wchan:
  futex_wait_queue   (x7)
  do_wait
  ep_poll
```

The container's rootfs was still mounted throughout (1 task mount present).

### Screenshots and logs

`SIGQUIT` with `GOTRACEBACK=all`. Goroutine 1 is the reset itself, parked waiting for gRPC response headers:

```
goroutine 1 gp=0x38dfa6d261e0 m=nil [select]:
runtime.gopark(...)
	runtime/proc.go:462
runtime.selectgo(...)
	runtime/select.go:351
google.golang.org/grpc/internal/transport.(*ClientStream).waitOnHeader(0x38dfa7b688c0)
	google.golang.org/grpc@v1.82.1/internal/transport/client_stream.go:130
google.golang.org/grpc/internal/transport.(*ClientStream).RecvCompress(...)
	google.golang.org/grpc@v1.82.1/internal/transport/client_stream.go:145
google.golang.org/grpc.(*csAttempt).recvMsg(...)
	google.golang.org/grpc@v1.82.1/stream.go:1156
google.golang.org/grpc.(*clientStream).RecvMsg.func1(...)
	google.golang.org/grpc@v1.82.1/stream.go:1021
...
github.com/k0sproject/k0s/pkg/container/runtime.(*CRIRuntime).StopContainer(...)
	github.com/k0sproject/k0s/pkg/container/runtime/cri.go:87
github.com/k0sproject/k0s/pkg/cleanup.(*containers).stopAllContainers(...)
	github.com/k0sproject/k0s/pkg/cleanup/containers.go:86
github.com/k0sproject/k0s/pkg/cleanup.(*containers).Run(...)
	github.com/k0sproject/k0s/pkg/cleanup/containers.go:54
github.com/k0sproject/k0s/pkg/cleanup.(*Config).Cleanup(...)
	github.com/k0sproject/k0s/pkg/cleanup/cleanup.go:33
```

So the wait is inside the gRPC client with no deadline available to break it.

On signals, one observation I can offer but not fully characterise: in an earlier unattended run a `SIGTERM` sent at the 4-minute mark did not take effect, the process was still running 21 minutes later, and a second `SIGTERM` then terminated it promptly. I have only that single observation and cannot say whether the first signal was lost, deferred, or handled by a path that itself blocked. `SIGQUIT` always produced the dump and exited.

### Additional context

**Relationship to earlier reports.** `k0s reset` hanging in this step has been reported before, and both reports were closed by the stale bot rather than fixed:

- #4211 "k0s reset hangs" (closed, `Stale`) — four reporters, last debug output before the hang is `* containers steps` / `started containerd successfully` / `trying to list all pods`, i.e. the same function. Two reporters were stuck permanently. Its author asked for exactly what is missing here: *"At least it should timeout with an error."*
- #4783 "k0s reset hangs when containerd is broken" (closed, `Stale`).

In #4211 a maintainer explained the general shape — reset must stop running containers before it can clean the data directory, because their mounts block deletion. What those threads never established is why the wait can be **unbounded**, which is what the stack above pins down.

**Relationship to #4434.** PR #4434 ("Make k0s reset fail if it can't reach containerd") addressed the neighbouring case: it bounded the gRPC dial retries and wrapped `ListPods` in retries so an absent socket no longer hangs forever. That fix stopped short of `StopContainer`, which still receives `context.TODO()` and is not retried or bounded. This report is that remaining half.

Note the same `context.TODO()` is also used for `managedContainerd.Init`/`Start` in `containers.Run`, so the exposure is not limited to `StopContainer`.

**Suggested fix.** Give the cleanup path's CRI calls a bounded context — a per-call timeout in `stopAllContainers()`, or plumb a cancellable context down from `k0s reset` so the command can be interrupted. Either turns an unbounded hang into a reportable failure.

Related: #8048 — a different defect in the same `k0s reset` cleanup, where a leftover mount under the run dir makes the run-dir delete fail. The reproduction here uses the same pinned-mount trick to make a container unstoppable, but the failure is in the `containers` step and is independent of that one. #5657 — `k0s stop` leaving containers running, which is what leaves a container around for the cleanup step to trip over.

jnummelin:
## Description

The context is the signal aware cli context so we'll respect terminating things properly. Signal is passed along to all context aware operations.

Fixes #8049

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/cleanup/cleanup.go
+++ b/pkg/cleanup/cleanup.go
@@ -32,8 +32,8 @@ func (c *Config) Cleanup(ctx context.Context) error {
 	for _, step := range c.cleanupSteps {
 		// Some steps are not-so context aware, before each step,
 		// check if the context is still valid, if not, abort the cleanup
-		if ctx.Err() != nil {
-			errs = append(errs, fmt.Errorf("cleanup aborted due to context error: %w", ctx.Err()))
+		if err := context.Cause(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("cleanup aborted due to context error: %w", err))
 			break
 		}
 		logrus.Info("* ", step.Name())
--- a/pkg/cleanup/cleanup.go
+++ b/pkg/cleanup/cleanup.go
@@ -4,6 +4,7 @@
 package cleanup
 
 import (
+	"context"
 	"errors"
 	"fmt"
 
@@ -25,12 +26,18 @@ func NewConfig(debug bool, k0sVars *config.CfgVars, systemUsers *k0sv1beta1.Syst
 	return &Config{cleanupSteps: steps}, nil
 }
 
-func (c *Config) Cleanup() error {
+func (c *Config) Cleanup(ctx context.Context) error {
 	var errs []error
 
 	for _, step := range c.cleanupSteps {
+		// Some steps are not-so context aware, before each step,
+		// check if the context is still valid, if not, abort the cleanup
+		if ctx.Err() != nil {
+			errs = append(errs, fmt.Errorf("cleanup aborted due to context error: %w", ctx.Err()))
+			break
+		}
 		logrus.Info("* ", step.Name())
-		if err := step.Run(); err != nil {
+		if err := step.Run(ctx); err != nil {
 			logrus.Debug(err)
 			errs = append(errs, err)
 		}
@@ -44,7 +51,7 @@ func (c *Config) Cleanup() error {
 // Step interface is used to implement cleanup steps
 type Step interface {
 	// Run impelements specific cleanup operations
-	Run() error
+	Run(ctx context.Context) error
 	// Name returns name of the step for conveninece
 	Name() string
 }
--- a/pkg/cleanup/containers.go
+++ b/pkg/cleanup/containers.go
@@ -10,6 +10,8 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/k0sproject/k0s/pkg/component/worker"
 	workerconfig "github.com/k0sproject/k0s/pkg/component/worker/config"
@@ -79,23 +81,39 @@ func (c *containers) stopAllContainers(ctx context.Context) error {
 		}
 	}
 
+	var wg sync.WaitGroup
+	var errsMu sync.Mutex
 	for _, pod := range pods {
-		logrus.Debugf("stopping container: %v", pod)
-		err := c.containerRuntime.StopContainer(ctx, pod)
-		if err != nil {
-			if strings.Contains(err.Error(), "443: connect: connection refused") {
-				// on a single node instance, we will see "connection refused" error. this is to be expected
-				// since we're deleting the API pod itself. so we're ignoring this error
-				logrus.Debugf("ignoring container stop err: %v", err.Error())
-			} else {
-				errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
+		wg.Go(func() {
+			logrus.Debugf("stopping container: %v", pod)
+			// StopContainer can take a long time if the container is not responding,
+			// IO pressure is high (umount can take 10+ secs) or other such cases,
+			// so we set a timeout to avoid hanging indefinitely
+			stopCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+			defer cancel()
+			err := c.containerRuntime.StopContainer(stopCtx, pod)
+			if err != nil {
+				if strings.Contains(err.Error(), "443: connect: connection refused") {
+					// on a single node instance, we will see "connection refused" error. this is to be expected
+					// since we're deleting the API pod itself. so we're ignoring this error
+					logrus.Debugf("ignoring container stop err: %v", err.Error())
+				} else {
+					errsMu.Lock()
+					errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
+					errsMu.Unlock()
+				}
 			}
-		}
-		err = c.containerRuntime.RemoveContainer(ctx, pod)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to remove pod %s: %w", pod, err))
-		}
+			removeCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+			defer cancel()
+			err = c.containerRuntime.RemoveContainer(removeCtx, pod)
+			if err != nil {
+				errsMu.Lock()
+				errs = append(errs, fmt.Errorf("failed to remove pod %s: %w", pod, err))
+				errsMu.Unlock()
+			}
+		})
 	}
+	wg.Wait()
 
 	pods, err = c.containerRuntime.ListContainers(ctx)
 	if err == nil && len(pods) == 0 {
--- a/pkg/cleanup/containers.go
+++ b/pkg/cleanup/containers.go
@@ -33,9 +33,8 @@ func (c *containers) Name() string {
 
 // Run removes all the pods and mounts and stops containers afterwards
 // Run starts containerd if custom CRI is not configured
-func (c *containers) Run() error {
+func (c *containers) Run(ctx context.Context) error {
 	if c.managedContainerd != nil {
-		ctx := context.TODO()
 		if err := c.managedContainerd.Init(ctx); err != nil {
 			logrus.WithError(err).Warn("Failed to initialize containerd, skipping container cleanup")
 			return nil
@@ -51,18 +50,17 @@ func (c *containers) Run() error {
 		}()
 	}
 
-	if err := c.stopAllContainers(); err != nil {
+	if err := c.stopAllContainers(ctx); err != nil {
 		logrus.Debugf("error stopping containers: %v", err)
 	}
 
 	return nil
 }
 
-func (c *containers) stopAllContainers() error {
+func (c *containers) stopAllContainers(ctx context.Context) error {
 	var errs []error
 
 	var pods []string
-	ctx := context.TODO()
 	err := retry.Do(func() error {
 		logrus.Debugf("trying to list all pods")
 		var err error
--- a/cmd/reset/reset.go
+++ b/cmd/reset/reset.go
@@ -27,7 +27,7 @@ func NewResetCmd() *cobra.Command {
 			if err != nil {
 				return err
 			}
-			err = uninstall.Run(uninstall.Options{
+			err = uninstall.Run(cmd.Context(), uninstall.Options{
 				Vars:      opts.K0sVars,
 				CriSocket: opts.CriSocket,
 				Debug:     debugFlags.IsDebug(),
--- a/pkg/cleanup/bridge_linux.go
+++ b/pkg/cleanup/bridge_linux.go
@@ -4,6 +4,7 @@
 package cleanup
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/vishvananda/netlink"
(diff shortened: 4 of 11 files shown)


pkg/cleanup/cleanup.go
```diff
@@ -25,12 +26,18 @@ func NewConfig(debug bool, k0sVars *config.CfgVars, systemUsers *k0sv1beta1.Syst
...
 
-func (c *Config) Cleanup() error {
+func (c *Config) Cleanup(ctx context.Context) error {
 	var errs []error
 
 	for _, step := range c.cleanupSteps {
+		// Some steps are not-so context aware, before each step,
+		// check if the context is still valid, if not, abort the cleanup
+		if ctx.Err() != nil {
+			errs = append(errs, fmt.Errorf("cleanup aborted due to context error: %w", ctx.Err()))
```
twz123: Let's use `context.Cause` instead:


```suggestion
		if err := context.Cause(ctx); err != nil {
			errs = append(errs, fmt.Errorf("cleanup aborted due to context error: %w", err))
```

pkg/cleanup/containers.go
```diff
@@ -83,7 +82,11 @@ func (c *containers) stopAllContainers() error {
 
 	for _, pod := range pods {
 		logrus.Debugf("stopping container: %v", pod)
-		err := c.containerRuntime.StopContainer(ctx, pod)
+		// StopContainer can take a long time if the container is not responding,
+		// IO pressure is high (umount can take 10+ secs) or other such cases,
+		// so we set a timeout to avoid hanging indefinitely
```
twz123: I wonder if it'd make sense to run this loop in parallel instead of hard-coding a per-pod timeout? Do we actually have a proof that we may be hanging indefinitely here? If yes, then I'd really opt for stopping pods concurrently, so users can decide how long they want to overall wait for the cleanup to finish, without having a single pod blocking the others from making progress.

pkg/cleanup/containers.go
```diff
@@ -83,7 +82,11 @@ func (c *containers) stopAllContainers() error {
...
+		podCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
twz123: Which of the two below commands (stop/remove) is actually the one that might take very long? Do we want to add a joint timeout to both of them, or should each of them get a fresh one instead?

twz123 (dismissed):

pkg/cleanup/containers.go
jnummelin: The original issue is not clear on which step it is hanging on unfortunately. My guess is the stop one, if there's something totally borked running in the container that does not willingly shutdown.

I don't have strong opinion either way, I'm ok with a joint timeout or a separate ones. Can't really judge which one is better than the other. 🤷

pkg/cleanup/containers.go
jnummelin: Yes, the original issue had stack trace where the goroutine was hanging indefinitely here. But that was before the context was wired in and things got stuck within the grpc calls over CRI

pkg/cleanup/containers.go
jnummelin: Split this now into 2 separate 30sec timeouts

jnummelin:
@twz123 PTAL again

pkg/cleanup/containers.go
```diff
@@ -81,23 +81,38 @@ func (c *containers) stopAllContainers() error {
...
+			// StopContainer can take a long time if the container is not responding,
+			// IO pressure is high (umount can take 10+ secs) or other such cases,
+			// so we set a timeout to avoid hanging indefinitely
+			stopCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+			defer cancel()
+			err := c.containerRuntime.StopContainer(stopCtx, pod)
+			if err != nil {
+				if strings.Contains(err.Error(), "443: connect: connection refused") {
+					// on a single node instance, we will see "connection refused" error. this is to be expected
+					// since we're deleting the API pod itself. so we're ignoring this error
```
twz123, line 98: Unrelated to this PR: I think this is looong outdated 🙈

pkg/cleanup/containers.go
```diff
@@ -81,23 +81,38 @@ func (c *containers) stopAllContainers() error {
...
-		err := c.containerRuntime.StopContainer(ctx, pod)
-		if err != nil {
-			if strings.Contains(err.Error(), "443: connect: connection refused") {
-				// on a single node instance, we will see "connection refused" error. this is to be expected
-				// since we're deleting the API pod itself. so we're ignoring this error
-				logrus.Debugf("ignoring container stop err: %v", err.Error())
-			} else {
-				errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
+		wg.Add(1)
+		go func(pod string) {
```
twz123: I think we can just use the outer pod and don't need to pass it along via a function arg.

pkg/cleanup/containers.go
```diff
@@ -81,23 +81,38 @@ func (c *containers) stopAllContainers() error {
...
+				} else {
+					errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
+				}
 			}
-		}
-		err = c.containerRuntime.RemoveContainer(ctx, pod)
-		if err != nil {
-			errs = append(errs, fmt.Errorf("failed to remove pod %s: %w", pod, err))
-		}
+			removeCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
twz123, line 106: Isn't this assignment to `cancel` overwriting the previous cancel function, and hence making line 96 cancel the remove context instead of the stop context? 🤔

pkg/cleanup/containers.go
```diff
@@ -81,23 +81,38 @@ func (c *containers) stopAllContainers() error {
...
+			stopCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+			defer cancel()
+			err := c.containerRuntime.StopContainer(stopCtx, pod)
+			if err != nil {
+				if strings.Contains(err.Error(), "443: connect: connection refused") {
+					// on a single node instance, we will see "connection refused" error. this is to be expected
+					// since we're deleting the API pod itself. so we're ignoring this error
+					logrus.Debugf("ignoring container stop err: %v", err.Error())
+				} else {
+					errs = append(errs, fmt.Errorf("failed to stop running pod %s: %w", pod, err))
```
twz123, line 102: Pretty sure this needs a mutex?

pkg/cleanup/containers.go
jnummelin: No, the defer registers the func at _that time_. So both will be called. See:
```
package main

import "fmt"

func main() {
	defFunc := func() {
		fmt.Println("defer 1 function called")
	}
	defer defFunc()

	defFunc = func() {
		fmt.Println("defer 2 function called")
	}
	defer defFunc()

	fmt.Println("Hello, I'm done. See the defers:")
}
```
Which outputs:
```
% go run defer-test.go 
Hello, I'm done. See the defers:
defer 2 function called
defer 1 function called
```

pkg/cleanup/containers.go
jnummelin: right, that was fixed in golang itself a while ago, just my muscle memory doing things the old way....

jnummelin: 1.25 also added waithgroup.Go to avoid the boilerplate wg.Add/Done stuff, will adjust to that too to simplify a bit

pkg/cleanup/containers.go
twz123: TIL! So this applies only when using something like `defer func() { /* use some local var here ... */ }()` I guess.

pkg/cleanup/containers.go
twz123: Right :ok_hand:

pkg/cleanup/containers.go
jnummelin: absolutely

twz123 (approved):

PR #8080, opened 2026-08-12
Pass kine endpoint via env var instead of arg to hide secrets

mback2k:
## Description

Since a kine endpoint can contain secrets like DB passwords, it should not be passed as an argument visible to any OS process and user. Instead using an environment variable which is supported since https://github.com/k3s-io/kine/commit/992a97f2e45812ea12350d803ed7efa1800aaddd (shipped since https://github.com/k3s-io/kine/releases/tag/v0.14.1) can avoid leaking such secrets.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/kine.go
+++ b/pkg/component/controller/kine.go
@@ -106,7 +106,6 @@ func (k *Kine) Start(ctx context.Context) error {
 	logrus.Info("Starting kine")
 
 	args := stringmap.StringMap{
-		"--endpoint": k.Config.DataSource,
 		// NB: kine doesn't parse URLs properly, so construct potentially
 		// invalid URLs that are understood by kine.
 		// https://github.com/k3s-io/kine/blob/v0.16.3/pkg/util/network.go#L5-L13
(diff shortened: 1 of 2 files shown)


pkg/component/controller/kine.go
```diff
@@ -132,7 +131,7 @@ func (k *Kine) Start(ctx context.Context) error {
 		UID:     k.uid,
 		GID:     kineGID,
 	}
-
+	os.Setenv("KINE_ENDPOINT", k.Config.DataSource)
```
jnummelin: I wonder if we could/should wire up the supervisor to take in envs. 🤔  That way we could push the env only to the kine process. Yes, it will get passed to the kine process too now but it will get passed to all sub-processes since supervisor copies all the envs. IMO that'd kinda follow the least exposure way.

mback2k: I thought about the same initially, but wasn't sure if I would be allowed to introduce such change without first discussing it. So, would that be accepted?

jnummelin: IMO that'd make sense!

mback2k: Okay, thanks. I tried to implement it now in a quite straightforward way. This way the Supervisor specific arguments take precedence over the os Environment, but still support being overridden via the prefix.

PR #8082, opened 2026-08-13
Add hack/list-images tool and make target

ncopa:
Prints every image (with version) that k0s may pull, across all platforms and configurations. Unlike `k0s airgap list-images`, it only touches pkg/airgap and pkg/apis/k0s/v1beta1, so it builds and runs on any host OS without needing a running node or a linux build of k0s.

## Description

Fixes # (issue)

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/hack/list-images/main.go
+++ b/hack/list-images/main.go
@@ -1,0 +1,46 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+// list-images prints every container image (and its version) that k0s may
+// pull, across all platforms and configurations. Unlike `k0s airgap
+// list-images`, it doesn't need a running node or a linux host to build/run,
+// since it only touches pkg/airgap and pkg/apis/k0s/v1beta1 directly.
+package main
+
+import (
+	"fmt"
+	"slices"
+
+	"github.com/k0sproject/k0s/pkg/airgap"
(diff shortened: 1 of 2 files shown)


hack/list-images/main.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+func main() {
+	spec := v1beta1.DefaultClusterSpec()
+
+	seen := make(map[string]struct{})
+	var uris []string
+	for _, platform := range []imagespecv1.Platform{
+		{OS: "linux", Architecture: "amd64"},
+		{OS: "windows", Architecture: "amd64"},
```
jnummelin, line 32: What about `arm64` and `arm/v7`?

ncopa: I assume we don't use different images/versions for those? I suppose it may be good to scan them anyway.

jnummelin (approved):

twz123:
What's the use case for this?

ncopa:
> What's the use case for this?

To simplify tooling that can answer: which images used by k0s are flagged as vulnerable by scanners? With this we can get the list of images from mac without need to build and run the k0s binary in a VM.

twz123:
Hmm, that's better off in the image-builder repo IMO.

PR #8085, opened 2026-08-13
Bump helm.sh/helm to v4.2.4 

soltysh:
## Description

This PR bumps helm to v4.2.4, replacing https://github.com/k0sproject/k0s/pull/6638

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

## Checklist


pschichtel:
Not sure if it matters much here, but I got recently surprised by this: https://github.com/helm/helm/issues/32522

pkg/component/controller/extensions_controller.go
```diff
@@ -495,35 +497,39 @@ func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv
 
 	if chart.Status.ReleaseName == "" {
 		isInstalling = true
-		status := release.StatusUnknown
+		status := common.StatusUnknown
+		version := -1
```
jnummelin: Does this have any deeper meaning on Helm internals? If so, is there some const we could use? Magic numbers a kinda meh 😄

pkg/component/controller/extensions_controller.go
```diff
@@ -44,9 +44,11 @@ import (
 	"github.com/Masterminds/sprig"
 	"github.com/bombsimon/logrusr/v4"
 	"github.com/sirupsen/logrus"
-	"helm.sh/helm/v3/pkg/registry"
-	"helm.sh/helm/v3/pkg/release"
-	"helm.sh/helm/v3/pkg/storage/driver"
+	ci "helm.sh/helm/v4/pkg/chart"
```
twz123: Not sure about `ci`. What about naming this `pkgchart`, or just `charts`? Same for the  usages of this package in other files.

pkg/component/controller/extensions_controller.go
```diff
@@ -427,7 +429,7 @@ func (cr *ChartReconciler) loadAndMergeRepositoryConfig(ctx context.Context, cha
 
 func (cr *ChartReconciler) updateOrInstallChart(ctx context.Context, chart helmv1beta1.Chart) (err error) {
 	var isInstalling bool
-	var chartRelease *release.Release
+	var chartReleaser release.Releaser
```
twz123: WDYT about keeping the old variable name?

pkg/component/controller/extensions_controller.go
```diff
@@ -681,21 +687,30 @@ func hashChartValues(chart *helmv1beta1.Chart) string {
...
-			updchart.Status.ReleaseName = chartRelease.Name
+			updchart.Status.ReleaseName = chartRelease.Name()
 		}
-		updchart.Status.Version = chartRelease.Chart.Metadata.Version
-		updchart.Status.AppVersion = chartRelease.Chart.AppVersion()
-		updchart.Status.Revision = int64(chartRelease.Version)
-		updchart.Status.Namespace = chartRelease.Namespace
+		metadata := ac.MetadataAsMap()
+		updchart.Status.Version = fmt.Sprintf("%s", metadata["Version"])
+		updchart.Status.AppVersion = fmt.Sprintf("%s", metadata["AppVersion"])
```
twz123: Just to harden things a bit, as the previous version could produce strings like `%!s(<nil>)` or `%!s(int=1)` in pathological cases:

```suggestion
		updchart.Status.Version = fmt.Sprint(cmp.Or(metadata["Version"], ""))
		updchart.Status.AppVersion = fmt.Sprint(cmp.Or(metadata["AppVersion"], ""))
```

pkg/component/controller/extensions_controller.go
```diff
@@ -681,21 +687,30 @@ func hashChartValues(chart *helmv1beta1.Chart) string {
...
+func (cr *ChartReconciler) updateStatus(ctx context.Context, chart helmv1beta1.Chart, chartReleaser release.Releaser, installed bool, err error) error {
 	nsn := types.NamespacedName{Namespace: chart.Namespace, Name: chart.Name}
 	var updchart helmv1beta1.Chart
 	if err := cr.Get(ctx, nsn, &updchart); err != nil {
 		return fmt.Errorf("can't get updated version of chart %s: %w", chart.Name, err)
 	}
 	chart.Spec.YamlValues() // XXX what is this function for ?
-	if chartRelease != nil {
+	if chartReleaser != nil {
+		chartRelease, accErr := release.NewAccessor(chartReleaser)
```
twz123: WDYT about just shadowing the outer `err`and rename `accErr` to `err` here?

pkg/component/controller/extensions_controller.go
```diff
@@ -681,21 +687,30 @@ func hashChartValues(chart *helmv1beta1.Chart) string {
...
+		if accErr != nil {
+			return fmt.Errorf("can't get updated version of chart %s: %w", chart.Name, accErr)
+		}
+		ac, accErr := ci.NewAccessor(chartRelease.Chart())
```
twz123: Maybe rename `ac` to `chartAccessor`?

pkg/component/controller/extensions_controller.go
```diff
@@ -681,21 +687,30 @@ func hashChartValues(chart *helmv1beta1.Chart) string {
...
 	var updchart helmv1beta1.Chart
 	if err := cr.Get(ctx, nsn, &updchart); err != nil {
 		return fmt.Errorf("can't get updated version of chart %s: %w", chart.Name, err)
 	}
 	chart.Spec.YamlValues() // XXX what is this function for ?
-	if chartRelease != nil {
+	if chartReleaser != nil {
+		chartRelease, accErr := release.NewAccessor(chartReleaser)
+		if accErr != nil {
+			return fmt.Errorf("can't get updated version of chart %s: %w", chart.Name, accErr)
```
twz123: We could replace `chart.Name` with `nsn` to include the chart namespace in the error message. Same for the other error returns in this method.

twz123: What about something along the lines of `"can't access updated release of chart %s: %w"`? Or use a more idiomatic wording `"failed to access updated release of chart %s: %w"`, and then, a few lines below `"failed to access updated chart %s: %w"`?

pkg/helm/helm.go
```diff
@@ -270,18 +276,24 @@ func (hc *Commands) initRepository(repoCfg Repository) error {
 	return nil
 }
 
-func (hc *Commands) downloadDependencies(chart *chart.Chart, chartPath string, registryClient *registry.Client) error {
-	if chart.Metadata.Dependencies == nil {
+func (hc *Commands) downloadDependencies(chart *chartv2.Chart, chartPath string, registryClient *registry.Client) error {
```
twz123: We just need the interface here.

```suggestion
func (hc *Commands) downloadDependencies(chart ci.Charter, chartPath string, registryClient *registry.Client) error {
```

pkg/helm/helm.go
twz123: Just a nit, but it somehow feels odd to directly reference a versioned type, now that we have all these accessors...

So what we could do is to replace this method by this function:

```go
func isChartInstallable(chartType string) bool {
	return chartType == "" || chartType == "application"
}
```

We could then use it at the call sites like so `isChartInstallable(loadedChart.Metadata.Type)`, and change the parameter type of `downloadDependencies` to `ci.Charter`.

pkg/helm/helm.go
```diff
@@ -292,6 +304,24 @@ func (hc *Commands) downloadDependencies(chart *chart.Chart, chartPath string, r
 	return nil
 }
 
+// resolveRegistryClient returns the registry client configured for chartName's
+// repository. For OCI charts without a configured repository (anonymous  access),
```
twz123: I found a whitespace typo 😬 

```suggestion
// repository. For OCI charts without a configured repository (anonymous access),
```

pkg/helm/helm.go
```diff
@@ -292,6 +304,24 @@ func (hc *Commands) downloadDependencies(chart *chart.Chart, chartPath string, r
...
+// it falls back to an anonymous registry client: Helm v4 requires a non-nil
+// RegistryClient to resolve any OCI chart reference.
+func (hc *Commands) resolveRegistryClient(chartName string) (*registry.Client, error) {
+	client, err := hc.registryManager.GetRegistryClient(chartName)
+	if err != nil {
+		return nil, fmt.Errorf("can't get registry client for chart `%s`: %w", chartName, err)
```
twz123: I usually tend to only include information in errors which are unknown to the caller. So instead of including the chart name here, I'd say it's up to the caller to add it, if needed.

```suggestion
		return nil, fmt.Errorf("failed to get registry client: %w", err)
```

Same for the next error in this method.

pkg/helm/helm.go
```diff
@@ -292,6 +304,24 @@ func (hc *Commands) downloadDependencies(chart *chart.Chart, chartPath string, r
...
+	}
+	if client == nil && registry.IsOCI(chartName) {
+		client, err = registry.NewClient(registry.ClientOptWriter(os.Stdout))
```
twz123: I reckon `ClientOptWriter` sets some sort of a logging backend? Then we should probably use `internal/pkg/log.NewWriter(...)` instead of stdout here.

pkg/helm/helm.go
twz123: Random find in untouched, but related code: This should probably go through a logger via `internal/pkg/log.NewWriter(...)`.

pkg/component/controller/extensions_controller.go
twz123: Why not just `var version int`?

pkg/helm/helm.go
```diff
@@ -378,15 +405,15 @@ func (hc *Commands) InstallChart(ctx context.Context, chartName string, version
...
 	if err != nil {
-		return nil, fmt.Errorf("can't get registry client for chart `%s`: %w", chartName, err)
+		return nil, err
 	}
 
 	install := action.NewInstall(cfg)
 	install.CreateNamespace = true
 	install.WaitForJobs = true
-	install.Wait = true
+	install.WaitStrategy = kube.LegacyStrategy
```
twz123: I wonder if we should switch to `kube.StatusWatcherStrategy`, along with `install.WaitOptions = []kube.WaitOption{kube.WithWaitContext(ctx)}` here and in the other Helm commands. I guess this will allow us to completely drop the stuff in `client.go` that was basically duct tape around that missing functionality in v3. IIUC our custom wait function implemetations aren't even called anymore in v4.

<details>
<summary>Asked the word salad machine about it, and it agrees.</summary>

> Short answer: **no, you don't need it anymore — but as the PR stands, the custom client isn't just redundant, it's silently dead code, and the interruptible waits it was added for are gone.**
> 
> ## What actually happens with the PR as written
> 
> In Helm v4 the action package never calls `KubeClient.Wait*` directly. Every wait site does this instead (`pkg/action/install.go:536-540`, `upgrade.go:483-487`, `uninstall.go:67-72`, `hooks.go:110-114`, `rollback.go:256-260`):
> 
> ```go
> if c, ok := i.cfg.KubeClient.(kube.InterfaceWaitOptions); ok {
>     waiter, err = c.GetWaiterWithOptions(i.WaitStrategy, i.WaitOptions...)
> } else {
>     waiter, err = i.cfg.KubeClient.GetWaiter(i.WaitStrategy)
> }
> waiter.WaitWithJobs(resources, i.Timeout)
> ```
> 
> k0s's `client` embeds `*kube.Client`, so `GetWaiterWithOptions` is promoted and Helm gets a stock waiter back. With `kube.LegacyStrategy` that's `&legacyWaiter{kubeClient: kc, ctx: c.WaitContext}` (`pkg/kube/client.go:194`) — and `WaitContext` is never set, so the wait runs on `context.Background()` + timeout (`wait.go:52`, `statuswait.go:314-318`). The `Wait`/`WaitWithJobs`/`WaitForDelete` overrides in `pkg/helm/client.go:43-105` are never reached by any Helm action. `client_test.go` still passes only because it calls `underTest.Wait(...)` directly.
> 
> Net effect: `pkg/helm/helm.go:400` cancels `actionCtx` on method exit, and the transport control will still kill in-flight requests, but a poll loop that's mid-`time.Sleep` no longer sees the cancellation — the legacy waiter treats the transport errors as retryable and retries up to 30× at 2s (`wait.go:62-72`) before giving up. Not catastrophic, but it's exactly the leak `84e88217ac9` was written to close.
> 
> ## What replaces it
> 
> Helm v4 grew the plumbing that the custom client was papering over:
> 
> - `kube.WithWaitContext(ctx)` (`pkg/kube/options.go:29`) → passed via `install.WaitOptions` / `upgrade.WaitOptions` / `uninstall.WaitOptions`.
> - Those options propagate to hooks (`hooks.go:36,60`) and to the rollback triggered by `RollbackOnFailure` (`upgrade.go:585-586`), so the whole action lifecycle honors the context.
> - The deprecated `kube.Client.WaitContext` field (`client.go:87-92`) is the only knob the **legacy** waiter reads — `LegacyStrategy` ignores `WaitOptions` entirely.
> 
> So two viable shapes:
> 
> **A. Minimal / behavior-preserving:** keep `LegacyStrategy`, delete the overrides, set `wrapped.WaitContext = ctx` in `newKubeClient`. The legacy waiter's semantics are near-identical to the k0s one — `isRetryableHTTPStatusCode` (`wait.go:109`) is literally the negation of the `stop` expression at `client.go:126`, and `PausedAsReady(true)` matches. Differences: 30-retry cap per resource instead of retry-until-deadline; `ctx.Err()` instead of `context.Cause`; `WaitForDelete` aborts on any non-NotFound error (`wait.go:126-128`) instead of polling on. Downside: relies on a field already marked deprecated.
> 
> **B. Recommended:** switch to `kube.StatusWatcherStrategy` (Helm 4's default for `--wait`) and set `WaitOptions: []kube.WaitOption{kube.WithWaitContext(actionCtx)}` on install/upgrade/uninstall. Then the whole `client.go` + `client_test.go` go away and `getActionCfg` just uses `kube.New(getter)`. Things to be aware of:
> - Readiness is kstatus-based watch rather than polling; `Failed` is treated as terminal (`statuswait.go:339-342`), so a crash-looping deployment fails fast instead of burning the whole timeout. Needs list/watch on the resource kinds — fine for k0s.
> - A zero timeout becomes 30s (`DefaultStatusWatcherTimeout`, `statuswait.go:60`). That bites `UninstallRelease` (`helm.go:523-527`), which only sets `Timeout` when the ctx has a deadline — and the reconcile ctx doesn't. With the legacy waiter a zero timeout means "no timeout". Either pass an explicit timeout there or accept the 30s cap.
> - The remaining `wrapped.SetLogger(...)` + transport control stay as they are; they're orthogonal to the waiter.
> 
> Either way, the PR should not merge with `client.go` in its current form — it reads as if interruptibility is preserved when it isn't. If you want, I can draft option B against this worktree so you can see the diff.

</details>

twz123 (commented):
Thanks for taking care of this! 🤟

Besides all the nits, we should definitely have a deeper look at the new v4 `kube.Client` behavior.

twz123:
> Not sure if it matters much here, but I got recently surprised by this: https://github.com/helm/helm/issues/32522

This is a v3 -> v4 regression concerning sub-charts, right? Well yes, if k0s is upgraded to v4, it will of course behave the same as upstream Helm v4. It's kinda expected that there'll be some migration pains. But we need to make the flip sooner than later, as we can't stay on v3 forever. I wouldn't wait for the fix for this particular issue to land before merging this PR. As soon as upstream releases a fixed version, k0s will inherit it. Helm version bumps are automated.

pkg/component/controller/extensions_controller.go
soltysh: Updated.

pkg/component/controller/extensions_controller.go
soltysh: Done.

pkg/component/controller/extensions_controller.go
soltysh: I wasn't sure, but after recently digging through helm code, I found they care about `version > 0` https://github.com/helm/helm/blob/c0039291c1654ea3178ca7aebc449a980b70a719/pkg/action/action.go#L596.

pkg/component/controller/extensions_controller.go
soltysh: Done.

pkg/component/controller/extensions_controller.go
soltysh: Done.

soltysh: Done.

pkg/helm/helm.go
soltysh: Updated.

pkg/helm/helm.go
soltysh: Done, I've also changed import to match from the other file, so `charts.Charter`.

pkg/helm/helm.go
soltysh: Fair point - updated accordingly.

pkg/helm/helm.go
soltysh: Done.

pkg/helm/helm.go
soltysh: Updated.

pkg/helm/helm.go
soltysh: Good call, also updated.

pkg/helm/helm.go
```diff
@@ -201,13 +206,13 @@ func (hc *Commands) getActionCfg(ctx context.Context, namespace string) (*action
...
-	driver.Log = log.WithField("helm", "driver").Debugf
+	sloglogger := slog.New(sloghook.NewHandler(log.WithField("helm", "driver").Logger, nil))
+	driver.SetLogger(sloglogger.Handler())
 
 	return &action.Configuration{
 		RESTClientGetter: getter,
 		Releases:         storage.Init(driver),
-		KubeClient:       newKubeClient(ctx, getter, log.WithField("helm", "client")),
-		Log:              log.WithField("helm", "action").Debugf,
+		KubeClient:       kube.New(getter),
```
twz123: Just realizing that both `kube.Client` _and_ `action.Configuration` hold their own logger, just like the `driver.Secrets` above. So we should call `SetLogger` on those, as well. We need to come up with the right field names for them, though.

pkg/config/cfgvars.go
```diff
@@ -46,6 +46,7 @@ type CfgVars struct {
 	// Helm config
 	HelmHome             string
 	HelmRepositoryCache  string
+	HelmContentCache     string
```
twz123, line 49: I don't find any usages of `HelmContentCache` anymore. There have been some in the previous version of this PR, right?

twz123 (commented):
Sorry for all the merge conflicts 🙈

Helm v4 apparently uses slog directly in some places. Could you add an slog default handler somewhere in `internal/pkg/log`?

pkg/config/cfgvars.go
soltysh: It it being used inside `pkg/helm/helm.go`, when constructing `ChartDownloader`, which currently requires both cache and content cache. This particular bits configuring helm aren't used anywhere in the codebase I could find, but I've noticed we're injecting them in several places like:
https://github.com/k0sproject/k0s/blob/a382a2497521563df401eb4fa82c3b8890ec1792/pkg/config/cli.go#L325
or 
https://github.com/k0sproject/k0s/blob/a382a2497521563df401eb4fa82c3b8890ec1792/cmd/install/controller.go#L43
So I figured them might be used indirectly and decided to also add the newly added config option. Should we drop the entire helm configuration in that case?

pkg/helm/helm.go
soltysh: Done. I went initially with just "log.WithField("helm", "action")`, but we could expand the `getActionCfg` with actual action instead, so that each invocation would also pass the name of the action. From what I see we call it in `GetRelease`, `InstallChart`, `UpgradeChart` and `UninstallRelease`. We could easily pass those name there, which will simplify debugging.

soltysh:
> Sorry for all the merge conflicts 🙈

That's part of the open-source life :wink: 

> Helm v4 apparently uses slog directly in some places. Could you add an slog default handler somewhere in `internal/pkg/log`?

Oh, I wasn't aware of that. Added a default slog logger in `internal/pkg/log/k0s.go`.

soltysh:
@twz123 should be ready for another round of reviews.

twz123 (approved):
Diff stats: `+192 -604` 💫

PR #8111, opened 2026-08-17
Fix NewCRDStack silently discarding an explicit stack name

### LINKED ISSUE #8103: etcdmember objects wiped upon join of a new controller node

### Platform

```shell
Linux 6.8.0-137-generic #137-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 17 20:28:23 UTC 2026 x86_64 GNU/Linux
PRETTY_NAME="Ubuntu 24.04.4 LTS"
```

### Version

v1.33.13+k0s.1

### Sysinfo

<pre>
Total memory: 21.8 GiB (pass)                                                                                                                                                                                                         [33/144]
File system of /var/lib/k0s: ext4 (pass)
Disk space available for /var/lib/k0s: 96.5 GiB (pass)
Relative disk space available for /var/lib/k0s: 83% (pass)
Name resolution: localhost: [::1 127.0.0.1] (pass)
  Linux kernel release: 6.8.0-137-generic (pass)
  AppArmor: active (pass)
      CONFIG_NF_NAT: module (pass)
      CONFIG_LLC: module (pass)
      CONFIG_STP: module (pass)
</pre>
</details>


### What happened?

After deployment of the second controller node, etcdmember objects disappear. We rely on these objects to determine whether it is safe to proceed with the deployment of a third controller.

```
$ kubectl get etcdmembers -A
No resources found
$ k0s etcd member-list
{"members":{"master-0":"https://10.100.91.52:2380","master-1":"https://10.100.91.50:2380"}}
```

### Steps to reproduce

1. deploy the first k0s controller node
2. deploy the second one


### Expected behavior

both etcdmember objects exist in kubernetes, the list corresponds to the etcd member-list output.

### Actual behavior

It seems that etcdmember CRD is being recreated upon the join of master-1:

```
$ kubectl get crd etcdmembers.etcd.k0sproject.io -o jsonpath='{.metadata.creationTimestamp}'
2026-08-17T01:54:01Z
```

master-0 has fully initialized around this timeframe
```
TASK [k0s : Wait for k8s-node availability] ************************************
task path: /root/lcm-ansible-0.31.0-59-812d58b5/roles/k0s/tasks/main.yml:234
ok: [master-0] => {
    "attempts": 3,
    "changed": false,
    "cmd": [
        "/usr/local/bin/kubectl",
        "get",
        "node",
        "master-0"
    ],
    "delta": "0:00:00.341634",
    "end": "2026-08-17 01:45:04.961757",
    "invocation": {
        "module_args": {
            "_raw_params": "/usr/local/bin/kubectl get node master-0",
            "_uses_shell": false,
            "argv": null,
            "chdir": null,
            "creates": null,
            "executable": null,
            "expand_argument_vars": true,
            "removes": null,
            "stdin": null,
            "stdin_add_newline": true,
            "strip_empty_ends": true
        }
    },
    "rc": 0,
    "start": "2026-08-17 01:45:04.620123"
}

STDOUT:

NAME       STATUS     ROLES    AGE   VERSION
master-0   NotReady   <none>   0s    v1.33.13+k0s
```

master-1 has fully initialized around this timeframe
```    
TASK [k0s : Wait for k8s-node availability] ************************************
task path: /root/lcm-ansible-0.31.0-59-812d58b5/roles/k0s/tasks/main.yml:234
ok: [master-1] => {
    "attempts": 8,
    "changed": false,
    "cmd": [
        "/usr/local/bin/kubectl",
        "get",
        "node",
        "master-1"
    ],
    "delta": "0:00:00.294705",
    "end": "2026-08-17 01:54:05.150578",
    "invocation": {
        "module_args": {
            "_raw_params": "/usr/local/bin/kubectl get node master-1",
            "_uses_shell": false,
            "argv": null,
            "chdir": null,
            "creates": null,
            "executable": null,
            "expand_argument_vars": true,
            "removes": null,
            "stdin": null,
            "stdin_add_newline": true,
            "strip_empty_ends": true
        }
    },
    "rc": 0,
    "start": "2026-08-17 01:54:04.855873"
}

STDOUT:

NAME       STATUS     ROLES    AGE   VERSION
master-1   NotReady   <none>   3s    v1.33.13+k0s
```

### Screenshots and logs

Attaching the k0scontroller logs from both machines

[master-0-k0scontroller.stripped.log](https://github.com/user-attachments/files/31141981/master-0-k0scontroller.stripped.log)
[master-1-k0scontroller.log](https://github.com/user-attachments/files/31141833/master-1-k0scontroller.log)

master-0 controller log is stripped to the first 40000 lines to allow upload.

### Additional context

Claude suggests it might be a race between starting up applier-manager and creating etcdmembers manifest during `CRD.Start()`.

Restart of k0scontroller service solves the problem.

pujitha24:
## Description

While investigating the `CRD.Start()` code path named in issue #8103, I found that
`NewCRDStack` only defaulted the stack name when `assetsDir` was also empty. Any
caller that supplies `WithStackName` but not `WithCRDAssetsDir` has its explicit
stack name silently overwritten back to the bundle name. The etcd-member CRD stack
hits exactly this case: it's constructed with `WithStackName(EtcdMemberStackName)`
but no `WithCRDAssetsDir`, so it ends up applied and labeled
`k0s.k0sproject.io/stack=etcd` instead of the intended `etcd-member`.

This PR defaults the stack name on its own, independent of `assetsDir`.
`WithCRDAssetsDir`/`assetsDir` has had no callers for a while now; its removal here
is dead-code cleanup, not a consequence of the fix.

Resources already applied under the old, wrong stack name kept their stale label,
since a checksum match alone used to be enough to skip an update. The stack now
compares the stack name too, so a stale label gets corrected through the normal
update/patch path.

The same defaulting bug (introduced in v1.34, before the in-memory CRD stack
landed in v1.35) previously made the filesystem-based CRD component write the
etcd CRD manifest on disk under the wrong, bundle-derived stack name (`etcd`)
instead of `etcd-member`. Clusters that ran one of those versions may still have
that leftover manifest and directory around, which would otherwise compete with
the in-memory etcd-member stack over the same CRD's stack-name label. This PR
also removes that leftover manifest/directory on controller start, before the
applier manager gets a chance to scan the manifests directory.

**This fix is not confirmed to resolve the CRD-wipe symptom reported in #8103.**
The reporter's cluster runs v1.33.13, which predates the defaulting bug itself
(introduced in v1.34) as well as the in-memory CRD-stack mechanism (introduced in
v1.35) — so none of the code paths described above apply to their version, and I
could not reproduce the reported wipe there. This is a genuine, independently
provable bug found while reading the code the issue points at, fixed on its own
merits. Given that, this PR does not close #8103; see #8103 for the original report.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

Updated `TestNewCRDStack_WithStackName` and `TestNewCRDStack_Defaults` in
`pkg/component/controller/crd_test.go` to use testify assertions. Added
`TestStack_StackNameChangeTriggersUpdateWithoutChecksumChange` in
`pkg/applier/stack_test.go`, which verifies a resource whose stack-name label
doesn't match the current stack gets updated even when its checksum is unchanged.

The leftover-manifest cleanup is inlined directly at its call site in
`cmd/controller/controller.go` rather than living in its own function, so it isn't
covered by a dedicated unit test; I verified locally that `os.Remove` on a
non-empty directory returns an error matching `syscall.ENOTEMPTY` via `errors.Is`,
which the non-empty-directory branch relies on.

Also ran:

- `go build ./cmd/controller/... ./pkg/component/controller/... ./pkg/applier/...`
- `go vet ./cmd/controller/... ./pkg/component/controller/... ./pkg/applier/...`
- `go test ./cmd/controller/... ./pkg/component/controller/... ./pkg/applier/...` (no regressions)
- `golangci-lint run ./cmd/controller/... ./pkg/component/controller/... ./pkg/applier/...` (0 issues)
- `gofmt -l` on changed files (clean)

Not run: `make codegen` (no generated files touched by this change) and no live
cluster / e2e reproduction (this is a pure unit-level fix plus a small filesystem
cleanup, not something that requires a live cluster to validate).

## Checklist

Relates to #8103

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -17,6 +17,7 @@ import (
 	"path/filepath"
 	"slices"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/k0sproject/k0s/cmd/internal"
@@ -366,6 +367,19 @@ func (c *command) start(ctx context.Context, runtimeConfig *config.RuntimeConfig
 		leaderElector = leaderelector.NewLeasePool(c.K0sVars.InvocationID, adminClientFactory, "k0s-endpoint-reconciler")
 	}
 	nodeComponents.Add(ctx, leaderElector)
+
+	// TODO: Remove in v1.38+. Clean up the bogus on-disk etcd CRD manifest
+	// left behind by affected versions of k0s, so it doesn't fight the
+	// in-memory etcd-member stack over the CRD's stack-name label.
+	{
+		bogusEtcdDir := filepath.Join(c.K0sVars.ManifestsDir, "etcd")
+		bogusEtcdManifest := filepath.Join(bogusEtcdDir, "etcd-crd-etcd.k0sproject.io_etcdmembers.yaml")
+		if err := os.Remove(bogusEtcdManifest); err != nil && !os.IsNotExist(err) {
+			logrus.WithError(err).Warn("Failed to remove bogus etcd CRD manifest")
+		} else if err := os.Remove(bogusEtcdDir); err != nil && !os.IsNotExist(err) && !errors.Is(err, syscall.ENOTEMPTY) {
+			logrus.WithError(err).Warn("Failed to remove bogus etcd CRD manifest directory")
+		}
+	}
 
 	if !slices.Contains(flags.DisableComponents, constant.ApplierManagerComponentName) {
 		nodeComponents.Add(ctx, &applier.Manager{
--- a/pkg/applier/stack.go
+++ b/pkg/applier/stack.go
@@ -132,8 +132,10 @@ func (s *Stack) Apply(ctx context.Context, prune bool) error {
 			continue
 		} else { // The resource already exists, we need to update/patch it
 			localChecksum := resource.GetAnnotations()[ChecksumAnnotation]
-			if serverResource.GetAnnotations()[ChecksumAnnotation] == localChecksum {
-				s.log.Debug("resource checksums match, no need to update")
+			checksumMatches := serverResource.GetAnnotations()[ChecksumAnnotation] == localChecksum
+			stackNameMatches := serverResource.GetLabels()[NameLabel] == s.Name
+			if checksumMatches && stackNameMatches {
+				s.log.Debug("resource checksum and stack name match, no need to update")
 				s.keepResource(resource)
 				continue
 			}
--- a/pkg/applier/stack_test.go
+++ b/pkg/applier/stack_test.go
@@ -50,3 +50,42 @@ func TestStack_StripsNamespaceFromClusterScopedResource(t *testing.T) {
 	require.NoError(t, err)
 	assert.Empty(t, appliedNs.Namespace)
 }
+
+func TestStack_StackNameChangeTriggersUpdateWithoutChecksumChange(t *testing.T) {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ns-relabel-test",
+		},
+	}
+	resources, err := applier.ToUnstructured(nil, ns)
+	require.NoError(t, err)
+
+	fakes := testutil.NewFakeClientFactory()
+
+	oldStack := applier.Stack{
+		Name:      "old",
+		Resources: []*unstructured.Unstructured{resources},
+		Clients:   fakes,
+	}
+	require.NoError(t, oldStack.Apply(t.Context(), true))
+
+	applied, err := fakes.Client.CoreV1().Namespaces().Get(t.Context(), "ns-relabel-test", metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Equal(t, "old", applied.Labels[applier.NameLabel])
+	checksumBefore := applied.Annotations[applier.ChecksumAnnotation]
+	require.NotEmpty(t, checksumBefore)
+
+	resources, err = applier.ToUnstructured(nil, ns)
+	require.NoError(t, err)
+	newStack := applier.Stack{
+		Name:      "new",
+		Resources: []*unstructured.Unstructured{resources},
+		Clients:   fakes,
+	}
+	require.NoError(t, newStack.Apply(t.Context(), false))
+
+	applied, err = fakes.Client.CoreV1().Namespaces().Get(t.Context(), "ns-relabel-test", metav1.GetOptions{})
+	require.NoError(t, err)
+	assert.Equal(t, "new", applied.Labels[applier.NameLabel])
+	assert.Equal(t, checksumBefore, applied.Annotations[applier.ChecksumAnnotation])
+}
--- a/pkg/component/controller/crd.go
+++ b/pkg/component/controller/crd.go
@@ -28,7 +28,7 @@ type CRDStack struct {
 }
 
 type crdOpts struct {
-	stackName, assetsDir string
+	stackName string
 }
 
 type CRDOption func(*crdOpts)
@@ -42,9 +42,8 @@ func NewCRDStack(clients kubernetes.ClientFactoryInterface, leaderElector leader
 		opt(&options)
 	}
 
-	if options.assetsDir == "" {
+	if options.stackName == "" {
 		options.stackName = bundle
-		options.assetsDir = bundle
 	}
 
 	return &CRDStack{
@@ -57,10 +56,6 @@ func NewCRDStack(clients kubernetes.ClientFactoryInterface, leaderElector leader
 
 func WithStackName(stackName string) CRDOption {
 	return func(opts *crdOpts) { opts.stackName = stackName }
-}
-
-func WithCRDAssetsDir(assetsDir string) CRDOption {
-	return func(opts *crdOpts) { opts.assetsDir = assetsDir }
 }
 
 // Init implements [manager.Component]. It does nothing.
@@ -76,7 +71,7 @@ func (c *CRDStack) Start(context.Context) error {
 	go func() {
 		defer close(done)
 		leaderelection.RunLeaderTasks(ctx, c.leaderElector.CurrentStatus, func(ctx context.Context) {
-			resources, err := applier.ReadUnstructuredDir(static.CRDs, c.assetsDir)
+			resources, err := applier.ReadUnstructuredDir(static.CRDs, c.bundle)
 			if err != nil {
 				logrus.WithError(err).Errorf("Failed to read %s CRD stack", c.bundle)
 				return
--- a/pkg/component/controller/crd_test.go
+++ b/pkg/component/controller/crd_test.go
@@ -1,0 +1,22 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewCRDStack_WithStackName(t *testing.T) {
+	stack := NewCRDStack(nil, nil, "the-bundle-name", WithStackName("the-stack-name"))
+
+	assert.Equal(t, "the-stack-name", stack.stackName, "stack name should be taken from WithStackName")
+}
+
+func TestNewCRDStack_Defaults(t *testing.T) {
+	stack := NewCRDStack(nil, nil, "the-bundle-name")
+
+	assert.Equal(t, "the-bundle-name", stack.stackName, "stack name should default to bundle name")
+}


jnummelin:
> This fix is not confirmed to resolve the CRD-wipe symptom reported in the issue

So maybe remove the `Fixes ...` from the description 😄

pujitha24:
Fair point, done — swapped `Fixes #8103` for `Relates to #8103` so merging this won't auto-close the issue, since I can't confirm it actually resolves the reported symptom.

pujitha24:
This is still green and rebased on main whenever you get a chance to take another look — happy to adjust anything that would make it easier to review.

pkg/component/controller/crd.go
```diff
@@ -42,8 +42,10 @@ func NewCRDStack(clients kubernetes.ClientFactoryInterface, leaderElector leader
 		opt(&options)
 	}
 
-	if options.assetsDir == "" {
+	if options.stackName == "" {
 		options.stackName = bundle
+	}
+	if options.assetsDir == "" {
```
twz123: AFAICT `WithCRDAssetsDir`is no longer used across the codebase, so maybe we should remove `assetsDir` completely?

twz123:
> One known limitation: on a cluster that already ran the buggy code, the live CRD
> object's stale stack=etcd label is not retroactively corrected by this fix alone,
> since the checksum comparison in pkg/applier/stack.go is computed before the stack
> label is applied, so an unchanged manifest skips the update. This does not create any
> new deletion/pruning risk (the object is still tracked by name during that apply
> pass), it's just cosmetic.

I'd say this is a problem, especially for pruning. Do you see any obstacles in including the stack name into the checksum?

pujitha24:
Removed `assetsDir` entirely in the latest commit — you're right it had no callers left.

On the checksum/pruning point: no fundamental obstacle, the fix would just be moving the `resourceChecksum` call in `prepareResource` to run after the stack label is set instead of before. But `Stack` is shared well beyond this CRD stack — autopilot, clusterconfig, extensions, systemrbac, workerconfig go through it directly, and `applier.Manager` routes every manifests-dir component (calico, coredns, kube-proxy, konnectivity-agent, etc.) through the same checksum code too. So changing what goes into the checksum means basically every stack-managed resource cluster-wide would see a one-time checksum mismatch and get re-patched on the first apply after upgrade, since existing server-side checksums were computed without the label baked in. That's a much bigger behavior change than this PR is meant to make on its own. Happy to send it as a follow-up PR if that seems worth doing, or fold it in here if you'd rather — let me know which you'd prefer.

twz123:
I think we should get it in here, as I think it's very important to get the stack name right in the cluster. Yes, during the k0s upgrade, all the existing resources would get updated once. That's more or less the same as the cluster bootstrap. Such an upgrade will result in a resource usage peak once, and that's it. You think this is a real problem? Otherwise, we could check the stack name in addition to the checksum in order to figure out if a resource needs to be updated ... ¯\\\_(ツ)_/¯

pkg/applier/stack.go
```diff
@@ -132,7 +132,9 @@ func (s *Stack) Apply(ctx context.Context, prune bool) error {
 			continue
 		} else { // The resource already exists, we need to update/patch it
 			localChecksum := resource.GetAnnotations()[ChecksumAnnotation]
-			if serverResource.GetAnnotations()[ChecksumAnnotation] == localChecksum {
+			checksumMatches := serverResource.GetAnnotations()[ChecksumAnnotation] == localChecksum
+			stackNameMatches := serverResource.GetLabels()[NameLabel] == s.Name
+			if checksumMatches && stackNameMatches {
 				s.log.Debug("resource checksums match, no need to update")
```
twz123: The log statement doesn't sound right anymore.

pkg/applier/stack_test.go
```diff
@@ -50,3 +50,43 @@ func TestStack_StripsNamespaceFromClusterScopedResource(t *testing.T) {
 	require.NoError(t, err)
 	assert.Empty(t, appliedNs.Namespace)
 }
+
+// A resource that was previously applied under the wrong stack name (e.g. due
+// to https://github.com/k0sproject/k0s/pull/8111) must get its stack label
+// corrected once it's applied under the right name, even if its content
+// (and thus its checksum) hasn't otherwise changed.
+func TestStack_CorrectsStaleStackNameLabel(t *testing.T) {
```
twz123: Let's not clutter the code with this. That's why we can annotate files in VCS and have commit messages.

```suggestion
func TestStack_StackNameChangeTriggersUpdate(t *testing.T) {
```

Do we also want to check that the checksum didn't change, or is this irrelevant here? If yes, then we might want to add the "WithoutChecksum" to the test function's name.

pkg/applier/stack_test.go
```diff
@@ -50,3 +50,43 @@ func TestStack_StripsNamespaceFromClusterScopedResource(t *testing.T) {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "ns-relabel-test",
+		},
+	}
+	resources, err := applier.ToUnstructured(nil, ns)
+	require.NoError(t, err)
+
+	fakes := testutil.NewFakeClientFactory()
+
+	wrongStack := applier.Stack{
```
twz123: Let's change the naming from wrong/corrected to old/new or foo/bar or the like. I'd also use those names for the stack name itself, instead of etcd/etcd-member.

pkg/applier/stack_test.go
```diff
@@ -50,3 +50,43 @@ func TestStack_StripsNamespaceFromClusterScopedResource(t *testing.T) {
...
+		Name:      "etcd",
+		Resources: []*unstructured.Unstructured{resources},
+		Clients:   fakes,
+	}
+	require.NoError(t, wrongStack.Apply(t.Context(), true))
+
+	applied, err := fakes.Client.CoreV1().Namespaces().Get(t.Context(), "ns-relabel-test", metav1.GetOptions{})
+	require.NoError(t, err)
+	require.Equal(t, "etcd", applied.Labels["k0s.k0sproject.io/stack"])
```
twz123: What about using the constant for the stack name label here?

pkg/component/controller/crd_test.go
```diff
@@ -0,0 +1,29 @@
...
+package controller
+
+import "testing"
+
+// TestNewCRDStack_WithStackName verifies that an explicitly provided stack
+// name via WithStackName is preserved. Previously, the defaulting logic in
+// NewCRDStack overwrote any explicitly set stack name back to the bundle
+// name whenever the (now removed) assets dir option was left unset, silently
+// discarding WithStackName.
+func TestNewCRDStack_WithStackName(t *testing.T) {
```
twz123: This is stating the obvious and can be figured out via VCS history.

```suggestion
func TestNewCRDStack_WithStackName(t *testing.T) {
```

pkg/component/controller/crd_test.go
```diff
@@ -0,0 +1,29 @@
...
+	stack := NewCRDStack(nil, nil, "etcd", WithStackName(EtcdMemberStackName))
```
twz123: Let's decouple it from etcd. Let's use "the-bundle-name"/"the-stack-name" or sth. around those lines.

pkg/component/controller/crd_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	if stack.stackName != EtcdMemberStackName {
+		t.Errorf("expected stack name %q, got %q", EtcdMemberStackName, stack.stackName)
```
twz123: Please use testify here.

pkg/component/controller/crd_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+}
+
+// TestNewCRDStack_Defaults verifies that the stack name defaults to the
+// bundle name when no options are given.
+func TestNewCRDStack_Defaults(t *testing.T) {
```
twz123: Better capture this in the assertion messages. Please use testify, and then you can add a message à la "stack name should default to bundle name".

```suggestion
func TestNewCRDStack_Defaults(t *testing.T) {
```

twz123 (commented):
Alright. Some nits in the tests, still, which I didn't look at too closely before. Mind to squash the commits into a single one and consolidate the commit message accordingly?

twz123:
Thanks for squashing. Two other things I noticed before this can go in. Turns out to be a deeper rabbit hole than I expected.

1. Could you take another pass on the commit message? A few points don't hold up. `WithCRDAssetsDir` didn't lose its callers because of this change. In fact, it hasn't been used in quite some time. It's dead code and not a consequence of the fix. The in-memory CRD stack is not "only present on unreleased main". It's in since 1.35. The defaulting bug itself is even older. it came in in v1.34, where it changed to which the CRD was written. The last paragraph is about the investigation and about what this change doesn't fix. That's fine for the PR description, but it doesn't belong in the commit message. It's also quite long for what the change does. Commit messages here usually explain what was wrong, what the change does, and why, in a few sentences each, without referencing source file paths unless it makes sense to do so. We usually use "Fixes:" and "See:" trailers for references and links.

2. Because of this bug, v1.34 wrote the etcd CRD to `<manifests-dir>/etcd/etcd-crd-etcd.k0sproject.io_etcdmembers.yaml` instead of `manifests/etcd-member/`. The applier manager's ignore list only covers etcd-member (which is right IMO), so on clusters that went through v1.34, that directory is still being applied as an on-disk etcd stack, in addition to the in-memory one. With this change, the two will now fight over the label. I think we need to address this here, as well. My proposal is to simply try to delete just that known file, and try to delete the etcd folder, if it's empty. There's prior art for removing manifests in the kube-proxy component.

pkg/component/controller/crd.go
```diff
@@ -117,3 +115,46 @@ func (c *CRDStack) Stop() error {
...
+
+// legacyEtcdManifestPath is the on-disk CRD manifest that versions of k0s
+// prior to v1.35 wrote for the etcd-member CRD stack under the wrong,
+// bundle-derived name "etcd", due to the stack-name defaulting bug fixed
+// alongside this function.
+func legacyEtcdManifestPath(manifestsDir string) (dir, file string) {
+	dir = filepath.Join(manifestsDir, "etcd")
+	file = filepath.Join(dir, "etcd-crd-etcd.k0sproject.io_etcdmembers.yaml")
+	return dir, file
+}
```
twz123: This is way too much ceremony. Let's just inline this.

pkg/component/controller/crd.go
```diff
@@ -117,3 +115,46 @@ func (c *CRDStack) Stop() error {
...
+// [legacyEtcdManifestPath], along with its parent directory if that becomes
+// empty. Both are optional: it's not an error if they're already gone.
+//
+// Left in place, the applier manager would keep picking up that directory
+// and applying it as its own on-disk "etcd" stack, competing with the
+// in-memory etcd-member stack over the same CRD's stack-name label. This
+// must run before the applier manager gets a chance to scan the manifests
+// directory, otherwise that on-disk stack may already be up and running by
+// the time the leftover manifest disappears out from under it.
+func RemoveLegacyEtcdManifest(manifestsDir string) error {
```
twz123: The comment is way too long. That's a story for the commit message, if at all. Maybe mix in some of the godoc of `legacyEtcdManifestPath` when inlining this and shorten this to a few lines only. I'd also add a "TODO: Remove in v1.38+" somewhere, as this is transitional cleanup code that we don't need to maintain forever. I'd rather name it `RemoveBogus...` instead of `RemoveLegacy...`, as removing the etcd folder is a bug cleanup, not some real functionality that gets phased out.

pkg/component/controller/crd.go
```diff
@@ -117,3 +115,46 @@ func (c *CRDStack) Stop() error {
...
+
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to read legacy etcd manifest directory: %w", err)
+	}
+	if len(entries) == 0 {
+		if err := os.Remove(dir); err != nil && !os.IsNotExist(err) {
```
twz123: Couldn't we do it the same way as we do it with the file, i.e. trying to remove the directory unconditionally, and then checking for `ENOTEMPTY` or something like that?

cmd/controller/controller.go
```diff
@@ -200,6 +200,13 @@ func (c *command) start(ctx context.Context, runtimeConfig *config.RuntimeConfig
 
 	perfTimer := performance.NewTimer("controller-start").Buffer().Start()
 
+	// This has to happen before the applier manager below gets a chance to
+	// scan the manifests directory, otherwise it may pick up the leftover
+	// on-disk etcd CRD manifest as its own stack before it's removed.
+	if err := controller.RemoveLegacyEtcdManifest(c.K0sVars.ManifestsDir); err != nil {
```
twz123: Maybe place this directly before the applier gets added to the node components? I'm seriously considering to simply inline the `RemoveLegacyEtcdManifest` function in here, instead of putting it into `crd.go`. It's transitional code anyways, and together with the `ENOTEMPTY` change, shouldn't be too long anyways. WDYT? (I guess we can spare the unit test if we verified once that the ENOTEMPTY stuff works out.)

If you disagree: Do you have other ideas where to put the function/test? Maybe to the etcd member reconciler or something?

pkg/component/controller/crd.go
```diff
@@ -117,3 +115,46 @@ func (c *CRDStack) Stop() error {
...
+// and applying it as its own on-disk "etcd" stack, competing with the
+// in-memory etcd-member stack over the same CRD's stack-name label. This
+// must run before the applier manager gets a chance to scan the manifests
+// directory, otherwise that on-disk stack may already be up and running by
+// the time the leftover manifest disappears out from under it.
+func RemoveLegacyEtcdManifest(manifestsDir string) error {
+	dir, file := legacyEtcdManifestPath(manifestsDir)
+
+	if err := os.Remove(file); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("failed to remove legacy etcd CRD manifest: %w", err)
```
twz123: This and the directory removal usually return `os.PathError`s, which are already expressive enough. Wrapping the errors repeats context the caller already has, as this is basically the "remove bogus etcd manifest" function. So I'd say we don't need to wrap the errors at all here.

twz123 (commented):
The PR description still has the old wording. Could you bring it in line with the commit message?

cmd/controller/controller.go
```diff
@@ -374,6 +368,20 @@ func (c *command) start(ctx context.Context, runtimeConfig *config.RuntimeConfig
...
 	nodeComponents.Add(ctx, leaderElector)
 
+	// TODO: Remove in v1.38+. A defaulting bug made affected versions of k0s
+	// write the etcd-member CRD manifest to disk under the wrong,
+	// bundle-derived stack name "etcd". Clean up that leftover file, and its
+	// directory if it's now empty, so the applier manager below doesn't pick
+	// it up as a stray on-disk "etcd" stack fighting the in-memory
+	// etcd-member stack over the CRD's stack-name label.
+	legacyEtcdDir := filepath.Join(c.K0sVars.ManifestsDir, "etcd")
+	legacyEtcdManifest := filepath.Join(legacyEtcdDir, "etcd-crd-etcd.k0sproject.io_etcdmembers.yaml")
```
twz123: Can we rename these to `bogus...`? Also can we place this in a block, so the variables aren't visible throughout the whole method?

twz123 (commented):
Could you please squash again?

twz123 (commented):
Thanks, could you please squash again?

pujitha24:
Sorry about that — squashed the rename/scoping commit back into the main one. Single commit now, code is unchanged from what you last saw, just the history is flattened.

twz123 (approved):

PR #8145, opened 2026-08-21
Fix lint findings surfaced by golangci-lint v2.13.1

kke:
Preparation for #8144. golangci-lint v2.13.1 pulls in Staticcheck 2026.2 and a newer modernize, which report three things on the current tree. None of them are new bugs, but all three fail the lint job.

modernize/reflecttypeassert flags two reflect type assertions that reflect.TypeAssert expresses directly. That function is available in Go 1.26, so this does not depend on the pending Go 1.27 bump.

unparam flags a func literal in watcher_test.go whose error result is always nil. Its signature is dictated by the mockProvider.watch field rather than chosen, so silence it in place.

SA4023 requires exclusion in linter config. Platform stubs such as `users.LookupUID` and `cleanup.buildSteps` return an unconditional error on the platforms they don't support, and Staticcheck 2026.2 now infers that, so every portable caller checking that error is reported as an always-true comparison: 10 sites on windows, 2 on darwin, none on linux. Per-site nolint directives cannot work here, because on the platforms where the stub is actually implemented the directive is unused and nolintlint fails the build instead. Exclude the check repository-wide instead, alongside the existing SA1019 rule. Note the pattern has to match "SA4023(related information):" as well as "SA4023:", as golangci-lint reports those as separate issues.

Verified with golangci-lint v2.13.1: 0 issues for GOOS=linux, darwin and windows.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -107,6 +107,15 @@ linters:
       - linters:
           - staticcheck
         text: '^SA1019:'
+      # Platform stubs, such as users.LookupUID or cleanup.buildSteps, return an
+      # unconditional error on the platforms they don't support. Portable
+      # callers still have to check that error, which trips SA4023 on those
+      # platforms only. Per-site nolint directives aren't an option: they'd be
+      # reported as unused by nolintlint on the platforms where the stub isn't
+      # in use.
+      - linters:
+          - staticcheck
+        text: '^SA4023[:(]'
     paths:
       - zz_*
       - build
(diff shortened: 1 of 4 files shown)


kke:
The lint runs are from when this was based on `main`. They fail there because linter is older.

The lint workflows are set to run only when base is `main` or `release-*`, so with the bump PR as base, they won't fire at all.

So, I think this needs to be merged to the bump PR, but that stops renovate from rebasing that PR.

PR #8203, opened 2026-08-31
Thread a real context through component readiness checks

magic-peach:
The `Ready` interface's `Ready()` method used `context.TODO()` internally in each implementation, so per-attempt timeouts and manager cancellation never reached an in-flight readiness probe. `Ready` now takes a `context.Context`, matching `Init`/`Start`/`Reconcile`, and every implementation (etcd, kine, apiserver, konnectivity, static_pods, nllb reconciler) threads the context the manager already has through to its probe.

### DIFF
--- a/pkg/component/controller/etcd.go
+++ b/pkg/component/controller/etcd.go
@@ -512,9 +512,9 @@ func (e *Etcd) setupCerts(ctx context.Context) error {
 }
 
 // Health-check interface
-func (e *Etcd) Ready() error {
+func (e *Etcd) Ready(ctx context.Context) error {
 	logrus.WithField("component", "etcd").Debug("checking etcd endpoint for health")
-	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
 	defer cancel()
 	err := etcd.CheckEtcdReady(ctx, e.K0sVars.CertRootDir, e.K0sVars.EtcdCertDir, e.Config)
 	return err
--- a/pkg/component/manager/component.go
+++ b/pkg/component/manager/component.go
@@ -41,7 +41,9 @@ type Component interface {
 
 type Ready interface {
 	// Ready performs a ready check and indicates that a component is ready to run.
-	Ready() error
+	// The given context may be used to bound and cancel the readiness check; it is
+	// derived from the context passed to the manager's Start method.
+	Ready(context.Context) error
 }
 
 // Reconciler defines the component interface that is reconciled based
--- a/pkg/component/manager/manager_test.go
+++ b/pkg/component/manager/manager_test.go
@@ -39,7 +39,7 @@ func (f *Fake) Stop() error {
 	f.StopCalled = true
 	return f.StopErr
 }
-func (f *Fake) Ready() error {
+func (f *Fake) Ready(_ context.Context) error {
 	f.HealthyCalled = true
 	return f.HealthyErr
 }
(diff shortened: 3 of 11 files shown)


twz123:
~~We definitely need to decouple the changes to the Ready interface to anything that we add for etcd into a separate PRs.~~

(Nevermind.)

pkg/component/controller/etcd.go
```diff
@@ -39,6 +39,14 @@ import (
...
 const etcdGID = 0
 
+// etcdReadyCheckTimeout bounds a single etcd readiness probe. The manager
+// polls Ready repeatedly against its own overall readiness deadline
+// (Manager.ReadyWaitDuration, 2 minutes by default), so this timeout only
+// needs to be long enough for etcd to answer a local status request and
+// short enough that one slow or unresponsive probe doesn't eat a large
+// chunk of that budget before the next poll attempt.
+const etcdReadyCheckTimeout = 1 * time.Second
+
```
ncopa:
```suggestion
```
I don't think this is necessary. Only adds noise.

pkg/component/controller/etcd.go
```diff
@@ -434,9 +442,9 @@ func (e *Etcd) setupCerts(ctx context.Context) error {
 }
 
 // Health-check interface
-func (e *Etcd) Ready() error {
+func (e *Etcd) Ready(ctx context.Context) error {
 	logrus.WithField("component", "etcd").Debug("checking etcd endpoint for health")
-	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, etcdReadyCheckTimeout)
```
ncopa:
```suggestion
	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
```

ncopa (changes requested):
Thank you for working on this! I think those changes are good. Just remove the unnecessary change and comment and we should be good to go.

Please fold the change into the relevant commit.

Rebasing it to current `main` should also make the integration tests pass. We have fixed some of the flakes recently.

twz123:
Also, please rewrite the PR description, there's nothing special about etcd in here.

magic-peach:
Rebased onto current main, removed the etcdReadyCheckTimeout constant and its comment (back to the inline 1*time.Second literal), and folded the ticker.Stop() fix into the main commit instead of leaving it separate. Also rewrote the PR description since the interface change touches every Ready implementation, not just etcd.

twz123:
> Also stops `waitForReady`'s polling ticker on return, which previously leaked its timer.

NB: unstopped timers are eligible for garbage collection in more recent Go versions. So, this is more of a cosmetic change than this comment suggests.

pkg/component/manager/component.go
```diff
@@ -41,7 +41,9 @@ type Component interface {
 
 type Ready interface {
 	// Ready performs a ready check and indicates that a component is ready to run.
-	Ready() error
+	// The given context may be used to bound and cancel the readiness check; it is
+	// derived from the context passed to the manager's Start method.
```
twz123:
```suggestion
```

pkg/component/manager/manager_test.go
```diff
@@ -39,7 +39,7 @@ func (f *Fake) Stop() error {
 	f.StopCalled = true
 	return f.StopErr
 }
-func (f *Fake) Ready() error {
+func (f *Fake) Ready(_ context.Context) error {
```
twz123: No need for the underscores, `Ready(context.Context) error` works just fine. Same in the other places.

PR #8204, opened 2026-08-31
Watch CSRs instead of polling for them every 10 seconds

magic-peach:
The CSR approver used a `time.Ticker` to poll for pending CertificateSigningRequests every 10 seconds, so a new node join could wait up to 10s for its kubelet-serving CSR to get approved.

Refactors to watch CertificateSigningRequest objects via an informer instead, so approval happens immediately on CSR creation. Approval logic/conditions are unchanged, only the trigger mechanism.

Added `TestCSRApproverWatch` covering the new watch path.


twz123:
Just to double-check, was this ever reviewed by a human before this PR was opened? The original PR template asks for a self-review, that means self-review by a human.

magic-peach:
yes, I did review it

pkg/component/controller/csrapprover.go
```diff
@@ -62,83 +64,97 @@ func (a *CSRApprover) Init(_ context.Context) error {
...
-				if err != nil {
-					a.log.WithError(err).Warn("CSR approval failed")
-				}
-			case <-ctx.Done():
-				a.log.Info("CSR Approver context done")
-				return
-			}
-		}
-	}()
+	go a.watchCSRs(ctx)
```
twz123: I'd rather use `leaderelection.RunLeaderTasks` here to run the watch only when this k0s process is the leader. This also closes a possible race with the current implementation in this PR which can miss to approve some certificates during leader transition periods.

pkg/component/controller/csrapprover_test.go
```diff
@@ -51,17 +60,93 @@ func TestBasicCRSApprover(t *testing.T) {
...
-	assert.NoError(t, c.approveCSR(ctx))
+	assert.NoError(t, c.approveCSR(ctx, newCsr))
 
 	csr, err := client.CertificatesV1().CertificateSigningRequests().Get(ctx, newCsr.Name, metav1.GetOptions{})
 	assert.NoError(t, err)
 	assert.NotNil(t, csr)
 	assert.Equal(t, newCsr.Name, csr.Name)
 	for _, c := range csr.Status.Conditions {
-		assert.True(t, c.Type == certv1.CertificateApproved && c.Reason == "Autoapproved by K0S CSRApprover" && c.Status == core.ConditionTrue)
+		assert.True(t, c.Type == certv1.CertificateApproved && c.Reason == "Autoapproved by K0s CSRApprover" && c.Status == core.ConditionTrue)
```
twz123: Let's not change this in this PR.

pkg/component/controller/csrapprover_test.go
```diff
@@ -51,17 +60,93 @@ func TestBasicCRSApprover(t *testing.T) {
...
+		},
+	}
+
+	newCsr, err := client.CertificatesV1().CertificateSigningRequests().Create(ctx, csrReq, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	// This would only succeed within the test's timeout if the approver
+	// reacts to the CSR being created, rather than waiting for its next poll
+	// tick (which no longer exists).
+	require.NoError(t, watch.CertificateSigningRequests(client.CertificatesV1().CertificateSigningRequests()).
```
twz123: You can probably wrap the whole test in a synctest bubble, and use a single get on the fake client instead.

twz123 (changes requested):

magic-peach:
@twz123 could youplease take a look now?

magic-peach:
Rebased onto current main, no other changes, the leader-gated watch fix from the last review round was already pushed and still applies cleanly.

Issue #8181, opened 2026-08-26
Upgrading a helm chart fails with context deadline exceeded


apreiml:
### Before creating an issue, make sure you've checked the following:

- [x] You are running the latest released version of k0s
- [x] Make sure you've searched for existing issues, both open and closed
- [x] Make sure you've searched for PRs too, a fix might've been merged already
- [x] You're looking at docs for the released version, "main" branch docs are usually ahead of released versions.

### Platform

```shell
Linux 6.12.94+deb13-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.94-1 (2026-06-20) x86_64 GNU/Linux
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.6
ID=debian
```

### Version

v1.36.3+k0s.2

### Sysinfo

<details><summary>`k0s sysinfo`</summary>
<pre>
Total memory: 3.7 GiB (pass)
File system of /var/lib/k0s: ext4 (pass)
Disk space available for /var/lib/k0s: 29.7 GiB (pass)
Relative disk space available for /var/lib/k0s: 79% (pass)
Name resolution: localhost: [::1 127.0.0.1] (pass)
Containerd config snippets in /etc/k0s/containerd.d/: no config snippets found (pass)
Operating system: Linux (pass)
  Linux kernel release: 6.12.94+deb13-cloud-amd64 (pass)
  Max. file descriptors per process: current: 524287 / max: 524288 (pass)
  AppArmor: active (pass)
  Executable in PATH: modprobe: /usr/sbin/modprobe (pass)
  Executable in PATH: mount: /usr/bin/mount (pass)
  Executable in PATH: umount: /usr/bin/umount (pass)
  /proc file system: mounted (0x9fa0) (pass)
  Control Groups: version 2 (pass)
    cgroup controller "cpu": available (is a listed root controller) (pass)
    cgroup controller "cpuacct": available (via cpu in version 2) (pass)
    cgroup controller "cpuset": available (is a listed root controller) (pass)
    cgroup controller "memory": available (is a listed root controller) (pass)
    cgroup controller "devices": available (device filters attachable) (pass)
    cgroup controller "freezer": available (cgroup.freeze exists) (pass)
    cgroup controller "pids": available (is a listed root controller) (pass)
[...]

PR #8205, opened 2026-08-31
Give Helm uninstalls their own wait timeout instead of relying on ctx

magic-peach:
`Commands.UninstallRelease` derived its Helm action Timeout only from `ctx.Deadline()`, unlike `InstallChart`/`UpgradeChart`, which both take an explicit timeout parameter. No deadline is ever attached to the context in the actual call path (`ExtensionsController.Start` uses a plain `context.WithCancel(context.Background())`), so `UninstallRelease` always got `Timeout = 0`. Since `Wait = true` is unconditional, Helm's `WaitForDelete` builds an already-expired context, so any uninstall with real resources to wait for fails almost instantly.

This reproduces the exact error text reported in #8181. Gives `UninstallRelease` an explicit timeout parameter (mirroring install/upgrade), with a small helper that still clamps to the context's own deadline if one is ever set and tighter.

Added `TestEffectiveTimeout` covering the previously-untested zero-timeout path.

Note: this closes a real bug in the uninstall timeout path regardless, but may not be the entire story behind #8181's original report -- the maintainers' stuck-LoadBalancer theory in that thread wasn't ruled out as a separate contributing factor for that specific repro.


pkg/helm/helm.go
```diff
@@ -494,14 +494,32 @@ func (hc *Commands) UninstallRelease(ctx context.Context, releaseName, namespace
...
+
+	// Mirror InstallChart/UpgradeChart and honor the caller-supplied
+	// timeout for the wait-for-delete phase below. The uninstall action
+	// doesn't derive its Timeout from ctx the way RunWithContext does for
+	// install/upgrade, so it has to be set explicitly here; callers such as
+	// the Chart reconciler don't normally attach a deadline to ctx, which
+	// used to leave Timeout at its zero value and made the subsequent wait
+	// fail instantly with "context deadline exceeded" instead of actually
+	// waiting for the release's resources to go away. Still clamp to ctx's
+	// own deadline when one happens to be set and it's the tighter bound.
```
twz123: That comment is way too long, and on the wrong level, tbh. It's, if at all, better placed in `extensions_controller.go`.

pkg/helm/helm.go
```diff
@@ -494,14 +494,32 @@ func (hc *Commands) UninstallRelease(ctx context.Context, releaseName, namespace
...
 	helmAction.Wait = true
 	helmAction.DeletionPropagation = string(metav1.DeletePropagationForeground)
 
 	_, err = helmAction.Run(releaseName)
 	return err
 }
+
+// effectiveTimeout returns timeout, clamped to ctx's remaining time when ctx
+// has a deadline that's sooner than timeout.
+func effectiveTimeout(ctx context.Context, timeout time.Duration) time.Duration {
```
twz123: Why would we need this? If the context gets cancelled, the context gets cancelled. No need to adjust the timeout, is there?

That said, I don't think there's anything in k0s that would call the helm uninstallation using a context that has a timeout, so this is effectively dead code, and speculatively added "just in case", or am I missing something, here?

pkg/component/controller/extensions_controller.go
```diff
@@ -318,7 +318,12 @@ func (cr *ChartReconciler) uninstall(ctx context.Context, chart helmv1beta1.Char
 	}
 	defer cleanup()
 
-	if err := helmCmd.UninstallRelease(ctx, chart.Status.ReleaseName, chart.Status.Namespace); err != nil {
+	timeout, err := time.ParseDuration(chart.Spec.Timeout)
+	if err != nil || timeout <= 0 {
+		timeout = defaultTimeout
+	}
+
+	if err := helmCmd.UninstallRelease(ctx, chart.Status.ReleaseName, chart.Status.Namespace, timeout); err != nil {
```
twz123, line 327: This is an interesting discussion point. This method is only called when a chart resource gets deleted, to clear the finalizer. Question is if we want to also add the chart's timeout here. The user won't be able to change this afterwards, since the chart resource itself is already marked as deleted.

pkg/component/controller/extensions_controller.go
```diff
@@ -318,7 +318,12 @@ func (cr *ChartReconciler) uninstall(ctx context.Context, chart helmv1beta1.Char
 	}
 	defer cleanup()
 
-	if err := helmCmd.UninstallRelease(ctx, chart.Status.ReleaseName, chart.Status.Namespace); err != nil {
+	timeout, err := time.ParseDuration(chart.Spec.Timeout)
+	if err != nil || timeout <= 0 {
+		timeout = defaultTimeout
+	}
```
twz123: Better use the same logic as in `updateOrInstallChart`.

twz123:
I don't see how this would fix #8181. This merely adds a timeout, but the user-observed error already indicates that a deadline exceeded.

magic-peach:
Rebased onto current main to resolve the conflict. This branch predated the Helm v4 SDK migration, which moved the release status constants from helm.sh/helm/v4/pkg/release to a new release/common subpackage and changed chartRelease's declared type to the release.Releaser interface. Kept those upstream changes as-is and reapplied both of this branch's own commits (the wait timeout addition, then the effectiveTimeout cleanup) on top. Verified: go build/go vet/go test clean on pkg/component/controller and pkg/helm.

PR #8196, opened 2026-08-28
Add code guidelines

twz123:
## Description

The contributor docs describe the mechanics of commits and pull requests, but say nothing about how the code should be written.

Add a code guidelines page covering these topics, and link to it from the contributor index, the testing guide, and the pull request template.

Extend the commit message section of the GitHub workflow accordingly. Be more explicit about their structure and purpose, so that it's clear what they're for, and clarify how trailers are used.

Adjust the pull request template accordingly, and replace the checklist item that asked for more comments with one that asks for the right kind of comments.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

- [x] My code follows the style [guidelines](https://docs.k0sproject.io/head/contributors/) of this project
- [x] My commit messages are [signed-off](https://docs.k0sproject.io/head/contributors/github_workflow/)
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

PR #8220, opened 2026-08-31
Do not prune a stack when its manifest directory is empty

### LINKED ISSUE #8214: Control plane version change wipes the coredns and metricserver stacks in the child cluster Template
### What happened?

Changing `K0smotronControlPlane.spec.version` from `v1.34.11-k0s.0` to `v1.35.6-k0s.0`
removed every object of the `coredns` and `metricserver` stacks from the child cluster:
Deployment, Service, ConfigMap, ServiceAccount, ClusterRoles/Bindings, APIService.
Nothing recreates them — the child cluster is left without DNS. Reproduced on three
clusters (two upgraded from v1.34.11, one from v1.34.8, all to v1.35.6).

The mechanism, from a run with `--debug`: a restarted CP pod starts with an empty
`/var/lib/k0s` (default `spec.persistence` is emptyDir). k0s' CoreDNS/MetricServer
components create their manifest directory in Init/Start but write the manifest file only
once the ClusterConfig is available — k0smotron adds `--enable-dynamic-config` by
default, and until then the log shows `failed to calculate metrics-server config:
cluster config not available yet`. When the restarted pod takes the applier lease, the
applier-manager runs the still-empty stack directories first:

```
time="2026-08-30 20:22:49" level=debug msg="applying with 0 resources" stack=coredns
time="2026-08-30 20:22:57" level=debug msg="deleting resource apps/Deployment:coredns@kube-system" stack=coredns
time="2026-08-30 20:23:01" level=debug msg="applying with 7 resources" stack=coredns
```

An apply with zero resources prunes everything labelled `k0s.k0sproject.io/stack=coredns`
(`pkg/applier/stack.go`); it is indistinguishable from the intentional stack deletion —
`Applier.Delete` is exactly an empty apply (`pkg/applier/applier.go`). The re-apply
twelve seconds later did not restore the objects: the deletions run with foreground
propagation and were still in progress, the applier is purely watch-based (initial run
plus fsnotify, no periodic resync, `pkg/applier/stackapplier.go`), and during the rolling
update each restarted replica that takes over the lease repeats the race. End state on
all three clusters: zero objects carrying the stack label.

A plain restart of the CP pods (a `controlPlaneFlags` patch, same k0s version) did not
trigger it; the version change did, three times out of three.

### Involved Areas

Cluster API integration

### What did you expect to happen?

A control plane version change does not remove system components from the child cluster.

### Steps to reproduce

1. Hosted control plane via Cluster API: `K0smotronControlPlane`, 3 replicas, default
   persistence (emptyDir), k0s v1.34.x; let it settle so the stacks exist in etcd.
2. Patch `spec.version` to `v1.35.6-k0s.0`.
3. After the rollout: `kubectl get all,cm,sa,clusterrole,clusterrolebinding -A -l k0s.k0sproject.io/stack=coredns`
   in the child cluster returns nothing.

### k0smotron version

v2.1.0

### k0s version

v1.34.11+k0s.0 and v1.34.8+k0s.0, upgraded to v1.35.6+k0s.0

### Anything else we need to know?

Same class upstream, different stack: k0sproject/k0s#8103 — etcdmember objects wiped when
a second controller joins; that report suspects the same race between the applier-manager
start and the component writing its manifest. k0sproject/k0s#8111 relates but is
explicitly not confirmed as a fix and is not in any release yet.

Child cluster: dual-stack, `network.provider: custom` (CNI via ClusterResourceSet),
`kubeProxy.disabled: true` — the kubeproxy stack removal in the logs is the intentional
path and not part of the problem. Storage: `spec.storage.type: etcd` (k0smotron-managed
etcd StatefulSet), which is why the objects predate the fresh pod filesystem.

pujitha24:
## Description

CoreDNS, metrics-server, Metrics, Calico, KubeRouter and KubeProxy used to
create their manifest subdirectory in `Init()`/`Start()`, before they had
anything to write into it. The applier-manager picks up a newly created stack
directory via fsnotify and fires an unconditional apply about a second after
it starts watching it. If that lands while the directory exists but is still
empty (e.g. right after a controller pod restart with fresh ephemeral
`/var/lib/k0s` storage, while the underlying Kubernetes objects from the
previous process instance still exist), `Applier.Apply` treats the empty
directory like an explicit `Applier.Delete` and prunes every object already
carrying that stack's label. Nothing recreates them afterwards, since the
applier is purely watch-based with no periodic resync.

We discussed a couple of alternatives before landing on this approach:
changing `Applier.Apply` so an empty directory is not treated as a delete
request was rejected as too big a behavioral change; moving these components
to in-memory applies (as done for some other stacks, e.g. #6739) would remove
the filesystem hand-off entirely but is a much larger, likely non-backportable
change.

Instead, each of the six components now creates its manifest subdirectory
immediately before writing its first manifest file, the same way
`konnectivityagent.go` already did. This doesn't eliminate the race — there's
still an async hop between creating the directory and writing into it — but it
shrinks the window from however long it takes for cluster config to become
available down to the gap between two statements in the same function, making
the wipe far less likely to hit in practice, and is simple enough to backport.
`Applier.Apply`'s prune-on-empty-directory behavior is unchanged.

Each manifest is now also rendered into an in-memory buffer before its
directory is created, rather than the directory being created ahead of a
straight write into a file within it. That way a template error leaves
neither a half-written file nor an empty, race-prone directory behind.
KubeRouter already worked this way; CoreDNS, metrics-server, Metrics and
KubeProxy now follow the same order.

Fixes #8214

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

Added `inttest/manifestrestart`: starts a controller with dynamic config
enabled and a worker, waits for CoreDNS and metrics-server to come up, records
their Deployment UIDs, stops the controller, wipes `/var/lib/k0s/manifests`,
restarts the controller, and asserts the UIDs are unchanged (i.e. the
Deployments were not deleted and re-created). Dynamic config is needed to
reproduce the race reliably: it depends on the apiserver being up and the
ClusterConfig CR round-tripping, which is what opens a gap wide enough to hit
in practice, whereas static config hands configuration to components over a
synchronous in-memory channel, so the manifest directory tends to get its
first write well within the applier's debounce window regardless of when the
directory was created.

`check-manifestrestart` passed in CI against an earlier commit on this branch,
before the commits were folded into one; it has not re-run against the current
commit yet. I have not bisected this locally (reverting the directory-creation
change and confirming the test then fails) to hard-prove it catches the
regression, since this environment has no Docker/privileged-container access
to run the inttest suite.

`go build`, `go vet` and `go test` on the affected packages
(`pkg/applier/...`, `pkg/component/controller/...`), `golangci-lint`, and
`gofmt` all passed against this diff.

## Checklist

### DIFF
--- a/inttest/manifestrestart/manifestrestart_test.go
+++ b/inttest/manifestrestart/manifestrestart_test.go
@@ -1,0 +1,76 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package manifestrestart
+
+import (
+	"context"
+	"testing"
+
+	"github.com/k0sproject/k0s/inttest/common"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type ManifestRestartSuite struct {
+	common.BootlooseSuite
+}
+
+// TestManifestDirNotPruned verifies that losing the local manifest directory
+// does not cause previously applied resources to be pruned.
+func (s *ManifestRestartSuite) TestManifestDirNotPruned() {
+	ctx := s.Context()
+
+	s.Require().NoError(s.InitController(0, "--enable-dynamic-config"))
+	s.Require().NoError(s.RunWorkers())
+
+	kc, err := s.KubeClient(s.ControllerNode(0))
+	s.Require().NoError(err)
+
+	for i := range s.WorkerCount {
+		s.Require().NoError(s.WaitForNodeReady(s.WorkerNode(i), kc))
+	}
+
+	s.Require().NoError(common.WaitForCoreDNSReady(ctx, kc))
+	s.Require().NoError(common.WaitForDeployment(ctx, kc, "metrics-server", metav1.NamespaceSystem))
+
+	coreDNSUID := s.deploymentUID(ctx, kc, "coredns")
+	metricsServerUID := s.deploymentUID(ctx, kc, "metrics-server")
+
+	s.Require().NoError(s.StopController(s.ControllerNode(0)))
+
+	ssh, err := s.SSH(ctx, s.ControllerNode(0))
+	s.Require().NoError(err)
+	defer ssh.Disconnect()
+	_, err = ssh.ExecWithOutput(ctx, "rm -rf /var/lib/k0s/manifests")
+	s.Require().NoError(err)
+
+	s.Require().NoError(s.StartController(s.ControllerNode(0)))
+	s.Require().NoError(s.WaitForKubeAPI(s.ControllerNode(0)))
+
+	s.Require().NoError(common.WaitForCoreDNSReady(ctx, kc))
+	s.Require().NoError(common.WaitForDeployment(ctx, kc, "metrics-server", metav1.NamespaceSystem))
+
+	s.Equal(coreDNSUID, s.deploymentUID(ctx, kc, "coredns"), "coredns Deployment must not have been deleted and re-created")
+	s.Equal(metricsServerUID, s.deploymentUID(ctx, kc, "metrics-server"), "metrics-server Deployment must not have been deleted and re-created")
+}
+
+func (s *ManifestRestartSuite) deploymentUID(ctx context.Context, kc kubernetes.Interface, name string) types.UID {
+	d, err := kc.AppsV1().Deployments(metav1.NamespaceSystem).Get(ctx, name, metav1.GetOptions{})
+	s.Require().NoError(err)
+	return d.UID
+}
+
+func TestManifestRestartSuite(t *testing.T) {
+	s := ManifestRestartSuite{
+		common.BootlooseSuite{
+			ControllerCount: 1,
+			WorkerCount:     1,
+		},
+	}
+	suite.Run(t, &s)
+}
--- a/pkg/component/controller/kuberouter.go
+++ b/pkg/component/controller/kuberouter.go
@@ -65,7 +65,7 @@ func NewKubeRouter(k0sVars *config.CfgVars, primaryAddressFamily v1beta1.Primary
 
 // Init implements [manager.Component].
 func (k *KubeRouter) Init(context.Context) error {
-	return dir.Init(filepath.Join(k.k0sVars.ManifestsDir, "kuberouter"), constant.ManifestsDirMode)
+	return nil
 }
 
 // Stop no-op as nothing running
@@ -169,6 +169,10 @@ func (k *KubeRouter) Reconcile(_ context.Context, clusterConfig *v1beta1.Cluster
 	err := tw.WriteToBuffer(output)
 	if err != nil {
 		return fmt.Errorf("error writing kube-router manifests, will NOT retry: %w", err)
+	}
+
+	if err := dir.Init(filepath.Join(k.k0sVars.ManifestsDir, "kuberouter"), constant.ManifestsDirMode); err != nil {
+		return err
 	}
 
 	if err := file.AtomicWithTarget(filepath.Join(k.k0sVars.ManifestsDir, "kuberouter", "kube-router.yaml")).
--- a/inttest/Makefile.variables
+++ b/inttest/Makefile.variables
@@ -62,6 +62,7 @@ smoketests := \
 	check-kubeletcertrotate \
 	check-kuberouter \
 	check-kuberouter-ipv6 \
+	check-manifestrestart \
 	check-metrics \
 	check-metricsscraper \
 	check-metricsscraper-singlenode \
--- a/pkg/component/controller/calico.go
+++ b/pkg/component/controller/calico.go
@@ -6,7 +6,6 @@ package controller
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io/fs"
 	"path"
@@ -127,10 +126,7 @@ func NewCalico(nodeConfig *v1beta1.ClusterConfig, manifestsDir string, hasWindow
 
 // Init implements [manager.Component].
 func (c *Calico) Init(context.Context) error {
-	return errors.Join(
-		dir.Init(filepath.Join(c.manifestsDir, "calico_init"), constant.ManifestsDirMode),
-		dir.Init(filepath.Join(c.manifestsDir, "calico"), constant.ManifestsDirMode),
-	)
+	return nil
 }
 
 // Start implements [manager.Component].
@@ -207,6 +203,10 @@ func (c *Calico) Start(context.Context) error {
 func (c *Calico) dumpCRDs(patches v1beta1.Patches) error {
 	var emptyStruct struct{}
 
+	if err := dir.Init(filepath.Join(c.manifestsDir, "calico_init"), constant.ManifestsDirMode); err != nil {
+		return err
+	}
+
 	// Write the CRD definitions only at "boot", they do not change during runtime
 	crds, err := fs.ReadDir(static.CalicoManifests, "CustomResourceDefinition")
 	if err != nil {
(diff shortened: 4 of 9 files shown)


twz123:
I'm not in favor of that solution.

pujitha24:
Fair enough — I'd rather not guess at what you'd prefer instead of proposing something that misses the mark. A couple of directions I can think of: have CoreDNS/metrics-server defer creating their manifest directory until they actually have something to write (narrows the race at the source instead of changing Apply's semantics), or have the applier-manager's initial apply wait for a first non-empty listing instead of firing unconditionally ~1s after start. Is either of those closer to what you had in mind, or is it something else?

jnummelin:
> have CoreDNS/metrics-server defer creating their manifest directory until they actually have something to write

This was my initial thought. I doesn't remove the race condition but makes it far less likely at least. And a simple enough to backport.

On the otherhand, I'm not really sure we can anyways completely remove the race with anything we do using the manifest dirs/files approach, there's always the async hop via filesystem in between. We do have the debouncer in between though to mitigate some of these problems.

One possible idea is to move to use in-memory applies for all the k0s owned things. That might be pretty big change and I'm not sure if that's really backportable. We've done that already for some other components/stacks, e.g. https://github.com/k0sproject/k0s/pull/6739

jnummelin:
I guess the main reason for "empty dir == no apply" idea rejection is the fact that it'd mean a significant behavioural change. @twz123 did you have some other reasons why you didn't prefer that?

@pujitha24 I think we need a inttest case for this too, bad enough bug we need to ensure get's properly covered. Basically something like:
- start controller (and a worker)
- wait till everything gets properly up
- record coredns etc. deployment object details (e.g. UUIDs etc.)
- stop controller
- remove manifest dir
- start controller
- verify coredns etc. did not get deleted nor re-created (by comparing UUIDs or what not)

jnummelin:
We need to also ensure there's no other components with the same pattern of creating the manifest dir in their `Init()`

pujitha24:
Pushed a follow-up commit for both points. Added `inttest/manifestrestart`: it starts a controller+worker, waits for CoreDNS/metrics-server to come up, records their Deployment UIDs, stops the controller, wipes `/var/lib/k0s/manifests`, restarts, and asserts the UIDs are unchanged.

On the audit: `dir.Init` on the addon's manifest subdir was also happening in `Init()`/`Start()` (before any config-dependent write) in Metrics, Calico, KubeRouter and KubeProxy, same pattern as CoreDNS/metrics-server. Moved all four to create the dir immediately before their first write instead. `konnectivityagent.go` already did this lazily. I checked the rest of `pkg/component/controller` and the other `dir.Init`/`MkdirAll` call sites in the repo (kine, konnectivity, etcd, ocibundle, helm, applier/manager's top-level `ManifestsDir`) — none of those are per-addon subdirectories under the fsnotify-watched manifest tree, so they're not subject to this race.

jnummelin:
Now with this approach, the PR description and title needs to be adjusted too

pujitha24:
Updated the description to reflect the wider scope (Metrics/Calico/KubeRouter/KubeProxy, not just CoreDNS/metrics-server) and the new inttest. I don't have a way to edit the title myself here, but something like "Defer manifest dir creation until first write to avoid pruning race on restart" would match the current approach better than the old "Do not prune a stack when its manifest directory is empty" — happy to have someone rename it, or let me know if you'd rather I not guess.

Also committed a fix for `check-manifestrestart`: the test called `InitController(0)` without `--enable-worker`/`--single`, so controller0 never registers as a Node, and `WaitForNodeReady(controller0)` hung until the deadline. Dropped those two waits; the suite already waits on the worker node and on CoreDNS/metrics-server readiness, which is what actually matters here. Confirmed kube-router and CoreDNS were applied fine within seconds in the failed run's logs, so this was a test bug, not a problem with the dir-creation fix itself.

The other failing check, `check-ap-controllerworker :: v1.36.4+k0s.0`, doesn't touch anything this PR changes (autopilot ControlNode plan monitoring) and the same job passes on recent main runs, so that one looks unrelated.

twz123:
This is moving in the right direction :ok_hand:

Regarding the integration test: Have you verified that it reliably detects the problem we are trying to fix? AFAIU this only occurs when dynamic configuration is used, but the test doesn't use it.

Could you also please take a look at the new code guidelines for k0s proposed in #8196? They haven't been merged yet, but this is the general direction in which we're heading.

jnummelin:
> AFAIU this only occurs when dynamic configuration is used, but the test doesn't use it.

I think one _can_ hit this even without if the disk is slow or something. Yes, the odds for hitting this with dynamic config is much higher for sure

pujitha24:
@twz123 pushed 5424af2 shortly after you asked, adding `--enable-dynamic-config` to the controller in the inttest. Static config hands config to components over a synchronous in-memory channel, so even with the old eager mkdir the manifest dir got its first write well inside the applier's 1s debounce window; dynamic config depends on the apiserver being up and the ClusterConfig CR round-tripping, which is what actually opens the gap — matches what @jnummelin said, the race can happen without it too (slow disk etc.), dynamic config just makes it far more likely, which is also why the test needed it to reproduce reliably. `check-manifestrestart` is passing in the latest CI run with this change. I haven't done a bisection (revert the dir-creation fix, confirm the test then fails) to hard-prove it catches the regression — no Docker here to try that myself — happy to if you'd want that before merging.

Took a look at #8196. Most of what I'd touched already lines up with it, but the log line I added in metricserver.go didn't: it copied the older `%s. will retry` style from the line right below it. Changed it to `m.log.WithError(err).Error("Failed to create metrics-server manifest dir, will retry")`, matching the guideline's logging section and the `WithError` pattern calico.go already uses. Left the untouched sibling line as-is.

pkg/component/controller/kuberouter.go
```diff
@@ -171,6 +171,10 @@ func (k *KubeRouter) Reconcile(_ context.Context, clusterConfig *v1beta1.Cluster
 		return fmt.Errorf("error writing kube-router manifests, will NOT retry: %w", err)
 	}
 
+	if err := dir.Init(filepath.Join(k.k0sVars.ManifestsDir, "kuberouter"), constant.ManifestsDirMode); err != nil {
+		return fmt.Errorf("error creating kube-router manifest dir: %w", err)
```
twz123: The `dir.Init` error isn't wrapped anywhere else. What's different about this case that it's wrapped here?

inttest/manifestrestart/manifestrestart_test.go
```diff
@@ -0,0 +1,89 @@
...
+// as an implicit request to prune the resources it previously applied.
+func (s *ManifestRestartSuite) TestManifestDirNotPruned() {
+	ctx := s.Context()
+
+	// Use dynamic config: the static config source hands the config to
+	// components over a local, in-memory channel, so the manifest dir tends
+	// to get its first write well within the applier's 1s debounce window
+	// regardless of when the dir was created. Dynamic config only becomes
+	// available once the apiserver is up and the ClusterConfig CR has been
+	// read back, which is what actually exposes this race in practice.
```
twz123: Those comments are _way_ too long and talk about things which are not important for the test. And they reference file names and line numbers.

inttest/manifestrestart/manifestrestart_test.go
twz123, line 1: I'm not sure we need this test at all? What does it cover that other unit and integration tests are missing?

twz123 (commented):
In general, we should try to render all the files into byte buffers, and should create the directories only after we have the content, so we catch potential templating errors early, before having created an empty directory.

twz123:
Could you please update the PR title and description accordingly and fold your commits, as outlined in #8196?

inttest/manifestrestart/manifestrestart_test.go
jnummelin: It covers the case where we restart a controller with empty manifest dir, it shouldn't go and delete nor re-create e.g. coredns. I proposed the testcase for regression control.

pujitha24:
Commits are already folded into one (`5cca994f2`). I've updated the description to match: it now covers the alternatives we discussed (empty-dir-means-delete change, in-memory applies) and why we landed on deferring directory creation, plus what was verified and what's deliberately left out. Note `check-manifestrestart` passed against an earlier, now-squashed commit on this branch but hasn't re-run against the folded one yet, so the description says that explicitly rather than claiming it's currently green.

For the title, I still don't have a way to rename the PR myself — "Defer manifest dir creation until first write to avoid pruning race" (the commit's subject) would match the current approach; happy to have someone apply it, or let me know if you'd rather word it differently.

pujitha24:
This is squashed into one commit, description updated, and green across CI — happy to rename the title myself if you can point me to how, or anything else that would help move this along.

jnummelin:
> happy to rename the title myself if you can point me to how

You should see the little pen icon next to the PR number:
image

PR #8230, opened 2026-09-02
Drop privileges for the k0s control API process

### LINKED ISSUE #8210: k0scontrolapi subprocess has no privilege-dropped user
pkg/component/controller/k0scontrolapi.go:35 has an existing TODO: "Make the api process to use some other user." The control API subprocess currently runs as whoever launched k0s (typically root) instead of a scoped-down user.

Filing to discuss the right UID/GID and file/socket permission approach before sending a fix, since a wrong change here could break the process's access to what it needs.

magic-peach:
## Description

The k0s control API subprocess (`k0s api`) was the only supervised controller component still running as whoever launched k0s, instead of a dropped-privilege user. `apiserver.go`, `controllermanager.go`, `etcd.go`, `konnectivity.go`, `kine.go`, and `scheduler.go` all already look up a dedicated system user and pass it to `supervisor.Supervisor{UID: ...}`; `k0scontrolapi.go` did not.

This mirrors that existing pattern: it looks up `constant.ApiserverUser` (falling back to root if the lookup fails, same as the other components) and wires it into the supervisor. `EnsureControllerUsers` already guarantees this user exists before controller components start, so no new dedicated user is needed.

Since the control API process reads `admin.conf` to talk to the kube-apiserver, and that file was previously owned by root with owner-only permissions, it also chowns `admin.conf` to the same user so it stays readable after the privilege drop. All other files the control API touches (`k0s-api.crt/key`, `sa.key`, `ca.key`, `apiserver-etcd-client.crt/key`) are already owned by this user or readable via the existing group permissions, so no further changes were needed there.

Fixes #8210

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [ ] Auto test added

Ran `go build ./...`, `go vet ./pkg/component/controller/... ./cmd/controller/...`, and the existing test suites for both affected packages (`go test ./pkg/component/controller/... ./cmd/controller/...`), all passing.

## Checklist


cmd/controller/certificates.go
```diff
@@ -96,7 +96,8 @@ func (c *Certificates) Init(ctx context.Context) error {
 			return err
 		}
 
-		if err := kubeConfig(c.K0sVars.AdminKubeConfigPath, kubeConfigAPIUrl, c.CACert, adminCert.Cert, adminCert.Key, users.RootUID, constant.OwnerOnlyMode); err != nil {
+		// owned by apiServerUID so the k0s control API process can read it
+		if err := kubeConfig(c.K0sVars.AdminKubeConfigPath, kubeConfigAPIUrl, c.CACert, adminCert.Cert, adminCert.Key, apiServerUID, constant.OwnerOnlyMode); err != nil {
```
twz123: Changing the owner of the admin kubeconfig feels like the wrong solution.

jnummelin:
@magic-peach We just merged a fix for the stubbornly flaky `check-ap-controllerworker` smoketest. Could you rebase so we see all CI passing smoothly

magic-peach:
Rebased onto current main to pick up the smoketest fix.

PR #8233, opened 2026-09-02
Log swallowed errors during resource pruning

magic-peach:
## Description
findPruneableResourceForGroupVersionKind and getPruneableResources both discarded errors outright instead of at least logging them, unlike the discovery error handling a few lines above in the same file. This logs them the same way instead of dropping them silently.

Fixes # (issue)
N/A, no existing issue, found while reading the pruning code.

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

Ran the existing pkg/applier unit tests plus go vet, gofmt, and golangci-lint (v2.13.2, the version this repo pins) locally, all clean. No new test added since this only adds logging to two already-existing error paths.

## Checklist

### DIFF
--- a/pkg/applier/stack.go
+++ b/pkg/applier/stack.go
@@ -375,8 +375,10 @@ func (s *Stack) clientForResource(mapper meta.ResettableRESTMapper, resource uns
 }
 
 func (s *Stack) findPruneableResourceForGroupVersionKind(ctx context.Context, mapper meta.ResettableRESTMapper, groupVersionKind *schema.GroupVersionKind) []unstructured.Unstructured {
-	mapping, _ := getRESTMapping(mapper, groupVersionKind)
-	// FIXME error handling...
+	mapping, err := getRESTMapping(mapper, groupVersionKind)
+	if err != nil {
+		s.log.Debugf("error getting REST mapping for pruning: %s", err.Error())
+	}
 	if mapping != nil {
 		client, err := s.Clients.GetDynamicClient()
 		if err != nil {
@@ -397,7 +399,7 @@ func (s *Stack) getPruneableResources(ctx context.Context, drClient dynamic.Reso
 	}
 	resourceList, err := drClient.List(ctx, listOpts)
 	if err != nil {
-		// FIXME why no error propagation !??!
+		s.log.Debugf("error listing resources for pruning: %s", err.Error())
 		return nil
 	}
 	for _, resource := range resourceList.Items {
--- a/pkg/applier/stack.go
+++ b/pkg/applier/stack.go
@@ -377,11 +377,12 @@ func (s *Stack) clientForResource(mapper meta.ResettableRESTMapper, resource uns
 func (s *Stack) findPruneableResourceForGroupVersionKind(ctx context.Context, mapper meta.ResettableRESTMapper, groupVersionKind *schema.GroupVersionKind) []unstructured.Unstructured {
 	mapping, err := getRESTMapping(mapper, groupVersionKind)
 	if err != nil {
-		s.log.Debugf("error getting REST mapping for pruning: %s", err.Error())
+		s.log.Warnf("error getting REST mapping for pruning: %s", err.Error())
 	}
 	if mapping != nil {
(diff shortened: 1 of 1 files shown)


twz123:
I think the FIXMEs aren't about logging, but more about "we need to have a closer look at the whole thing, to figure out what's the right thing to do." Many (all) TODOs/FIXMEs aren't replaceable by one or two lines of code. If they were, there'd be no reason to add them instead of adding those one or two lines in the first place.

We can add the log statements, but those aren't the solution to the FIXMEs. Did you have a deeper look at that topic?

magic-peach:
@twz123 I made the changes

If you'd rather see actual propagation so findPruneableResources can decide whether a per-kind failure should fail the whole prune, happy to take a pass at that as a separate change, it's a bigger behavior decision than this PR was scoped for.

twz123:
As said, these FIXMEs aren't just replaceable by some logs. The log statements are arguably a good addition, so those errors won't go unnoticed, at least. So please re-add the FIXMEs, then we can merge this one. That said, the logs should really use `WithError`. There's some code guidelines that are about to be merged in #8196. Maybe those help.

Concerning the "real" fix for the FIXMEs: I don't know what's best. I would need to dig deep in the code to get an idea about that. You've filed this PR, so it seemed that you wanted to come up with an analysis and propose a solution?

magic-peach:
Done, re-added both FIXMEs and switched both new log lines plus the one I missed on GetDynamicClient to WithError, going by the guidelines in #8196.

On the real fix: I don't have a confident answer either, and I'd rather not guess my way into a design decision this PR wasn't scoped for. The tradeoff I keep landing on is that this whole function fans out in parallel per GVK and already tolerates one bad group failing discovery without aborting the rest, so propagating these two would need to decide whether one bad kind should now fail the entire prune, which feels like it deserves its own issue and a real look at how prune failures should be surfaced overall, not something to decide as a side effect of adding log lines. Happy to file that issue if useful, or take a pass at it separately, but for this PR I think logging plus the FIXMEs staying as an open marker is the honest state of things.

pkg/applier/stack.go
```diff
@@ -375,11 +375,15 @@ func (s *Stack) clientForResource(mapper meta.ResettableRESTMapper, resource uns
 }
 
 func (s *Stack) findPruneableResourceForGroupVersionKind(ctx context.Context, mapper meta.ResettableRESTMapper, groupVersionKind *schema.GroupVersionKind) []unstructured.Unstructured {
-	mapping, _ := getRESTMapping(mapper, groupVersionKind)
+	mapping, err := getRESTMapping(mapper, groupVersionKind)
 	// FIXME error handling...
+	if err != nil {
```
twz123, line 380: Now that we're checking for the error, we should probably also sanitize the control flow:

(`getRESTMapping` never returns a mapping together with an error, so the `if mapping != nil` nesting isn't needed anymore)

```go
mapping, err := getRESTMapping(mapper, groupVersionKind)
if err != nil {
      // FIXME error handling...
      log.WithError(err).WithField("gvk", groupVersionKind).Warn("Failed to get REST mapping, cannot prune this kind")
      return nil
}

client, err := s.Clients.GetDynamicClient()
if err != nil {
      log.WithError(err).WithField("gvk", groupVersionKind).Warn("Failed to get dynamic client, cannot prune this kind")
      return nil
}
```

twz123 (commented):
Thanks, getting closer. A few things about the log messages themselves, as per the new [logging guidelines]:

* Start logs with a capital letter.
* Name the consequences in the log message, e.g. by appending "cannot prune this kind" or similar.
* We might want to add the concrete kind to the logs. It's probably included in most of the error messages that can realistically reach this code, but when stating the outcome, as suggested above, it would be very weird if they didn't. One way to do this is to add, say, a `gvk` field to the logger.

[logging guidelines]: https://github.com/k0sproject/k0s/blob/775515627549e8dff9f384718dfa7ced6346a0ab/docs/contributors/code_guidelines.md#logging

Issue #8254, opened 2026-09-07
Move component stack applies to in-memory


jnummelin:
The async FS hop in between creating the manifests and applier actually triggering to apply is cause of pain. The latest is https://github.com/k0sproject/k0s/issues/8214. In that cases there's evidence where in some cases there might be 10-15sec time between when creating the stack dir and finally applying it. Now the applier treats and empty dir as empty stack, prunes all the things away, essentially nuking the cluster.

So let's move all the internal components to use in-memory stack applies. This mitigates the async FS ho entirely. We've done that already for some components, e.g. https://github.com/k0sproject/k0s/pull/6739.

Components we need to refactor:
- [ ] coredns; https://github.com/k0sproject/k0s/pull/8255
- [ ] calico
- [ ] kube-router
- [ ] kube-proxy
- [ ] konnectivity-agent
- [ ] metrics-server
- [ ] push-gateway

PR #8255, opened 2026-09-07
Use in-memory stack apply in coredns component

jnummelin:
## Description

This is to avoid the async hop via FS which has know timing etc. issues.

Part of #8254

Needs to land AFTER #8220 so we can backport that across all needed branches

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

### DIFF
--- a/pkg/component/controller/coredns.go
+++ b/pkg/component/controller/coredns.go
@@ -4,15 +4,18 @@
 package controller
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"math"
-	"path/filepath"
 	"reflect"
 	"time"
 
+	"github.com/k0sproject/k0s/internal/sync/value"
+	"github.com/k0sproject/k0s/pkg/applier"
+	"github.com/k0sproject/k0s/pkg/component/controller/leaderelector"
 	"github.com/k0sproject/k0s/pkg/component/manager"
-	"github.com/k0sproject/k0s/pkg/config"
+	"github.com/k0sproject/k0s/pkg/leaderelection"
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
@@ -20,10 +23,8 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/metadata"
 
-	"github.com/k0sproject/k0s/internal/pkg/dir"
 	"github.com/k0sproject/k0s/internal/pkg/templatewriter"
 	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
-	"github.com/k0sproject/k0s/pkg/constant"
 	k8sutil "github.com/k0sproject/k0s/pkg/kubernetes"
 )
 
@@ -262,6 +263,9 @@ spec:
     protocol: TCP
 `
 
+// CoreDNSStackName is the name of the in-memory applier stack managing the CoreDNS resources.
+const CoreDNSStackName = "coredns"
+
 const HostsPerExtraReplica = 10.0
 
 var _ manager.Component = (*CoreDNS)(nil)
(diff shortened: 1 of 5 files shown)


pkg/component/controller/coredns.go
```diff
@@ -312,14 +313,15 @@ func NewCoreDNS(k0sVars *config.CfgVars, clientFactory k8sutil.ClientFactoryInte
...
 		client:        client,
+		clientFactory: clientFactory,
+		leaderElector: leaderElector,
 		log:           logrus.WithField("component", "coredns"),
-		manifestDir:   filepath.Join(k0sVars.ManifestsDir, "coredns"),
 	}, nil
 }
 
-// Init does nothing
+// Init does nothing as there's nothing to initialize
```
twz123, line 327: :trollface: 😅

pkg/component/controller/coredns.go
```diff
@@ -418,9 +420,21 @@ func (c *CoreDNS) Stop() error {
 	return nil
 }
 
+// CoreDNSStackName is the name of the in-memory applier stack managing the CoreDNS resources.
+const CoreDNSStackName = "coredns"
```
twz123: Would probably move this more to the top of the file.

pkg/component/controller/coredns.go
```diff
@@ -418,9 +420,21 @@ func (c *CoreDNS) Stop() error {
...
+
 // Reconcile detects changes in configuration and applies them to the component
 func (c *CoreDNS) Reconcile(ctx context.Context, clusterConfig *v1beta1.ClusterConfig) error {
 	logrus.Debug("reconcile method called for: CoreDNS")
+
+	// Needed regardless of leadership, so the node-count ticker in Start can
+	// retry once this instance becomes the leader.
+	c.lastKnownClusterConfig = clusterConfig
```
twz123: Oh, I think this write races with the read in the goroutine in `Start`. So this is a data race that predates this PR. Maybe we want to fix this first in a preliminary PR that we can backport?

That said, given that we now do a full cluster sync here compared to just writing a file into the file system, I'd probably push the reconciliation to the `Start` goroutine. You could just save the current cluster config here (or just the `coreDNSConfig` distilled out of that), and then send a wakeup signal to the reconciling goroutine. There's `value.Latest` that you could use, if you don't want to hand-roll this pattern here.

pkg/component/controller/coredns.go
```diff
@@ -418,9 +420,21 @@ func (c *CoreDNS) Stop() error {
...
+
+	if status, _ := c.leaderElector.CurrentStatus(); status != leaderelection.StatusLeading {
+		return nil
+	}
```
twz123: In order to react more promptly to lost leases, you could set up a context:

```go
ctx, cancel := context.WithCancelCause(ctx)
go func() {
	select {
	case <-statusChanged:
		cancel(leaderelection.ErrLostLead)
	case <-ctx.Done():
	}
}()
```

pkg/component/controller/coredns_test.go
```diff
@@ -36,6 +43,51 @@ func TestCoreDNS_RenderWithPatch(t *testing.T) {
...
+
+	require.NoError(t, c.Reconcile(t.Context(), v1beta1.DefaultClusterConfig()))
+
+	_, err := clients.Client.CoreV1().ConfigMaps(metav1.NamespaceSystem).Get(t.Context(), "coredns", metav1.GetOptions{})
+	assert.True(t, apierrors.IsNotFound(err), "a non-leader must not apply the coredns stack, got: %v", err)
+	assert.NotNil(t, c.lastKnownClusterConfig, "lastKnownClusterConfig must still be tracked so a later leadership change gets picked up")
+}
+
+// pendingLeaderElector is a [leaderelector.Interface] that never leads.
+type pendingLeaderElector struct{}
```
twz123: Just an idea: If you change the `leaderelector.Interface` in `NewCoreDNS` to `func() (leaderelection.Status, <-chan struct{})`, you wouldn't need this helper struct.

pkg/component/controller/coredns.go
jnummelin: You mean for cases to cancel an "in-flight" reconcile when we loose the lease? I did think of that but IMO that might have unwanted side effects, we might apply stuff "halfway" and then cancel abruptly. That might leave things in weird state.

jnummelin: I created `LeaderContext` in https://github.com/k0sproject/k0s/pull/8255/commits/4b606fc42a5ed56f60accc7519c21b7ae9519875 in the anticipation that his will be needed also in other components when we transition to in-memory applies. So basically once we lose the leadership, we'll cancel the context passed on to any of the operations happening within `Reconcile`

jnummelin:
@twz123 Reworked the concurrency as we discussed in the community call, care to have another pass

PR #8260, opened 2026-09-07
Add unit tests for DecodeJoinToken, JoinEncode, and GetTokenType

magic-peach:
## Description

`pkg/token/joindecode.go` (DecodeJoinToken, GetTokenType) and `pkg/token/joinencode.go` (JoinEncode) had no direct unit tests. They were only exercised indirectly through `pkg/token/joinclient_test.go` (which uses JoinEncode as test setup, not as the thing under test) and `inttest/etcdlearner/etcdlearner_test.go`, which needs a live cluster. Neither exercises DecodeJoinToken's error paths.

Adds direct unit tests for DecodeJoinToken, JoinEncode, and GetTokenType covering the round trip, the invalid base64 and invalid gzip error paths in DecodeJoinToken, and both branches of GetTokenType.

Fixes # (issue)
N/A, no existing issue, found while looking for untested error paths.

## Type of change

- [ ] Documentation update

This is a test only change, none of the above categories quite fit; no production code is touched.

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist

### DIFF
--- a/pkg/token/joindecode_test.go
+++ b/pkg/token/joindecode_test.go
@@ -1,0 +1,83 @@
+// SPDX-FileCopyrightText: 2026 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package token_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/k0sproject/k0s/pkg/token"
+
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDecodeJoinToken_RoundTrip(t *testing.T) {
+	t.Parallel()
+
+	encoded, err := token.JoinEncode(bytes.NewReader([]byte("the-payload")))
+	require.NoError(t, err)
+
+	decoded, err := token.DecodeJoinToken(encoded)
+	require.NoError(t, err)
+	assert.Equal(t, []byte("the-payload"), decoded)
+}
+
+func TestDecodeJoinToken_InvalidBase64(t *testing.T) {
+	t.Parallel()
+
+	decoded, err := token.DecodeJoinToken("not-valid-base64!!!")
+	assert.ErrorContains(t, err, "illegal base64 data")
+	assert.Zero(t, decoded)
+}
+
+func TestDecodeJoinToken_InvalidGzip(t *testing.T) {
+	t.Parallel()
+
+	// Valid base64, but not gzip data underneath.
+	decoded, err := token.DecodeJoinToken("bm90LWd6aXA=")
+	assert.ErrorContains(t, err, "unexpected EOF")
+	assert.Zero(t, decoded)
+}
+
+func TestGetTokenType(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		contexts map[string]*clientcmdapi.Context
+		want     []string // map iteration order is unspecified, so any of these is acceptable
+	}{
+		{
+			name: "zero contexts",
+			want: []string{""},
+		},
+		{
+			name: "one context",
+			contexts: map[string]*clientcmdapi.Context{
+				"the-context": {AuthInfo: "the-auth-info"},
+			},
+			want: []string{"the-auth-info"},
+		},
+		{
+			name: "two contexts",
+			contexts: map[string]*clientcmdapi.Context{
+				"context-a": {AuthInfo: "auth-a"},
+				"context-b": {AuthInfo: "auth-b"},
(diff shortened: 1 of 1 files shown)


pkg/token/joindecode_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Valid base64, but not gzip data underneath.
+	_, err := token.DecodeJoinToken("bm90LWd6aXA=")
+	assert.Error(t, err)
+}
+
+func TestGetTokenType(t *testing.T) {
+	t.Parallel()
+
+	assert.Equal(t, "", token.GetTokenType(&clientcmdapi.Config{}))
```
twz123: Might make sense to have a table test with zero, one and two contexts.

pkg/token/joindecode_test.go
```diff
@@ -0,0 +1,55 @@
...
+	decoded, err := token.DecodeJoinToken(encoded)
+	require.NoError(t, err)
+	assert.Equal(t, []byte("the-payload"), decoded)
+}
+
+func TestDecodeJoinToken_InvalidBase64(t *testing.T) {
+	t.Parallel()
+
+	_, err := token.DecodeJoinToken("not-valid-base64!!!")
+	assert.Error(t, err)
```
twz123: Let's check the actual error. Not sure if there's a concrete type or value that we can check for, but at least a simple `assert.ErrorContains(t, err, "some meaningful error message substring")` would make sense.

Also, for completeness, we should check if the returned value satisfies `assert.Zero(...)`

pkg/token/joindecode_test.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func TestDecodeJoinToken_InvalidGzip(t *testing.T) {
+	t.Parallel()
+
+	// Valid base64, but not gzip data underneath.
+	_, err := token.DecodeJoinToken("bm90LWd6aXA=")
+	assert.Error(t, err)
```
twz123: Let's check the actual error. Not sure if there's a concrete type or value that we can check for, but at least a simple `assert.ErrorContains(t, err, "some meaningful error message substring")` would make sense.

Also, for completeness, we should check if the returned value satisfies `assert.Zero(...)`

pkg/token/joindecode_test.go
magic-peach: Turned this into a table test covering zero, one, and two contexts. Since map iteration order is unspecified, the two-context case asserts the result is one of the two AuthInfo values present rather than a single fixed one.

pkg/token/joindecode_test.go
magic-peach: Added assert.ErrorContains(t, err, "illegal base64 data") and assert.Zero(t, decoded).

pkg/token/joindecode_test.go
magic-peach: Added assert.ErrorContains(t, err, "unexpected EOF") and assert.Zero(t, decoded). The gzip.NewReader call fails with unexpected EOF here since the decoded payload is only 8 bytes, too short for it to even read a gzip header.

pkg/token/joindecode_test.go
```diff
@@ -0,0 +1,83 @@
...
+			want: []string{"auth-a", "auth-b"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			cfg := &clientcmdapi.Config{Contexts: test.contexts}
+			assert.Contains(t, test.want, token.GetTokenType(cfg))
```
twz123: Shouldn't this rather be `assert.Equal(t, test.want, token.GetTokenType(cfg))` with a `test.want` of type `string`? Not sure why you're passing in a slice here.

pkg/token/joindecode_test.go
```diff
@@ -0,0 +1,83 @@
...
+	assert.Zero(t, decoded)
+}
+
+func TestGetTokenType(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		contexts map[string]*clientcmdapi.Context
+		want     []string // map iteration order is unspecified, so any of these is acceptable
```
twz123, line 52: Right. Just had a short look at the code to be tested and saw the for range loop, without realizing that it's looping over a map rather than a slice. So I asked for the table test, so we can test that only the first context is picked up. With a map, though, the two context test case is not super meaningful. But well, now we have it.

twz123 (commented):
Could you please squash your commits?

magic-peach:
Squashed, thanks.

twz123 (approved):

Issue #8285, opened 2026-09-11
k0s reset leaves containerd shim state under the shared /run/containerd, breaking same-ID container creation after reinstall


vikramhh:
### What happened?

The shims of k0s's managed containerd keep their runtime state under the host-shared `/run/containerd`, not under k0s's own run dir: with a single-node cluster running, a container's runc state is at `/run/containerd/runc/k8s.io/<id>` and its IO fifos under `/run/containerd/fifo/<n>/`. `k0s reset` cleans `/run/k0s`, the data dir and the kubelet root dir, but never `/run/containerd` — understandably, since that path can be shared with a host containerd or Docker and cannot be deleted blindly.

Container state created outside the CRI plugin (e.g. via `ctr`/`nerdctl` in the `k8s.io` namespace, or a leftover shim) is not cleaned by the CRI-based `containers` step, so it survives a **successful** reset. After reset + reinstall on the same booted host, recreating a container with the same ID then fails:

```
Error: failed to create shim task: OCI runtime create failed: runc create failed: container with given ID already exists
```

A reboot clears it (`/run` is tmpfs), matching reset's "a node reboot is recommended" hint — but automated stop → reset → reinstall flows that do not reboot hit this.

### Steps to reproduce (released v1.36.3+k0s.0, verified on a fresh Ubuntu 22.04 EC2 instance)

```bash
k0s install controller --single && k0s start          # wait Ready
k0s ctr -n k8s.io run -d docker.io/library/busybox:1.36 samename sleep 100000
# /run/containerd/runc/k8s.io/samename now exists (shared path, not under /run/k0s)
k0s stop
# release the task mount by hand so reset does not fail on #8048:
umount /run/k0s/containerd/io.containerd.runtime.v2.task/k8s.io/samename/rootfs
k0s reset --data-dir /var/lib/k0s --kubelet-root-dir /var/lib/kubelet   # succeeds (rc=0)
# /run/containerd/runc/k8s.io/samename and a /run/containerd/fifo/* dir remain
k0s install controller --single && k0s start
k0s ctr -n k8s.io run -d docker.io/library/busybox:1.36 samename sleep 100000
#   -> Error: ... container with given ID already exists
```

Verified outcome: `reset` exits 0, the same-ID create exits 1 with the error above; k0s's own directories (`/run/k0s`, `/var/lib/k0s`) were removed while `/run/containerd/runc/k8s.io/samename` persisted.

### Expected behavior

[...]

PR #8286, opened 2026-09-11
Unmount leftover mounts under the run dir on reset

### LINKED ISSUE #8048: k0s reset fails on leftover container mounts under the run dir, and the failed attempt makes further resets impossible
**Before creating an issue, make sure you've checked the following**
- [x] You're looking at docs for the released version

### Platform
```
Ubuntu 22.04.5 LTS, kernel 6.8.0-1061-aws, x86_64
AWS EC2 t3.medium (2 vCPU, 3.7 GiB), single node
```

### Version
```
v1.36.3+k0s.0
```

Also reproduced identically on `v1.35.4+k0s.0`. `pkg/cleanup` is byte-identical between the two — `git diff v1.35.4+k0s.0 v1.36.3+k0s.0 -- pkg/cleanup/` is empty.

### Sysinfo

<details><summary><code>k0s sysinfo</code></summary>

```
Total memory: 3.7 GiB (pass)
File system of /var/lib: ext4 (pass)
Disk space available for /var/lib/k0s: 26.8 GiB (pass)
Relative disk space available for /var/lib/k0s: 92% (pass)
Name resolution: localhost: [127.0.0.1] (pass)
Containerd config snippets in /etc/k0s/containerd.d/: no config snippets found (pass)
  Linux kernel release: 6.8.0-1061-aws (pass)
  AppArmor: active (pass)
      CONFIG_NF_NAT: module (pass)
      CONFIG_LLC: module (pass)
      CONFIG_STP: module (pass)
```
</details>

### What happened?

For context, the three directories the reset cleanup deals with:

- **`dataDir`** — the k0s data directory, `/var/lib/k0s` by default (`constant.DataDirDefault`), overridable with `--data-dir`. Holds the bundled binaries in `bin/` (`runc`, `containerd`, `containerd-shim-runc-v2`, `etcd`, …), the containerd content store and overlayfs snapshots, etcd data and PKI.
- **`kubeletRootDir`** — `<dataDir>/kubelet` by default, overridable with `--kubelet-root-dir`. (The reproduction below passes `/var/lib/kubelet` explicitly, which is not the default.)
- **`runDir`** — runtime state: supervised PID files and sockets, `containerd.sock`, `containerd-cri.toml`, and **containerd's task state under `io.containerd.runtime.v2.task/<ns>/<id>/`, which includes each running container's mounted `rootfs` overlay**. This is `/run/k0s` whenever k0s runs as root and `<dataDir>/run` otherwise; there is no flag to change it.

Two separate problems, the second more serious than the first.

**1. The reset cleanup never unmounts under `runDir`, but then deletes it.**

In `pkg/cleanup/directories.go` the unmount loop is guarded by:

```go
if isUnderPath(v.Path, d.kubeletRootDir) || isUnderPath(v.Path, d.dataDir) {
```

`runDir` is not included. Container task rootfs overlays live under it — `/run/k0s/containerd/io.containerd.runtime.v2.task/<ns>/<id>/rootfs` — so any mount that survives the `containers` cleanup step is still mounted when the same function reaches:

```go
if err := os.RemoveAll(d.runDir); err != nil {
    return fmt.Errorf("failed to delete %s: %w", d.runDir, err)
}
```

`UnmountLazy` (`MNT_DETACH`) already exists in the package but is only reachable from the dataDir/kubeletRootDir loop, and the `errorIsUnlinkat` mountpoint tolerance is likewise applied only to `dataDir`. Mount-awareness is present in one half of the function and absent in the other.

**2. A failed reset cannot be retried.**

The same step deletes `dataDir` **before** `runDir`. Once the `runDir` delete fails, `/var/lib/k0s/bin` is already gone, so the next `k0s reset` cannot extract `runc`/`containerd-shim`, cannot start containerd, skips container cleanup entirely, and fails on the same mount in ~140 ms. The first failure destroys what any retry needs; the node then requires a manual unmount or a reboot.

This matters for any automated stop → reset → restore flow, which cannot recover by retrying.

### Steps to reproduce

```bash
k0s install controller --single && k0s start        # wait for the node to be Ready
kubectl run holder --image=busybox:1.36 -- sleep 100000

# Any container the CRI-based cleanup cannot see will do. Simplest -- note this
# uses k8s.io, the containerd namespace the CRI plugin itself uses
# (containerd's constants.K8sContainerdNamespace), so the straggler sits in the
# same namespace as every real pod on the node:
k0s ctr -n k8s.io images pull docker.io/library/busybox:1.36
k0s ctr -n k8s.io run -d docker.io/library/busybox:1.36 straggler sleep 100000

k0s stop
k0s reset --data-dir /var/lib/k0s --kubelet-root-dir /var/lib/kubelet   # fails on /run/k0s
k0s reset --data-dir /var/lib/k0s --kubelet-root-dir /var/lib/kubelet   # ~140ms, containerd cannot start
```

A control run **without** the extra container succeeds and leaves zero task mounts, confirming the `containers` step is what normally clears `runDir`.

### Expected behavior

`k0s reset` releases mounts under `runDir` before deleting it, as it already does for `dataDir` and `kubeletRootDir`. A failed reset leaves the node in a state where reset can be retried.

### Actual behavior

```
Error: errors occurred during clean-up: failed to delete /run/k0s: unlinkat
/run/k0s/containerd/io.containerd.runtime.v2.task/<ns>/<id>: directory not empty
```

and on every subsequent attempt:

```
level=warning msg="Failed to initialize containerd, skipping container cleanup"
  error="open /var/lib/k0s/bin/.runc.<random>.tmp: no such file or directory"
```

### Screenshots and logs

**Three sequential resets after a single surviving mount**

`v1.36.3+k0s.0`:

| Attempt | Duration | Outcome |
|---|---|---|
| 1 | 1603 ms | `failed to delete /run/k0s: unlinkat …: directory not empty`; `dataDir` and `dataDir/bin` now gone |
| 2 | 142 ms | `Failed to initialize containerd, skipping container cleanup`; same delete failure |
| 3 | 139 ms | same |

`v1.35.4+k0s.0`: 2300 ms / 148 ms / 143 ms, identical messages (`rc=1` on all three).

**Single-straggler case, both versions, 2 runs each**

```
v1.36.3+k0s.0  reset_rc=1  ENOTEMPTY  mounts_after_stop=7   mounts_after_reset=1
v1.35.4+k0s.0  reset_rc=1  ENOTEMPTY  mounts_after_stop=11  mounts_after_reset=1
```

The surviving mount in both cases:

```
/run/k0s/containerd/io.containerd.runtime.v2.task/k8s.io/straggler/rootfs
```

The straggler is in the **same containerd namespace as the running cluster**, not an
isolated one -- `k0s ctr -n k8s.io containers ls` immediately before `k0s stop`:

```
CONTAINER                IMAGE                                                       RUNTIME
249c9fe6f...             quay.io/k0sproject/kube-proxy:v1.36.3                       io.containerd.runc.v2
41645d479...             quay.io/k0sproject/kube-router:v2.10.0-iptables1.8.13-k0s.0 io.containerd.runc.v2
8b117797a...             quay.io/k0sproject/pause:3.10.2-0                           io.containerd.runc.v2
9c5c4adc4...             quay.io/k0sproject/cni-node:1.8.0-k0s.0                     io.containerd.runc.v2
a5197f4b4...             docker.io/library/busybox:1.36                              io.containerd.runc.v2   <- the "holder" pod
straggler                docker.io/library/busybox:1.36                              io.containerd.runc.v2   <- survives reset
```

So the namespace is not what makes the container invisible to the cleanup. The
`containers` step enumerates via the CRI `ListPodSandbox` RPC
(`pkg/container/runtime/cri.go`), which returns **pod sandboxes registered with the
CRI plugin** -- not the contents of a containerd namespace. Anything created outside
the CRI plugin is missed no matter where it lives, which is why a leftover shim from
a crashed/upgraded runtime, or anything created with `ctr`/`nerdctl` in `k8s.io`,
produces this.

In every run the `containers` step cleaned every CRI-visible mount and left exactly the one straggler; the delete then failed while unlinking it:

```
Error: errors occurred during clean-up: failed to delete /run/k0s: unlinkat
/run/k0s/containerd/io.containerd.runtime.v2.task/k8s.io/straggler: directory not empty
```

**Controls**

With no surviving mount, reset exits 0 and leaves zero task mounts (3/3 runs on `v1.36.3+k0s.0`, 2/2 on `v1.35.4+k0s.0`). With the `containers` step disabled — `--cri-socket` pointing at a non-existent socket, so k0s does not start its own containerd — reset fails on `/run/k0s` in 3/3 runs at each version.

Whether the delete surfaces `ENOTEMPTY` or `EBUSY` varies between identical runs (readdir order); it is the same defect either way.

### Additional context

Suggested fixes:

1. Include `runDir` in the unmount loop, or extend the existing `errorIsUnlinkat` tolerance to the `runDir` delete.
2. Make reset idempotent under failure — delete `runDir` before `dataDir`, or otherwise preserve the binaries a later attempt needs.

**FYI on overlap with #6472.** PR #6472 ("Don't unmount kubelet-root-dir on k0s reset") modifies the same `(d *directories) Run()` function: it adds `kubeletRootDirMounted`/`kubeletRootDirSeparate`, changes the same `procMounts` loop (also refactoring it to `slices.Backward`), and extends the `errorIsUnlinkat` tolerance to `kubeletRootDir`. That is the same pattern suggested above for `runDir`, so the two changes look complementary — but they will conflict textually, and merging them will likely need coordinating.

Related: **#5657** — `k0s stop` leaving containers running is what creates the precondition. That is expected containerd behaviour (shims are deliberately decoupled) and is tracked there, not part of this report. **#2852** — same family: reset not fully cleaning up.

vikramhh:
## Description

`k0s reset`'s directories step unmounts everything under the data dir and the kubelet root dir before deleting them, but not under the run dir — where containerd's task state lives, including each container's mounted `rootfs` overlay. Any mount that survives the containers cleanup step (anything created outside the CRI plugin, or a leftover shim) makes `os.RemoveAll(runDir)` fail with `ENOTEMPTY`. Because the data dir is deleted before the run dir, that first failure also destroys the bundled binaries (`runc`, `containerd-shim`, …) that a retried reset needs to start containerd — so every subsequent reset fails in ~140 ms and the node needs a manual unmount or a reboot.

Three changes in `pkg/cleanup/directories.go`:

- **Mounts under the run dir are unmounted** in the same reverse-order loop that already covers the data dir and the kubelet root dir, with the existing lazy-unmount fallback — container task `rootfs` overlays live there.
- **A run dir that is itself a mount point is emptied rather than removed**, with the same `unlinkat` tolerance the other two directories already get — completing the mount-awareness parity the issue describes.
- **The run dir is deleted before the data dir**, so a failure at that point leaves the bundled binaries a retried reset needs. A comment records why the order matters.

Regression tests cover all three changes (mounts under every managed dir are unmounted while unrelated mounts are left alone; the data dir is deleted last; a failed run-dir deletion preserves the data dir; the non-root layout where the run dir is nested inside the data dir; and the mounted-run-dir `unlinkat` tolerance, see below). Two small seams (`newMounter`, `removeAll`) make the step testable with the fake mounter from `k8s.io/mount-utils`.

The second change's k0s-side handling is unit-tested: `TestRunToleratesUnlinkatWhenRunDirIsAMountPoint` and its two negative cases inject a synthetic `*os.PathError{Op: "unlinkat"}` through the `removeAll` seam and assert the tolerance is gated on the run dir being a mount point and on the error actually being an `unlinkat` on that directory. What unit tests still cannot reproduce is the *kernel* behavior that produces that error — `os.RemoveAll` partially emptying a real mount point and then failing `unlinkat` on the mountpoint itself — so the end-to-end "run dir as its own mount point" scenario in Manual Testing remains the only confirmation that the real error takes this path.

While verifying this fix, a related but separate gap surfaced: containerd shim state (runc state, IO fifos) lives under the host-shared `/run/containerd`, outside every directory reset cleans, and survives even a successful reset. Filed as #8285; it is out of scope here.

Fixes #8048

## Type of change

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

Manual verification, control vs. patched **on the same host** (Ubuntu 22.04.5, kernel 6.8, x86_64, AWS EC2, single-node `k0s install controller --single`), following the reproduction from #8048.

- **Setup, identical for both runs**: start k0s and wait for the node to be Ready; run a `holder` pod through the API server; create one non-CRI container in the `k8s.io` containerd namespace (`k0s ctr -n k8s.io run -d docker.io/library/busybox:1.36 <id> sleep 100000`); `k0s stop`; then `k0s reset --data-dir /var/lib/k0s --kubelet-root-dir /var/lib/kubelet`, twice.
- **Control (released `v1.36.3+k0s.0`)**: reset attempt 1 exits 1 after ~1.6 s with `failed to delete /run/k0s: unlinkat …/io.containerd.runtime.v2.task/k8s.io/<id>: directory not empty`; `/var/lib/k0s/bin` is already gone. Attempt 2 exits 1 in ~210 ms with `Failed to initialize containerd, skipping container cleanup` and recommends a reboot. One task mount remains.
- **Patched (this branch)**: reset attempt 1 exits 0 and releases the mount; attempt 2 also exits 0 (~94 ms); zero task mounts remain.

Additional scenarios, each on its own fresh single-node instance with the patched binary:

- **Failure-injected retry**: with an immutable file under the kubelet root dir, reset attempt 1 fails but `/var/lib/k0s/bin` is left intact, and attempt 2 then completes cleanly (rc=0, no task mounts) — the retryability the ordering change protects.
- **Run dir as its own mount point**: reset empties it and leaves it mounted rather than failing — the live check for the second change above.
- **Busy task mount**: a process holding the task `rootfs` forces the lazy-unmount fallback, and the reset still succeeds.
- **Decoy mounts**: bind and tmpfs mounts outside the k0s directories survive a reset untouched.
- **Default layout**: with the kubelet root under the data dir, reset completes cleanly.

`go test ./pkg/...` passes on Linux (51 packages, including the new tests), as do `GOOS=linux go build` and `go vet`.

## Checklist

### DIFF
--- a/pkg/cleanup/directories_linux_test.go
+++ b/pkg/cleanup/directories_linux_test.go
@@ -8,6 +8,7 @@ package cleanup
 import (
 	"context"
 	"errors"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -30,6 +31,28 @@ func stubSeams(t *testing.T, mounter mount.Interface, removeErrFor string) *[]st
 		removed = append(removed, path)
 		if removeErrFor != "" && path == removeErrFor {
 			return errors.New("injected removal failure")
+		}
+		return nil
+	}
+	return &removed
+}
+
+// stubSeamsErr is stubSeams with a caller-supplied error returned for the
+// removal of one specific path, so a test can inject a synthetic *os.PathError
+// and exercise the mounted-directory tolerance logic (errorIsUnlinkat) without
+// a real mount point.
+func stubSeamsErr(t *testing.T, mounter mount.Interface, errPath string, errVal error) *[]string {
+	t.Helper()
+	var removed []string
+
+	prevMounter, prevRemove := newMounter, removeAll
+	t.Cleanup(func() { newMounter, removeAll = prevMounter, prevRemove })
+
+	newMounter = func() mount.Interface { return mounter }
+	removeAll = func(path string) error {
+		removed = append(removed, path)
+		if path == errPath {
+			return errVal
 		}
 		return nil
 	}
@@ -131,4 +154,46 @@ func TestRunHandlesRunDirNestedInDataDir(t *testing.T) {
 
 	assert.Contains(t, unmountedPaths(fake), taskMount)
 	assert.Equal(t, []string{d.kubeletRootDir, d.runDir, d.dataDir}, *removed)
+}
+
+
+// The run dir can itself be a mount point (e.g. a tmpfs on /run/k0s). os.RemoveAll
+// then empties it and returns an unlinkat error on the mountpoint directory,
+// which the cleanup tolerates rather than failing. This exercises the k0s
+// decision logic with a synthetic *os.PathError; the kernel behavior that
+// actually produces it is covered by a live test (see the PR).
+func TestRunToleratesUnlinkatWhenRunDirIsAMountPoint(t *testing.T) {
+	d := testDirs(t)
+	fake := mount.NewFakeMounter([]mount.MountPoint{{Path: d.runDir}})
+	removed := stubSeamsErr(t, fake, d.runDir, &os.PathError{Op: "unlinkat", Path: d.runDir, Err: errors.New("directory not empty")})
+
+	require.NoError(t, d.Run(context.Background()),
+		"an unlinkat on a run dir that is itself a mount point must be tolerated")
+	assert.Equal(t, []string{d.kubeletRootDir, d.runDir, d.dataDir}, *removed,
+		"tolerating the run dir removal must not stop the data dir cleanup")
+}
+
+// A non-unlinkat error (or an unlinkat on a different path) on a mounted run dir
+// is a real failure, not the empty-a-mount-point case, so it must surface.
+func TestRunFailsOnNonUnlinkatErrorForMountedRunDir(t *testing.T) {
+	d := testDirs(t)
+	fake := mount.NewFakeMounter([]mount.MountPoint{{Path: d.runDir}})
+	removed := stubSeamsErr(t, fake, d.runDir, &os.PathError{Op: "remove", Path: d.runDir, Err: errors.New("permission denied")})
+
+	err := d.Run(context.Background())
+	require.ErrorContains(t, err, "contents of mounted run-dir")
+	assert.NotContains(t, *removed, d.dataDir, "a hard run dir failure must not proceed to the data dir")
+}
+
+// The unlinkat tolerance is gated on the run dir being a mount point. The same
+// error when the run dir is NOT mounted is a genuine failure and must surface.
+func TestRunFailsOnUnlinkatWhenRunDirNotMounted(t *testing.T) {
+	d := testDirs(t)
+	removed := stubSeamsErr(t, mount.NewFakeMounter(nil), d.runDir, &os.PathError{Op: "unlinkat", Path: d.runDir, Err: errors.New("directory not empty")})
+
+	err := d.Run(context.Background())
+	require.ErrorContains(t, err, d.runDir)
+	assert.NotContains(t, err.Error(), "contents of mounted run-dir",
+		"an unmounted run dir failure is the hard-delete error, not the mounted-tolerance path")
+	assert.NotContains(t, *removed, d.dataDir)
 }
--- a/pkg/cleanup/directories_linux_test.go
+++ b/pkg/cleanup/directories_linux_test.go
@@ -1,0 +1,134 @@
+//go:build linux
+
+// SPDX-FileCopyrightText: 2025 k0s authors
+// SPDX-License-Identifier: Apache-2.0
+
+package cleanup
+
+import (
+	"context"
+	"errors"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/mount-utils"
+)
+
+// stubSeams replaces the host-touching seams for the duration of a test and
+// returns a recorder of the directory removals in call order.
+func stubSeams(t *testing.T, mounter mount.Interface, removeErrFor string) *[]string {
+	t.Helper()
+	var removed []string
+
+	prevMounter, prevRemove := newMounter, removeAll
+	t.Cleanup(func() { newMounter, removeAll = prevMounter, prevRemove })
+
+	newMounter = func() mount.Interface { return mounter }
+	removeAll = func(path string) error {
+		removed = append(removed, path)
+		if removeErrFor != "" && path == removeErrFor {
+			return errors.New("injected removal failure")
+		}
+		return nil
+	}
+	return &removed
+}
+
+func testDirs(t *testing.T) *directories {
+	t.Helper()
+	base := t.TempDir()
+	return &directories{
+		dataDir:        filepath.Join(base, "data"),
+		kubeletRootDir: filepath.Join(base, "kubelet"),
+		runDir:         filepath.Join(base, "run"),
+	}
+}
+
+func unmountedPaths(fake *mount.FakeMounter) []string {
+	var paths []string
+	for _, action := range fake.GetLog() {
+		if action.Action == mount.FakeActionUnmount {
+			paths = append(paths, action.Target)
+		}
+	}
+	return paths
+}
+
+// Regression test for the first half of https://github.com/k0sproject/k0s/issues/8048:
+// mounts under the run dir (containerd task rootfs overlays) were never
+// unmounted before the run dir was deleted, unlike mounts under the data dir
+// and the kubelet root dir.
+func TestRunUnmountsLeftoverMountsUnderEveryManagedDir(t *testing.T) {
+	d := testDirs(t)
+	taskMount := filepath.Join(d.runDir, "containerd/io.containerd.runtime.v2.task/k8s.io/straggler/rootfs")
+	dataMount := filepath.Join(d.dataDir, "containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs")
+	kubeletMount := filepath.Join(d.kubeletRootDir, "pods/x/volumes/kubernetes.io~projected/kube-api-access")
+	unrelatedMount := filepath.Join(t.TempDir(), "unrelated")
+
+	fake := mount.NewFakeMounter([]mount.MountPoint{
+		{Path: dataMount},
+		{Path: kubeletMount},
+		{Path: taskMount},
+		{Path: unrelatedMount},
+	})
+	stubSeams(t, fake, "")
+
+	require.NoError(t, d.Run(context.Background()))
+
+	unmounted := unmountedPaths(fake)
+	assert.Contains(t, unmounted, taskMount,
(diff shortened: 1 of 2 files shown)


vikramhh:
Thanks for the review. Addressed in 61af534: the package-level seams (`newMounter`, `removeAll`) are now explicitly documented as serial-only test overrides that must not be used with `t.Parallel()`. Left as package-level rather than struct-level DI to avoid carrying Linux-only mounter types on the cross-platform `directories` type, as discussed. The other two points (run-dir-as-mountpoint coverage, and #8285 for host `/run/containerd` state) are already documented in the PR description.

jnummelin:
> Thanks for the review.

Which review? 🤔  😂

jnummelin:
> The second change is not unit-tested: its unlinkat tolerance only triggers when os.RemoveAll hits a real mount point, which the removeAll seam cannot faithfully model (a stubbed error cannot reproduce the kernel's partial-removal-then-unlinkat-on-the-mountpoint **behavior)**

I get that we can't repro the _actual_ kernel behaviour in tests, but why couldn't we make the `removeAll` stub return `&os.PathError{Op: "unlinkat", ...}` in the tests? That way we can exercise k0s logic at least?

vikramhh:
You are right above. While the kernel behavior (partial removal then unlinkat on the mountpoint) can only be tested "live", the k0s-side handling is testable exactly as you suggest. I added unit tests based on that in [452ef24](https://github.com/k0sproject/k0s/pull/8286/changes/452ef242e02b152b1c55623500b162bdac0f2ba4) and modified the PR description accordingly, please take a look.

vikramhh:
Hey @jnummelin - sorry that `thanks for the review` was confusing because this repo likely has no automated reviews and it was related to an offline review. 

> > Thanks for the review.
> 
> Which review? 🤔 😂

pkg/cleanup/directories_linux_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	fake := mount.NewFakeMounter([]mount.MountPoint{
+		{Path: dataMount},
+		{Path: kubeletMount},
+		{Path: taskMount},
+		{Path: unrelatedMount},
+	})
+	stubSeams(t, fake, "")
+
+	require.NoError(t, d.Run(context.Background()))
```
jnummelin: Why background context? Shouldn't we use `t.Context()` so timeouts etc. gets wired in properly?

s3rj1k: we should, see how linters would have helped here

vikramhh: What you are suggesting is essentially equivalent because `Run` takes the context only to satisfy the `cleanup.Step` interface and the parameter is unnamed. But it should be done anyway to use the idiomatic form and to account for the possibility that `Run` could start using the context in the future. Changed in 68431818c.

ncopa (commented):
> Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Can you please remove this from the commit message?

vikramhh:
> > Co-Authored-By: Claude Fable 5 [noreply@anthropic.com](mailto:noreply@anthropic.com)
> 
> Can you please remove this from the commit message?

@ncopa - done

PR #8291, opened 2026-09-13
Add unit tests for reconcileMember

### LINKED ISSUE #8211: etcd.go and etcd_member_reconciler.go have no unit test coverage
A sweep of pkg/component/controller/ for files missing a matching _test.go turned up 15 files with none. etcd.go and etcd_member_reconciler.go stand out given they manage the single most critical stateful component in the cluster (membership, join/leave, backup) with zero coverage.

Filing to scope this properly -- happy to take a pass at it if there's interest, this is a large-ish effort so wanted to check first rather than show up with an unreviewed test suite.

magic-peach:
## Description

Adds unit tests for `reconcileMember` in `etcd_member_reconciler.go`, per the scope we agreed on in #8211: invalid memberID, a member no longer in the cluster (both prior Joined states), a member not marked to leave, and the last controller refusing to leave. These exercise the decision logic directly against a fake `EtcdMemberInterface`, without needing a live etcd process.

Fixes #8211

## Type of change

- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

## Checklist


pkg/component/controller/etcd_member_reconciler_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"testing"
+
+	etcdv1beta1 "github.com/k0sproject/k0s/pkg/apis/etcd/v1beta1"
+	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+	fakeclientset "github.com/k0sproject/k0s/pkg/client/clientset/fake"
+	"github.com/k0sproject/k0s/pkg/etcd"
+	"github.com/stretchr/testify/assert"
+	ktesting "k8s.io/client-go/testing"
```
twz123: We already have imports that name this package `clientgotesting`, `k8stesting`, or just `testing`. Let's not add a fourth variant 🙈

pkg/component/controller/etcd_member_reconciler_test.go
```diff
@@ -0,0 +1,99 @@
...
+				Conditions: []etcdv1beta1.JoinCondition{{Type: etcdv1beta1.ConditionTypeJoined, Status: etcdv1beta1.ConditionFalse}},
+			}},
+			wantResult: true,
+			check: func(t *testing.T, _ *etcdv1beta1.EtcdMember, actions []ktesting.Action) {
+				assert.Empty(t, actions)
+			},
+		},
+		{
+			name:       "member gone, previously joined",
+			member:     &etcdv1beta1.EtcdMember{Status: etcdv1beta1.Status{MemberID: "1"}},
```
twz123: This member has no `Joined` condition at all, so the case covers "join status unset", not "previously joined". The code treats both the same, but the name promises something that isn't checked for.

Let's have both the unset and joined cases covered, in two sibling test cases.

twz123 (commented):
Overall, I don't think it makes sense to have this as table tests. The loop body is not doing much, the assertions themselves are already closures, and future tests may diverge in the required test setup.

So what about converting this into "regular" subtests, and just share a little bit of common setup via functions, à la

```go
func TestReconcileMember(t *testing.T) {
	// One or more shared helpers for test setup and so on...
	reconcile := func(t *testing.T, ...) (...) {
		// shared stuff across test cases
	}

	// Here go the test cases...
	t.Run("marks member as failed when member ID is invalid", func(t *testing.T) {
		// Test setup goes first
		member := ...

		// Then call the reconciler
		... := reconcile(t, ...)

		// Finally some assertions, preferrably with messages
		// stating the expectation in a way that adds value.
		assert.False(t, ...)
		assert.Equal(t, ...)
 	})

	// ...and so on.
}
```

Some general recommendations:

* Consider adding messages to the testify assertions whenever that adds meaningful context. Adding a message that states the expectation is a good rule of thumb. So, picking one of the added cases as an example: `assert.Empty(t, actions, "A member that isn't marked to leave shouldn't be touched")` would be good. Brevity is a plus, as long as this doesn't lose important context or reads awkward.
* Try to let the names describe the behavior, instead of the input values, e.g. say "marks member as failed when member ID is invalid" instead of  "invalid member id". Only "last controller refuses to leave" states an outcome. Same here: Brevity is a plus, as long as this doesn't lose important context or reads awkward. It might be necessary to make a compromise and have a shorter name and an additional comment in the code.

pkg/component/controller/etcd_member_reconciler_test.go
```diff
@@ -0,0 +1,99 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"testing"
+
+	etcdv1beta1 "github.com/k0sproject/k0s/pkg/apis/etcd/v1beta1"
+	"github.com/k0sproject/k0s/pkg/apis/k0s/v1beta1"
+	fakeclientset "github.com/k0sproject/k0s/pkg/client/clientset/fake"
```
twz123: The only other import in the code base calls this package `k0sfake`.

pkg/component/controller/etcd_member_reconciler_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/stretchr/testify/assert"
+	k8stesting "k8s.io/client-go/testing"
+)
+
+func TestReconcileMember(t *testing.T) {
+	reconcile := func(t *testing.T, member *etcdv1beta1.EtcdMember, clusterMembers []etcd.Member, peerAddress string, controllerCount uint) (bool, []k8stesting.Action) {
+		t.Helper()
+		clientset := k0sfake.NewSimpleClientset(member)
+		client := clientset.EtcdV1beta1().EtcdMembers()
+		e := &EtcdMemberReconciler{
```
twz123: I'd rather pass in the whole reconciler here, instead of adding all its fields as function parameters.

twz123 (commented):
The tests currently assert on the member struct that's passed into `reconcileMember`, and that's the same object the reconciler mutates before it calls `UpdateStatus`. So the tests verify the in-memory mutation, but not that anything gets written to the API. The write is the behavior that matters here. So the tests should at least read back the written object from the fake client. That also means you need to add a proper `ObjectMeta`, with a name, so getting the object actually works.

Also, check that the write went to the status subresource. The fake clients don't model this accurately, and if the members get updated as a whole or only their status gets updated is an important nuance. You could e.g. add an assertion helper function that takes the `actions` as parameter for that.

On the commit message: Issue references [go into the PR description](https://github.com/k0sproject/k0s/blob/775515627549e8dff9f384718dfa7ced6346a0ab/docs/contributors/github_workflow.md#commit--push), so the reference to #8211 can be dropped from the commit body.

PR #8292, opened 2026-09-13
fix(autopilot): respect custom status-socket path in autopilot controllers

### LINKED ISSUE #6750: Non-default --status-socket or --run-dir break autopilot

Tyagiquamar:
## Summary
Fixes #6750 by passing the configured `StatusSocketPath` from `CfgVars` down to the autopilot controller's `k0sVersionHandler` instead of using hardcoded `status.DefaultSocketPath`.

## Description
When running `k0s` with a non-default `--status-socket` or `--run-dir`, autopilot failed because `getK0sVersion` in `pkg/autopilot/controller/signal/k0s/init.go` always attempted to read status from `status.DefaultSocketPath` (`/run/k0s/status.sock`).

This change:
1. Adds `StatusSocketPath` to `aproot.RootConfig`.
2. Passes `a.K0sVars.StatusSocketPath` from `pkg/component/controller/autopilot.go` and `pkg/component/worker/autopilot.go` into `aproot.RootConfig`.
3. Passes `StatusSocketPath` through `signal.RegisterControllers` down to `k0s.RegisterControllers`.
4. Uses `statusSocketPath` in `k0sVersionHandler` (falling back to `status.DefaultSocketPath` if empty).
5. Adds unit test `TestGetK0sVersionWithSocketPath`.


pkg/autopilot/controller/signal/k0s/init.go
```diff
@@ -59,7 +59,11 @@ func RegisterControllers(ctx context.Context, logger *logrus.Entry, mgr crman.Ma
 	logger.Infof("Using effective hostname = '%v'", hostname)
 
 	k0sVersionHandler := func() (string, error) {
-		return getK0sVersion(status.DefaultSocketPath)
+		socket := statusSocketPath
+		if socket == "" {
+			socket = status.DefaultSocketPath
+		}
+		return getK0sVersion(socket)
```
twz123, line 66: Personal preference: Use `cmp.Or` to fall back to a default value.

pkg/autopilot/controller/signal/k0s/init_test.go
```diff
@@ -58,3 +58,9 @@ func TestSignalDataUpdateCommandK0sPredicate(t *testing.T) {
 		})
 	}
 }
+
+func TestGetK0sVersionWithSocketPath(t *testing.T) {
```
twz123, line 62: Not sure what's being proved by this test 😅 I'd rather drop it in its current form.

PR #8304, opened 2026-09-15
Add a metadata client to the client factory

twz123:
## Description

Components that need a metadata client have been constructing their own from the client factory's REST config. Those clients aren't cached like the other ones, and they can't be faked, as the fake client factory has no REST config. The `CanUpdate` check test had to spin up an HTTP server and wrap the fake client factory to work around that.

Add a metadata client getter to the client factory and use it wherever a metadata client was previously constructed ad hoc. Accordingly, add a fake metadata client to the fake client factory that shares the object tracker with the other fake clients, so objects created via any of them are visible through it. Rewrite the `CanUpdate` test on top of that.

Fix a few other things in the fake client along the way:

* The transforming tracker's delete passed the namespace instead of the name.
* The fake discovery data advertised empty group versions, which a real API server never does.
* The fake clientset constructor now requires a fake client at compile time instead of asserting it at runtime,
* and it now forwards watch list options to the tracker like client-go's generated fakes do since Kubernetes 1.33.

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [ ] Auto test added

## Checklist

- [x] My code follows the style [guidelines](https://docs.k0sproject.io/head/con


ncopa (approved):
LGTM

PR #8308, opened 2026-09-15
Add JSON Schema for k0s configuration

### LINKED ISSUE #5668: Feature Request:  JSON Schema to validate k0s config

Hey k0 maintainers,

Right now, if I create or edit a `k0s.yaml` file there isn't a way for my IDE to give me hints about the what each key/value expects and I have to refer back to the config file reference https://docs.k0sproject.io/stable/configuration/#configuration-file-reference each time.

Ideally, I'd like my IDE to be able to validate the schema as I'm typing it and give me hints like field descriptions and type warnings.

I was wondering if you'd be interested in having a jsonschema (https://json-schema.org/) generated and checked into the repo for the `k0s.yaml` config file.

It has integrations with a lot of tools to validate the structure of the config file and can be used by IDE's for code-completion.

There's a couple of golang libraries / build tools that could generate this for you using the go structs within the codebase: https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=languages&licenses=&languages=&drafts=&toolingTypes=&environments=&showObsolete=false#go

I've looked for VS Code extensions to support the k0s config file syntax for code completion / validation but it doesn't exist in the marketplace as of today.

Alternatively, I tried the k0s validate command via the CLI. Which works for a high-level validation, but it doesn't assist with actually writing or editing the file moving forward. Just checks if it's a valid config.

### Additional context

This might be a pre-requisite for this issue to be completed: https://github.com/k0sproject/k0sctl/issues/860

seanoseanohay:
## Description

Adds a JSON Schema for editing k0s configuration, generated from the existing ClusterConfig CRD. Editors can suggest fields and allowed values, display descriptions, and flag wrong types, missing required fields, invalid patterns, and values outside declared bounds.

The generator translates the CRD's OpenAPI features into JSON Schema draft-07. It preserves source constraints and annotations, translates integer-or-string and signed 32-bit bounds, and rejects unsupported keywords or formats. It contains no configuration-path rules or config-file defaulting exceptions.

The ImageSpec Go marker uses an equivalent hexadecimal character class supported by both Go and JavaScript. Required fields and stored-object validation constraints are preserved. Arbitrary worker-profile values remain open. Unknown fields are not rejected where the CRD leaves additionalProperties unspecified.

The schema is deliberately stricter than the config-file parser: omit a port to request its default, rather than supplying zero; omit optional fields rather than setting them to null. Defaults are editor annotations, not an implementation of config-file defaulting. Continue using k0s config validate for semantic validation.

## Differences still needing discussion

- **Partial image overrides:** config-file loading merges overrides with defaults, but ImageSpec requires both image and version in the CRD. Making those fields optional would relax stored-object validation, so both remain required.
- **Default annotations:** CRD and config-parser defaults can differ, including Calico's flexVolumeDriverPath. The generator retains the CRD annotations without choosing a different default.
- **addressLabel:** the Go field is uint32, but the generated CRD declares int32. The schema follows that CRD range; no field-specific exception was added. Correcting the source constraint needs a separate decision.
- **Artifact distribution:** schemas/k0s.json remains checked in for use without Go or code generation. The question about tracking it versus publishing a generated artifact remains open.

Kubernetes-specific list constraints, pruning, cross-field rules, and embedded worker/Helm/patch contents are outside the editor schema's coverage.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [x] Manual test
- [x] Auto test added

Validation after rebasing onto upstream main (77c63ce57):

- make config-schema DOCKER=, with byte-for-byte artifact consistency checked by the Go tests.
- go test -count=1 ./hack/config-schema ./pkg/apis/k0s/v1beta1.
- node --test hack/config-schema/regex.test.cjs; also added to codegen CI. No npm dependencies.
- Pinned golangci-lint v2.13.2 on ./hack/config-schema/....
- Official Red Hat YAML language server: field and enum completion, hover, diagnostics, image digests, arbitrary worker values, and int32 boundaries checked over JSON-RPC.
- Regression sensitivity checked locally before rebasing: restoring the Go-only regex fails the JavaScript test; removing feature-name minLength fails the corrected Go fixture.
- go mod tidy -diff and git diff --check.

Full repository tests and live cluster integration tests were not run.

## Checklist

- [x] My code follows the code guidelines of this project
- [x] My commit messages are signed off
- [x] I have commented non-obvious constraints
- [x] My changes generate no new warnings in scoped lint
- [x] I have added tests that prove the feature works
- [ ] New and existing unit tests pass locally with my changes (targeted suites passed; full repository suite not run)
- [x] Any dependent changes have been merged and published in downstream modules (no dependency version changes relative to upstream)
- [x] I have checked my code and corrected misspellings

Fixes #5668


schemas/k0s.json
```diff
@@ -0,0 +1,2374 @@
+{
+  "$comment": "Generated by make config-schema. Do not edit this generated file directly. Use k0s config validate for additional semantic validation.",
```
ncopa, line 2: Do we really need to store `k0s.json` in git, since it is generated?

seanoseanohay: It doesn’t have to be stored in Git. I included it so users can get a schema matching their k0s release without installing Go or running code generation, as described in the editor setup. But it could make more sense to just generate it during release instead of committing the generated file. Does that sound better to you guys? I can easily take this out.

twz123: I'd publish it along with the docs.

twz123:
Thanks for working on this! It's a really thorough approach. However, when I tagged #5668 as a good first issue, I unfortunately didn't realize that the severe mismatch between what k0s accepts as a config file and what is allowed when storing a `ClusterConfig` object within the Kubernetes API would have such a bad impact on the JSON schema generation.

Frankly, I'm hesitant to move forward with this as is. You've put a lot of effort into it, but you were cornered by some suboptimal design choices from the past that turn this feature into technical debt when it should fall out of the CRDs quite naturally. Codifying those rules in the schema generator is at odds with rest of the configuration validation logic, which is wrinkled enough already... It basically duplicates the defaulting behavior, which makes the extra test necessary in the first place, to prevent drift.

Before moving forward, I'd like to explore how far we can close the gap between config files and stored Kubernetes objects. One easy decision could be to consider the laxness of the current config parser a bug rather than a feature, which we'll keep for backwards compatibility. That would eliminate the gaps between absent and zero values. An editor saying "port: 0 isn't valid, omit the field to get the default" would be good UX, not a false positive, even if k0s would still accept such a config. Some more differences (e.g. ImageSpec) look like they could be fixed in the Go type markers directly?

The goal would be a generator that only translates the CRD's OpenAPI dialect into JSON Schema and knows nothing about k0s semantics. I'm not sure how much of the gap we can close, but I'd really like to look into that first. Would you be willing to work in that direction?

seanoseanohay:
Thank you for the explanation I'm going to work in that direction. I can see encoding defaulting exceptions in the generator creates more places that code has to be maintained.

I wanted to accept anything k0s accepts, but stricter editor guidance makes sense. Users can omit a port to get the default instead of setting it to zero.

I’ll review the differences, remove the generator’s special cases, and check what we can fix in the Go type markers without weakening validation. I’ll raise anything unresolved before adding exceptions. The goal is a generator that only translates between schema formats.

schemas/k0s.json
seanoseanohay: @twz123 and @ncopa I'm fine with either way. Should I resolve this comment, or remove k0s.json.

seanoseanohay:
Hi @twz123, I’ve pushed the requested dialect-only revision as a separate second commit. It removes the configuration-file and defaulting exceptions, so the generator now translates the CRD OpenAPI schema to Draft 7 JSON Schema only.

The fork PR checks are awaiting maintainer approval. When you have a chance, could you approve CI and let me know whether this direction matches what you had in mind? I’ve left the generated-artifact discussion open pending that feedback.

PR #8342, opened 2026-09-19
Add --allow-cached-config to start workers with the cached profile

### LINKED ISSUE #7379: Support offline worker startup with cached kubelet configuration

gshklovs:
## Description

Workers load their worker profile from the `worker-config-<profile>-<k8s-minor>` ConfigMap at startup. When the Kubernetes API is unreachable, `LoadProfile` retries for roughly four minutes and then returns an error, `k0s worker` exits, and under a service manager it crash-loops. The kubelet never starts, so nothing on the node recovers, including static pods that would otherwise run without a control plane.

This matters for edge nodes that boot on intermittent connectivity. After a power loss, such a node's worker often comes up before its uplink does, and today that means the node stays down until someone intervenes, even though it already has everything it needs on disk.

The worker already caches the full profile in `<data-dir>/worker-profile.yaml` after every successful load, but until now only used that cache to seed API server addresses for node-local load balancing.

This change adds an opt-in `--allow-cached-config` worker flag. With the flag:

- k0s still loads the profile from the API first, with the existing retry behaviour. A fresh server response always wins and always rewrites the cache.
- Only if that fails because the API server was unreachable does k0s start with the cached profile, and it logs a warning. `loadProfile` marks that one case with a dedicated error type, so a fetch that succeeded and then failed to be parsed or persisted is never downgraded to a stale profile. Rejected or forbidden credentials and a shutting-down context are excluded too.
- The cached profile is refused if it was written for another `--profile` name or another Kubernetes minor version. To support the version check, the cache file now records `kubernetesVersion`. A cache written before that field existed is refused as well; it is rewritten in the current format on the next successful start.
- The fallback path never writes the cache, so a stale profile cannot be promoted.
- Without a usable cache, the worker exits with both the API error and the reason the cache was unusable.

Without the flag, behaviour is unchanged.

Deliberately out of scope: shortening the initial retry window, re-fetching the profile in the background after an offline start, and hot-reloading the kubelet configuration. A worker that started from the cache keeps that configuration until it restarts, which matches how a running worker already behaves when the API goes away.

Related: #7685 by @azmifarih proposes the same `--allow-cached-config` flag and has been open since May with merge conflicts and no review. I opened this as a fresh, rebased take rather than a competing one: same user-facing shape, but the fallback decision keys off a dedicated error type from the loader instead of classifying error strings, HTTP-level rejections stay fatal, the cache carries a Kubernetes version guard, and the docs are included. Happy to coordinate with @azmifarih or defer to #7685 if they want to pick it back up.

Fixes #7379

## Follow-ups

- A worker that started from the cache keeps that kubelet configuration until it restarts. The only runtime watcher of the worker profile is the node-local load balancer reconciler, which refreshes the on-disk cache and its own config once the API is back but does not re-render the kubelet configuration. Re-fetching and reloading after an offline start is left for a later change.
- Before the fallback kicks in, the worker still runs the full retry window of roughly four minutes. That is long for a node rebooting after a power loss. Making the initial attempt bounded or the retry tunable would be a natural follow-up.

## Type of change

- [x] Documentation update

## How Has This Been Tested?

- [ ] Manual test
- [x] Auto test added

New unit tests in `pkg/component/worker/config/loader_test.go`:

- `TestLoadProfileFromCache`: round-trip, missing cache reported as `fs.ErrNotExist`, corrupt cache, wrong profile name, wrong Kubernetes version, cache without a recorded Kubernetes version.
- `TestLoadProfile_Cache`: a successful load overwrites a differing cache; an aborted load leaves the cache untouched.
- `TestShouldFallBackToCachedProfile` in `cmd/worker`: the fallback predicate accepts an unreachable API server and rejects invalid credentials, insufficient permissions, a cancelled or expired context, and unrelated failures, each also when wrapped.

Ran locally on Windows with Go 1.27: `go build ./...`, `gofmt -l`, `go test -tags=hack` for `pkg/component/worker/config`, `cmd/worker`, `cmd/controller` and `pkg/config`, and `golangci-lint v2.13.2 run --build-tags=hack` on the changed packages (0 issues). The golden help-text tests, which are Linux-only, were also run in a `golang:1.27` container and pass. Not run: `-race` (no cgo toolchain on the Windows machine).

## Checklist

- [x] My code follows the [code guidelines](https://docs.k0sproject.io/head/contributors/code_guidelines/) of this project
- [x] I have commented my code, particularly explaining any non-obvious constraints

🤖 Generated with [Claude Code](https://claude.com/claude-code)


azmifarih:
Thanks for picking this up @gshklovs! Happy for #8342 to supersede #7685 — please go ahead, I'll close mine.

gshklovs:
No problem @azmifarih! Thank you for handing it off.

